PR #50, opened 2022-04-07
Proposal for expectations testing

Morriar:
### Motivation

Simplify the way we write tests.

### How to add a new test for an existing request handler

(edited for the last iteration)

1. Add a new fixture `my_fixture.rb` file under `test/fixtures`
2. (optional) Add new expectations under `test/expectations/$HANDLER` for the request handlers you're concerned by
3. Profit by running `bin/test test/requests/$HANDLER_expectations_test my_fixture`
    * Handlers for which you added expectations will be checked with `assert_expectations`
    * Handlers without expectations will be ran against your new test to check that nothing breaks
    
### How to add a new request handler

1. Add a new file under `test/requests` that subclasses `ExpectationsTestRunner` and call `expectations_tests $HANDLER, "$EXPECTATION_DIR"` where: `$HANDLER` is the fully qualified name or your handler class and `$EXPECTATION_DIR` is the directory name where you want to store the expectation files.

2. (optional) Override the `run_expectations` and `assert_expectations` methods if needed. See the different request handler expectation runner for examples.

4. (optional) Add new fixtures for your handler under `test/fixtures`

5. (optional) Add new expectations under `test/expectations/$HANDLER`
   * No need to write the expectations by hand, just run the test with an empty expectation file and copy from the output.

7. Profit by running, `bin/test test/expectations_test $HANDLER`
    * Tests with expectations will be checked with `assert_expectations`
    * Tests without expectations will be ran against your new $HANDLER to check that nothing breaks

### TODO

* [ ] I think we can simplify the test output a bit
* [ ] We may be able to simplify how we write the `run_x`, `assert_x` to not have to duplicate code
* [ ] We need to somehow catch the stderr of `Formatting` to avoid cluttering the output
* [ ] Migrate `SemanticHighlighting` but I wanted to show that just adding the mechanics already caught two errors:

   image


vinistock:
@st0012 the PR description is probably outdated now, we changed the folder for fixtures.

I agree that there might be a learning curve in terms of knowing how to read the tests. However, we had decided moving forward with this approach because
- We can re-use Ruby syntax fixtures across different tests
- All requests are forced to test every fixture, which means that adding new fixtures increases coverage for all requests and assures that all of them can handle that piece of syntax

Also, you don't need that many steps to check what a test does. Each fixture is a different test case and then you can look into the corresponding expectation file for a specific request to find out what is expected for that piece of syntax.

vinistock (approved):

Morriar:
I updated the pull-request description to match the new behavior.

As Vini pointed out, the new framework greatly simplifies the way we write tests and allows us to reuse and share the Ruby fixtures across expectations runners for all the handlers which will hopefully catch more errors than if we need to duplicate the tests by hand every time.

In a normal setting, when you know how the framework words, you only have two files to look at:

1. The fixture file
2. The expectation file for your handler

We added a new section in the README under `Contributing` to explain how to use the framework.

If you need to create new fixtures, expectations or runners, you'll notice that most steps are actually optional.

Issue #87, opened 2022-04-28
Add document highlight to the `ruby-lsp`


vinistock:
The document highlight request highlights a range of interest depending on the position of the cursor.

For example, we can highlight all occurrences of the same local variable. Or highlight the instance variable associated to an `attr_reader`.

Documentation: https://microsoft.github.io/language-server-protocol/specification#textDocument_documentHighlight

PR #91, opened 2022-05-03
Add document highlight

st0012:
### Motivation

Incremental support of https://github.com/Shopify/ruby-lsp/issues/87: 

- Global variables
- Constants (without scoping)
- Instance variables (without scoping)
- Class variables (without scoping)
- Local variables (without scoping)
- Method identifiers (without scoping)

### Implementation

~~There are 2 parts that are unique to this feature (which may due to my unfamiliarity to the tools):~~

1. ~~Because VSCode only provides the cursor's location in the request (a single integer instead of a range), we need to resolve the full token ourselves. And to do that, I use the `Ripper` class to tokenize the line, and comparing the cursor location with each token's start position its the next token's in the `find_target_token` method.~~
2. ~~The specification supports highlighting reading/writing to a variable differently by assigning each highlight a different `kind` attribute. But because there's no link from a node to its parent, we can't know if the `@foo` in `@foo = 1` is inside an assignment from the node itself. Therefore, based on @kddnewton's advice, I use a stack in `visit_assign` to keep tracking if we're inside an assignment.~~

I applied @kddnewton's [suggestion](https://github.com/Shopify/ruby-lsp/pull/91#issuecomment-1117794988) and it solved the above 2 problems 🎉 

### Demo

https://user-images.githubusercontent.com/5079556/166711285-14efe443-a105-447e-9a52-ffc447ba409e.mov

### Automated Tests

Written in the `test/requests/document_highlight_test.rb` file.

### Manual Tests

**On `vscode-ruby-lsp`**

1. Switch to the branch `add-document-highlight`
2. Rebuild the extension from the branch

**On `ruby-lsp`**

1. Turn off the [Ruby extension](https://marketplace.visualstudio.com/items?itemName=rebornix.Ruby) and reload
2. Switch to the branch: `add-document-highlight`
4. Install the `ruby-lsp` extension from the newly built source
5. Restart the Ruby LSP
6. In any Ruby file, click on a variables that have read/write occurrences in the same file.
7. Check:
    1. There are messages like these in the `Ruby LSP` output
      ```
      [Trace - 6:13:06 PM] Sending request 'textDocument/documentHighlight - (17)'.
      [Trace - 6:13:06 PM] Received response 'textDocument/documentHighlight - (17)' in 3ms.
      ```
    2. The read/write occurrences are highlighted differently, like the `@highlights` variable in this screenshot:
  
  Screenshot 2022-05-04 at 18 29 28

### DIFF
--- a/lib/ruby_lsp/requests/document_highlight.rb
+++ b/lib/ruby_lsp/requests/document_highlight.rb
@@ -25,18 +25,41 @@ module RubyLsp
       end
 
       def initialize(document, position)
+        @highlights = []
         position = Document::Scanner.new(document.source).find_position(position)
-        @visitor = find(document.tree, position)
+        @target = find(document.tree, position)
 
         super(document)
       end
 
       def run
-        # no visitor means the target is not highlightable
-        return [] unless @visitor
+        # no @target means the target is not highlightable
+        return [] unless @target
 
-        @visitor.visit(@document.tree)
-        @visitor.highlights
+        visit(@document.tree)
+        @highlights
+      end
+
+      def visit_var_field(node)
+        if matches_target?(node.value)
+          add_highlight(
+            node.value,
+            LanguageServer::Protocol::Constant::DocumentHighlightKind::WRITE
+          )
+        end
+
+        super
+      end
+
+      def visit_var_ref(node)
+        if matches_target?(node.value)
+          add_highlight(
+            node.value,
+            LanguageServer::Protocol::Constant::DocumentHighlightKind::READ
+          )
+        end
+
+        super
       end
 
       private
@@ -52,88 +75,20 @@ module RubyLsp
           end
 
         case matched
-        when SyntaxTree::GVar
-          GVarHighlightVisitor.new(matched)
-        when SyntaxTree::Ident
-          IdentHighlightVisitor.new(matched)
-        when SyntaxTree::IVar
-          IVarHighlightVisitor.new(matched)
-        when SyntaxTree::Const
-          ConstHighlightVisitor.new(matched)
-        when SyntaxTree::CVar
-          CVarHighlightVisitor.new(matched)
+        when SyntaxTree::GVar, SyntaxTree::Ident, SyntaxTree::IVar, SyntaxTree::Const, SyntaxTree::CVar
+          matched
         when SyntaxTree::Node
           find(matched, position)
         end
       end
 
-      class HighlightVisitor < SyntaxTree::Visitor
-        WRITE_KIND = 3
-        READ_KIND = 2
-
-        attr_reader :highlights
-
-        def initialize(target_node)
-          @highlights = []
-          @target = target_node.value
-
-          super()
-        end
-
-        def visit_var_field(node)
-          highlight(node.value, WRITE_KIND) if matches_target?(node.value)
-        end
-
-        def visit_var_ref(node)
-          highlight(node.value, READ_KIND) if matches_target?(node.value)
-        end
-
-        private
-
-        def highlight(node, kind)
-          range = range_from_syntax_tree_node(node)
-          @highlights << LanguageServer::Protocol::Interface::DocumentHighlight.new(range: range, kind: kind)
-        end
-
-        def range_from_syntax_tree_node(node)
-          loc = node.location
-
-          LanguageServer::Protocol::Interface::Range.new(
-            start: LanguageServer::Protocol::Interface::Position.new(line: loc.start_line - 1,
-              character: loc.start_column),
-            end: LanguageServer::Protocol::Interface::Position.new(line: loc.end_line - 1, character: loc.end_column),
-          )
-        end
+      def matches_target?(node)
+        node.is_a?(@target.class) && node.value == @target.value
       end
 
-      class IVarHighlightVisitor < HighlightVisitor
-        def matches_target?(node)
-          node.is_a?(SyntaxTree::IVar) && node.value == @target
-        end
-      end
-
-      class GVarHighlightVisitor < HighlightVisitor
-        def matches_target?(node)
-          node.is_a?(SyntaxTree::GVar) && node.value == @target
-        end
-      end
-
-      class IdentHighlightVisitor < HighlightVisitor
-        def matches_target?(node)
-          node.is_a?(SyntaxTree::Ident) && node.value == @target
-        end
-      end
-
-      class ConstHighlightVisitor < HighlightVisitor
-        def matches_target?(node)
-          node.is_a?(SyntaxTree::Const) && node.value == @target
-        end
-      end
-
-      class CVarHighlightVisitor < HighlightVisitor
-        def matches_target?(node)
-          node.is_a?(SyntaxTree::CVar) && node.value == @target
-        end
+      def add_highlight(node, kind)
+        range = range_from_syntax_tree_node(node)
+        @highlights << LanguageServer::Protocol::Interface::DocumentHighlight.new(range: range, kind: kind)
       end
     end
   end
--- a/lib/ruby_lsp/requests/base_request.rb
+++ b/lib/ruby_lsp/requests/base_request.rb
@@ -17,6 +17,16 @@ module RubyLsp
       def run
         raise NotImplementedError, "#{self.class}#run must be implemented"
       end
+
+      def range_from_syntax_tree_node(node)
+        loc = node.location
+
+        LanguageServer::Protocol::Interface::Range.new(
+          start: LanguageServer::Protocol::Interface::Position.new(line: loc.start_line - 1,
+            character: loc.start_column),
+          end: LanguageServer::Protocol::Interface::Position.new(line: loc.end_line - 1, character: loc.end_column),
+        )
+      end
     end
   end
 end
--- a/lib/ruby_lsp/requests/document_symbol.rb
+++ b/lib/ruby_lsp/requests/document_symbol.rb
@@ -210,16 +210,6 @@ module RubyLsp
 
         symbol
       end
-
-      def range_from_syntax_tree_node(node)
-        loc = node.location
-
-        LanguageServer::Protocol::Interface::Range.new(
-          start: LanguageServer::Protocol::Interface::Position.new(line: loc.start_line - 1,
-            character: loc.start_column),
-          end: LanguageServer::Protocol::Interface::Position.new(line: loc.end_line - 1, character: loc.end_column),
-        )
-      end
     end
   end
 end
--- a/test/integration_test.rb
+++ b/test/integration_test.rb
@@ -60,7 +60,7 @@ class IntegrationTest < Minitest::Test
     )
 
     range = response[:result].first
-    assert_equal(RubyLsp::Requests::DocumentHighlight::HighlightVisitor::WRITE_KIND, range[:kind])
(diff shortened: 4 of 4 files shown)


st0012:
@kddnewton Thanks for all the suggestions! But I hit another obstacle right away 😂 

I'm trying to implement another binary search with line number (the sample only considered character positions). And I found that `bodystmt`'s `start_line` and `end_line` doesn't seem to be correct.

I have a simple script like this:
```rb
class Foo # line 1
  def initialize
    @foo = 10
  end

  def foo
    @foo
  end
end # line 9
```

After parsing it, the `SyntaxTree::ClassDeclaration` node has the right start/end line (1, 9):

```
(rdbg) node
(class (const_ref (const "Foo")) (bodystmt (statements ((void_stmt), (def (ident "initialize") (params) (bodystmt (statements ((assign (var_field (ivar "@foo")) (int "10")))))), (def (ident "foo") (params) (bodystmt (statements ((var_ref (ivar "@foo"))))))))))
(ruby) node.location
#<SyntaxTree::Location:0x0000000106d5aa28 @end_char=76, @end_column=3, @end_line=9, @start_char=0, @start_column=0, @start_line=1>
```

However, the `SyntaxTree::BodyStmt` of it, has the same start/end line (9, 9), even though it covers the entire body of the class:

```
(ruby) node.child_nodes.last
(bodystmt (statements ((void_stmt), (def (ident "initialize") (params) (bodystmt (statements ((assign (var_field (ivar "@foo")) (int "10")))))), (def (ident "foo") (params) (bodystmt (statements ((var_ref (ivar "@foo")))))))))
(ruby) node.child_nodes.last.location
#<SyntaxTree::Location:0x0000000106d60db0 @end_char=73, @end_column=0, @end_line=9, @start_char=9, @start_column=9, @start_line=9>
```

That means the binary search will fail here because it failed to recognize that it `SyntaxTree::BodyStmt` actually covers the range it needs to dig into.

Is this an intentional design? If so, how can a get around it?

kddnewton:
I’m not sure why the line numbers are wrong. I can look into that. But the
sample is using start_char, not start_column. The former is the overall
position in the document, so it should be fine to use that.

kddnewton:
Looks like there's already something in this gem that does that (https://github.com/Shopify/ruby-lsp/blob/main/lib/ruby_lsp/document.rb#L80-L97). 

The issue is, you're still just tokenizing, which doesn't contain semantic meaning only lexical. You're going to run into issues where you're highlighting an ident within a method name, for instance. This can be mitigated by checking if the previous token is `def`, but there are so many cases that you're going to end up recreating the entire parser anyway. Since `lex` only gives you a string, you can't know if you're highlighting something you actually want to be highlighting or not. 

For example, consider this:

```ruby
Ripper.lex("foo; :foo")
# => [[[1, 0], :on_ident, "foo", CMDARG], [[1, 3], :on_semicolon, ";", BEG], [[1, 4], :on_sp, " ", BEG], [[1, 5], :on_symbeg, ":", FNAME], [[1, 6], :on_ident, "foo", ENDFN]]
```

Right now with your lexing approach, if the position is at line 0 column 5, it's going to say that it's hovering over an `on_ident` token with a value of `"foo"`. Then it's going to highlight every `Ident` node with `"foo"` as the value, which means it's going to highlight the `foo` method call as well. This is because it doesn't know that the second `on_ident` is actually the value of a symbol and not something that should be highlighted.

st0012:
Yeah what you said makes sense. Thank you for the explanation 👍 Using parser node will enable more sophisticated operation when implementing advanced features.

But to continue on the Syntax Tree approach, I need to resolve the incorrect start/end line issue first. I've created an isolated script to reproduce that:

```rb
require "syntax_tree"

code = <<~RUBY
  class Foo
    def initialize
      @foo = 10
    end

    def foo
      @foo
    end
  end
RUBY

tree = SyntaxTree.parse(code)
class_node = tree.child_nodes.last.child_nodes.last
body_stmt = class_node.child_nodes.last
pp(body_stmt)
puts([body_stmt.location.start_line, body_stmt.location.end_line])
```


**result**
```
(bodystmt
  (statements
    ((void_stmt),
      (def (ident "initialize") (params) (bodystmt (statements ((assign (var_field (ivar "@foo")) (int "10")))))),
      (def (ident "foo") (params) (bodystmt (statements ((var_ref (ivar "@foo")))))))))
9
9
```

vinistock:
So, a few comments based on the current conversations:
1. The lines are correct. SyntaxTree is one based and VS Code is zero based, we just subtract one from the lines everywhere
2. In terms of finding the correct token first and then finding the matches, what do you think about something like this? It would maintain the style of current requests
```ruby
class DocumentHighlight
  def initialize(document, position)
    ...
    @target_token = nil
  end

  def run
    visit(@tree) # first visit, looking for the target_token
    visit(@tree) # second visit, finds occurrences
    @highlights
  end

  def visit_global_var(node)
    if @target_token
      # check that node matches the target token to add a highlight
    else
      # we haven't found the target token yet, so check if positions match and set it
      @target_token = node
    end
  end
end
```

vinistock:
@kddnewton by splitting up visitors, do you mean having a visitor for finding the token and another one to do the highlights? I like that idea, we'll likely need the visitor for finding tokens in other requests.

kddnewton:
I don't think you should find the token using a visitor. Check out the binary search thing I posted above. I think that will be a lot better than trying to use a visitor since it'll be faster because it won't have to walk through every node in the tree at worst case.

I think you should have one visitor for each kind of highlighting you would want to do. One for global variables, one for instance variables, etc. Then you'll just be visiting specific nodes and it'll be easier to test without having it all smooshed together.

lib/ruby_lsp/requests/document_highlight.rb
```diff
@@ -0,0 +1,153 @@
...
+        def matches_target(node)
+          node.is_a?(SyntaxTree::Ident) && node.value == @target.value
+        end
+      end
+
+      class ConstHighlightVisitor < HighlightVisitor
+        def matches_target(node)
+          node.is_a?(SyntaxTree::Const) && node.value == @target.value
+        end
+      end
```
st0012, line 150: Right now the each sub-visitor and their `matches_target` methods look duplicated. But in the follow-up PRs each of them will have different scope-resolving implementations.

lib/ruby_lsp/requests/document_highlight.rb
```diff
@@ -0,0 +1,153 @@
...
+        end
+
+        def visit_assign(node)
+          @assign_node_stack << node
+          super
+        ensure
+          @assign_node_stack.pop
+        end
+
+        def visit_var_field(node)
```
st0012, line 87: @kddnewton  Can I say `var_field` always represents a "write" to the variable, while `var_ref` is always "read"? Is there any exception?

lib/ruby_lsp/requests/document_highlight.rb
```diff
@@ -0,0 +1,94 @@
...
+    #
+    # ```ruby
+    # FOO = 1 # should be highlighted as "write"
+    #
+    # def foo
+    #   FOO # should be highlighted as "read"
+    # end
+    # ````
+    class DocumentHighlight < BaseRequest
+      WRITE_KIND = 3
```
vinistock, line 23: Let's use the constants already exposed by the [language server gem](https://github.com/mtsmfm/language_server-protocol-ruby/blob/main/lib/language_server/protocol/constant/document_highlight_kind.rb).

lib/ruby_lsp/requests/document_highlight.rb
```diff
@@ -0,0 +1,94 @@
...
+        when SyntaxTree::Node
+          find(matched, position)
+        end
+      end
+
+      def matches_target?(node)
+        node.is_a?(@target.class) && node.value == @target.value
+      end
+
+      def highlight(node, kind)
```
vinistock, line 78: NIT: can we rename it to `add_highlight`?

lib/ruby_lsp/requests/document_highlight.rb
```diff
@@ -0,0 +1,94 @@
...
+        range = range_from_syntax_tree_node(node)
+        @highlights << LanguageServer::Protocol::Interface::DocumentHighlight.new(range: range, kind: kind)
+      end
+
+      def range_from_syntax_tree_node(node)
```
vinistock, line 83: We have this method in `DocumentSymbol` too. Can we extract it into `BaseRequest` and remove it from both `DocumentHighlight` and `DocumentSymbol`?

vinistock (approved):

lib/ruby_lsp/requests/document_highlight.rb
```diff
@@ -0,0 +1,87 @@
...
+
+        visit(@document.tree)
+        @highlights
+      end
+
+      def visit_var_field(node)
+        add_highlight(
+          node.value,
+          LanguageServer::Protocol::Constant::DocumentHighlightKind::WRITE
+        ) if matches_target?(node.value)
```
paracycle, line 47: This is super nitpicky, but can we flip this around, since I think post conditionals are really hard to read on multiline statements?
```suggestion
        return unless matches_target?(node.value)
        
        add_highlight(
          node.value,
          LanguageServer::Protocol::Constant::DocumentHighlightKind::WRITE
        )
```

paracycle: Also, should it be `matches_target?(node)` or `matches_target?(node.value)`?

lib/ruby_lsp/requests/document_highlight.rb
```diff
@@ -0,0 +1,87 @@
...
+      end
+
+      def visit_var_ref(node)
+        add_highlight(
+          node.value,
+          LanguageServer::Protocol::Constant::DocumentHighlightKind::READ
+        ) if matches_target?(node.value)
```
paracycle, line 54: Same comment as above.

lib/ruby_lsp/requests/document_highlight.rb
```diff
@@ -0,0 +1,87 @@
...
+
+      def run
+        # no @target means the target is not highlightable
+        return [] unless @target
+
+        visit(@document.tree)
+        @highlights
+      end
+
+      def visit_var_field(node)
```
paracycle, line 43: Do we need to call `super` in these visitors so that their children can be visited as well? I would assume we could have nested `VarField` and `VarRef`s within each other, and if we don't call super, then we would not be able to match the nested ones.

lib/ruby_lsp/requests/document_highlight.rb
st0012: It's `matches_target?(node.value)` because the node here is a `VarField` or `VarRef`. And their value (e.g. `Ident`) is what we try to match.

lib/ruby_lsp/requests/document_highlight.rb
```diff
@@ -0,0 +1,95 @@
...
+        if matches_target?(node.value)
```
st0012, line 44: @paracycle In this case, I think the plain if condition is easier to understand than `return unless matches_target?(node.value)`, which requires an extra step when thinking.

paracycle: 👍

paracycle (approved):

st0012:
I will merge this after `0.0.3` is released.

PR #96, opened 2022-05-12
Use the correct path for reading the VERSION file

vinistock:
### Motivation

When the gem is installed and used in projects, trying to read `VERSION` without making it relative to the gem file paths will attempt to read the incorrect path.


Morriar (approved):
Why do we need the version to be in a separate `VERSION` file rather than in `lib/ruby-lsp/version.rb` for example?

wildmaples (approved):

PR #97, opened 2022-05-13
Add documentation publish workflow

### LINKED ISSUE #89: Publish our documentation as a static website
It would be nice to have our yard documentation published as a static website and linked from our README. Investigate the best practices around this and add a GitHub action to publish the website when something is merged to `main`.

To generate the documentation, simply run
```shell
bundle exec rake yard
```

vinistock:
### Motivation

Closes #89

Add GitHub action automation to publish our website to GH pages.

### Implementation

According to the [checkout action documentation](https://github.com/actions/checkout#push-a-commit-using-the-built-in-token), we should be able to push new commits using this mechanism. The action does the following:
1. Configures the git user and email
2. Resets the gh-pages branch to match main
3. Generates the documentation and moves it to a folder that is not ignored by git
4. Commits the folder using the message `Publish website LAST_SHA`
5. Pushes to gh-pages

If this action works and we're able to push the generates docs to the branch, getting the website up should just be a matter of changing the repo's settings.

### Automated Tests

N/A.

### Manual Tests

N/A.

### DIFF
--- a/.github/workflows/publish_docs.yml
+++ b/.github/workflows/publish_docs.yml
@@ -1,0 +1,39 @@
+name: Publish docs
+
+on:
+  push:
+    branches: [main]
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    name: Publish documentation website
+    steps:
+      - uses: actions/checkout@v3
+
+      - name: Set up Ruby
+        uses: ruby/setup-ruby@v1
+        with:
+          ruby-version: 3.1.1
+          bundler-cache: true
+
+      - name: Configure git
+        run:
+          - git config user.name github-actions
+          - git config user.email github-actions@github.com
+
+      - name: Reset gh-pages branch
+        run:
+          - git switch gh-pages
+          - git reset --hard origin/main
+
+      - name: Generate documentation
+        run:
+          - bundle exec rake yard
+          - mv doc docs
+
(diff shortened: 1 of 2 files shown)


egiurleo (approved):

.github/workflows/publish_docs.yml
```diff
@@ -0,0 +1,39 @@
...
+
+      - name: Reset gh-pages branch
+        run:
+          - git switch gh-pages
+          - git reset --hard origin/main
+
+      - name: Generate documentation
+        run:
+          - bundle exec rake yard
+          - mv doc docs
```
wildmaples, line 33: I think you may be able to use `yard --output-dir docs/` to generate the documents directly into the folder we want.

.github/workflows/publish_docs.yml
```diff
@@ -0,0 +1,39 @@
...
+
+      - name: Commit to gh-pages
+        run:
+          - git add docs
+          - git commit -m "Publish website $(git log --format=format:%h -1)"
+          - git push origin +gh-pages
```
wildmaples, line 39: Instead of doing the reset and push to `gh-pages` branch jobs manually, another option is to use a Github Action for it: https://github.com/peaceiris/actions-gh-pages

I think that'll simplify the code here, but the drawback is that we're relying on a Github Action.

PR #99, opened 2022-05-13
Fix publish_docs run syntax

vinistock:
In order to use multiple commands in run, we have to use the script syntax using a pipe. Listing multiple items is not allowed and makes the action fail.


egiurleo (approved):
Makes sense to me

PR #101, opened 2022-05-13
Avoid switching branches in publish_docs

vinistock:
### Motivation

Our action is [failing](https://github.com/Shopify/ruby-lsp/runs/6428844612?check_suite_focus=true), this time when trying to switch branches.

### Implementation

I've tried simplifying the action a little further. Instead of switching to the `gh-pages` branch and resetting, we just create the commit locally in `main`, but push it to a different branch.

### DIFF
--- a/.github/workflows/publish_docs.yml
+++ b/.github/workflows/publish_docs.yml
@@ -22,11 +22,6 @@ jobs:
           git config user.name github-actions
           git config user.email github-actions@github.com
 
-      - name: Reset gh-pages branch
-        run: |
-          git switch gh-pages
-          git reset --hard origin/main
-
       - name: Generate documentation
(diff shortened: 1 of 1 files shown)


egiurleo (approved):
❤️ I definitely agree with simplifying the workflow.

One question -- why do we have to force push? Is it because the source and destination branches are different?

vinistock:
@egiurleo The force push is just to avoid having to rebase the `gh-pages` branch and deal with conflicts. Whatever docs are generated in the current `main`, we want to push. The purpose of the branch is just for GH pages, so there shouldn't be any "real" commits there.

egiurleo:
@vinistock Okay that makes sense.

PR #109, opened 2022-05-16
Always push telemetry/event notifications

vinistock:
### Motivation

This work is necessary to help address the problem described in https://github.com/Shopify/vscode-ruby-lsp/pull/85.

It's brittle to let the plugin tell the server if `telemetry/event` notifications should be pushed or not. It's more robust to always send the notifications and just let the plugin decide whether to discard or do something with them.

### Implementation

Removed the check for whether telemetry was enabled or not and also removed the ignored requests logic. It's much simpler to just send everything to the plugin and let it decide.

### Automated Tests

Added the extra assertions expected from always sending telemetry notifications.

### Manual Tests

1. Start the plugin in debug mode using https://github.com/Shopify/vscode-ruby-lsp/pull/85 and this PR's branch for the server
2. Verify that telemetry events are printed in the debug console of the plugin window


KaanOzkan (approved):

vinistock:
@paracycle in the second commit, I have refactored the `with_telemetry` method a bit. Let me know what you think.

Morriar (approved):

lib/ruby_lsp/handler.rb
```diff
@@ -162,29 +158,36 @@ def respond_with_document_highlight(uri, position)
...
-    end
-
     def with_telemetry(request)
-      params = { request: request[:method], lspVersion: RubyLsp::VERSION }
       result = nil
+      error = nil
 
       request_time = Benchmark.realtime do
         result = yield
       rescue StandardError => e
```
paracycle, line 167: Question: Should we catch all kinds of errors here? For example, what if we get a `LoadError` if we've messed up an autoload path or something? Maybe we should rescue `Exception` instead?

paracycle (approved):

lib/ruby_lsp/handler.rb
vinistock: So, I tested and it looks like `Interrupt` is an `Exception`. I'm afraid swallowing it might make VS Code unable to stop the server if necessary. What do you think about adding `LoadError` to the rescued errors instead?

paracycle: Right, that's a good argument against handling `Exception`. Ok, let's leave it as is in that case, we should be able to catch `LoadError`s and others during test, hopefully.

PR #112, opened 2022-05-24
Implement token modifiers in SemanticTokenEncoder

wildmaples:
### Motivation

We want to utilize the token modifiers API from the LSP for better semantic highlighting. 

https://github.com/Shopify/ruby-dev-exp-issues/issues/518#issuecomment-1112681273

### Implementation

This is just implementing the code to encode the modifiers. We're not using it yet. 

The changes can be seen more clearly in the test cases. But essentially, we're encoding modifiers to each tokens. 

### ⭐️ Walk Through The Entire Flow! 

For example:

A token can be marked with the `:definition` and `:abstract` modifiers (maybe not in Ruby, lol). 

We first map these modifiers using `TOKENS_MODIFIERS` hash to get the indices of the each of the modifiers. This results in `[1, 5]`.

Using the array of indices, we pass that onto `SemanticTokenEncoder` which then encodes these indices into the appropriate bit map to be passed with the request.  

### Automated Tests

I've added the appropriate tests that covers all the cases. 

### Manual Tests

No manual tests yet because we aren't using the changes yet.

### DIFF
--- a/lib/ruby_lsp/requests/semantic_highlighting.rb
+++ b/lib/ruby_lsp/requests/semantic_highlighting.rb
@@ -21,7 +21,19 @@ module RubyLsp
         :variable,
         :method,
       ].freeze
-      TOKEN_MODIFIERS = [].freeze
+
+      TOKEN_MODIFIERS = {
+        declaration: 0,
+        definition: 1,
+        readonly: 2,
+        static: 3,
+        deprecated: 4,
+        abstract: 5,
+        async: 6,
+        modification: 7,
+        documentation: 8,
+        default_library: 9,
+      }.freeze
 
       SemanticToken = Struct.new(:location, :length, :type, :modifier)
 
@@ -90,9 +102,10 @@ module RubyLsp
         add_token(node.value.location, :method)
       end
 
-      def add_token(location, type)
+      def add_token(location, type, modifiers = [])
         length = location.end_char - location.start_char
-        @tokens.push(SemanticToken.new(location, length, TOKEN_TYPES.index(type), 0))
+        modifiers_indices = modifiers.map { |modifier| TOKEN_MODIFIERS[modifier] }
+        @tokens.push(SemanticToken.new(location, length, TOKEN_TYPES.index(type), modifiers_indices))
       end
     end
   end
--- a/lib/ruby_lsp/requests/semantic_highlighting.rb
+++ b/lib/ruby_lsp/requests/semantic_highlighting.rb
@@ -104,7 +104,7 @@ module RubyLsp
 
       def add_token(location, type, modifiers = [])
         length = location.end_char - location.start_char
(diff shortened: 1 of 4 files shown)


lib/ruby_lsp/requests/semantic_highlighting.rb
```diff
@@ -89,9 +101,10 @@ def visit_vcall(node)
         add_token(node.value.location, :method)
       end
 
-      def add_token(location, type)
+      def add_token(location, type, modifiers = [])
         length = location.end_char - location.start_char
-        @tokens.push(SemanticToken.new(location, length, TOKEN_TYPES.index(type), 0))
+        modifiers_indices = modifiers.map { |modifier| TOKEN_MODIFIERS[modifier] }
```
wildmaples, line 106: This assumes that the modifier we pass will exist in the `TOKEN_MODIFIERS` hash map. Currently it returns `nil` if the key doesn't exist. We should probably silently warn if the key doesn't exist.

st0012: I think that'll only happen when VSCode updates modifier names, since all 10 digits are all taken so there won't be new ones? Just curious.
But yeah it'll be helpful for debugging if such situation is logged. I wonder if there's a way to get notified whenever the specification changed though 🤔 

Debugger's [Debug Adapter Protocol](https://microsoft.github.io/debug-adapter-protocol/specification) has a [json schema](https://microsoft.github.io/debug-adapter-protocol/debugAdapterProtocol.json) that we can use to validate the format in tests. I can't find anything similar for LSP though.

vinistock: Let's change `map` to `filter_map` to automatically ignore modifiers we don't know about and avoid having to handle `nil`.

lib/ruby_lsp/requests/support/semantic_token_encoder.rb
```diff
@@ -38,11 +38,22 @@ def compute_delta(token)
...
         end
+
+        # Encode an array of modifiers to positions onto a bit flag
+        # For example, [:default_library] will be encoded as
+        # 0b1000000000, as :default_library is the 10th bit according
+        # to the token modifiers index map.
+        def encode_modifiers(modifiers)
+          modifiers.inject(0) do |encoded_modifiers, modifier|
+            encoded_modifiers |= (1 << modifier)
+            encoded_modifiers
```
vinistock, line 54: I think the previous line already returns the result, so you may not need this one.

paracycle: Yeah this can just be:
```ruby
          modifiers.inject(0) do |encoded_modifiers, modifier|
            encoded_modifiers | (1 << modifier)
          end
```
There is no point in mutating `encoded_modifiers` since it is just a local variable. The return value from the block is what is significant and becomes the value of the `encoded_modifiers` for the next loop.

Issue #62, opened 2022-04-08
Evaluate if we should add Sorbet


vinistock:
Evaluated the advantages and disadvantages to add Sorbet to the `ruby-lsp`, so that we can decide on whether we want to do it or not.

If the decision is to add Sorbet, add it using our default tooling.

egiurleo:
Awesome, thank you!

Morriar:
Let's close this since we already decided to add it: https://github.com/Shopify/ruby-lsp/pull/119.

PR #119, opened 2022-05-31
Add Sorbet to the Ruby LSP

vinistock:
### Motivation

First step in #62

Add Sorbet with all sigils as typed false.

I'll follow this one with a PR to change the sigils and add more typing.

### Implementation

I tried ordering the commits, so that the changes are easier to review. But it's basically
- Add the gems
- Add configuration for rubocop, Tapioca
- Add typecheck to CI
- Generate RBIs and fix things


ruby-lsp.gemspec
```diff
@@ -21,5 +21,6 @@ Gem::Specification.new do |s|
 
   s.add_dependency("language_server-protocol")
   s.add_dependency("rubocop", ">= 1.0")
+  s.add_dependency("sorbet-static-and-runtime")
```
paracycle, line 24: I am not a fan of adding `sorbet-static-and-runtime` as a dependency since the gem does not really have a runtime dependency on sorbet static at all.

I'd rather that we declare this as:
```suggestion
  s.add_dependency("sorbet-runtime")
```
and add a dependency on `sorbet-static-and-runtime` in the `Gemfile` which would do the right dependabot thing that we want while still limiting our runtime dependency footprint.

vinistock: Fair point. I've changed the dependency and added missing requires for `sorbet-runtime`.

test/test_helper.rb
```diff
@@ -1,7 +1,9 @@
+# typed: false
 # frozen_string_literal: true
 
 $LOAD_PATH.unshift(File.expand_path("../lib", __dir__))
 
+require "sorbet-runtime"
 require "ruby-lsp"
 require "syntax_tree"
 require "ruby_lsp/cli"
```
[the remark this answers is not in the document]
vinistock: We wanted to not require anything on the default require, to avoid polluting loaded features. I created an internal file like we do in Tapioca do group all necessary requires.

Morriar (approved):

RyanBrushett (approved):
Thanks Vini

PR #122, opened 2022-06-01
Fix text synchronization for multibyte characters

vinistock:
### Motivation

Our previous approach to finding character positions in the source was counting bytes - not characters. This causes a bug with multibyte characters, where we apply the edits in the wrong index, messing up the entire document.

### Implementation

A simple solution would be to replace `scanner.pos` with `scanner.charpos`. The first one returns the byte position of the string scanner and the second returns the character position (which is what we want).

However, I played around a bit and benchmarked rolling out our own loop vs using the string scanner. On benchmark-ips, rolling out our own loop was 11x faster than using string scanner, so I switched the implementation.

We now split the source we're analyzing in characters and then keep track of the position ourselves.

### Automated Tests

Added a test for pushing edits and for auto-formatting as well.

### Manual Tests

1. Using this branch, open up a file
2. Add any multibyte character
3. Continue editing the file, like adding more method definitions
4. Verify that diagnostics don't show up unexpectedly
5. Verify that auto-formatting doesn't mess up the file with incorrect text

### DIFF
--- a/lib/ruby_lsp/document.rb
+++ b/lib/ruby_lsp/document.rb
@@ -1,7 +1,5 @@
 # frozen_string_literal: true
 
-require "strscan"
-
 module RubyLsp
   class Document
     attr_reader :tree, :source, :syntax_error_edits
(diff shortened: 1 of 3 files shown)


lib/ruby_lsp/document.rb
```diff
@@ -71,19 +69,19 @@ def apply_edit(source, range, text)
...
         @current_line = 0
+        @pos = 0
+        @chars = source.chars
       end
 
       def find_position(position)
-        # Move the string scanner counting line breaks until we reach the right line
         until @current_line == position[:line]
-          @scanner.scan_until(/\R/)
+          @pos += 1 until /\R/.match?(@chars[@pos])
```
paracycle, line 79: Do we really need `@chars`? Can't we do `@source[@pos]` here instead?

paracycle (approved):

Morriar (approved):

PR #127, opened 2022-06-06
Move codebase to typed strict

vinistock:
### Motivation

The final PR in our efforts to adopt Sorbet. This PR enforces typed strict in our files (except for test files, which are enforced as typed true).

### Implementation

Split by commit
1. Move the codebase to typed strict
2. Add generic for response type in RuboCopRequest
3. Add generic for response type in BaseRequest
4. Enhance some block related types

Let me know what you think about the generics. I don't love it, but they did improve typing and actually helped me find an incorrect signature in `Handler`.

### DIFF
--- a/lib/ruby_lsp/requests/base_request.rb
+++ b/lib/ruby_lsp/requests/base_request.rb
@@ -1,24 +1,31 @@
-# typed: true
+# typed: strict
 # frozen_string_literal: true
 
 module RubyLsp
   module Requests
     # :nodoc:
     class BaseRequest < SyntaxTree::Visitor
+      extend T::Sig
+      extend T::Helpers
+
+      abstract!
+
+      sig { overridable.params(document: Document).returns(T.untyped) }
       def self.run(document)
         new(document).run
       end
 
+      sig { params(document: Document).void }
       def initialize(document)
         @document = document
 
         super()
       end
 
-      def run
-        raise NotImplementedError, "#{self.class}#run must be implemented"
-      end
+      sig { abstract.returns(T.untyped) }
+      def run; end
 
+      sig { params(node: SyntaxTree::Node).returns(LanguageServer::Protocol::Interface::Range) }
       def range_from_syntax_tree_node(node)
         loc = node.location
 
--- a/lib/ruby_lsp/requests/base_request.rb
+++ b/lib/ruby_lsp/requests/base_request.rb
@@ -7,10 +7,13 @@ module RubyLsp
     class BaseRequest < SyntaxTree::Visitor
       extend T::Sig
       extend T::Helpers
+      extend T::Generic
+
+      Response = type_template { { upper: T.untyped } }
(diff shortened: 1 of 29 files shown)


lib/ruby_lsp/requests/base_request.rb
```diff
@@ -1,24 +1,34 @@
...
       def initialize(document)
         @document = document
 
         super()
       end
 
-      def run
-        raise NotImplementedError, "#{self.class}#run must be implemented"
-      end
+      sig { abstract.returns(T.untyped) }
```
Morriar, line 28: Isn't it always returning a `Response`?

```suggestion
      sig { abstract.returns(Response) }
```

vinistock: Yes, but Sorbet doesn't allow us to use a `type_template` in instance methods. Because the `run` method must be overridden by child classes, we get the correct typing without the need of adding a second generic as a `type_member`.

paracycle: I am not sure if `self.run` methods are adding any value to our code, to be honest. We instantiate and run these request classes in one location in our codebase and a few locations at most in our tests. Is it worth going through a `self.run` method that is hard to type?

vinistock: I agree, `self.run` is not providing any value. I removed those and removed the generics in the latest commit.

The reason `CodeActions` doesn't inherit from `BaseRequest` is because all it does is run `Diagnostics` and then filter the results.

paracycle: Could we not have the generic still but as a `type_member` now so that we can use it as the return type of `run`?

Morriar: > The reason CodeActions doesn't inherit from BaseRequest is because all it does is run Diagnostics and then filter the results.

Still, doesn't it implement the same interface than `BaseRequest`? What makes it incompatible with `BaseRequest`?

lib/ruby_lsp/requests/base_request.rb
```diff
@@ -1,24 +1,26 @@
...
```
Morriar, line 20: I think `Object` would work better here:

```suggestion
      sig { abstract.returns(Object) }
```

It is still type safe and even safer since Sorbet will yell if you try to call non-existing methods `Object` rather than be silent because of the `T.untyped`.

vinistock: Some requests return `T.any`, which is incompatible with `Object`. Sorbet throws a runtime error. I think we'll need to keep this one as untyped.

vinistock: Got it to work in the latest commit.

Morriar (approved):

wildmaples (approved):

PR #133, opened 2022-06-07
Add semantic token for name in a method definition

wildmaples:
### Motivation

Define a new semantic token for method names using the modifiers implemented in https://github.com/Shopify/ruby-lsp/pull/112

### Implementation

Similar to the previous semantic tokens, add a visit method for it. 

### Automated Tests

Most of the tests contain a method definition, which means an additional token is added to the array of expected tokens. 

### Manual Tests

- Check out this branch and use the tool `Inspect Editor Token and Scope`. 
- Click on a method name

Screen Shot 2022-06-07 at 2 01 55 PM

### DIFF
--- a/lib/ruby_lsp/requests/semantic_highlighting.rb
+++ b/lib/ruby_lsp/requests/semantic_highlighting.rb
@@ -52,6 +52,12 @@ module RubyLsp
         @encoder.encode(@tokens)
       end
 
+      def visit_def(node)
+        add_token(node.name.location, :method, [:declaration])
+        visit(node.params)
+        visit(node.bodystmt)
+      end
+
       def visit_m_assign(node)
         node.target.parts.each do |var_ref|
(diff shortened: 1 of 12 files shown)


lib/ruby_lsp/requests/semantic_highlighting.rb
```diff
@@ -52,6 +52,12 @@ def run
         @encoder.encode(@tokens)
       end
 
+      def visit_def(node)
```
vinistock, line 55: Can we do singleton and endless definition methods too?

wildmaples: https://github.com/Shopify/ruby-lsp/pull/134

vinistock: Thank you!

PR #138, opened 2022-06-09
Add semantic token for constants

wildmaples:
### Motivation

Inconsistent highlighting in constants: 

Screen Shot 2022-06-08 at 8 39 31 PM

### Implementation

Add new semantic token type, `namespace`, and add them as a token for constants. 

Irrelevant to the primary issue but I also rearranged some of the visit methods in alphabetical order.  

### Automated Tests

- Fix the existing tests
- Add a new test file `const.exp`
- Fix the integration test -- the code contains a constant, so the assertion needed to be updated

### Manual Tests

1. Checkout this branch
2. Open `test/fixtures/const.rb`
3. See consistent highlighting: 

Screen Shot 2022-06-08 at 8 44 15 PM


lib/ruby_lsp/requests/semantic_highlighting.rb
```diff
@@ -20,6 +20,7 @@ class SemanticHighlighting < BaseRequest
       TOKEN_TYPES = [
         :variable,
         :method,
+        :namespace,
```
vinistock, line 23: Since we can't differentiate between a module and class reference, should we use namespace or class?

wildmaples: I thought calling both module and classes a class token may be confusing, and so settled with namespace. Any opinions?

vinistock (approved):

test/integration_test.rb
```diff
@@ -78,7 +78,7 @@ def test_semantic_highlighting
     assert_telemetry("textDocument/didOpen")
 
     response = make_request("textDocument/semanticTokens/full", { textDocument: { uri: "file://#{__FILE__}" } })
-    assert_empty(response[:result][:data])
+    assert_equal([0, 6, 3, 2, 0], response[:result][:data])
```
Morriar, line 81: We should reconsider how we test this (in another PR) because this is surprisingly cryptic.

Morriar (approved):

PR #140, opened 2022-06-10
Move code actions to new framework

vinistock:
### Motivation

Continuing the work to migrate all our tests to the new test framework. This one is for code actions and includes a minor refactor.

### Implementation

Some tests have to re-map their results to make asserting things easier. I have added a `map_result` method that tests can override to re-map their expectations into the right format. This allowed us to simplify the diagnostics test a bit.

Other than that, I just added a new expectation runner for code actions and migrated our test to the new format.


test/expectations/expectations_test_runner.rb
```diff
@@ -115,4 +115,8 @@ def json_expectations(expected_json_string)
 
     JSON.parse(expected_json_string)["result"]
   end
+
+  def map_result(result)
```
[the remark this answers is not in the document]
vinistock: Agreed. I tried moving outside to a new method `initialize_params`. Let me know what you think.

test/requests/formatting_expectations_test.rb
```diff
@@ -22,4 +22,7 @@ def assert_expectations(source, expected)
     assert_empty(stdout)
     assert_equal(expected, result)
   end
+
+  def initialize_params(_expected)
```
Morriar, line 26: Why do we need to redefine this? Shouldn't the default implementation work (or no-op)?

RyanBrushett (approved):

test/requests/formatting_expectations_test.rb
vinistock: I'll explore this in a follow up PR.

PR #142, opened 2022-06-10
Add semantic token for keywords and keyword rest parameters

wildmaples:
### Motivation

Add semantic token for keyword and keyword rest to make it consistent between other params in the method definition 

### Implementation

Add a visitor method for `SyntaxTree::Params` 

### Automated Tests

Yes. I added a new test fixture for method parameters. 

### Manual Tests

A visual test (or you can use the Developer Inspect Editor Scope mode) 

Before:
Screen Shot 2022-06-10 at 2 41 01 PM

After: 
Screen Shot 2022-06-10 at 2 40 45 PM

### DIFF
--- a/lib/ruby_lsp/requests/semantic_highlighting.rb
+++ b/lib/ruby_lsp/requests/semantic_highlighting.rb
@@ -148,19 +148,7 @@ module RubyLsp
       def visit_params(node)
         node.keywords.each do |keyword,|
           location = keyword.location
-
-          # We exclude the ":" symbol from the keyword
-          # for consistency with the rest of the parameters
-          modified_location = SyntaxTree::Location.new(
-            start_line: location.start_line,
-            start_column: location.start_column,
-            start_char: location.start_char,
-            end_char: location.end_char - 1,
-            end_column: location.end_column - 1,
-            end_line: location.end_line,
-          )
-
-          add_token(modified_location, :variable)
+          add_token(location_without_colon(location), :variable)
         end
 
         add_token(node.keyword_rest.name.location, :variable) if node.keyword_rest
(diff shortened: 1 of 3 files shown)


vinistock:
I noticed that the `:` is a part of the keyword parameter location. Would we be able to not include the `:`?

vinistock:
For most other pieces of syntax involving symbols, the `:` is not highlighted as a part of it. Even if it's a part of the identifier, we can adjust it only for keyword parameter, since we know it's always going to be one character longer. What do you think?

wildmaples:
Screen Shot 2022-06-10 at 3 39 49 PM

@vinistock https://github.com/Shopify/ruby-lsp/pull/142/commits/12f7132928e9847b1a8b50cb0a6ee038efe69e5d

lib/ruby_lsp/requests/semantic_highlighting.rb
```diff
@@ -144,6 +144,28 @@ def visit_m_assign(node)
...
+          location = keyword.location
+
+          # We exclude the ":" symbol from the keyword
+          # for consistency with the rest of the parameters
+          modified_location = SyntaxTree::Location.new(
+            start_line: location.start_line,
+            start_column: location.start_column,
+            start_char: location.start_char,
+            end_char: location.end_char - 1,
+            end_column: location.end_column,
```
vinistock, line 159: For consistency sake, let's subtract one here too.

vinistock (approved):

lib/ruby_lsp/requests/semantic_highlighting.rb
```diff
@@ -144,6 +144,28 @@ def visit_m_assign(node)
...
+          # We exclude the ":" symbol from the keyword
+          # for consistency with the rest of the parameters
+          modified_location = SyntaxTree::Location.new(
+            start_line: location.start_line,
+            start_column: location.start_column,
+            start_char: location.start_char,
+            end_char: location.end_char - 1,
+            end_column: location.end_column - 1,
+            end_line: location.end_line,
+          )
```
Morriar, line 161: Should this be extracted as a private method?

The name of the extracted method would allow for self-documentation at the call-site, the current comment can be moved as part of the method documentation and the method could be reused for symbols as hash keys and for keyword arguments in call-sites if they have the same issue.

st0012 (approved):

lib/ruby_lsp/requests/semantic_highlighting.rb
wildmaples: https://github.com/Shopify/ruby-lsp/pull/142/commits/666abf163fbc6b6ee71c29566e75911272a0dd0b

kddnewton:
> @kddnewton out of curiosity, is it expected that the `:` is considered a part of the `Ident` for keyword parameters? If not, we can change this upstream and simplify the code.

Ahh sorry I didn't respond to this sooner! It _is_ expected just because that's what comes back from ripper.

PR #147, opened 2022-06-13
Update the doc -> docs in gitignore

st0012:
`rake yard` puts documents under `/docs` instead of `/doc`. So we should ignore `/docs` instead.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,7 +2,7 @@
 /**/.yardoc
 /**/_yardoc/
 /**/coverage/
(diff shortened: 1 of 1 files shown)


vinistock:
We changed the default output folder to `docs`, exactly to avoid it being gitignored for the [publish docs](https://github.com/Shopify/ruby-lsp/blob/main/.github/workflows/publish_docs.yml) action. Maybe we can use some environment variable that is only present in GitHub actions to change the folder?

Issue #146, opened 2022-06-13
Formatting sometimes hangs indefinitely


vinistock:
We have received reports of RuboCop formatting hanging indefinitely. Notice that it does not seem like the issue is performance on large files, which is a separate issue.

It seems that somehow the LSP either gets stuck or returns something invalid to VS Code (which in turn makes the "Running Ruby LSP formatter" dialogue never disappear).

Both reports we have received mentioned that there's nothing in the output tab, other than the default message of "Starting the Ruby LSP...".

### Suggestion

We should consider making [format on save](https://github.com/Shopify/vscode-shopify-ruby/blob/2fe3903426724bdb351f2e207805fc7af788b417/src/configuration.ts#L27) `false` in the extension pack, while we work on improving the situation.

### Suggestion on how to debug

The difficult part of this is reproducing the issue, given that we do not know what causes it. Some possible steps:
- Make sure `"ruby lsp.trace.server": "messages"` is turned on in `.vscode/settings.json`
- When formatting hangs, check the output tab and see if any other requests are still being made. If you manage to get requests going for any file, then it's unlikely that the server is stuck - which probably means we returned something invalid to VS Code (or didn't return anything when it was expecting something)

### Definition of done

This ticket is considered closed when the root cause of the bug (hanging formatting) is identified and addressed.

st0012:
> Make sure "ruby lsp.trace.server": "messages" is turned on in .vscode/settings.json

Is this 

```
    "ruby.lsp.trace.server": true,
```

or 

```
    "[ruby]": {
        "lsp.trace.server": true,
```

VSCode says the former is unknown and the later is not per-language setting 🤔

PR #148, opened 2022-06-13
Upgrade rubocop requirement to 1.30.0

mihyaeru21:
### Motivation
From version 0.0.4, formatting uses `--autocorrect` for RuboCop. This was added in RuboCop [1.30.0](https://github.com/rubocop/rubocop/releases/tag/v1.30.0) (https://github.com/rubocop/rubocop/pull/10547). Therefore, `textDocument/formatting` will not work with RuboCop versions less than 1.30.0.

### Implementation
Upgraded requirement version in gemspec.

### Automated Tests
I think it is not needed to add test. CI has been passed with RuboCop 1.30.1.

### Manual Tests
There is no change in behavior.

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -3,7 +3,7 @@ PATH
   specs:
     ruby-lsp (0.0.4)
       language_server-protocol
-      rubocop (>= 1.0)
+      rubocop (>= 1.30)
       sorbet-runtime
       syntax_tree (>= 2.3)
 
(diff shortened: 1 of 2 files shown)


st0012:
Hey @mihyaeru21 thanks for the PR, it helped us debug https://github.com/Shopify/ruby-lsp/issues/146 tremendously!
And after discussion, we decided to rollback the Rubocop flag to `--auto-correct` (PR: https://github.com/Shopify/ruby-lsp/pull/149). We're aware of the deprecation warning but we think it's a good tradeoff for supporting more users. Therefore, I'm closing this PR now. Thanks again 😄

PR #153, opened 2022-06-15
Remove the word "simple" from gem description

RyanBrushett:
### Motivation

It's not really that simple anymore. Maybe `s/simple/powerful` ? Or just keep it short and sweet?

### DIFF
--- a/ruby-lsp.gemspec
+++ b/ruby-lsp.gemspec
@@ -7,8 +7,8 @@ Gem::Specification.new do |s|
   s.email = ["ruby@shopify.com"]
   s.metadata["allowed_push_host"] = "https://rubygems.org"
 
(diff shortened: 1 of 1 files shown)


ruby-lsp.gemspec
```diff
@@ -7,8 +7,8 @@ Gem::Specification.new do |s|
   s.email = ["ruby@shopify.com"]
   s.metadata["allowed_push_host"] = "https://rubygems.org"
 
-  s.summary = "A simple language server for ruby"
-  s.description = "A simple language server for ruby"
+  s.summary = "A language server for ruby"
```
paracycle, line 10: How about:
```suggestion
  s.summary = "An opinionated language server for Ruby"
```

paracycle (approved):
Thanks!

PR #160, opened 2022-06-20
Return error response when exceptions are raised

st0012:
### Motivation

LSP [supports `error` attribute in response](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#responseMessage), which will print the error in the extension output **even if tracing is not enabled**. I think this will help use debug issues a lot. 

**Example**

```
[Error - 2:48:24 PM] Request textDocument/formatting failed.
  Message: #<OptionParser::InvalidOption: --auto-foo>
  Code: -32603 
{:jsonrpc=>"2.0", :id=>17, :method=>"textDocument/formatting", :params=>{:textDocument=>{:uri=>"file:///Users/hung-wulo/src/github.com/Shopify/ruby-lsp/lib/ruby_lsp/requests/formatting.rb"}, :options=>{:tabSize=>2, :insertSpaces=>true, :trimTrailingWhitespace=>true, :insertFinalNewline=>true}}}
```

### Implementation

When there's an exception inside `Handler#handle`, we write back the response with `error` attribute instead of `result`.

### Automated Tests

Based on the response to my question in https://github.com/Shopify/ruby-lsp/pull/149 (none), I think we're ok with not covering error handling with a new category of tests for now. 

### Manual Tests

1. Change the `--auto-correct` flag in `formatting.rb` to something like `--foo`
2. Restart LSP
3. Open a Ruby file, mess up indentations, and save
4. The error should show up in the `Ruby LSP` output and look like

```
[Error - 2:48:24 PM] Request textDocument/formatting failed.
  Message: #<OptionParser::InvalidOption: --auto-foo>
  Code: -32603 
{:jsonrpc=>"2.0", :id=>17, :method=>"textDocument/formatting", :params=>{:textDocument=>{:uri=>"file:///Users/hung-wulo/src/github.com/Shopify/ruby-lsp/lib/ruby_lsp/requests/formatting.rb"}, :options=>{:tabSize=>2, :insertSpaces=>true, :trimTrailingWhitespace=>true, :insertFinalNewline=>true}}}
```


lib/ruby_lsp/handler.rb
```diff
@@ -62,7 +62,20 @@ def handle(request)
...
 
-          @writer.write(id: request[:id], result: result) unless result == VOID
+          if error
+            @writer.write(
+              {
+                id: request[:id],
+                error: {
+                  code: LanguageServer::Protocol::Constant::ErrorCodes::INTERNAL_ERROR,
+                  message: error.inspect,
+                  data: request.to_s,
```
st0012, line 72: If we don't turn `request` into String here, it'll be shown as `[Object object]` in the output.

test/handler_test.rb
```diff
@@ -0,0 +1,52 @@
...
+    mocked_stdout = mock_stdout do
+      handler.send(:handle, request)
+    end
+
+    headers = mocked_stdout.gets("\r\n\r\n")
+    raw_response = mocked_stdout.read(T.must(headers)[/Content-Length: (\d+)/i, 1].to_i)
+    JSON.parse(T.must(raw_response), symbolize_names: true)
+  end
+
+  sig { params(block: T.proc.returns(T.untyped)).returns(Tempfile) }
```
st0012, line 35: @Morriar I want to declare the return type as `IO` but `Tempfile`'s superclass `DelegateClass(File)` instead of `File`. Is there a way to declare it as anything with the same interfaces as `IO`? Something like `interface` in Golang?

Morriar: Could you `T.any(IO, Tempfile)`?

st0012:
I changed the label from `enhancement` to `bug` because according to the LSP specification, the current behavior of returning `result` even when errored is incorrect:

```
/**
* The result of a request. This member is REQUIRED on success.
* This member MUST NOT exist if there was an error invoking the method.
*/
result?: string | number | boolean | object | null;
```

lib/ruby_lsp/handler.rb
```diff
@@ -62,7 +62,20 @@ def handle(request)
             error = e
           end
 
-          @writer.write(id: request[:id], result: result) unless result == VOID
+          if error
+            @writer.write(
+              {
+                id: request[:id],
+                error: {
+                  code: LanguageServer::Protocol::Constant::ErrorCodes::INTERNAL_ERROR,
```
vinistock, line 70: We have the `LanguageServer::Protocol::Constant` assigned to `Constant`, so we can make this a little shorter and maybe even fit the hash into a single line.

vinistock (approved):

lib/ruby_lsp/handler.rb
```diff
@@ -62,7 +62,14 @@ def handle(request)
             error = e
           end
 
-          @writer.write(id: request[:id], result: result) unless result == VOID
+          if error
+            @writer.write(
+              { id: request[:id],
+                error: { code: Constant::ErrorCodes::INTERNAL_ERROR, message: error.inspect, data: request.to_json }, }
+            )
```
Morriar, line 69: This style looks weird:
```suggestion
              {
                id: request[:id],
                error: { code: Constant::ErrorCodes::INTERNAL_ERROR, message: error.inspect, data: request.to_json },
              }
            )
```

test/handler_test.rb
```diff
@@ -0,0 +1,52 @@
...
+    mocked_stdout = mock_stdout do
+      handler.send(:handle, request)
+    end
+
+    headers = mocked_stdout.gets("\r\n\r\n")
+    raw_response = mocked_stdout.read(T.must(headers)[/Content-Length: (\d+)/i, 1].to_i)
+    JSON.parse(T.must(raw_response), symbolize_names: true)
+  end
+
+  sig { params(block: T.proc.returns(T.untyped)).returns(T.any(IO, Tempfile)) }
```
Morriar, line 35: Why do we need `returns(T.any(IO, Tempfile))` in the signature if we're always returning a `Tempfile.new`?
```

st0012: I didn't want to restrict the type to `Tempfile` because it's an implementation detail and shouldn't be depended on. I hope it can be just `IO` instead, as that already provides all the interfaces we'll need. 
However, because `Tempfile`'s super class is `DelegateClass(File)`, it's not recognized as `IO` even it has all the interfaces. 

But since it's just a helper for one test, it's not worth wrestling on it. I'll change it to `Tempfile`.

Morriar (approved):

PR #161, opened 2022-06-20
Add semantic tokens for heredoc identifiers

wildmaples:
~**Blocked 🛑 Tests are failing because this PR is contingent on the next release of syntax_tree that [implements `SyntaxTree::HeredocEnd`](https://github.com/ruby-syntax-tree/syntax_tree/pull/95)**~

PR is ready to be merged and the syntax_tree gem bump is part of the PR. 

### Motivation

Add semantic tokens for HEREDOCS 

Screen Shot 2022-06-20 at 2 51 07 PM

Screen Shot 2022-06-20 at 2 50 40 PM

### Implementation

Add semantic tokens to ensure there's no ambiguity between the delimiters and the string value

The token type I chose to add was :variable and the modifier, `:default_library`. The reasoning is because we're trying to keep it simple for now and not add any more new tokens than we need to. It also aligns with how other themes highlight their HEREDOCs, e.g. in Ruby API this is the colouring used: 

Screen Shot 2022-06-20 at 2 55 28 PM

### Automated Tests

I've added the appropriate tests for this case. 

### Manual Tests

Use this source code to test before and after applying this branch: 

```ruby
<<~HEREDOC
  some text
HEREDOC

foo = <<~HEREDOC
  some text
HEREDOC
```

### DIFF
--- a/lib/ruby_lsp/requests/semantic_highlighting.rb
+++ b/lib/ruby_lsp/requests/semantic_highlighting.rb
@@ -129,6 +129,16 @@ module RubyLsp
         visit(node.arguments)
       end
 
+      sig { params(node: SyntaxTree::HeredocBeg).void }
+      def visit_heredoc_beg(node)
+        add_token(node.location, :variable, [:default_library])
+      end
+
+      sig { params(node: SyntaxTree::HeredocEnd).void }
+      def visit_heredoc_end(node)
+        add_token(node.location, :variable, [:default_library])
+      end
+
       sig { params(node: SyntaxTree::Kw).void }
       def visit_kw(node)
         case node.value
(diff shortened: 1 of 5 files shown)


lib/ruby_lsp/requests/semantic_highlighting.rb
```diff
@@ -129,6 +129,16 @@ def visit_fcall(node)
         visit(node.arguments)
       end
 
+      sig { params(node: SyntaxTree::HeredocBeg).void }
+      def visit_heredoc_beg(node)
+        add_token(node.location, :variable, [:default_library])
```
Morriar, line 134: I wonder if `label` would make sense as the token type.

PR #171, opened 2022-06-27
Sanitize home directory from telemetry URI

vinistock:
### Motivation

We were sending the full URI for telemetry, which includes `file:///Users/myusername/....`. We don't really care (and it makes harder to analyze the data) about the beginning of the URI.

### Implementation

This PR just removes the `file://$HOME_DIR` part of the URI.

Before: `file:///Users/john/path/to/project/my_file.rb`
After: `~/path/to/project/my_file.rb`

### Automated Tests

Adapted the integration test.

### DIFF
--- a/lib/ruby_lsp/handler.rb
+++ b/lib/ruby_lsp/handler.rb
@@ -227,7 +227,7 @@ module RubyLsp
         params[:errorMessage] = error.message
       end
 
-      params[:uri] = uri if uri
+      params[:uri] = uri.sub(%r{.*://#{Dir.home}}, "~") if uri
       params
     end
   end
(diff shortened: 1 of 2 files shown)


lib/ruby_lsp/handler.rb
```diff
@@ -227,7 +227,7 @@ def telemetry_params(request, request_time, error)
         params[:errorMessage] = error.message
       end
 
-      params[:uri] = uri if uri
+      params[:uri] = uri.sub(%r{.*://#{Dir.home}}, "~") if uri
```
Morriar, line 230: What if I open the source of a gem located in `/opt/rubies/`? 

I wonder if we should restrict the metrics to files inside the project or something like this?

PR #173, opened 2022-06-28
Respect RuboCop file exclusion

### LINKED ISSUE #170: VSCode extension does not respect Rubocop ignored files
## Expected

When opening a file that's ignored in `.rubocop.yml` 

eg. 

```yml
AllCops:
  Exclude:
    - ignored.rb
```

It should hide the errors

## Actual

It runs Rubocop on that file, and displays any errors it finds.

DM me on Slack for an example repo to reproduce on :)

vinistock:
### Motivation

Closes #170

RuboCop doesn't respect the exclusion configuration if an excluded path is passed explicitly as an argument (which is what we do in diagnostics and formatting).

### Implementation

Pass the `--force-exclusion` flag to RuboCop, which makes it respect ignored paths even if requested explicitly.

### Manual Tests

2. Open an included file, such as `diagnostics.rb`
3. Make some edits
4. Verify diagnostics and auto-formatting work
5. Open an excluded file, such as any Ruby file under `test/fixtures`
6. Make some edits
7. Verify neither diagnostics or formatting are available

### DIFF
--- a/lib/ruby_lsp/requests/rubocop_request.rb
+++ b/lib/ruby_lsp/requests/rubocop_request.rb
@@ -15,6 +15,7 @@ module RubyLsp
 
       COMMON_RUBOCOP_FLAGS = T.let([
         "--stderr", # Print any output to stderr so that our stdout does not get polluted
+        "--force-exclusion",
         "--format",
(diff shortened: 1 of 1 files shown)


paracycle (approved):
Nice catch!

Morriar (approved):

PR #181, opened 2022-07-04
Include the last line in folding ranges

### LINKED ISSUE #154: Proposal to change how we do folding ranges
Currently Ruby LSP folds ranges by including the closing delimiter into the folded range. For example, the following Ruby code:
```ruby
def foo
  42
end

def bar
  "hello"
end
```
folds to:
```ruby
def foo ⋯

def bar ⋯
```

However, that introduces an ergonomic problem if I want to add a new line after the definition of the `foo` method, for example. If I place the cursor at the end of the `def foo⋯` line and press `Enter`, I start adding a line inside the `foo` method (above `42`). There is actually no way to add a line right after the `def foo` block without starting to type on the empty line between the `foo` and `bar` methods.

I propose that we start folding ranges by leaving the closing delimiters unfolded, so that the original code gets folded as:
```ruby
def foo ⋯
end

def bar ⋯
end
```
This way, if I want to add a new line inside the `foo` method, I put the cursor at the end of the `def foo` line and press `Enter`. If I want to add a new line after the `foo` method, I put the cursor at the end of the matching `end` of the `foo` method and press `Enter`. Thus, this folding style is more flexible for editing code in folded form.

adisonlampert:
### Motivation

Closes #154 
Now, folding ranges look like this:
```ruby
def foo ⋯
end

def bar ⋯
end
```

### Implementation
Updated the range of relevant foldable code (braced regions, functions, etc.)  to `end - 1` so the last line would be left out of the fold.

### Automated Tests

Updated the existing tests.


vinistock (approved):

lib/ruby_lsp/requests/folding_ranges.rb
```diff
@@ -77,7 +77,8 @@ def visit(node)
 
         case node
         when *SIMPLE_FOLDABLES
-          add_node_range(T.must(node))
+          location = T.must(node).location
+          add_lines_range(location.start_line, location.end_line - 1)
```
paracycle, line 81: Can we encapsulate this logic inside add_lines_range method. From what I can see, it always takes a start and and end line number and the end is always 1 less than the end of the location. We could instead always send a location object into the method and it can be responsible for subtracting 1 from the end line numbers. It would be better for encapsulation.

paracycle: Ah ok, makes sense.

Morriar (commented):
How does this behave with method rescues:

```rb
def foo
  # some code
rescue => e
  # more code
end
```

Are we folding this as:

```rb
def foo
rescue => e
  # more code
end
```

Or:

```rb
def foo
end
```

The behavior before this PR was to only fold the method body, not the rescue which in my experience was never what I wanter. I'm using the fold to hide the methods I am not interested in and it generally means I'll have to fold twice: the body, then the rescue(s).

What do you think?

paracycle:
> ...
> and you can also fold it on the `rescue` to get this:
> ```ruby
> def foo
>   # some code
> rescue => e ...
> ```

Hmmm, I would have expected this to be:

```ruby
  def foo
    # some code
  rescue => e ...
  end
```

Is that not the case? If not, the issue I reported in my original GH issue would still persist.

adisonlampert:
@paracycle Yeah, I noticed that too and realized it gets further messed up if you go from 
```ruby 
def foo
  # some code
rescue => e ...
```
and fold on the `def foo` to get:
```ruby 
def foo ...
```
I messaged Vini about it and I'll also look into that now!

adisonlampert:
Fixed the issue & added that example into the tests:
```ruby
def foo
  # some code
rescue => e
  # more code
end
```
folds on `rescue` to:
```ruby 
def foo
  # some code
rescue => e ...
end
```
which also resolves the function folding issue. 

Additionally to match the way `if` folds, `begin` now folds completely so:
```ruby
begin
  puts "begin"
rescue StandardError => e
  puts "stderror"
ensure
  puts "ensure"
end
```
folds to:
```ruby
begin ...
end

KaanOzkan (approved):

st0012 (approved):

RyanBrushett (approved):
👏

PR #184, opened 2022-07-05
Remove RuboCop dependency and fallback to SyntaxTree formatting

### LINKED ISSUE #76: Fallback to SyntaxTree formatting if RuboCop is not used by the application
If a project does not use RuboCop, we can still fallback and offer auto-formatting through SyntaxTree. The interface used to format the file was introduced [here](https://github.com/ruby-syntax-tree/syntax_tree/pull/52).

As a part of this work, investigate if the best solution is to remove `rubocop` from our dependencies, check if it's a part of the application's bundle and then decide which auto-formatting to use.

Alternatively, we can keep `rubocop` as a dependency and check for a `.rubocop.yml` file.

vinistock:
### Motivation

Closes #76 
Building on top of #183

If RuboCop is not present in the current application, we can fallback to formatting using SyntaxTree. Additionally, this allows us to drop the dependency on RuboCop and simply try to require it when available.

### Implementation

Both the diagnostics and formatting runners will now exit early without defining the classes if RuboCop is not present. If the runner classes don't exist, then we can exit early in diagnostics and fallback to SyntaxTree on formatting.

### Automated Tests

Honestly, I'm not sure about the best way to test this. If you have ideas on how to fake that RuboCop is not present in the bundle, please let me know.

### Manual Tests

It's a bit of work, but here are the steps if you're interested in testing it out.

**With RuboCop**
1. On the Ruby LSP project itself (which uses RuboCop), start the server
2. Edit some files
3. Verify that RuboCop diagnostics are showing up
4. Verify that formatting the file fixes the RuboCop violations

**Without RuboCop**
1. Create a minimal project structure (`Gemfile`, `dev.yml` and an `example.rb` file)
2. Configure `dev.yml` with the desired Ruby version and dev up
3. In the `Gemfile`, add the Ruby LSP pointing to this branch (e.g.: `gem "ruby-lsp", path: "../ruby-lsp"`)
4. Start the plugin
5. Edit `example.rb`
6. Verify no diagnostics show up
7. Verify that saving the file still formats (this time using SyntaxTree)

### DIFF
--- a/test/requests/formatting_test.rb
+++ b/test/requests/formatting_test.rb
@@ -26,7 +26,7 @@ class FormattingTest < Minitest::Test
   end
 
   def test_formats_with_syntax_tree_when_rubocop_is_not_present
-    with_disabled_rubocop do
+    with_uninstalled_rubocop do
       assert_equal(<<~RUBY, formatted_document)
         class Foo
           def foo
@@ -38,12 +38,11 @@ class FormattingTest < Minitest::Test
 
   private
 
-  def with_disabled_rubocop(&block)
+  def with_uninstalled_rubocop(&block)
     rubocop_paths = $LOAD_PATH.select { |path| path.include?("gems/rubocop") }
     rubocop_paths.each { |path| $LOAD_PATH.delete(path) }
     $LOADED_FEATURES.delete_if { |path| path.include?("ruby_lsp/requests") || path.include?("gems/rubocop") }
-    RubyLsp.send(:remove_const, :Requests)
-    Object.send(:remove_const, :RuboCop)
+    unload_constants
 
     block.call
   ensure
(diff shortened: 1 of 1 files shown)


paracycle (commented):
We can test this by modifying the `$LOAD_PATH` like so:
```ruby
$ irb
> require "rubocop"
true

$ irb
> $LOAD_PATH.delete_if {|path| path.start_with?(Gem.loaded_specs["rubocop"].full_gem_path) }
> require "rubocop"
LoadError: cannot load such file -- rubocop
```
(of course, we would need to recover the original `$LOAD_PATH` afterwards)

test/requests/formatting_test.rb
```diff
@@ -0,0 +1,54 @@
...
+
+      class Foo
+        def foo
+        end
+      end
+    RUBY
+  end
+
+  def test_formats_with_syntax_tree_when_rubocop_is_not_present
+    rubocop_paths = $LOAD_PATH.select { |path| path.include?("gems/rubocop") }
```
paracycle, line 29: Nice test, but could we encapsulate the bits around the actual test into something like:
```ruby
def with_disabled_rubocop(&blk)
  # Do the LOAD_PATH magic
  blk.call
ensure
  # Do cleanup
end
```
so that the test itself reads more meaningfully?

ruby-lsp.gemspec
```diff
@@ -20,7 +20,6 @@ Gem::Specification.new do |s|
   s.require_paths = ["lib"]
 
   s.add_dependency("language_server-protocol")
-  s.add_dependency("rubocop", ">= 1.0")
```
paracycle, line 23: 🎉

paracycle (approved):

PR #187, opened 2022-07-06
Extract document highlight target to allow more granular handling

### LINKED ISSUE #165: Document highlight does not highlight definitions
Because we're checking if the node type matches exactly, we are not highlighting some legitimate occurrences of symbols. For example,

```ruby
class Foo; end

Foo # <-- clicking here will not highlight the Foo in the class definition, but it should
```

The same thing happens for other types of definitions, like methods and modules.

vinistock:
### Motivation

Closes #165

In the previous structure, it was difficult to account for highlights that needed to happen between different nodes (e.g.: a method invocation and definition). This refactor aims at making it more flexible.

### Implementation

This refactor extracts a supporting class `HighlightTarget`, which knows how to extract the string value of each node of interest as well as matching the target with nodes that should be highlighted.

For example, if the target is a method invocation, then we want to highlight other invocations of the same method plus the definition.

After that extraction, the request class now only overrides the `visit` method and delegates to `HighlightTarget` to determine whether to highlight a piece of syntax depending on the target type.

### Automated Tests

Added a few tests that were missing.

### Manual Tests

Use the added fixture files to test.

2. Put the cursor on a method invocation
3. Verify that the invocation and definition are highlighted
4. Put the cursor on a class
5. Verify that the reference and definition are highlighted

### DIFF
--- a/lib/ruby_lsp/requests/document_highlight.rb
+++ b/lib/ruby_lsp/requests/document_highlight.rb
@@ -53,34 +53,26 @@ module RubyLsp
         @highlights
       end
 
-      sig { params(node: SyntaxTree::VarField).void }
-      def visit_var_field(node)
-        if matches_target?(node.value)
-          add_highlight(
-            node.value,
-            LanguageServer::Protocol::Constant::DocumentHighlightKind::WRITE
-          )
-        end
+      sig { params(node: T.nilable(SyntaxTree::Node)).void }
+      def visit(node)
+        return if node.nil?
 
-        super
-      end
-
-      sig { params(node: SyntaxTree::VarRef).void }
-      def visit_var_ref(node)
-        if matches_target?(node.value)
-          add_highlight(
-            node.value,
-            LanguageServer::Protocol::Constant::DocumentHighlightKind::READ
-          )
-        end
+        type = T.must(@target).highlight_type(node)
+        add_highlight(node, type) if type
 
         super
       end
 
       private
 
-      sig { params(node: SyntaxTree::Node, position: Integer).returns(T.nilable(VarNodes)) }
-      def find(node, position)
+      sig do
+        params(
+          node: SyntaxTree::Node,
+          position: Integer,
+          parent: T.nilable(SyntaxTree::Node)
+        ).returns(T.nilable(Support::HighlightTarget))
+      end
+      def find(node, position, parent = nil)
         matched =
           node.child_nodes.compact.bsearch do |child|
             if (child.location.start_char...child.location.end_char).cover?(position)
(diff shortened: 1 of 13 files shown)


lib/ruby_lsp/requests/support/highlight_target.rb
```diff
@@ -0,0 +1,85 @@
...
+              HighlightMatch.new(type: WRITE, node: other)
+            when SyntaxTree::VarRef
+              HighlightMatch.new(type: READ, node: other)
+            when SyntaxTree::ClassDeclaration, SyntaxTree::ModuleDeclaration
+              HighlightMatch.new(type: WRITE, node: other.constant)
+            when SyntaxTree::ConstPathRef
+              HighlightMatch.new(type: READ, node: other.constant)
+            when SyntaxTree::Params
+              params = other.child_nodes.compact
+              match = params.find { |param| value(param) == @value }
```
st0012, line 57: Is it possible to have both `@value` and `value(param)` being `nil`? And in that case, should we consider them as matched?

lib/ruby_lsp/requests/support/highlight_target.rb
```diff
@@ -0,0 +1,85 @@
...
+        end
+
+        private
+
+        # Match the target type (where the cursor is positioned) with the `other` type (the node we're currently
+        # visiting)
+        sig { params(other: SyntaxTree::Node).returns(T.nilable(HighlightMatch)) }
+        def matched_highlight(other)
+          case @node
+          when SyntaxTree::VCall, SyntaxTree::FCall, SyntaxTree::Call, SyntaxTree::Command,
```
st0012, line 36: Is it possible to categorize these node types in the `when` statement, Like moving them into a descriptive constant? It'll increase readability a lot.
I know that it may not be possible because of some edge cases though.

lib/ruby_lsp/requests/support/highlight_target.rb
vinistock: Good catch. I added an extra check for `nil`, since we don't want to highlight anything if we couldn't figure out its value.

lib/ruby_lsp/requests/document_highlight.rb
```diff
@@ -91,22 +72,24 @@ def find(node, position)
...
-          matched
+        when SyntaxTree::GVar,
+             SyntaxTree::IVar,
+             SyntaxTree::Const,
+             SyntaxTree::CVar,
+             SyntaxTree::VarField
+          Support::HighlightTarget.new(matched)
+        when SyntaxTree::Ident
+          relevant_node = node.is_a?(SyntaxTree::Params) ? matched : node
+          Support::HighlightTarget.new(relevant_node)
```
adisonlampert, line 83: Could you give an example of an ident? I don't understand why this separate case is necessary

adisonlampert (approved):

lib/ruby_lsp/requests/support/highlight_target.rb
```diff
@@ -0,0 +1,85 @@
...
+        READ = LanguageServer::Protocol::Constant::DocumentHighlightKind::READ
+        WRITE = LanguageServer::Protocol::Constant::DocumentHighlightKind::WRITE
+
+        class HighlightMatch < T::Struct
+          const :type, Integer
+          const :node, SyntaxTree::Node
+        end
+
+        sig { params(node: SyntaxTree::Node).void }
+        def initialize(node)
```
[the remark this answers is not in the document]
vinistock: Which `type`? Do you mean the one for the `HighlightMatch` class?

st0012 (approved):

PR #214, opened 2022-07-19
Do not use the Sorbet runtime in the default load path

vinistock:
### Motivation

If people forget to use `require: false` when adding the `ruby-lsp` to their `Gemfile`, then `lib/ruby-lsp.rb` will be required by Bundler.

We shouldn't use anything related to the Sorbet runtime in this file, because that would mean we have to require `sorbet-runtime` and possibly pollute the application in development mode.

This is currently breaking in apps that do not use `sorbet-runtime`, because `T.let` is not defined.

### Implementation

Ignored this file for the strict sigil cop and remove the `T.let`.

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -37,3 +37,4 @@ Sorbet/StrictSigil:
   Exclude:
     - "**/*.rake"
     - "test/**/*.rb"
+    - "lib/ruby-lsp.rb"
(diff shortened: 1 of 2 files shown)


paracycle (approved):

Morriar (commented):
I'm surprised we do not have to change any `require` anywhere to not pull `sorbet-runtime` in the mix?

vinistock:
Our default require (lib/ruby-lsp.rb) doesn't actually require anything (not even dependencies like sorbet-runtime). All of our requires are in `internal.rb`, which is why when bundler does the default require, it breaks saying that `T` is undefined.

Morriar (approved):

PR #216, opened 2022-07-22
Return nil if document is already formatted

vinistock:
### Motivation

If the documented is already formatted, we should not return text edits to VS Code as it may delay saving the file.

### Implementation

If the formatted text matches the source, then return early.

### Automated Tests

Added an example.

### DIFF
--- a/lib/ruby_lsp/requests/formatting.rb
+++ b/lib/ruby_lsp/requests/formatting.rb
@@ -33,7 +33,7 @@ module RubyLsp
         formatted_text = formatted_file
         return unless formatted_text
 
-        return if formatted_text == @document.source
+        return if formatted_text.size == @document.source.size && formatted_text == @document.source
 
         size = @document.source.size
 
--- a/lib/ruby_lsp/requests/formatting.rb
+++ b/lib/ruby_lsp/requests/formatting.rb
@@ -33,9 +33,8 @@ module RubyLsp
         formatted_text = formatted_file
         return unless formatted_text
 
-        return if formatted_text.size == @document.source.size && formatted_text == @document.source
-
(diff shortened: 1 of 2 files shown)


lib/ruby_lsp/requests/formatting.rb
```diff
@@ -33,6 +33,8 @@ def run
         formatted_text = formatted_file
         return unless formatted_text
 
+        return if formatted_text == @document.source
```
paracycle, line 36: A slight optimization here for large documents might be to check the size of each of these strings before comparing them byte-for-byte:
```suggestion
        return if formatted_text.size == @document.source.size && formatted_text == @document.source
```
This should speed up the cases where the formatted text is different from the original by catching the difference in the sizes as early as possible.

paracycle (approved):

lib/ruby_lsp/requests/formatting.rb
vinistock: Very nice!

lib/ruby_lsp/requests/formatting.rb
```diff
@@ -33,6 +33,8 @@ def run
         formatted_text = formatted_file
         return unless formatted_text
 
+        return if formatted_text.size == @document.source.size && formatted_text == @document.source
```
Morriar, line 36: Shouldn't this be using `size` declared just bellow?

vinistock: I changed to use the `size` variable, I totally had missed it.

Let's wait to move it into `Document` if the need to re-use this arises. I feel like this won't be needed elsewhere.

st0012 (approved):

PR #221, opened 2022-07-25
Centralize request handling logic in server.rb

st0012:
### Motivation

Currently, the steps to add a new request are:

1. Define new provider in `handler.rb`'s `respond_with_capabilities` method
1. Define the related `respond_with_*` method in `handler.rb`
1. Define the supported request type (which usually just calls the above method) in `cli.rb`
1. Add a new request class and call it in the `respond_with_*` method in `handler.rb`

IMO the jumping between files is confusing. And I think in group pairings we also mention that these steps can be shorten if we don't define the `respond_with_*` methods.

After this PR, the flow will become

1. Define new provider in `server.rb`'s `initialize` request
1. Define the supported request type in `server.rb`
1. Add a new request class and call it in the above block `server.rb`

It has less steps and the contributor can focus on only `server.rb`.

### Implementation

- Move all `respond_with_*` methods into `Handler#on` blocks.
- Move the request definitions to `server.rb` and remove `cli.rb`.

### Automated Tests

Don't need any new tests but all old tests should pass.

### Manual Tests

1. Restart RubyLSP
2. Check there's no error in ruby-lsp's extension output
3. Check features like document highlight works as expected


vinistock:
I like this refactor, but can we also please get rid of the CLI class altogether? It doesn't provide any value at all. We could move the entire logic inside the `self.start` method to the top level of the file and rename the file to `routes.rb`. What do you think?

st0012:
@vinistock That's even better 👍 But I prefer `server.rb` because it does a bit more (e.g. caching) than routing. Wdyt?

st0012:
@vinistock I got rid of the intermediate file completely (so not even `server.rb`). Let me know if that's too aggressive?

vinistock:
@st0012 I think I'd rather still have a `server.rb` file, to serve as the central point for configuring requests.

lib/ruby_lsp/server.rb
```diff
@@ -0,0 +1,157 @@
+# typed: strict
+# frozen_string_literal: true
+
+require_relative "internal"
```
vinistock, line 4: Can we use a regular require here? I like to avoid relative requires as much as possible.

lib/ruby_lsp/server.rb
```diff
@@ -0,0 +1,157 @@
...
+
+module RubyLsp
```
vinistock, line 6: We should be able to remove this module.

lib/ruby_lsp/server.rb
st0012: I'm fine with both but why did we had `require_relative "handler"` before?

lib/ruby_lsp/server.rb
st0012: If we remove the module, most of the constants below will need to have `RubyLsp` namespace, which I think doesn't help the readability.

vinistock (approved):

PR #222, opened 2022-07-25
Prevent breaking on nameless keyword rest parameters

vinistock:
### Motivation

Keyword rest parameters can be nameless, which means trying to access `keyword_rest.name.location` may try to invoke a method on `nil` and break.

### Implementation

Improve the guard clause and added a new test fixture to make sure it does not break requests.

### DIFF
--- a/lib/ruby_lsp/requests/semantic_highlighting.rb
+++ b/lib/ruby_lsp/requests/semantic_highlighting.rb
@@ -179,7 +179,8 @@ module RubyLsp
           add_token(location_without_colon(location), :variable)
         end
 
-        add_token(node.keyword_rest.name.location, :variable) if node.keyword_rest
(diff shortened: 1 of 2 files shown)


lib/ruby_lsp/requests/semantic_highlighting.rb
```diff
@@ -179,7 +179,8 @@ def visit_params(node)
           add_token(location_without_colon(location), :variable)
         end
 
-        add_token(node.keyword_rest.name.location, :variable) if node.keyword_rest
+        name = node.keyword_rest&.name
+        add_token(node.keyword_rest.name.location, :variable) if name
```
st0012, line 183: Could be just `name.location`?

st0012 (approved):

PR #223, opened 2022-07-25
Fix FCall within chained invocation in folding ranges

vinistock:
### Motivation

When processing chained invocations, if the first receiver is an `FCall`, we keep trying to invoke `receiver` recursively, but this method doesn't exist on that class (since `FCall`s don't have receivers and are invoked directly on `self`).

### Implementation

If we found an `FCall`, it means we found the final receiver and can stop looking.

### Automated Tests

Added a new fixture and expectation.


lib/ruby_lsp/requests/folding_ranges.rb
```diff
@@ -195,7 +195,8 @@ def add_call_range(node)
             receiver = receiver.receiver
           when SyntaxTree::MethodAddBlock
             visit(receiver.block)
-            receiver = receiver.call.receiver
+            receiver = receiver.call
+            receiver = receiver.receiver unless receiver.is_a?(SyntaxTree::FCall)
```
st0012, line 199: Will `if receiver.respond_to?(:receiver)` be more flexible?

Morriar: I don't think it will type check properly tho

Morriar: I wonder if we do the reverse and list explicitly the cases where we're going to look for `receiver`?

Something like:

```rb
case receiver
when CallKind1, CallKind2
  receiver.receiver
end
```

To be more robust if `syntax-tree` add new node kinds.

vinistock: I like the idea of an allow list approach. Using this made me realize that `Command` is also a possible node, but doesn't implement `receiver` because it's invoked on `self`.

I have started to check for the nodes we care about specifically.

lib/ruby_lsp/requests/folding_ranges.rb
```diff
@@ -195,7 +195,11 @@ def add_call_range(node)
             receiver = receiver.receiver
           when SyntaxTree::MethodAddBlock
             visit(receiver.block)
-            receiver = receiver.call.receiver
+            receiver = receiver.call
+
+            if receiver.is_a?(SyntaxTree::Call) || receiver.is_a?(SyntaxTree::CommandCall)
```
st0012, line 200: I really hope we can use pattern matching here (doesn't work with Ruby 2.7):

```rb
if receiver in SyntaxTree::Call | SyntaxTree::CommandCall
  receiver = receiver.receiver
end

```

st0012 (approved):

PR #224, opened 2022-07-25
Recover from files with initial syntax errors

vinistock:
### Motivation

We are getting some errors on user's machines of the `document` instance being `nil`. The only possibility for this, is if we weren't able to parse the first version of the file due to syntax errors and never recovered.

### Implementation

This PR attempts to make this process more robust by always inserting the document into the store regardless of syntax errors. If a syntax error occurs, the `tree` will be `nil`, which we have to handle in requests.

When edits are made to the file with syntax errors, `push_edits` will take care of re-parsing.

### Automated Tests

Added a couple of examples and changed some existing ones.

### Manual Tests

If you wanna try this manually
2. Created a new file and save it with a syntax error (e.g.: `def Foo`)
3. Close the file to clear its state from the store
4. Re-open the file (this re-open will fail to parse)
5. Fix the syntax error by adding `end`
6. Verify that the LSP recovered and features are now available


lib/ruby_lsp/document.rb
```diff
@@ -32,6 +32,13 @@ def ==(other)
       @source == other.source
```
st0012, line 32: Why don't we call `parse` here instead? I think we'd always want to parse it?

lib/ruby_lsp/document.rb
```diff
@@ -32,6 +32,13 @@ def ==(other)
...
     end
 
+    sig { void }
+    def parse
+      @tree = SyntaxTree.parse(@source)
+    rescue SyntaxTree::Parser::ParseError
+      # Do not raise if we failed to parse
```
Morriar, line 39: Should we log something in the console in this case?

Morriar (commented):
I wonder if we should let the exception bubble up to to `handle` and catch it there rather than  making all those types nilable?

lib/ruby_lsp/document.rb
vinistock: I think it's valuable, but I'll suggest doing this in a follow up PR. If we're going to log something, we should include the URI, which is currently not passed to `Document`. I can create a follow up as soon as this one goes in.

Morriar:
> So we need to save the initial document, regardless of whether there are errors or not.

Ah I see. Ok then 👍

lib/ruby_lsp/document.rb
st0012: Sorry I meant parse in `Document#initialize` like we currently do. This PR does 2 things:

1. Extract and postpone document parsing
2. Rescue the parsing error

And I'm not sure why we need to do 1 because it adds another operation we need to keep in mind when dealing with documents.

vinistock: Oh, I see what you mean. Indeed, we don't need the `parse` method. I've moved rescuing inside `initialize` and simplified these changes.

st0012 (commented):
~~I think this may not be the best solution for the problem. Currently, successful document parsing and document processing logic (including caching) go hand-in-hand. And that's why it's safe for the latter because it saves us the need to handle `nil` in a lot of places.~~

~~If we want to make parsing part failable, we shouldn't compromise the safety we have on document processing. Instead, we should probably consider handling requests in stages, like:~~

~~1. Document parsing + storing~~
~~2. Parsing result check + early return~~
~~3. Document processing~~

~~It'll likely require structural change on the `Handler` class but I think it'd be better.~~

I gave the above a try and it doesn't work because requests like `didOpen` do need to handle the text -> document conversion. So if we separate a request into stages, it'll just make those requests hard to understand.


Another concern is that VSCode will hang if it doesn't receive a request's response (see https://github.com/Shopify/ruby-lsp/pull/149). So we may need to pay more attention on the early return's implementation and make sure it still returns something?

lib/ruby_lsp/store.rb
```diff
@@ -14,20 +14,19 @@ def initialize
       @state = T.let({}, T::Hash[String, Document])
     end
 
-    sig { params(uri: String).returns(Document) }
+    sig { params(uri: String).returns(T.nilable(Document)) }
```
st0012, line 17: Why is this `nilable` now? It looks like `set` will still always store a document?

lib/ruby_lsp/store.rb
```diff
@@ -56,10 +55,13 @@ def delete(uri)
...
           block: T.proc.params(document: Document).returns(T.type_parameter(:T))
-        ).returns(T.type_parameter(:T))
+        ).returns(T.nilable(T.type_parameter(:T)))
     end
     def cache_fetch(uri, request_name, &block)
-      get(uri).cache_fetch(request_name, &block)
+      document = get(uri)
+      return if document.nil? || document.tree.nil?
+
+      document.cache_fetch(request_name, &block)
```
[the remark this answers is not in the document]
st0012: I gave it a try in another branch. Here's the [comparison](https://github.com/Shopify/ruby-lsp/compare/224-alternative?expand=1)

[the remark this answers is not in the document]
vinistock: I'm not sure I love having to return early on every request rather than having `cache_fetch` handle this. This method now only returns `nil` if the document is not parsable.

Every other return is the result - either the cached one or the one we get by executing the block.

vinistock:
@st0012 the void vs nil returns have been addressed in #100. Returning `nil` for all of our current requests are valid responses and VS Code handles them properly.

PR #230, opened 2022-07-28
Alert users about non auto-correctable diagnostics

### LINKED ISSUE #215: Alert users about non auto-correctable diagnostics
Some developers are left wondering why some violations aren't automatically fixed on formatting. Sometimes, it's just because those violations are not auto-correctable (or because their auto-correct is considered unsafe).

We should include some text in our diagnostics to call attention to that, so that devs know it's expected.

adisonlampert:
### Motivation

Closes #215 

### Implementation
 
If the offense is not correctable, 
- changes severity so that the squiggle underneath is red and 
- includes `"This offense is not auto-correctable."` at the top of the offense message.

### Automated Tests

Updated existing unit tests. All tests passing.

### Manual Tests

1. Open the Ruby LSP repo on this branch
2. Create a new file in `lib/ruby_lsp`
3. Create a non auto-correctable issue (e.g., `RubyLsp.constants`)
4. Check that the offense appears red and the error message includes `"This offense is not auto-correctable."`

### DIFF
--- a/lib/ruby_lsp/requests/support/rubocop_diagnostic.rb
+++ b/lib/ruby_lsp/requests/support/rubocop_diagnostic.rb
@@ -61,11 +61,18 @@ module RubyLsp
 
         sig { returns(LanguageServer::Protocol::Interface::Diagnostic) }
         def to_lsp_diagnostic
+          if @offense.correctable?
+            severity = RUBOCOP_TO_LSP_SEVERITY[@offense.severity.name]
+            message = @offense.message
+          else
+            severity = LanguageServer::Protocol::Constant::DiagnosticSeverity::WARNING
+            message = "#{@offense.message}\n\nThis offense is not auto-correctable.\n"
+          end
           LanguageServer::Protocol::Interface::Diagnostic.new(
-            message: @offense.message,
+            message: message,
             source: "RuboCop",
             code: @offense.cop_name,
-            severity: RUBOCOP_TO_LSP_SEVERITY[@offense.severity.name],
+            severity: severity,
             range: LanguageServer::Protocol::Interface::Range.new(
               start: LanguageServer::Protocol::Interface::Position.new(
                 line: @offense.line - 1,
--- a/test/expectations/diagnostics/if_inside_else.exp
+++ b/test/expectations/diagnostics/if_inside_else.exp
@@ -11,10 +11,10 @@
           "character": 14
         }
       },
-      "severity": 3,
+      "severity": 2,
       "code": "Sorbet/TrueSigil",
       "source": "RuboCop",
-      "message": "Sorbet/TrueSigil: Sorbet sigil should be at least `true` got `false`."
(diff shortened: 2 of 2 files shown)


lib/ruby_lsp/requests/support/rubocop_diagnostic.rb
```diff
@@ -61,11 +61,17 @@ def to_lsp_code_action
 
         sig { returns(LanguageServer::Protocol::Interface::Diagnostic) }
         def to_lsp_diagnostic
+          severity, message = if @offense.correctable?
+            [RUBOCOP_TO_LSP_SEVERITY[@offense.severity.name], @offense.message]
+          else
+            [LanguageServer::Protocol::Constant::DiagnosticSeverity::ERROR,
+             "This offense is not auto-correctable.\n#{@offense.message}",]
```
vinistock, line 68: Can we invert this an put the message first with two line breaks?
```suggestion
             "#{@offense.message}\n\nThis offense is not auto-correctable."]
```

lib/ruby_lsp/requests/support/rubocop_diagnostic.rb
```diff
@@ -61,11 +61,17 @@ def to_lsp_code_action
 
         sig { returns(LanguageServer::Protocol::Interface::Diagnostic) }
         def to_lsp_diagnostic
+          severity, message = if @offense.correctable?
+            [RUBOCOP_TO_LSP_SEVERITY[@offense.severity.name], @offense.message]
+          else
+            [LanguageServer::Protocol::Constant::DiagnosticSeverity::ERROR,
```
vinistock, line 67: Since this is just a linting violation that can't be auto-corrected, let's use warning instead of error. Errors are typically for things that will really break your program, like a syntax error.

```suggestion
            [LanguageServer::Protocol::Constant::DiagnosticSeverity::WARNING,
```

lib/ruby_lsp/requests/support/rubocop_diagnostic.rb
```diff
@@ -61,11 +61,17 @@ def to_lsp_code_action
 
         sig { returns(LanguageServer::Protocol::Interface::Diagnostic) }
         def to_lsp_diagnostic
+          severity, message = if @offense.correctable?
```
vinistock, line 64: I'm not a huge fan of the multi assignment here. What do you think about doing
```ruby
if @offense.correctable?
  severity = ...
  message = ...
else
  severity = ...
  message = ...
end
```

lib/ruby_lsp/requests/support/rubocop_diagnostic.rb
vinistock: Yeah, that's what I mean with the two line breaks. It's easier to read when the violation message is split from this new note in my opinion.

vinistock (approved):

lib/ruby_lsp/requests/support/rubocop_diagnostic.rb
adisonlampert: Thanks! I fixed it by adding a new line to the end of the message :)

PR #231, opened 2022-07-29
Include Kernel.methods as part of special Ruby methods

vinistock:
### Motivation

The method `extend` and others are actually present in `Kernel.methods` instead of `Kernel.instance_methods`. This means that we were accidentally pushing semantic tokens for `extend` when we didn't mean to.

Our test expectation was actually asserting the wrong behaviour and that's why the test passed.

### Implementation

Just added `Kernel.methods(false)` to the list and split it into multiple lines.

### Automated Tests

Fixed the expectation for special Ruby methods.

### DIFF
--- a/lib/ruby_lsp/requests/semantic_highlighting.rb
+++ b/lib/ruby_lsp/requests/semantic_highlighting.rb
@@ -60,10 +60,13 @@ module RubyLsp
         default_library: 9,
       }.freeze, T::Hash[Symbol, Integer])
 
-      SPECIAL_RUBY_METHODS = T.let((Module.instance_methods(false) +
-        Kernel.methods(false) + Bundler::Dsl.instance_methods(false) +
-        Module.private_instance_methods(false))
-        .map(&:to_s), T::Array[String])
+      SPECIAL_RUBY_METHODS = T.let((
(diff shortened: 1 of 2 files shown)


lib/ruby_lsp/requests/semantic_highlighting.rb
```diff
@@ -60,10 +60,13 @@ class SemanticHighlighting < BaseRequest
         default_library: 9,
       }.freeze, T::Hash[Symbol, Integer])
 
-      SPECIAL_RUBY_METHODS = T.let((Module.instance_methods(false) +
-        Kernel.methods(false) + Bundler::Dsl.instance_methods(false) +
```
KaanOzkan, line 64: `Kernel.methods(false)` was already on the list, did you mean `Kernel.instance_methods(false)` in the PR description?

paracycle (approved):

lib/ruby_lsp/requests/semantic_highlighting.rb
vinistock: Yes, that was a mistake on my part. I meant `Kernel.instance_methods(false)`.

PR #233, opened 2022-07-29
Better specify gemspec files

vinistock:
### Motivation

The previous implementation for listing files was based on a deny list approach (listing everything and then rejecting). This made us accidentally distribute the gem with the `sorbet` and `bin` folder and some other irrelevant files for the packaged gem.

### Implementation

Use an allow list approach and only specify `exe`, `lib` plus important files.

### DIFF
--- a/ruby-lsp.gemspec
+++ b/ruby-lsp.gemspec
@@ -12,11 +12,9 @@ Gem::Specification.new do |s|
   s.homepage = "https://github.com/Shopify/ruby-lsp"
   s.license = "MIT"
 
-  s.files = Dir.chdir(File.expand_path(__dir__)) do
-    %x(git ls-files -z).split("\x0").reject { |f| f.match(%r{\A(?:test|spec|features|misc)/}) }
-  end
+  s.files = Dir.glob("lib/**/*.rb") + ["README.md", "CHANGELOG.md", "VERSION", "LICENSE.txt"]
   s.bindir = "exe"
-  s.executables = s.files.grep(/\Aexe/) { |f| File.basename(f) }
+  s.executables = ["ruby-lsp"]
   s.require_paths = ["lib"]
 
   s.add_dependency("language_server-protocol")


ruby-lsp.gemspec
```diff
@@ -12,9 +12,7 @@ Gem::Specification.new do |s|
   s.homepage = "https://github.com/Shopify/ruby-lsp"
   s.license = "MIT"
 
-  s.files = Dir.chdir(File.expand_path(__dir__)) do
-    %x(git ls-files -z).split("\x0").reject { |f| f.match(%r{\A(?:test|spec|features|misc)/}) }
-  end
+  s.files = Dir.glob("{lib,exe}/**/*") + ["README.md", "CHANGELOG.md", "Gemfile", "VERSION", "LICENSE.txt"]
```
paracycle, line 15: I don't think we should bundle the `Gemfile`, it is not used in deployed gems at all and will be just dead code. I see now that I've done the same mistake of bundling it in Tapioca, which I will fix.

ruby-lsp.gemspec
```diff
@@ -12,9 +12,7 @@ Gem::Specification.new do |s|
...
   s.bindir = "exe"
   s.executables = s.files.grep(/\Aexe/) { |f| File.basename(f) }
```
Morriar, line 17: I think you could be passing just `ruby-lsp` here.

ruby-lsp.gemspec
Morriar, line 15: I don't think you need to pass the exe directory as it's already the binddir:

```suggestion
  s.files = Dir.glob("lib/**/*") + ["README.md", "CHANGELOG.md", "Gemfile", "VERSION", "LICENSE.txt"]
```

paracycle: Also, do we want to include all kinds of files in `lib` or just `.rb` files?

vinistock: I thought we needed to include the `exe` folder there (where the files have no extension). I added the `.rb` to the end of the pattern now (although we only have `rb` files in `lib` anyway).

rafaelfranca (approved):

PR #236, opened 2022-08-02
Fix call processing when message is a :call symbol literal

vinistock:
### Motivation

The `Call#message` method can return the symbol literal `:call` when using this Ruby syntax:

```ruby
foo.(bar)
```

We weren't properly processing this and the code would throw an exception.

### Implementation

Skipped doing anything when the message is `:call`, since that's not the part of the syntax we care about. If you check the test expectations, `foo` and `bar` are still properly highlighted.

### Automated Tests

Added two new fixtures and one new expectation file.


lib/ruby_lsp/requests/support/highlight_target.rb
```diff
@@ -76,7 +76,8 @@ def value(node)
           when SyntaxTree::VarField, SyntaxTree::VarRef, SyntaxTree::VCall, SyntaxTree::FCall
             node.value&.value
           when SyntaxTree::Call, SyntaxTree::Command, SyntaxTree::CommandCall
-            node.message.value
+            message = node.message
+            node.message != :call ? message.value : nil
```
st0012, line 80:
```suggestion
            message != :call ? message.value : nil
```

st0012 (approved):
Can use a small change but looks good 👍

PR #237, opened 2022-08-03
Resolve bundler and version paths

vinistock:
### Motivation

In https://github.com/Shopify/tapioca/pull/1094, we implement a generic URI handler for source location comments. We need to start handling it in a similar fashion on the side of the Ruby LSP to find the right file paths and produce the jump to gem source links.

### Implementation

- Started trying to match the URIs inside the existing gem specifications for the project

Also, re-generated the RBIs with the changes in Tapioca to make it easier to test.

### Automated Tests

Added another expectation and changed existing tests.


dirceu (approved):

lib/ruby_lsp/requests/support/source_uri.rb
```diff
@@ -0,0 +1,82 @@
+# typed: strict
+# frozen_string_literal: true
+
+require "uri/file"
+
+module URI
+  # Must be kept in sync with the one in Tapioca
```
KaanOzkan, line 7: Why does it need to be kept in sync? Why can't we omit the methods that we don't use in `ruby-lsp`?

lib/ruby_lsp/requests/document_link.rb
```diff
@@ -1,30 +1,79 @@
...
+
+            Gem::Specification.default_stubs.each do |stub|
+              spec = stub.to_spec
+              lookup[spec.name] = {}
+              lookup[spec.name][spec.version.to_s] = {}
+
+              spec.files.each do |file|
+                prefix_matchers = [//]
+                prefix_matchers.concat(spec.require_paths.map { |rp| Regexp.new("^#{rp}/") })
+                prefix_matcher = Regexp.union(prefix_matchers)
```
st0012, line 57: Since these 3 lines don't need access to `file`, maybe they can be moved out of the loop?

wildmaples (approved):

PR #243, opened 2022-08-10
Add request cancellation

### LINKED ISSUE #238: Parallelism support
For very large files, requests start taking a considerable amount of time, sometimes reaching 5+ seconds. However, our metrics show that processing requests is never that slow.

I believe that, because requests take a little longer on large files, they start piling up and create a backlog while users are rapidly editing. This means we waste time serving old requests while trying to catch up to the latest edit on the file. This would explain why individual request time metrics aren't so bad, but the perceived performance on large files is.

For this issue, explore the possibility of adding parallelism to the LSP, so that we can support [request cancellation](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#cancelRequest).

This feature can only work in a multi process/threaded environment, where the main process orchestrates requests (and possibly executes critical requests that need to run before others, like `textDocument/didChange`) and workers serve other requests that can be processes later (like document symbols, folding ranges, etc).

In order to support cancellation, we would need to
- delegate requests to workers
- easily configure which requests run in parallel (on workers) and which ones must be sequential
- stop a worker in the middle and make it throw away previous progress when a cancellation request is received. The worker can't be paused, it must completely drop what it's doing (which would be a stale request) and receive a new request to handle
- if we can't stop workers in the middle, does removing requests from the queue deliver a performance improvement that justifies the effort?

### Definition of done

The definition of done is not the implementation itself, but an investigation outlining the best path forward. Other questions to answer other than whether it works or not:
- Does this work with processes or only Ractors?
- If only Ractors could support something like this, can we fallback to sequential processing for Ruby < 3?

Note: the addition of workers benefits #199 since there is some intersection.

vinistock:
### Motivation

Closes #238. More context on the issue.

Process requests in a `Thread` worker to allow the main thread to manage a queue of requests.

### Implementation

The first commit adds the concurrency and queue mechanism. The second commit refactors the code to separate running requests from writing the result to the IO. We can only cancel while processing requests, but writing to the IO must always happen.

This PR
1. Allows us to configure which requests can run in parallel and which ones must run in the main thread
2. Adds a `Queue` to process requests in a worker
3. Adds the capability to cancel requests

### Automated Tests

Added tests for the queue. After the refactors, it is not longer possible to test the example that existed in `handler_test`, so I removed that file.

### Manual Tests

There is not a lot of specific things to check. Basically just a smoke test trying things out and making sure nothing is breaking.
2. Play around with editing files, using folding ranges, document symbols, etc
3. Verify that requests continue to work as expected


lib/ruby_lsp/queue.rb
```diff
@@ -0,0 +1,186 @@
+# typed: strict
+# frozen_string_literal: true
+
+require "benchmark"
```
chrisseaton, line 4: What's this used for?

lib/ruby_lsp/queue.rb
```diff
@@ -0,0 +1,186 @@
...
+          @mutex.synchronize do
+            @jobs.delete(request[:id])
+            @current_job = job
+          end
+
+          begin
+            next if job.cancelled
+
+            result = execute(request)
+            finalize_request(result, request)
```
chrisseaton, line 150: Should the finalization happen outside of the cancel handler? It's not cancellable is it?

test/queue_test.rb
```diff
@@ -0,0 +1,95 @@
...
+
+      assert_equal(:started, @running_job_response.pop)
+      @queue.cancel("job_to_cancel")
+      @running_job_signal.push(:unblock)
+
+      @queue.shutdown
+
+      assert_empty(@running_job_response)
+    end
+  end
```
chrisseaton, line 77: Can we write a test for cancel during finalization - it shouldn't have an effect.

lib/ruby_lsp/queue.rb
vinistock: We benchmark requests for telemetry, so that we can measure the server's performance.

test/queue_test.rb
vinistock: How would that test differ from `test_after_processing`? Is there any way to guarantee that it's in the middle of finalizing?

lib/ruby_lsp/queue.rb
chrisseaton: Right - it's just timing a block. I guess it seemed weird to me to see the word 'benchmark' in production code, but it's fine.

lib/ruby_lsp/queue.rb
```diff
@@ -0,0 +1,189 @@
...
+    private
+
+    sig { returns(Thread) }
+    def new_worker
+      Thread.new do
+        loop do
+          # Thread::Queue#pop is thread safe and will wait until an item is available
+          job = T.let(@job_queue.pop, T.nilable(Job))
+          # The only time when the job is nil is when the queue is closed and we can then terminate the thread
+          break if job.nil?
```
st0012, line 141: Maybe using `while` can simplify this?

```rb
while job = T.let(@job_queue.pop, T.nilable(Job))
end
```

lib/ruby_lsp/queue.rb
```diff
@@ -0,0 +1,189 @@
...
+        handlers: T::Hash[String, Handler::RequestHandler]
+      ).void
+    end
+    def initialize(writer, handlers)
+      @writer = writer
+      @handlers = handlers
+      # The job queue is the actual list of requests we have to process
+      @job_queue = T.let(Thread::Queue.new, Thread::Queue)
+      # The jobs hash is just a way of keeping a handle to jobs based on the request ID, so we can cancel them
+      @jobs = T.let({}, T::Hash[T.any(String, Integer), Job])
```
st0012, line 42: Can we enforce job ids to have the same type? So we don't need to handle 2 types of them.

lib/ruby_lsp/queue.rb
```diff
@@ -0,0 +1,189 @@
...
+
+      # Remember a handle to the job, so that we can cancel it
+      @mutex.synchronize do
+        @jobs[request[:id]] = job
+      end
+
+      @job_queue << job
+    end
+
+    sig { params(id: T.any(String, Integer)).void }
```
st0012, line 63: Same here on the job id's type

lib/ruby_lsp/queue.rb
chrisseaton: I think the LSP interface allows them to be either?

lib/ruby_lsp/queue.rb
vinistock: `Thread::Queue#pop` automatically makes the thread sleep if there are no items in the queue and then wakes it up once there's a new item there.

It only returns `nil` when the queue has been closed, which allows us granular control on how to deterministically shutdown the worker.

lib/ruby_lsp/queue.rb
vinistock: The [specification](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#cancelRequest) mentions ID can be a string or integer, so I figured it would be better to match it.

lib/ruby_lsp/queue.rb
st0012: So we may do more work in the future than just call `break`?

lib/ruby_lsp/queue.rb
st0012: I see. Yeah although it's unlikely but there could be both `1` and `"1"` being sent.

lib/ruby_lsp/queue.rb
st0012: Yes that's what I meant 🙂

lib/ruby_lsp/queue.rb
```diff
@@ -0,0 +1,187 @@
...
+      Thread.main.priority = 1
+    end
+
+    sig { params(request: T::Hash[Symbol, T.untyped]).void }
+    def push(request)
+      job = Job.new(request: request, cancelled: false)
+
+      # Remember a handle to the job, so that we can cancel it
+      @mutex.synchronize do
+        @jobs[request[:id]] = job
```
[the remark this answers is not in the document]
vinistock: It's an incremental number, it should never be reused.

lib/ruby_lsp/queue.rb
```diff
@@ -0,0 +1,187 @@
...
+      # The current job is a handle to cancel jobs that are currently being processed
+      @current_job = T.let(nil, T.nilable(Job))
+      @mutex = T.let(Mutex.new, Mutex)
+      @worker = T.let(new_worker, Thread)
+
+      Thread.main.priority = 1
+    end
+
+    sig { params(request: T::Hash[Symbol, T.untyped]).void }
+    def push(request)
```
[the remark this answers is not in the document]
vinistock: We get that for free with `Thread::Queue`. It's not possible to push things to a closed queue.

st0012 (approved):

Issue #249, opened 2022-08-15
Inform users of RuboCop infinite loops


vinistock:
RuboCop keeps trying to run auto-corrects until there are no more changes made to a file. However, depending on how cops play together, it may end up in an infinite loop.

When this happens, RuboCop will raise `RuboCop::Runner::InfiniteCorrectionLoop`, causing formatting to halt and just return an error.

This can be confusing since - without looking at VS Code's output tab - the user has no idea why formatting failed. Let's display a window message when this occurs.

PR #252, opened 2022-08-18
Show error message when there's a InfiniteCorrectionLoop exception

wildmaples:
### Motivation

Resolves: https://github.com/Shopify/ruby-lsp/issues/249

When RuboCop raises this error, any formatting runners just halt without any errors. If the users aren't looking at the output, they won't notice it - making the entire experience confusing.  

This change makes it obvious to the user that there's an infinite loop error from RuboCop through a windowed error so they can investigate. 

### Implementation

This rescue must be in two requests, `/formatting` and `/diagnostic`. Both these requests use a support runner that inherit from RuboCop::Runner module (i.e. the [format runner](https://github.com/Shopify/ruby-lsp/blob/a404a0809d9a175c21d7adffe1223c5a3d16e844/lib/ruby_lsp/requests/support/rubocop_formatting_runner.rb#L17) and [diagnostics runner](https://github.com/Shopify/ruby-lsp/blob/a404a0809d9a175c21d7adffe1223c5a3d16e844/lib/ruby_lsp/requests/support/rubocop_diagnostics_runner.rb#L17)). 

The [RuboCop::Runner::InfiniteCorrectionLoop exception](https://github.com/rubocop/rubocop/blob/89b5599d5d9d1527613cc79596ebded65404e0d4/lib/rubocop/runner.rb#L11
) may be raised in any of those runs.

### Automated Tests

Unfortunately it isn't easy to reproduce this error. I've looked into how we test another instance of a rescued exception in Server where the Rubocop config file isn't valid - that exception is simple to replicate. Would be open to ideas on how we can test this. 

### Manual Tests

Same

### DIFF
--- a/lib/ruby_lsp/server.rb
+++ b/lib/ruby_lsp/server.rb
@@ -148,6 +148,9 @@ module RubyLsp
       uri = request.dig(:params, :textDocument, :uri)
 
       Requests::Formatting.new(uri, store.get(uri)).run
+    rescue RuboCop::Runner::InfiniteCorrectionLoop => e
+      show_message(Constant::MessageType::ERROR, "Error from RuboCop: #{e.message}")
+      nil
     end
 
     on("textDocument/documentHighlight") do |request|


st0012 (approved):

wildmaples:
>It may be possible to replicate integration tests that are similar to [this one](https://github.com/Shopify/ruby-lsp/blob/a404a0809d9a175c21d7adffe1223c5a3d16e844/test/integration_test.rb#L250-L271), with [these setup](https://github.com/rubocop/rubocop/blob/00e7ac78fb83383ef537e9e08ecf2301a1427578/spec/rubocop/runner_spec.rb#L187-L244) to repro the error. Wdyt?

I looked into it more and gave this a shot but there's no no trivial to replicate and we cannot modify code that runs in the integration tests.

PR #259, opened 2022-08-22
[RFC] Add path completion support inside `require` calls

paracycle:
### Motivation

I got inspired by the recently released [Markdown Language Server](https://code.visualstudio.com/blogs/2022/08/16/markdown-language-server) doing completion for references to other files. I realized that we can match `require` strings against a list of files in the `$LOAD_PATH` and suggest completions like that.

### Implementation

1. Register a completion provider that is triggered on the `/` character
2. Handle the completion request and process it if (and only if) it is originating from the string argument of a `require` call.
3. Match the completion prefix to the list of paths in `$LOAD_PATH` and offer completions.

### Automated Tests

Added a small automated test


vinistock:
This is awesome ❤️ . I wonder if it would be worth finding files in the current directory as well? For example, if you create a brand new file and want to require it, it won't be in the load path yet.

paracycle:
@Morriar and I reworked this PR so that it uses a more efficient Trie (which is also called a Prefix Tree) data structure to index files in the load path and brought it up to work against the changes in latest `main`.

@vinistock Currently we reindex the files on the load path for every request to completion, so we would be able to discover any newly added files. I know this is not the most efficient way of doing it, but based on my usage testing the responsiveness seems to be just fine for now.

lib/ruby_lsp/requests/path_completion.rb
```diff
@@ -0,0 +1,99 @@
...
+
+        @tree = T.let(Support::PrefixTree.new(collect_load_path_files), Support::PrefixTree)
+
+        char_position = document.create_scanner.find_char_position(position)
+        @target = T.let(find(T.must(document.tree), char_position), T.nilable(SyntaxTree::TStringContent))
+      end
+
+      sig { override.returns(T.all(T::Array[LanguageServer::Protocol::Interface::CompletionItem], Object)) }
+      def run
+        # # no @target means the we are not inside a `require` call
```
Morriar, line 32:
```suggestion
        # no @target means the we are not inside a `require` call
```

lib/ruby_lsp/requests/path_completion.rb
```diff
@@ -0,0 +1,99 @@
...
+          return unless matched.message.value == "require"
+
+          args = matched.arguments
+          if args.is_a?(SyntaxTree::ArgParen)
+            args = args.arguments
+          end
+
+          path_node = args.parts.first.parts.first
+          return unless path_node
+          return unless (path_node.location.start_char..path_node.location.end_char).cover?(position)
```
Morriar, line 79: Didn't we already check this when computing `matched`?

Morriar: Maybe we should extract this as a method so we can attach it to a more explicit name

lib/ruby_lsp/requests/path_completion.rb
```diff
@@ -0,0 +1,99 @@
...
+          end
+
+        case matched
+        when SyntaxTree::Command, SyntaxTree::CommandCall, SyntaxTree::CallNode
+          return unless matched.message.value == "require"
+
+          args = matched.arguments
+          if args.is_a?(SyntaxTree::ArgParen)
+            args = args.arguments
+          end
```
Morriar, line 75:
```suggestion
          args = args.arguments if args.is_a?(SyntaxTree::ArgParen)
```

lib/ruby_lsp/requests/support/prefix_tree.rb
```diff
@@ -0,0 +1,81 @@
...
+# frozen_string_literal: true
+
+module RubyLsp
+  module Requests
+    module Support
+      class PrefixTree
+        extend T::Sig
+
+        sig { returns(Node) }
+        attr_reader :root
```
Morriar, line 11: Do we need to expose `root` here? This seems like an internal implementation detail and people should be reaching for `insert` and `search` instead.

paracycle: Removed!

lib/ruby_lsp/executor.rb
```diff
@@ -257,6 +259,15 @@ def semantic_tokens_range(uri, range)
       )
     end
 
+    sig do
+      params(uri: String, position: Document::PositionShape).returns(T.nilable(T::Array[Interface::CompletionItem]))
+    end
+    def completion(uri, position)
+      @store.cache_fetch(uri, :completion) do |document|
```
vinistock, line 266: Since this request is position dependent, I don't think we should be caching the results.

lib/ruby_lsp/requests/path_completion.rb
```diff
@@ -0,0 +1,96 @@
...
+        end
+      end
+
+      sig do
+        params(
+          node: SyntaxTree::Node,
+          position: Integer,
+        ).returns(T.nilable(SyntaxTree::TStringContent))
+      end
+      def find(node, position)
```
vinistock, line 57: We actually got rid of our previous implementation that used binary search in favour of `BaseRequest#locate`. It had the issue of not always returning the most specific node as possible.

To avoid duplication, can we please change `locate` to accept a target node (with its default being `nil`) and use it?

We can do something like
```ruby
possible_completion = locate(position, [SyntaxTree::Command, SyntaxTree::CallNode])
# perform argument and other checks
```

lib/ruby_lsp/requests/support/prefix_tree.rb
```diff
@@ -0,0 +1,78 @@
...
+        def initialize(items)
+          @root = T.let(Node.new(""), Node)
+
+          items.each do |item|
+            insert(item)
+          end
+        end
+
+        sig { params(item: String).void }
+        def insert(item)
```
vinistock, line 20: Should this be private?

lib/ruby_lsp/requests/path_completion.rb
```diff
@@ -0,0 +1,96 @@
...
+          node.child_nodes.compact.bsearch do |child|
+            if (child.location.start_char..child.location.end_char).cover?(position)
+              0
+            else
+              position <=> child.location.start_char
+            end
+          end
+
+        case matched
+        when SyntaxTree::Command, SyntaxTree::CommandCall, SyntaxTree::CallNode
```
vinistock, line 68: When would a require be a `CommandCall`? I think it can only be `Command` (for no parenthesis) and `CallNode` (for parenthesis).

vinistock (commented):
This works really well! Some comments about the code structure

lib/ruby_lsp/requests/path_completion.rb
```diff
@@ -0,0 +1,96 @@
...
+        @target = T.let(find(T.must(document.tree), char_position), T.nilable(SyntaxTree::TStringContent))
+      end
+
+      sig { override.returns(T.all(T::Array[LanguageServer::Protocol::Interface::CompletionItem], Object)) }
+      def run
+        # no @target means the we are not inside a `require` call
+        return [] unless @target
+
+        text = @target.value
+        @tree.search(text).sort.map do |path|
```
vinistock, line 35:
```suggestion
        @tree.search(text).sort.map! do |path|
```

lib/ruby_lsp/requests/path_completion.rb
paracycle: I think it can be a `CommandCall` if you call it as `Kernel.require`, which we don't have to support, but it wasn't hard to include.

lib/ruby_lsp/requests/path_completion.rb
paracycle: Ah nice, I can change it to be that. Thanks for the pointer.

paracycle: It turns out `locate` wasn't accepting a list of node types to match, so I added it as an optional kwarg to the `locate` method and used it.

PR #260, opened 2022-08-24
Handle class/module declarations as a `class` token with `declaration` modifier

paracycle:
### Motivation

The current semantic highlighting was not handling `class Foo < Bar` or `module Baz` constructs and skipping them in semantic highlighting. This PR makes them semantic `class` tokens with the `declaration` modifier.

### Implementation

Handle `visit_class` and `visit_module` callbacks and emit the correct tokens.

### Automated Tests

Updated automated tests.

### DIFF
--- a/lib/ruby_lsp/requests/semantic_highlighting.rb
+++ b/lib/ruby_lsp/requests/semantic_highlighting.rb
@@ -217,6 +217,19 @@ module RubyLsp
         add_token(node.value.location, :method) unless special_method?(node.value.value)
       end
 
+      sig { params(node: SyntaxTree::ClassDeclaration).void }
+      def visit_class(node)
+        add_token(node.constant.location, :class, [:declaration])
+        add_token(node.superclass.location, :class) if node.superclass
+        visit(node.bodystmt)
+      end
+
+      sig { params(node: SyntaxTree::ModuleDeclaration).void }
+      def visit_module(node)
+        add_token(node.constant.location, :class, [:declaration])
+        visit(node.bodystmt)
+      end
+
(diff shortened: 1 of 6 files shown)


lib/ruby_lsp/requests/semantic_highlighting.rb
```diff
@@ -217,6 +217,23 @@ def visit_vcall(node)
         add_token(node.value.location, :method) unless special_method?(node.value.value)
       end
 
+      sig { params(node: SyntaxTree::ClassDeclaration).void }
+      def visit_class(node)
+        loc = node.constant.location
+        # let's treat the whole `Foo < Bar` span as a single class declaration token
+        # if there is a superclass.
```
vinistock, line 224: This means that the `<` will be included and highlighted differently than other punctuation. Could we emit two tokens for the class and superclass instead?

paracycle: This is on purpose, since I've noticed that `<` is already included in the current highlighting. If I split `constant` and `superclass`, then the `<` looks odd.

paracycle: It is punctuation **and** class:
image
compared to:
Screenshot 2022-08-25 at 17 03 04

vinistock: Punctuation is higher up in the hierarchy, so the classification of the token is correct and we should not change it. Maybe the theme you're using doesn't define a colour for punctuation, which ends up inheriting from the wrong token type.

Issue #55, opened 2022-03-29
Add inlay hints support to the `ruby-lsp`


vinistock:
Inlay hints display "invisible" information to the user to assist with the understanding of the code. For example:

```ruby
class User < ApplicationRecord
  (self.)has_many :posts
#^ inlay hint displaying the receiver (doesn't actually exist in the real code)
end
```

We can also display the name of positional arguments as inlay hints.

Documentation: https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#textDocument_inlayHint

st0012:
@vinistock I want to give this a try. I'll start by listing common cases that can use the hint.

vinistock:
There's no built-in way by checking the method definition straight. We'd need to keep track of which section we're currently inside. Something like this in pseudo code

```ruby
class InlayHint < BaseRequest
  def initialize(...)
    @current_visibility = :public # Maybe use a Sorbet enum?
  end

  def visit_call(node) # not sure if it's call, command, fcall...
    if this_thing_is_private_public_or_protected
      @current_visibility = the_new_visibility
    end
  end

  def visit_def(node)
    add_hint_using_current_visibility
  end
end
```
Two thoughts:
- I think we shouldn't add anything for public methods, since that's the default
- If it's possible, I think this inlay hint would look better on top of the method name (not on the left or right)

RyanBrushett:
Ooo neat, thanks for the pointers!

> I think we shouldn't add anything for public methods, since that's the default

Completely agree. That would be really messy.

> If it's possible, I think this inlay hint would look better on top of the method name (not on the left or right)

Should be entirely possible. This hint in particular really runs the risk of looking messy, I agree that above is a good place to start. If I get a chance to hack on it, I'll be sure to propose a few options with screenshots for easy evaluation.

PR #261, opened 2022-08-25
Add basic inlay hint support

RyanBrushett:
### Motivation

Part of https://github.com/Shopify/ruby-lsp/issues/55

Inlay hints are neat and it seemed like an easy issue to get started with. Started with adding a hint for the implied StandardError when a user makes a bare rescue.

### Implementation

- Largely similar to the other request types
- ~~Inlay Hints were added in LSP spec 3.17 so it's not as well-typed as I'd like. The gem is still at 3.16~~
- Checks for `node.is_a?(SyntaxTree::Rescue) && node.exception.nil?`

### Automated Tests

I added an expectation test, can certainly add more if necessary.

### Manual Tests

- I modified the gemfile of a local project (Tapioca) to point to my local copy of Ruby-LSP
- I modified the client to support inlay hints (see https://github.com/Shopify/vscode-ruby-lsp/pull/202)
- I opened Tapioca when running the client in debug mode so it was using the new LSP version and the client that could support it.
- I typed the following code to ensure the inlay hint popped up:

```ruby
begin
  puts "hello"
rescue
  puts "bye-bye"
end
```


lib/ruby_lsp/server.rb
```diff
@@ -172,6 +177,14 @@ module RubyLsp
       end
     end
 
+    on("textDocument/inlayHint", parallel: true) do |request|
+      document = store.get(request.dig(:params, :textDocument, :uri))
+
+      if document.parsed?
+        Requests::InlayHints.new(document).run
+      end
```
RyanBrushett, line 185: Feedback on this? Is this the right thing to put here?

RyanBrushett:
CI failing on doc check. I'll dig into that shortly.

RyanBrushett:
I need to do a gif demo. I'll do that in the next few days.

lib/ruby_lsp/server.rb
vinistock: The `document.parsed?` aspect is indeed correct, but I think we're missing the range parameter. The [spec](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#textDocument_inlayHint) mentions that inlay hints are computed for the current visible range.

lib/ruby_lsp/requests/inlay_hints.rb
```diff
@@ -0,0 +1,55 @@
...
+      def run
+        visit(@document.tree)
+        @hints
+      end
+
+      sig { params(node: T.nilable(SyntaxTree::Node)).void }
+      def visit(node)
+        return if node.nil?
+
+        if node.is_a?(SyntaxTree::Rescue) && node.exception.nil?
```
vinistock, line 39: Instead of checking the node type, we can just define `visit_rescue` and it'll only be invoked on rescue nodes.

Additionally, you can use the range parameter sent from the client to skip over rescue nodes that are not in the visible range.

Something like this
```ruby
sig { params(node: SyntaxTree::Rescue).void }
def visit_rescue(node)
  return unless node.exception.nil?

  loc = node.location
  return unless loc.start_line - 1 >= @range.dig(:start, :line) && loc.end_line - 1 <= @range.dig(:end, :line)

  @hints << ...
end
```

vinistock (commented):
This is great ❤️

lib/ruby_lsp/server.rb
RyanBrushett: Right, OK! I'll add that, my bad.

paracycle:
@RyanBrushett The 3.17 gem upgrade was done by dependabot already, and I did the RBI update and the TODO resolution that was waiting for it in https://github.com/Shopify/ruby-lsp/pull/268. Feel free to rebase on `main` once that is merged so that you can get rid of your LSP related commits.

As for your question about `super`: Yes, my understanding is that you still need it so that you can catch `rescue`s inside other `rescue` blocks like:
```ruby
begin
  #
rescue
  # Not sure why anyone would do this, but it is legal so should be supported:
  begin
  rescue # Without `super` in `visit_rescue`, this won't be processed.
  end
end
```
Without calling `super`, you would not be processing the child nodes of the `rescue` node and you would be missing the inner one.

Might be good to add that as a test case, btw.

test/requests/inlay_hints_expectations_test.rb
```diff
@@ -0,0 +1,14 @@
...
+
+require "test_helper"
+require "expectations/expectations_test_runner"
+
+class InlayHintsExpectationsTest < ExpectationsTestRunner
+  expectations_tests RubyLsp::Requests::InlayHints, "inlay_hints"
+
+  def run_expectations(source)
+    document = RubyLsp::Document.new(source)
+    RubyLsp::Requests::InlayHints.new(document, 0..5).run
```
vinistock, line 12: If we use `default_args`, you shouldn't need to override `run_expectations`. Similar to [here](https://github.com/Shopify/ruby-lsp/blob/29c1d14020ca74f1f92f46708245b386630b91ec/test/requests/selection_ranges_expectations_test.rb#L21), we can do
```ruby
def default_args
  [0..5]
end
```

lib/ruby_lsp/requests/inlay_hints.rb
```diff
@@ -0,0 +1,53 @@
...
+    # # Example
+    #
+    # ```ruby
+    # begin
+    #   puts "do something that might raise"
+    # rescue # Label "StandardError" goes here as a bare rescue implies rescuing StandardError
+    #   puts "handle some rescue"
+    # end
+    # ```
+
```
Morriar, line 21:
```suggestion
```

lib/ruby_lsp/requests/inlay_hints.rb
```diff
@@ -0,0 +1,53 @@
...
+      sig { params(node: SyntaxTree::Rescue).void }
+      def visit_rescue(node)
+        return unless node.exception.nil?
+
+        loc = node.location
+        return unless @range.cover?(loc.start_line - 1) && @range.cover?(loc.end_line - 1)
+
+        @hints << LanguageServer::Protocol::Interface::InlayHint.new(
+          position: { line: loc.start_line - 1, character: loc.start_column + "rescue".length },
```
Morriar, line 45: Should we extract `"rescue".length` as a constant?

lib/ruby_lsp/server.rb
```diff
@@ -188,6 +193,17 @@ module RubyLsp
...
 
+    on("textDocument/inlayHint", parallel: true) do |request|
+      document = store.get(request.dig(:params, :textDocument, :uri))
+      range = request.dig(:params, :range)
+      start_line = range.dig(:start, :line)
+      end_line = range.dig(:end, :line)
+
+      if document.parsed?
+        Requests::InlayHints.new(document, start_line..end_line).run
+      end
```
Morriar, line 204:
```suggestion
      return unless document.parsed?
      
      Requests::InlayHints.new(document, start_line..end_line).run
```

vinistock: You can't directly from this block or it results in a `LocalJumpError`. We've been bitten by this before 😕

paracycle: But you can do `next unless document.parsed?` 😉

Morriar: Urgh

paracycle: On this point, I actually want to make the LSP gem more typed so that instead of registering blocks, you start overriding methods that know what parameters of what type will be passed in for the request that you are handling. See:
- https://github.com/mtsmfm/language_server-protocol-ruby/issues/46, and
- https://github.com/mtsmfm/language_server-protocol-ruby/issues/47

lib/ruby_lsp/requests/inlay_hints.rb
RyanBrushett: Ah good eye! Sure!

lib/ruby_lsp/server.rb
```diff
@@ -55,6 +55,10 @@ module RubyLsp
         )
       end
 
+      inlay_hint_provider = if enabled_features.include?("inlayHint")
+        { resolveProvider: false }
```
vinistock, line 59: Now with the 3.17, we can use [inlay hint option](https://github.com/mtsmfm/language_server-protocol-ruby/blob/main/lib/language_server/protocol/interface/inlay_hint_options.rb) instead of the hash.

vinistock (approved):

lib/ruby_lsp/server.rb
RyanBrushett: Ooo good eye

RyanBrushett:
Implemented PR feedback, added a test for nested begin/rescues, and updated my code to be 3.17-friendly! Rebased.

I'll squash all this down like a neat little sandwich if it looks good to you all!

Screenshot 2022-08-29 at 14 45 53

Morriar (approved):

PR #269, opened 2022-08-29
Add runtime dependency on version 3.17.x of LSP gem

paracycle:
### Motivation

Since we are currently supporting v3.17 of the spec, made the runtime dependency to only consider 3.17.x.y versions of the LSP gem, in case there are breaking changes to the protocol with the 3.18 version of the spec.


vinistock:
Thanks!

PR #275, opened 2022-08-31
Make sure stdout is clean during all request processing

paracycle:
### Motivation

RuboCop is still too chatty and breaks the protocol by printing to stdout despite our best attempts. Today I was having problems with the formatting on Tapioca and realized it was because of the:
```
An error occurred while Layout/BlockEndNewline cop was inspecting /Users/ufuk/src/github.com/Shopify/tapioca/lib/tapioca/static/rbs_converter.rb:166:8.
To see the complete backtrace run rubocop -d.
```
problem (which is [in per-1.34 versions of RuboCop](https://github.com/rubocop/rubocop/pull/10878)). 

I am not exactly sure why this was causing a problem, since that specific message should be going to stderr, but the LSP communication was still broken because of it.

Regardless, I realized that we never want any request to print to stdout, willingly or by mistake, so I decided to resolve that at the handler layer.

### Implementation

Implement a `capture_stdout` and use it inside the `Queue#execute` method.

### Automated Tests

I haven't added any but it might be nice to maybe add one.

### Manual Tests

Not sure how, but if you can recreate the conditions of https://github.com/rubocop/rubocop/pull/10878 on a repo that is using RuboCop 1.33 or earlier, then it is enough to trigger the broken behaviour and to see that this PR fixes it.


vinistock:
TIL about the `STDERR` difference. To add a unit test for this, you can register a handler in the [queue_test](https://github.com/Shopify/ruby-lsp/blob/e5c3fd05531a5614ba5f54c468a7f809bd997f25/test/queue_test.rb#L32) that prints something and verify that it didn't actually print anything.

paracycle:
@st0012 Yeah, I know about the difference between `STDERR`/`STDOUT` and `$stderr`/`$stdout`. However, my goal is to catch legitimate uses of `puts` in general, which end up going to `$stdout`.

In any case, I changed my approach and implementation to limit the capture to only RuboCop and to capture only `$stdout`, which seems to working well for me. FYI @vinistock.

paracycle:
Ok, it ended up being much more tricky and I had to change the `$VERBOSE` global as well. That flag controls if `Kernel#warn` is going to have an output or not and I propose that we set `$VERBOSE = nil` for the `ruby-lsp` executable overall in any case. I don't think we want random warning messages being sent to the `Output` tab.

WDYT?

st0012:
>  I propose that we set $VERBOSE = nil for the ruby-lsp executable overall in any case

I think that makes sense 👍 

Regarding the implementation, I was a bit worried about swapping `$stderr/out` will affect the other thread's output as we can run [2 requests at a time](https://github.com/Shopify/ruby-lsp/blob/e5c3fd05531a5614ba5f54c468a7f809bd997f25/lib/ruby_lsp/handler.rb#L68-L73) now. But right now it's not going to happen because [language_server-protocol-ruby](https://github.com/mtsmfm/language_server-protocol-ruby) uses [`STDOUT`](https://github.com/mtsmfm/language_server-protocol-ruby/blob/e9a357f576d5bef81de192dd98e7ddfe098b06e3/lib/language_server/protocol/transport/stdio/writer.rb#L7) instead of `$stdout`.

However, it's still quite dangerous because if one day the `STDOUT` thing is changed, we can start experiencing weird race condition issues.

lib/ruby_lsp/requests/support/rubocop_runner.rb
```diff
@@ -0,0 +1,90 @@
...
+
+          $VERBOSE = nil
+          $stderr = StringIO.new
+          $stdout = StringIO.new
+
+          block.call
+        ensure
+          $stdout = orig_stdout
+          $stderr = orig_stderr
+          $VERBOSE = original_verbosity
```
Morriar, line 85: Nit, `VERBOSE` is first in the two previous related blocks.

```suggestion
          $VERBOSE = original_verbosity
          $stdout = orig_stdout
          $stderr = orig_stderr
```

lib/ruby_lsp/requests/support/rubocop_runner.rb
```diff
@@ -0,0 +1,90 @@
...
+        end
+
+        sig { params(block: T.proc.void).void }
+        def capture_output(&block)
+          original_verbosity = $VERBOSE
+          orig_stdout = $stdout
+          orig_stderr = $stderr
+
+          $VERBOSE = nil
+          $stderr = StringIO.new
```
Morriar, line 78: How hard/bad would it be to monkey patch the classes from Rubocop that still print so we noop `puts` rather than changing the global variable?

lib/ruby_lsp/requests/support/rubocop_runner.rb
```diff
@@ -0,0 +1,90 @@
...
+          args += DEFAULT_ARGS
+          rubocop_options = ::RuboCop::Options.new.parse(args).first
+          super(rubocop_options, ::RuboCop::ConfigStore.new)
+        end
+
+        sig { params(path: String, contents: String).void }
+        def run(path, contents)
+          @errors = []
+          @warnings = []
+          @offenses = []
```
vinistock, line 43: Can we use `@offenses.clear` instead?

lib/ruby_lsp/requests/support/rubocop_runner.rb
```diff
@@ -0,0 +1,90 @@
...
+
+          args += DEFAULT_ARGS
+          rubocop_options = ::RuboCop::Options.new.parse(args).first
+          super(rubocop_options, ::RuboCop::ConfigStore.new)
+        end
+
+        sig { params(path: String, contents: String).void }
+        def run(path, contents)
+          @errors = []
+          @warnings = []
```
vinistock, line 42: Why do we need these?

lib/ruby_lsp/requests/support/rubocop_runner.rb
paracycle: I was explicitly reverting changes in the reverse order :)

lib/ruby_lsp/requests/support/rubocop_diagnostics_runner.rb
```diff
@@ -14,46 +11,31 @@ module RubyLsp
...
+          diagnostics(uri)
         end
 
         private
 
-        sig { params(_file: String, offenses: T::Array[RuboCop::Cop::Offense]).void }
-        def file_finished(_file, offenses)
-          @diagnostics = offenses.map { |offense| Support::RuboCopDiagnostic.new(offense, T.must(@uri)) }
+        sig { params(uri: String).returns(T::Array[Support::RuboCopDiagnostic]) }
+        def diagnostics(uri)
```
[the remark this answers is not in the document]
paracycle: Sure, we can do that!

lib/ruby_lsp/requests/support/rubocop_runner.rb
```diff
@@ -0,0 +1,90 @@
...
+
+        sig { returns(String) }
+        def formatted_source
+          @options[:stdin]
+        end
+
+        private
+
+        sig { void }
+        def display_handled_errors
```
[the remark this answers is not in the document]
paracycle: This is so that people can see and/or report problems that any formatting/diagnostic operation has encountered. I feel like this is needed more for troubleshooting than making people aware. After all, there is no action to be taken on any of these, since they are already handled.

This is what the output looks like:
```
[RuboCop] Encountered and handled errors:
[RuboCop]   - An error occurred while Layout/BlockEndNewline cop was inspecting /Users/ufuk/src/github.com/Shopify/tapioca/lib/tapioca/static/rbs/parameter_converter.rb:4:17.
```

PR #279, opened 2022-09-01
Handle formatting errors in on_error block

### LINKED ISSUE #277: Uninitialized constant RubyLsp::RuboCop before formatting or diagnostics run for the first time
We only load RuboCop inside formatting and diagnostics, but we rescue RuboCop errors in [server.rb](https://github.com/Shopify/ruby-lsp/blob/e5c3fd05531a5614ba5f54c468a7f809bd997f25/lib/ruby_lsp/server.rb#L163)

Sometimes, the order of operations leads to an uninitialized constant error. We may need to try to require RuboCop in `internal` or `server` to prevent this from happening.

vinistock:
### Motivation

Closes #277 

Everything inside of the `on` block is cancellable by our queue mechanism. Therefore, we cannot perform any IO operations inside of it. We were using `show_message` inside of a rescue, instead of using our `on_error` handler, which means we were writing to IO on a cancellable block.

### Implementation

Just switch to using `on_error`, which is guaranteed to run and cannot be cancelled in the middle of writing to IO.

### Automated Tests

I don't have any clever ideas for a test that will catch regression, but if you do let me know!

### DIFF
--- a/lib/ruby_lsp/requests/formatting.rb
+++ b/lib/ruby_lsp/requests/formatting.rb
@@ -19,6 +19,8 @@ module RubyLsp
     # end
     # ```
     class Formatting < BaseRequest
+      class Error < StandardError; end
+
       extend T::Sig
 
       sig { params(uri: String, document: Document).void }
--- a/lib/ruby_lsp/server.rb
+++ b/lib/ruby_lsp/server.rb
@@ -160,9 +160,8 @@ module RubyLsp
       uri = request.dig(:params, :textDocument, :uri)
 
       Requests::Formatting.new(uri, store.get(uri)).run
-    rescue RuboCop::Runner::InfiniteCorrectionLoop => e
-      show_message(Constant::MessageType::ERROR, "Error from RuboCop: #{e.message}")
-      nil
+    end.on_error do |error|
+      show_message(Constant::MessageType::ERROR, "Formatting error: #{error.message}")
     end
 
     on("textDocument/onTypeFormatting", parallel: true) do |request|
--- a/lib/ruby_lsp/requests/support/rubocop_runner.rb
+++ b/lib/ruby_lsp/requests/support/rubocop_runner.rb
@@ -48,6 +48,8 @@ module RubyLsp
           @options[:stdin] = contents
 
           super([path])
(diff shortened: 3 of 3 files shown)


lib/ruby_lsp/server.rb
```diff
@@ -160,9 +160,10 @@ module RubyLsp
       uri = request.dig(:params, :textDocument, :uri)
 
       Requests::Formatting.new(uri, store.get(uri)).run
-    rescue RuboCop::Runner::InfiniteCorrectionLoop => e
-      show_message(Constant::MessageType::ERROR, "Error from RuboCop: #{e.message}")
-      nil
+    end.on_error do |error|
+      if error.is_a?(RuboCop::Runner::InfiniteCorrectionLoop)
```
paracycle, line 164: Should we not handle this error at the `RuboCopRunner` layer?

paracycle: Sorry, that wasn't what I meant to say. I was trying to say that seeing `RuboCop::Runner` here is leaky and might end up leading to bigger failures on projects that don't have RuboCop loaded (our runtime dependency on RuboCop is conditional, remember?).

So, I think the `RuboCopRunner` class should handle this error and maybe raise a `RubyLsp::Requests::Formatting::Error` which can be handled at this layer. We can always get the original RuboCop exception using the `cause`: https://ruby-doc.org/core-3.1.0/Exception.html#cause-method

st0012 (approved):

lib/ruby_lsp/requests/formatting.rb
```diff
@@ -56,6 +58,8 @@ def formatted_file
         else
           SyntaxTree.format(@document.source)
         end
+      rescue RuboCop::Runner::InfiniteCorrectionLoop => error
```
paracycle, line 61: I am afraid this is still not the correct place. In this file, you are still not guaranteed that `RuboCop` will exist as a constant.

The error handling needs to live inside the new `RuboCopRunner`, since that's the only location that does a require and bails if the require fails.

paracycle (changes requested):

paracycle (approved):

Issue #258, opened 2022-08-22
Use the document link request to link to Rails documentation


egiurleo:
We implemented the DocumentLink request in #58 and used it to go to method definition from gem RBIs.

As a next step, let's use this request to link to the correct Rails documentation for methods that come from the rails gem.

When a user hovers over a method that's defined in rails, there should be a link that says "go to documentation." We can use information about the method, such as the class that defines it and its ancestor chain, as well as information about the rails gem (version number) to link to the correct documentation page for that method.

st0012:
I'm not sure if the implementation in my head is the best solution, so I want to discuss it here:

### Link format

**Class (or module)**
- Example: https://api.rubyonrails.org/v6.1.4/classes/ActiveSupport/Callbacks.html
- `"https://api.rubyonrails.org/#{version}/classes/#{namespaces.join("/")}.html"`

**Method**
- Example: https://api.rubyonrails.org/v6.1.4/classes/ActiveSupport/Callbacks.html#method-i-run_callbacks
- `"https://api.rubyonrails.org/#{version}/classes/#{namespaces.join("/")}.html#method-#{instance_or_class}-#{method_name}"`

### Information required for making the link

- Gem name (e.g. `activesupport`): file name (e.g. `activesupport@7.0.3.1.rbi`)
    - Need it to check if it's a Rails component
- Gem version (e.g. `7.0.3.1`): file name (e.g. `activesupport@7.0.3.1.rbi`)
    - Need it to form the link
- Class/Module name
    - Need it to form the link
- (For methods) method name
    - Need it to form the link's anchor
- (For methods) if it's a class or instance method
    - Need it to form the link's anchor

### Implementation

1. When the request comes in, extract the gem's name and version from it.
2. Check if it's a Rails gem. Return if it's not.
3. Create a namespace stack
4. Parse the document and
    1. When we enter a class/module definition, create a link and push it to the stack
    2. When we visit a method definition, create a link with the class/module name on the stack + the method name

paracycle:
Rails API guides have a search mechanism which relies on the data stored in the `search_index.js` [file](https://api.rubyonrails.org/v6.1.4/js/search_index.js). The data is basically JSON, which we can use to match method names to discover the full doc URL.

For example, the `info` hash contains the following for `run_callbacks`:
```json5
  // ...
  [
    "run_callbacks",
    "ActiveSupport::Callbacks",
[...]

PR #281, opened 2022-09-02
Pass path to expectation tests instead of the file content

st0012:
### Motivation

When implementing #258, we'll decide whether to display links based on the file's path. And to test that mechanism, we need to pass real uri to the tests. But under the current testing structure, it's not possible to access that information in expectation tests.

It's also worth noting that **not** every test needs or can use real uri. For example, rubocop related tests can't use real uri because the fixture path is excluded in `.rubocop.yml`.

With that being said, I think tests should opt-in to real uri when possible so the testing setup can be more realistic.

### Implementation

Instead of passing `source` to `run_expectations`, passing the file's path instead. 

### Automated Tests

Existing tests


st0012:
Thanks for the suggestions. I'll just assign it to `@_path` 👍

PR #282, opened 2022-09-02
Fix on type formatting pipe completion for regular or expressions

vinistock:
### Motivation

We were always automatically adding another pipe if the user typed in one. However, that's not correct, we only want to duplicate the pipe for block arguments (pipes that are preceded with `do` or a curly brace). Always adding two pipes is an odd experience.

### Implementation

Started checking if `do` or a curly brace is present in the same line of the original pipe.

### Automated Tests

Added a test that fails on the original implementation.


egiurleo (approved):
😅 Good catch

KaanOzkan (approved):

PR #283, opened 2022-09-02
Add inlay hint for hash arguments with no braces

RyanBrushett:
### Motivation

I like this inlay hint, at least in theory. It makes it clear what's actually going on when you can sometimes omit braces from method calls. Could be helpful for newer ruby users too.

Definitely feel free to boot this one up and have a look at it. As it is, these will also be added in sorbet signatures because `params` is called with BareAssocHash arguments. Maybe this could be refined to only add the hint if there's a mix of positional arguments and hash arguments? I'm open to suggestions.

### Implementation

When visiting arguments, if it has a child BareAssocHash, process the inlay hints around that hash.

**Note:** I still need to get a gif of it working for docs. Or maybe the StandardError example is sufficient to demonstrate inlay hints? Your call.

### Automated Tests

Added a bunch of different expectation tests. Any cases I'm missing?

### Manual Tests

- I modified the gemfile of a local project (Tapioca) to point to my local copy of Ruby-LSP
- I opened Tapioca when running the client in debug mode so it was using my branch LSP version
- I typed the following code to ensure the inlay hint popped up and get a clearer idea of what it looks like and how it `feels` to use it.

```ruby
  def foo(positional, **args)
    puts positional
    puts args
  end

  foo("positional", bar: "one", baz: "two")

  foo(
    "positional",
    bar: "three",
    baz: "four",
  )

  foo("positional", bar: "five",
    baz: "six")

  puts bar: "one"

  foo "positional", bar: "one", baz: "two"
```

### DIFF
--- a/lib/ruby_lsp/requests/inlay_hints.rb
+++ b/lib/ruby_lsp/requests/inlay_hints.rb
@@ -17,6 +17,9 @@ module RubyLsp
     # rescue # Label "StandardError" goes here as a bare rescue implies rescuing StandardError
     #   puts "handle some rescue"
     # end
+    #
+    # foo(positional_arg, key: value)
+    # #                  ^          ^ these two locations will have brace labels added
     # ```
     class InlayHints < BaseRequest
       RESCUE_STRING_LENGTH = T.let("rescue".length, Integer)
@@ -40,7 +43,7 @@ module RubyLsp
         return unless node.exception.nil?
 
         loc = node.location
-        return unless @range.cover?(loc.start_line - 1) && @range.cover?(loc.end_line - 1)
+        return unless visible?(loc)
 
         @hints << LanguageServer::Protocol::Interface::InlayHint.new(
           position: { line: loc.start_line - 1, character: loc.start_column + RESCUE_STRING_LENGTH },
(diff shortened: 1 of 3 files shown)


vinistock (approved):
I like it, even if it appears on sig params, but I'm curious to know what others think. Concerning the GIF, I think it's fine to leave the current one as is. The idea of the documentation is to show an example of each feature so that people know what they are, but not necessarily show everything it does.

Morriar (commented):
I'm am not convinced by the usefulness of this and it seems to add a lot of clutter.

How does that look in a file using Rails DSLs for example?

If we go ahead with this feature, can we make it possible to disable it?

rafaelfranca:
Using your example:

```
>> foo 1, a: 1
1
{:a=>1}   
=> nil    
>> foo 1, { a: 1 }
(irb):1:in `foo': wrong number of arguments (given 2, expected 1) (ArgumentError)
        from (irb):7:in `<main>'
        from /opt/rubies/ruby-3.1.2/lib/ruby/gems/3.1.0/gems/irb-1.4.1/exe/irb:11:in `<top (required)>'
        from /usr/local/bin/irb:25:in `load'
        from /usr/local/bin/irb:25:in `<main>'
```

RyanBrushett:
I haven't had a lot of time to hack on this but I'm glad it exists so if anyone wants to pick it back up and nudge this over the line, maybe this will be a nice starting point.

I think Alex is right, digging into configuring which hints are active is probably a more valuable thing to tackle before adding more.

Thank you all for the feedback and the help! ❤️

PR #284, opened 2022-09-02
Add semantic tokens for local arguments

vinistock:
### Motivation

Tracking the local environment enables a lot of features that we can't otherwise implement. For example, knowing all occurrences of a specific variable or keeping track of parameters that only exist inside a specific scope.

This RFC PR proposes a way in which we can keep track of the environment in a simple way that is consistent with our existing requests.

### Implementation

It's significantly easier to review per commit.
1. Create a new request parent class `RequestWithEnvironment` and `Support::Environment`. The `Support::Environment` class keeps track of the current environment for a scope and the `RequestWithEnvironment` automatically pushes and pops new environments as needed (as well as declaring local arguments)
2. Add support for semantic highlighting method parameters consistently inside their current scope

For reference, this is what local variables vs local arguments look like to me in this branch.
Screen Shot 2022-09-02 at 1 38 30 PM

### Automated Tests

Added tests for the new type of request. Adapted and added more examples for semantic highlighting.

### Manual Tests

1. Fire up the LSP on this branch
2. Visit any file that has methods with arguments
3. Verify they are highlighted consistently

Note: depending on your theme configurations, there may actually not be any difference between highlighting of variables and parameters. This won't work for everybody, but here are my theme customizations to help
```json
{
  "editor.semanticTokenColorCustomizations": {
    "[YOUR_THEME_NAME_GOES_HERE]": {
      "rules": {
        "variable": {
          "foreground": "#3cc098",
        }
      }
    }
  },
  "editor.tokenColorCustomizations": {
    "[YOUR_THEME_NAME_GOES_HERE]": {
      "textMateRules": [
        {
          "scope": [
            "variable.parameter",
          ],
          "settings": {
            "foreground": "#ff9a3b",
          }
        },
      ]
    }
  }
}
```


lib/ruby_lsp/requests/semantic_highlighting.rb
```diff
@@ -182,37 +183,45 @@ def visit_m_assign(node)
...
 
       sig { params(node: SyntaxTree::VarRef).void }
       def visit_var_ref(node)
-        case node.value
+        value = node.value
+
+        case value
         when SyntaxTree::Ident
-          add_token(node.value.location, :variable)
+          type = current_environment.find_argument(value.value) ? :parameter : :variable
```
st0012, line 221: Same as the above

lib/ruby_lsp/requests/semantic_highlighting.rb
```diff
@@ -102,7 +102,8 @@ def run
 
       sig { params(node: SyntaxTree::ARefField).void }
       def visit_a_ref_field(node)
-        add_token(node.collection.value.location, :variable)
+        type = current_environment.find_argument(node.collection.value) ? :parameter : :variable
```
st0012, line 105: Maybe we can be a helper of `Environment`? Like

```rb
type = current_environment.determine_value_type(node.collection)
```

lib/ruby_lsp/requests/semantic_highlighting.rb
```diff
@@ -182,37 +183,45 @@ def visit_m_assign(node)
...
 
       sig { params(node: SyntaxTree::VarField).void }
       def visit_var_field(node)
-        case node.value
+        value = node.value
+
+        case value
         when SyntaxTree::Ident
-          add_token(node.value.location, :variable)
+          type = current_environment.find_argument(value.value) ? :parameter : :variable
```
st0012, line 208: Same as the above

lib/ruby_lsp/requests/semantic_highlighting.rb
st0012, line 221: Same as the above

lib/ruby_lsp/requests/request_with_environment.rb
```diff
@@ -0,0 +1,64 @@
...
+      attr_reader :current_environment
+
+      sig { params(document: Document).void }
+      def initialize(document)
+        @current_environment = T.let(Support::Environment.new, Support::Environment)
+        super
+      end
+
+      sig { override.overridable.returns(Object) }
+      def run; end
```
st0012, line 24: Question: why is this needed?

lib/ruby_lsp/requests/request_with_environment.rb
```diff
@@ -0,0 +1,64 @@
...
+        node.keywords.each { |param| @current_environment.declare_argument(param.first) }
+        node.optionals.each { |param| @current_environment.declare_argument(param.first) }
+      end
+
+      sig { params(block: T.proc.void).void }
+      def with_new_environment(&block)
+        previous_environment = @current_environment
+        @current_environment = Support::Environment.new(@current_environment)
+        block.call
+        @current_environment = previous_environment
```
st0012, line 60: Do we need to put this in an `ensure` block to make sure environment is restored? I know in the case of exception that probably won't matter though.

st0012:
Forgot to say, I love this 👍

RyanBrushett:
Could this be used also to keep track of method visibility, e.g. private/protected/public?

vinistock:
We've chatted in Slack and we'll look into implementing this directly in syntax_tree. Concerning method visibilities, yes we could also use it to keep track of it.

vinistock:
Okay, this should be good for a round of reviews.

lib/ruby_lsp/requests/semantic_highlighting.rb
```diff
@@ -172,37 +173,59 @@ def visit_kw(node)
...
         end
 
         node.requireds.each do |required|
-          add_token(required.location, :variable)
+          add_token(required.location, :parameter)
         end
 
         rest = node.keyword_rest
-        return if rest.nil? || rest.is_a?(SyntaxTree::ArgsForward)
+        unless rest.nil? || rest.is_a?(SyntaxTree::ArgsForward)
```
paracycle, line 184: Nitpick: This conditional would be so much easier to read if flipped:
```suggestion
        if rest && !rest.is_a?(SyntaxTree::ArgsForward)
```

lib/ruby_lsp/requests/semantic_highlighting.rb
```diff
@@ -172,37 +173,59 @@ def visit_kw(node)
...
+        case value
         when SyntaxTree::Ident
-          add_token(node.value.location, :variable)
+          local = current_environment.find_local(value.value)
+
+          type = if local.nil? || local.type == :variable
+            :variable
+          else
+            :parameter
+          end
```
paracycle, line 204: Nitpick:
```suggestion
          type = :variable 
          type = :parameter if local && local.type != :variable
```

lib/ruby_lsp/requests/semantic_highlighting.rb
```diff
@@ -172,37 +173,59 @@ def visit_kw(node)
...
```
paracycle, line 224: Same nitpick:
```suggestion
          type = :variable 
          type = :parameter if local && local.type != :variable
```

lib/ruby_lsp/requests/semantic_highlighting.rb
paracycle: Actually, it might even be better to extract a method:
```ruby
def type_for_local(value)
  local = current_environment.find_local(value.value)

  if local && local.type != :variable
    :parameter
  else
    :variable
  end
end
```

paracycle (approved):
This looks good, I mostly have super minor comments.

PR #285, opened 2022-09-02
Support document link for Rails documentation

### LINKED ISSUE #258: Use the document link request to link to Rails documentation

st0012:
### Motivation

Closes #258

### TODOs

- [x] Replace Rails version number with a placeholder
- [x] Mock the Rails document request in tests
- [x] Activate the feature based on file path, e.g. Only check `activerecord`'s documents when the file is under a `models/` folder.
- [x] Exclude `activesupport` and `actionview`
- [x] Add tests for `actionpack`, `activemodel`, and `activestorage`
- [x] Use better way for searching
- [x] Documentation


lib/ruby_lsp/requests/document_link.rb
```diff
@@ -114,6 +186,24 @@ def resolve_version(uri)
 
         GEM_TO_VERSION_MAP[uri.gem_name]
       end
+
+      sig { params(node: SyntaxTree::Node, name: String).void }
+      def add_rails_document_link(node, name)
+        # ignore non-Ruby files
+        return unless @uri.match?(/\.rb$/)
```
vinistock, line 193: The LSP's scheme configuration is already set to Ruby files, so we should never get one that isn't Ruby.

lib/ruby_lsp/requests/document_link.rb
```diff
@@ -114,6 +186,24 @@ def resolve_version(uri)
...
+
+        candidate = self.class.rails_documents.bsearch { |ary| ary[0] <=> name }
```
vinistock, line 195: To avoid having to use a binary search on every node that matches, can we switch the implementation to using a hash where the key is the thing we want to find documentation for?

I mean pre-building and then doing straight access on something like
```ruby
@rails_documents = {
  "belongs_to" => "url_for_belongs_to",
  "ActionController::Base" => "url_for_action_controller_base"
}
```

lib/ruby_lsp/requests/document_link.rb
st0012: I want to exclude files like `Gemfile` or `Rakefile`...etc.

lib/ruby_lsp/requests/document_link.rb
```diff
@@ -63,6 +72,54 @@ def gem_paths
...
+            end
+          rescue StandardError => e
+            $stderr.puts("Exception occurred when fetching Rails document index: #{e.inspect}")
+            []
+          end, T.nilable(T::Array[T.untyped]))
+        end
+        sig { returns(T::Array[T.untyped]) }
+        def rails_documents
+          if @rails_documents
+            @rails_documents
```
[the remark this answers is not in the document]
st0012: Ah nvm. I think having it `nilable` is actually better for checking if `fetch_rails_documents` has been performed yet.

lib/ruby_lsp/requests/document_link.rb
```diff
@@ -63,6 +72,53 @@ def gem_paths
             lookup
           end, T.nilable(T::Hash[String, T::Array[String]]))
         end
+
+        sig { void }
+        def fetch_rails_documents
+          @rails_documents = T.let(begin
```
vinistock, line 78: To get around the nilable thing, you can probably rename this method to `rails_documents` and use conditional assignment, which Sorbet now supports. Something like
```ruby
sig { returns(T::Hash[String, String]) }
def rails_documents
  @rails_documents ||= T.let(..., T.nilable(T::Hash[String, String]))
end
```
We know we'll always set it to a hash in fail cases, so it should only execute once.

lib/ruby_lsp/requests/document_link.rb
```diff
@@ -82,6 +139,20 @@ def run
         @links
       end
 
+      sig { params(node: SyntaxTree::Command).void }
+      def visit_command(node)
+        method_name = node.message.value
+        add_rails_document_link(node.message, method_name)
```
vinistock, line 145:
```suggestion
        message = node.message
        add_rails_document_link(message, message.value)
```

lib/ruby_lsp/requests/document_link.rb
```diff
@@ -114,6 +185,29 @@ def resolve_version(uri)
...
+
+        doc_path = T.must(self.class.rails_documents)[name]
+
+        if doc_path
+          url = "#{RAILS_DOC_HOST}/v#{RAILTIES_VERSION}/#{doc_path}"
+
+          @links << LanguageServer::Protocol::Interface::DocumentLink.new(
+            range: range_from_syntax_tree_node(node),
+            target: url,
+            tooltip: "Browse the Rails documentation for: #{name}"
```
vinistock, line 207: NIT
```suggestion
            tooltip: "Jump to #{name}"
```

lib/ruby_lsp/requests/document_link.rb
```diff
@@ -114,6 +185,29 @@ def resolve_version(uri)
...
+        return unless @uri.match?(/\.rb$/)
+
+        if RAILTIES_VERSION && self.class.rails_documents.nil?
+          $stderr.puts "Fetching Rails Documents..."
+          self.class.fetch_rails_documents
+        end
+
+        doc_path = T.must(self.class.rails_documents)[name]
+
+        if doc_path
```
vinistock, line 201: NIT: can we `return unless doc_path` to avoid the extra nesting?

lib/ruby_lsp/requests/document_link.rb
st0012: I think jump is not accurate because we're not going to the source code?
And given that the implementation could give wrong suggestion, I think explicitly mentioning it's to Rails doc is still necessary.

vinistock:
This is looking great. One final detail: do we need to pull in VCR and webmock for testing it? Would we be able to use a Minitest stub and avoid adding the dependencies? I don't foresee the LSP having too many API requests, so I'm not sure adding the dependencies will pay off.

lib/ruby_lsp/requests/document_link.rb
st0012: One problem with Gemfile is that `AcitveRecord`'s `group` method will be mapped to the Gemfile's `group :test do`, which is incorrect. And there are 2 ways to implement it:

1. Exclude `Gemfile`, which is I'm doing for now
2. Use `path: namespace` pair to for filtering. For example, `"models/": /ActiveRecord/, /ActiveModel/` will work as well. I'm more keen to this but I'm still thinking what data structure will fit this nested filtering better.

vinistock: I like the second idea. If we can have our hash somehow store a path regex as well we can check the `uri` in a more general manner.

vinistock:
Agreed that there's a lot of value in that, but you can also achieve it with a stub, no?
````ruby
Net::Http.stub(:get_response, js_content) do
  ...
end
```

test/requests/document_link_expectations_test.rb
```diff
@@ -20,20 +20,31 @@ def map_expectations(expectations)
     end
   end
 
+  class FakeHTTPResponse < Struct.new(:code, :body)
+  end
+
   def run_expectations(source)
     document = RubyLsp::Document.new(source)
-    RubyLsp::Requests::DocumentLink.new("file://fake/path/without_version.rb", document).run
+    js_content = File.read(File.join(TEST_FIXTURES_DIR, "rails_search_index.js"))
```
vinistock, line 28: I think we always automatically read every file under `test/fixtures`. Maybe we can change it to read only ruby files?

lib/ruby_lsp/requests/document_link.rb
```diff
@@ -63,6 +101,61 @@ def gem_paths
...
           end, T.nilable(T::Hash[String, T::Array[String]]))
         end
+
+        sig { returns(T.nilable(T::Hash[String, T::Hash[Symbol, String]])) }
+        def rails_documents
+          @rails_documents ||= T.let(begin
+            return {} unless RAILTIES_VERSION
+
+            $stderr.puts "Fetching Rails Documents..."
+            response = Net::HTTP.get_response(URI("#{RAILS_DOC_HOST}/v7.1.1/js/search_index.js"))
```
vinistock, line 111: Should the version here be `RAILTIES_VERSION`?

lib/ruby_lsp/requests/document_link.rb
```diff
@@ -63,6 +101,61 @@ def gem_paths
...
+
+            # If the version's doc is not found, e.g. Rails main, it'll be redirected
+            # In this case, we just fetch the latest doc
+            if response.code == "302"
```
vinistock, line 115: Instead of checking for the response of the http request, is there a way to determine through the gem's version that it's pointing to main?

vinistock (commented):
Two questions that came up while testing this
- Is there an option for placing the link on top of the hover documentation? These Rails methods have a lot of documentation and we end up having to scroll through a lot before finding the links
- Is there a way of disabling cmd click for document links only? Maybe some VS Code configuration? Right now, if you cmd click a method with links you jump to the browser and Sorbet jumps to the RBI file at the same time

lib/ruby_lsp/requests/document_link.rb
```diff
@@ -63,6 +101,61 @@ def gem_paths
             lookup
           end, T.nilable(T::Hash[String, T::Array[String]]))
         end
+
+        sig { returns(T.nilable(T::Hash[String, T::Hash[Symbol, String]])) }
+        def rails_documents
```
egiurleo, line 106: I personally don't think that this functionality belongs in the `DocumentLink` class. How would folks feel about creating a `RailsDocumentsClient` class or something along those lines that can wrap the calls to the Rails documentation and then expose them to the LSP via an API? It would make the code in this class cleaner and then we could even write separate tests for the client to make sure it has good coverage.

(cc: @vinistock would also love your opinion)

lib/ruby_lsp/requests/document_link.rb
```diff
@@ -114,6 +222,25 @@ def resolve_version(uri)
...
+        owner = doc[:owner]
+
+        return unless RAILS_DOC_PATHS_MAP.any? do |folder, patterns|
+          @file_dir.match?(folder) && T.must(owner).match?(patterns)
+        end
+
+        @links << LanguageServer::Protocol::Interface::DocumentLink.new(
+          range: range_from_syntax_tree_node(node),
+          target: "#{RAILS_DOC_HOST}/v#{RAILTIES_VERSION}/#{doc[:path]}",
+          tooltip: "Browse the Rails documentation for: #{name}"
```
egiurleo, line 241: How would you feel about adding `owner` somewhere in here? e.g. `Browse the Rails documentation for: #{owner}##{name}` or something? One thing that always confuses me about Rails is that sometimes different components can have methods with the same names, so it's not always obvious which ones you're looking at.

lib/ruby_lsp/requests/document_link.rb
egiurleo: AND maybe as we expand the functionality of document links we'll have more of these clients and they can share behavior?

egiurleo (commented):
This is really cool so far! I left some questions and a suggestion!

lib/ruby_lsp/requests/document_link.rb
st0012: I was testing something and accidentally committed it 🤦

lib/ruby_lsp/requests/base_request.rb
```diff
@@ -30,6 +30,25 @@ def range_from_syntax_tree_node(node)
           end: LanguageServer::Protocol::Interface::Position.new(line: loc.end_line - 1, character: loc.end_column),
         )
       end
+
+      sig { params(node: SyntaxTree::ConstPathRef).returns(String) }
+      def full_constant_name(node)
+        name = +node.constant.value
```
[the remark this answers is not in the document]
st0012: It returns a mutable duplication of the string: https://rubyapi.org/3.1/o/string#method-i-2B-40

lib/ruby_lsp/requests/document_link.rb
st0012: Sure, but I'd use `#{name} (#{owner})` as the format instead. Because many class methods are defined as instance methods under modules like `ClassMethods` and then extended to the class ([example](https://api.rubyonrails.org/v7.0.3.1/classes/ActiveRecord/Associations/ClassMethods.html#method-i-has_many)). This makes it hard to determine if we should use `.` or `#` to indicate the method type.

lib/ruby_lsp/requests/document_link.rb
```diff
@@ -83,6 +177,7 @@ def run
       end
 
       sig { override.params(node: SyntaxTree::Comment).void }
+
```
vinistock, line 180: Let's please remove this space between sig and method def.

test/requests/document_link_expectations_test.rb
```diff
@@ -20,20 +20,37 @@ def map_expectations(expectations)
...
+      @body = body
+    end
+  end
+
   def run_expectations(source)
     document = RubyLsp::Document.new(source)
-    RubyLsp::Requests::DocumentLink.new("file://fake/path/without_version.rb", document).run
+    js_content = File.read(File.join(TEST_FIXTURES_DIR, "rails_search_index.js"))
+    fake_response = FakeHTTPResponse.new("200", js_content)
+    T.unsafe(Net::HTTP).stub(:get_response, fake_response) do
```
vinistock, line 36: I think if you add `require "minitest/mock"` to `sorbet/tapioca/require.rb` and re-generate the RBI for minitest, you won't need the unsafe.

lib/ruby_lsp/requests/document_link.rb
```diff
@@ -28,6 +50,22 @@ class DocumentLink < BaseRequest
         T::Hash[String, String],
       )
 
+      RAILS_DOC_HOST = "https://api.rubyonrails.org"
+      RAILS_DOC_PATHS_MAP = T.let(
+        {
+          "controllers" => Regexp.union(/ActionController/, /AbstractController/),
```
vinistock, line 56: We should add `ActiveRecord` here too. A lot of queries happen in controllers.

lib/ruby_lsp/requests/document_link.rb
```diff
@@ -28,6 +50,22 @@ class DocumentLink < BaseRequest
...
+          "models" => Regexp.union(/ActiveRecord/, /ActiveModel/, /ActiveStorage/),
+          "config" => /ActionDispatch/,
+          "jobs" => /ActiveJob/,
```
vinistock, line 59: Let's add `ActiveRecord` here as well.

lib/ruby_lsp/requests/document_link.rb
```diff
@@ -28,6 +50,22 @@ class DocumentLink < BaseRequest
         T::Hash[String, String],
       )
 
+      RAILS_DOC_HOST = "https://api.rubyonrails.org"
+      RAILS_DOC_PATHS_MAP = T.let(
+        {
+          "controllers" => Regexp.union(/ActionController/, /AbstractController/),
+          "models" => Regexp.union(/ActiveRecord/, /ActiveModel/, /ActiveStorage/),
```
vinistock, line 57: We may want to add `ActionText`.

lib/ruby_lsp/requests/document_link.rb
vinistock: IMO, we should stick with `Owner#method`. It's a convention used everywhere in Ruby and the owner being at the back is a bit odd.

vinistock:
I tophatted this again and it's looking great, I'm really excited for it.

The one thing I think we need to find a decent solution for is the overlap of this feature with go to definition by Sorbet. Right now, CMD + click will perform both operations at the same time. You get redirected to the browser with the documentation and when you come back to the editor you're suddenly in the RBI file.

Two possible ideas, but we need to validate that the experience is decent
1. If it's possible, turn off CMD + click for document links
2. Instead of attaching the document link to the entire word, use only the first character. This will make it harder to hit the conflict, but may degrade the document links experience

vinistock (approved):

st0012:
Close in favour of https://github.com/Shopify/ruby-lsp/pull/313

PR #293, opened 2022-09-06
Make the diagnostic request parallel

vinistock:
### Motivation

Make the diagnostic request parallel and avoid referencing RuboCop classes in `server.rb`, given that it's an optional dependency.

### Implementation

Added `parallel: true`, moved a comment that was accidentally misplaced in `queue.rb` and create a custom error class similar to the one for `Formatting::Error`.

### Automated Tests

This is a refactor and the existing test for displaying RuboCop validation errors should continue to pass.

### DIFF
--- a/lib/ruby_lsp/server.rb
+++ b/lib/ruby_lsp/server.rb
@@ -207,7 +207,7 @@ module RubyLsp
       Handler::VOID
     end
 
-    on("textDocument/diagnostic") do |request|
+    on("textDocument/diagnostic", parallel: true) do |request|
       uri = request.dig(:params, :textDocument, :uri)
       response = store.cache_fetch(uri, :diagnostics) do |document|
         Requests::Diagnostics.new(uri, document).run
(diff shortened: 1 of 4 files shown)


lib/ruby_lsp/server.rb
```diff
@@ -207,15 +207,15 @@ module RubyLsp
...
+    on("textDocument/diagnostic", parallel: true) do |request|
       uri = request.dig(:params, :textDocument, :uri)
       response = store.cache_fetch(uri, :diagnostics) do |document|
         Requests::Diagnostics.new(uri, document).run
       end
 
       { kind: "full", items: response.map(&:to_lsp_diagnostic) } if response
     end.on_error do |error|
-      if error.is_a?(RuboCop::ValidationError)
+      if error.is_a?(Requests::Support::RuboCopRunner::ConfigurationError)
```
paracycle, line 218: This won't work as intended since if `RuboCop` is not available then `RuboCopRunner` won't be available either.

st0012 (approved):

Morriar (approved):

PR #295, opened 2022-09-07
Improve worker stability

vinistock:
### Motivation

We have been receiving reports of the LSP being unstable when performing edits. After debugging, it looks like the issue is that the thread worker is being killed when trying to cancel requests midway of processing.

I believe we should remove the ability to cancel requests while they are already being processed. We can keep request cancellation to remove items from the queue if processing has not started, but trying to stop the thread worker midway has proven difficult and flaky.

### Implementation

I split the changes in 2 commits
- Added a mutex synchronization block when writing to IO to guarantee that we won't try to concurrently perform IO
- Removed the functionality for cancelling requests midway. We will no longer try to raise on the thread and rescue

### Automated Tests

Removed tests that were related to cancelling requests as they were being processed.

### Manual Tests

Basically a smoke test on this branch.
2. Edit files and play around
3. Verify that the LSP does not get in a state of hanging. For example, if the thread worker was accidentally killed, the formatting pop up created by VS Code would never go away (it only closes when it receives a response)

### DIFF
--- a/lib/ruby_lsp/queue.rb
+++ b/lib/ruby_lsp/queue.rb
@@ -109,25 +109,27 @@ module RubyLsp
       ).void
     end
     def finalize_request(result, request)
-      error = result.error
-      if error
-        T.must(@handlers[request[:method]]).error_handler&.call(error, request)
+      @mutex.synchronize do
+        error = result.error
+        if error
+          T.must(@handlers[request[:method]]).error_handler&.call(error, request)
 
-        @writer.write(
-          id: request[:id],
-          error: {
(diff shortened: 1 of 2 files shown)


Morriar (approved):
That also means we renounce the ability to implement cancellation for our own requests right?

vinistock:
@Morriar It means that we can't cancel a request if it has already started. If it's sitting in the queue, we can still cancel it. We can also look into bringing this functionality back later, but for now, I couldn't figure out a way of making this stable.

KaanOzkan (approved):
:tophat:

PR #296, opened 2022-09-09
Prevent older versions of Rubocop from breaking LSP

### LINKED ISSUE #286: ruby-lsp fails with older versions of Rubocop
Hello! I was trying to run ruby-lsp with an old version of Rubocop, and it seems to fail due to the `--stderr` option:
```
Starting Ruby LSP...
[Error - 11:16:52 AM] Request textDocument/diagnostic failed.
Message: #<OptionParser::InvalidOption: --stderr>
Code: -32603
{"jsonrpc":"2.0","id":1,"method":"textDocument/diagnostic","params":{"textDocument":{"uri":"file:///…”}}}
```

I think it might be because the `--stderr` option was only added in [1.4.0](https://github.com/rubocop/rubocop/blob/master/CHANGELOG.md#140-2020-11-23), so it might be good to set a minimum version of Rubocop.

ruby-lsp v0.3.0, used the VS Code extension.

adisonlampert:
### Motivation

Closes #286 

### Implementation

- Successfully replicated the error with `rubocop` versions less than `1.4.0`.
- When testing, I noticed that `rubocop-shopify (~> 2.9)` needs `rubocop` to be `~> 1.33` so I went ahead and updated it in the gemfile for development.
- Added in logic to `rubocop_runner` so that the lsp would also run without `rubocop` if the version is less than `1.4.0`

### Tests

- Looked into writing a regression test to catch the error, but it ended up being difficult since it would involve running a test with a `rubocop` version that is different than the project's which meant the test would have to be written separately from the current `rubocop` testing.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -8,7 +8,7 @@ gem "debug", "~> 1.6", require: false
 gem "minitest", "~> 5.16"
 gem "minitest-reporters", "~> 1.5"
 gem "rake", "~> 13.0"
-gem "rubocop", ">= 1.0.0"
+gem "rubocop", "~> 1.33"
 gem "rubocop-shopify", "~> 2.10", require: false
 gem "rubocop-minitest", "~> 0.22.1", require: false
 gem "rubocop-rake", "~> 0.6.0", require: false
--- a/lib/ruby_lsp/requests/support/rubocop_runner.rb
+++ b/lib/ruby_lsp/requests/support/rubocop_runner.rb
@@ -7,6 +7,12 @@ rescue LoadError
   return
 end
 
+begin
+  gem("rubocop", ">= 1.4.0")
+rescue LoadError
+  raise StandardError, "Incompatible RuboCop version. Ruby LSP requires >= 1.4.0"
+end
+
 module RubyLsp
   module Requests
     module Support
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -81,7 +81,7 @@ GEM
       sorbet (>= 0.5.9204)
       sorbet-runtime (>= 0.5.9204)
       thor (>= 0.19.2)
-    syntax_tree (3.5.0)
+    syntax_tree (3.6.0)
       prettier_print
     tapioca (0.10.1)
       bundler (>= 1.17.3)
@@ -114,7 +114,7 @@ DEPENDENCIES
   minitest (~> 5.16)
   minitest-reporters (~> 1.5)
   rake (~> 13.0)
-  rubocop (>= 1.0.0)
+  rubocop (~> 1.33)
   rubocop-minitest (~> 0.22.1)
   rubocop-rake (~> 0.6.0)
   rubocop-shopify (~> 2.10)


lib/ruby_lsp/requests/support/rubocop_runner.rb
```diff
@@ -3,6 +3,7 @@
 
 begin
   require "rubocop"
+  return if Gem.loaded_specs["rubocop"].version < Gem::Version.create("1.4.0")
```
paracycle, line 6: Can we instead use `Kernel.gem` here to do this?

```suggestion
  return unless gem("rubocop", ">= 1.4.0")
  require "rubocop"
```

lib/ruby_lsp/requests/support/rubocop_runner.rb
```diff
@@ -14,17 +15,22 @@ module Support
       class RuboCopRunner < RuboCop::Runner
         extend T::Sig
 
-        class ConfigurationError < StandardError; end
+        class ConfigurationError < StandardError
+        end
```
paracycle, line 19: This and other layout changes below seem like they are coming from Syntax Tree formatting and not RuboCop. Can we remove those from this PR?

paracycle (commented):
Good catch

Gemfile
```diff
@@ -8,7 +8,7 @@ gem "debug", "~> 1.6", require: false
 gem "minitest", "~> 5.16"
 gem "minitest-reporters", "~> 1.5"
 gem "rake", "~> 13.0"
-gem "rubocop", ">= 1.0.0"
+gem "rubocop", "~> 1.33"
```
vinistock, line 11: Should we do `>= 1.33` instead? Not sure if we need to lock into `1.33`.

lib/ruby_lsp/requests/support/rubocop_runner.rb
adisonlampert: When I tried this, it errored:
```
can't activate rubocop (>= 1.4.0), already activated rubocop-1.3.0. Make sure all dependencies are added to Gemfile.
```
I think this is because `gem(gem_name, *requirements)` is used to activate a specific version of `gem_name` and not to check for it? [source code](https://github.com/rubygems/rubygems/blob/cbcbcbbef98514b96c65309bc933363afd5879a6/lib/rubygems/core_ext/kernel_gem.rb)
I looked around on RubyGems for a cleaner solution, but couldn't find anything aside from solutions that were similar (e.g., using `Gem::Requirement` with the rubocop version through `Gem.loaded_specs` and `satisfied_by?`).

vinistock:
Rebasing on the current main should fix the CI errors related to trailing commas.

lib/ruby_lsp/requests/support/rubocop_runner.rb
adisonlampert: Oh I see! Thanks for the clarification 😄 I updated the PR

PR #300, opened 2022-09-12
Remove rebornix.ruby and recommend Shopify's extension pack

wildmaples:
### Motivation

We shouldn't recommend an extension we're replacing xD  

### Implementation

Remove it from extensions.json and recommend our own extension pack

### Automated Tests

N/A 

### Manual Tests

N/A

### DIFF
--- a/.vscode/extensions.json
+++ b/.vscode/extensions.json
@@ -1,5 +1,5 @@
 {
   "recommendations": [
-    "rebornix.ruby"
+    "shopify.ruby-extensions-pack",
   ]
 }


Morriar (commented):
Should we also add https://github.com/Shopify/vscode-ruby-lsp?

wildmaples:
> Should we also add https://github.com/Shopify/vscode-ruby-lsp?

`Shopify.ruby-lsp` is part of `vscode-shopify-ruby`, so we don't need to add it separately.

PR #301, opened 2022-09-14
Include params and backtrace for error telemetry

### LINKED ISSUE #276: Include file name and line number in error telemetry
Sometimes, error messages are generic and don't include any information about the code that triggered the exception. E.g.: `cannot convert nil into Integer`.

This makes it very difficult to figure out where exactly in the LSP the error occurred. We should include the file path and the line number of where the exception occurred in telemetry to help us debug.

vinistock:
### Motivation

Closes #276 

Some errors are currently very difficult to trace because we don't know the backtrace and the parameters used in the request.

For example, if an error happens in `textDocument/didChange`, we don't know which edit caused the error (or where the error happened in our files).

### Implementation

Started including the parameters and the backtrace as optional string fields for our telemetry.

### Automated Tests

Added a new test for error telemetry and did some clean up of things that weren't being used in `queue_test.rb`.

### DIFF
--- a/lib/ruby_lsp/queue.rb
+++ b/lib/ruby_lsp/queue.rb
@@ -107,7 +107,7 @@ module RubyLsp
             id: request[:id],
             error: {
               code: LanguageServer::Protocol::Constant::ErrorCodes::INTERNAL_ERROR,
-              message: result.error.inspect,
+              message: error.inspect,
               data: request.to_json,
             },
           )
@@ -117,7 +117,7 @@ module RubyLsp
 
         request_time = result.request_time
         if request_time
-          @writer.write(method: "telemetry/event", params: telemetry_params(request, request_time, result.error))
+          @writer.write(method: "telemetry/event", params: telemetry_params(request, request_time, error))
         end
       end
     end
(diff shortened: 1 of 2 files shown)


st0012 (approved):
I don't want to complicate things, but since we're improving error reporting, do you consider reporting cause as well?

vinistock:
@st0012 what do you mean by cause?

st0012:
`Exception#cause` can return another `Exception`, which can has a cause too.
The chain of exceptions can be extracted [like this](https://github.com/getsentry/sentry-ruby/blob/d8724882b67c3107d062456e4fc16237aa5e46f9/sentry-ruby/lib/sentry/utils/exception_cause_chain.rb#L6-L17).

st0012:
To name a few:

1. When a library catches errors and throw their custom errors, like what we do [here](https://github.com/Shopify/ruby-lsp/blob/9e6ba1a60d5a7749c77d9d878b04a99927bff491/lib/ruby_lsp/requests/support/rubocop_runner.rb#L53-L54). In this case we don't care about the cause because the rescued error is specific. But some libraries do the same with `StandardError`, and then having the `cause` available will be important.
2. When there's issue in our rescuing logic:

    ```rb
    begin
      begin
        1/0
      rescue => StandardError
        oop_a_typo
      end
    rescue => e
      puts e.inspect #=> NoMethodError
      puts e.cause.inspect #=> ZeroDivisionError, which is the root issue
    end
    ```

    In this case, if we don't report `cause` too, we'll need to fix and release the rescuing logic first, and then wait for the original issue to occur again.

st0012:
They'll have different backtrace because when the `NoMethodError` is raised, it already lost the frames that triggered `ZeroDivisionError`

```
========Exception Reported=========
#<NameError: undefined local variable or method `oop_a_typo' for main:Object

    oop_a_typo
    ^^^^^^^^^^>
foo.rb:5:in `rescue in <main>'
foo.rb:2:in `<main>'
========Original Exception=========
#<ZeroDivisionError: divided by 0>
foo.rb:3:in `/'
foo.rb:3:in `<main>'
```

You can see that `NoMethodError`'s backtrace doesn't have anything about the original error.

PR #303, opened 2022-09-16
Add check valid visits rake task

vinistock:
### Motivation

To prevent something like #302 from happening again, I propose we add a new rake task that verifies all visit overrides are correct.

### Implementation

SyntaxTree already has a list of all valid visits, so we're just eager loading requests and checking against that list. I also added a message to `check_docs` to make sure that it prints something on success.

Finally, I added a `ci` command to dev.yml, so that we can run all checks locally without having to type them manually.

### DIFF
--- a/lib/ruby_lsp/requests/document_highlight.rb
+++ b/lib/ruby_lsp/requests/document_highlight.rb
@@ -43,7 +43,7 @@ module RubyLsp
         @highlights
       end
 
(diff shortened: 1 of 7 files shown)


vinistock:
Okay, changed the task to check that all visits are marked as overrides instead, based on https://github.com/Shopify/ruby-lsp/pull/302#discussion_r973285288.

st0012 (approved):

PR #309, opened 2022-09-22
Add GitHub action to create release

vinistock:
### Motivation

As part of our effort to simplify our releases, we should try to automate as many steps as possible. Creating GitHub releases through the UI can be fully automated with an action, so I propose we do it.

### Implementation

Removed CHANGELOG.md and its entry in the gemspec files.

The new action is fired whenever a new tag starting with the name `v` is pushed, which is done by Shipit for us. Then we use github-scripts to create a new release using the current owner and repository. The `${{ github.ref }}` is the tag that triggered the workflow (e.g.: `v0.1.2`), which we use as both the tag for the release and for the title. Finally, we use the auto-generated release notes to create the release body.

### Manual Tests

I really wish we could test this and make sure it works before committing to it. I have this idea, but I'd like to hear if anyone has a better one

1. Create a fake version tag starting with `v` and push it
2. Verify if this actually works
3. Delete the release
4. Delete the tag


dirceu:
> Delete the release
> Delete the tag

I don't know what are the implications of doing this; wouldn't Dependabot create PRs for the new releases? 

We do have the `ruby/debug` bugfix to get out... could we do a patch release and then, if there are any problems, do another patch release manually?

vinistock:
Dependabot creates the PRs based on new versions in rubygems and not tags or releases on GitHub, since you're not required to do those, you can just release to rubygems straight away.

The manual test idea is to create a fake tag for a version that was never pushed to rubygems just to test if the release is created. Deleting the tag and the release shouldn't do anything.

About `ruby/debug`, would we be able to use it for a test? It's an external repo, so I don't think we're using shipit (or control the GitHub actions).

egiurleo (approved):
Very nice! I support trying this.

dirceu (approved):

PR #313, opened 2022-09-29
Provide links to Rails DSL/constant document with hover request

st0012:
### Motivation

Implements #258 with `hover` instead of `documentLink` request due to conflicting `cmd+click` behaviour with `GoToDefinition`, which is not configurable.

### Implementation

- Added `Request::Hover` to handle hover requests.
- Added `Support::RailsDocumentClient` for providing Rails document urls based on the method/class/module name.
    - It fetches `search_index.js` (an artifact generated by `rdoc`) when first invoked and then memoizes the data set.
- When hovering over certain nodes, currently `Command` (DSL call) and `ConstPathRef` (`A::B` constant invocation), it does the following things:
    1. Fetching possible candidates from the memoized dataset.
    2. Filtering out some candidates based on the request's uri. 
        - For example, if it's a `root :something` call under `models/post.rb`, we don't return a link to `ActionDispatch::Routing::Mapper::Resources#root`, as it's unlikely to be the `root` method the user is using.
    3. Generating markdown link(s) like 

        ```markdown
        [Rails Document: `ActiveRecord::Associations::ClassMethods#belongs_to`](https://api.rubyonrails.org/v7.0.3.1/classes/ActiveRecord/Associations/ClassMethods.html#method-i-belongs_to)
        ```

### Automated Tests

- Under `test/expectations/hover/`
- A new case in `test/integration_test.rb`

### Additional Notes:

- It doesn't trigger `GoToDefinition` when markdown-rendered links in the hover modal.
- The data set and generated url's version will be determined by the `railties` version the user has in `Gemfile.lock`. 

### Demo

rails_document_link_hover


lib/ruby_lsp/requests/hover.rb
```diff
@@ -0,0 +1,99 @@
...
+          end
+        end
+
+        case matched
+        when SyntaxTree::Command, SyntaxTree::ConstPathRef
+          matched
+        when SyntaxTree::Node
+          find(matched, position)
+        end
+      end
```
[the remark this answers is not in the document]
kddnewton: Yeah big fan of doing this in syntax tree. I'd be super amenable to adding this. Only thing I would change is that this should say `while matched in SyntaxTree::Node` instead of recursing.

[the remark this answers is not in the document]
paracycle: @kddnewton I know that you want to keep the Syntax Tree light and thus didn't have a position to node map in place, but I wonder if clients could ask to have such a map when calling the `parse` method and Syntax Tree could start keeping track of such a thing to make operations like this to be constant time?

lib/ruby_lsp/requests/hover.rb
```diff
@@ -0,0 +1,109 @@
...
+      sig { override.params(node: SyntaxTree::FCall).void }
+      def visit_fcall(node)
+        return unless node == @target
+
+        message = node.value
+        generate_rails_document_link_hover(message.value, message)
+      end
+
+      sig { override.params(node: SyntaxTree::ConstPathRef).void }
+      def visit_const_path_ref(node)
```
vinistock, line 60: I understand we need the `find` technique for document highlight, but I'm not sure it's necessary for this one.

Instead of inspecting the AST twice (once during `find` and another one during `visit`), couldn't we do something like this?

```ruby
def visit_fcall(node)
  # We found an fcall, which is one of the nodes we're interested in. As long as its location covers the position,
  # we should be able to push the hover, no?
  location = node.location
  return unless (location.start_line - 1..location.end_line - 1).cover?(@position[:line]) && (location.start_column..location.end_column).cover?(@position[:character])
end
```
We can even put it in the `BaseRequest` given that it's common for requests to check if a node covers the position. E.g.:
```ruby
def visit_fcall(node)
  return unless node_covers_position?(node, @position)
end
```

lib/ruby_lsp/requests/hover.rb
```diff
@@ -0,0 +1,109 @@
...
+
+      private
+
+      sig { params(name: String, node: SyntaxTree::Node).void }
+      def generate_rails_document_link_hover(name, node)
+        urls = Support::RailsDocumentClient.generate_rails_document_urls(name, @file_dir)
+
+        unless urls.empty?
+          values = urls.map do |url|
+            "[Rails Document: `#{url[:name]}`](#{url[:url]})"
```
vinistock, line 75: NIT: Can we extract the `Rails Document` part so that it's not a part of the links? Also, what do you think about joining the different entries with pipes instead of new lines?

Example result
```
Rails documentation: SomeClass#method | OtherClass#other_method
```

vinistock (commented):
Tophat looks great. The experience is much nicer

vinistock:
We can probably remove the filtering heuristics now and just present all possible documents. I think it makes sense to simplify it. 👍

lib/ruby_lsp/requests/hover.rb
st0012: > Also, what do you think about joining the different entries with pipes instead of new lines?

Given that Rails components' namespaces usually aren't that short, I feel using `|` will make the line too long. Based on what I see, I think newline is still better.

Multiple links

lib/ruby_lsp/requests/base_request.rb
```diff
@@ -30,6 +30,51 @@ def range_from_syntax_tree_node(node)
...
+      end
+
+      sig do
+        params(
+          parent: SyntaxTree::Node,
+          target_nodes: T::Array[T.class_of(SyntaxTree::Node)],
+          position: Integer,
+        ).returns(T::Array[SyntaxTree::Node])
+      end
+      def locate_node_and_parent(parent, target_nodes, position)
```
st0012, line 60: @vinistock I've extracted the node locating logic here. We need to return both the matched and its parent because in `DocumentHighlight` we need the parent as well.

lib/ruby_lsp/requests/hover.rb
```diff
@@ -0,0 +1,74 @@
...
+
+      sig { params(document: Document, uri: String, position: Document::PositionShape).void }
+      def initialize(document, uri, position)
+        super(document)
+
+        @file_dir = T.let(Pathname.new(uri).dirname.to_s, String)
+        @hover = T.let(nil, T.nilable(LanguageServer::Protocol::Interface::Hover))
+        position = Document::Scanner.new(document.source).find_position(position)
+
+        target, _ = locate_node_and_parent(
```
st0012, line 33: @vinistock Now it's just one tree traversal and it's happen in `#initialize`. We don't even need the visitor methods anymore.

st0012: Actually, I decided to move this part to `#run` instead.

lib/ruby_lsp/requests/hover.rb
```diff
@@ -0,0 +1,74 @@
...
+          generate_rails_document_link_hover(constant_name, target)
+        end
+
+        @hover
+      end
+
+      private
+
+      sig { params(name: String, node: SyntaxTree::Node).void }
+      def generate_rails_document_link_hover(name, node)
```
paracycle, line 58: Is there are reason why this method is not directly communicating with its caller but is instead doing so via an ivar. I am assuming it is a left-over from the previous visitor implementation.

It would be much better to just return the `LanguageServer::Protocol::Interface::Hover` item that was constructed.

lib/ruby_lsp/requests/support/rails_document_client.rb
```diff
@@ -0,0 +1,117 @@
...
+
+              if response.code == "200"
+                process_search_index(response.body)
+              else
+                $stderr.puts("Response failed: #{response.inspect}")
+                nil
+              end
+            rescue StandardError => e
+              $stderr.puts("Exception occurred when fetching Rails document index: #{e.inspect}")
+            end, T.nilable(T::Hash[String, T::Array[T::Hash[Symbol, String]]]))
```
paracycle, line 49: I still think this huge `begin`/`end` block inside a `T.let` is very unreadable, and this method mixes up the building of the rails documents with memoization of it.

So, I will still suggest that the building is split out to a specific builder method.

lib/ruby_lsp/requests/base_request.rb
st0012, line 60: @paracycle @kddnewton @vinistock Following the earlier discussion around node-locating API: In the current use cases we have in `ruby-lsp`, I'd expect that API to:

- Locate certain types of nodes with a given position.
- Return both the node and its parent. Having the parent allows us to locate nodes (like `(vcall (ident "foo"))`) based on a child node (like `(ident "foo")`). We use this in `documentHighlight`.

paracycle (approved):
Thanks for the latest changes. This looks good to me!

lib/ruby_lsp/requests/hover.rb
```diff
@@ -0,0 +1,72 @@
...
+
+      private
+
+      sig do
+        params(name: String, node: SyntaxTree::Node).returns(T.nilable(LanguageServer::Protocol::Interface::Hover))
+      end
+      def generate_rails_document_link_hover(name, node)
+        urls = Support::RailsDocumentClient.generate_rails_document_urls(name)
+
+        unless urls.empty?
```
egiurleo, line 59: I personally would prefer an early return so we don't have to indent the remaining lines of the method but I'm fine merging without.

lib/ruby_lsp/requests/support/rails_document_client.rb
```diff
@@ -0,0 +1,123 @@
...
+
+            # An entry looks like this:
+            #
+            # ["belongs_to",                                                              # method or module/class
+            #  "ActiveRecord::Associations::ClassMethods",                                # method owner
+            #  "classes/ActiveRecord/Associations/ClassMethods.html#method-i-belongs_to", # path to the document
+            #  "(name, scope = nil, **options)",                                          # method's parameters
+            #  "<p>Specifies a one-to-one association with another class..."]             # document preview
+            #
+            info.each do |ary|
```
egiurleo, line 72: I should have written this on the previous PR as well but thank you for using comments to make this code so easy to understand. The documentation is so valuable!

lib/ruby_lsp/requests/support/rails_document_client.rb
```diff
@@ -0,0 +1,123 @@
...
+        RAILTIES_VERSION = T.let(
+          [*::Gem::Specification.default_stubs, *::Gem::Specification.stubs].find do |s|
+            s.name == "railties"
+          end&.version&.to_s, T.nilable(String)
+        )
+
+        class << self
+          extend T::Sig
+          sig { returns(T.nilable(T::Hash[String, T::Array[T::Hash[Symbol, String]]])) }
+          def search_index
```
egiurleo, line 28: Question: does this have to be a public method? Also non-blocking, I'm just curious what you have in mind.

egiurleo (approved):
🎉 This looks so good!

lib/ruby_lsp/requests/hover.rb
```diff
@@ -0,0 +1,72 @@
...
+      end
+
+      sig { override.returns(T.nilable(LanguageServer::Protocol::Interface::Hover)) }
+      def run
+        target, _ = locate_node_and_parent(
+          T.must(@document.tree), [SyntaxTree::Command, SyntaxTree::FCall, SyntaxTree::ConstPathRef], @position
+        )
+
+        case target
+        when SyntaxTree::Command
```
vinistock, line 39: The reason I had suggested checking the node locations in the visits is because it would conform to the style we're using in most requests and because the code is a bit easier to read.

Is there a reason why we can't use the visitor pattern here? What's the advantage of using this locate node strategy?

st0012: Because I agree with what @paracycle said [here](https://github.com/Shopify/ruby-lsp/pull/313#discussion_r984056705), it's more clear to just return the result than mutate and return the `@hover` ivar.

In other requests, we go through nodes to collect data (highlights, tokens...etc.), push them to a state, and return the final state. So that pattern makes sense.

But in this case, we don't have a bunch of data to collect , nor do we have multiple nodes to check. We only care about one node, which we can locate right away. This meant having that `@hover` state becomes unnecessary. Also, based on how other requests work, seeing a visitor pattern would give me an impression that we may match multiple nodes, which it's not the case. 

Therefore, I actually think the current implementation is easier to read and understand.

lib/ruby_lsp/requests/support/rails_document_client.rb
```diff
@@ -0,0 +1,123 @@
...
+
+            # An entry looks like this:
+            #
+            # ["belongs_to",                                                              # method or module/class
+            #  "ActiveRecord::Associations::ClassMethods",                                # method owner
+            #  "classes/ActiveRecord/Associations/ClassMethods.html#method-i-belongs_to", # path to the document
+            #  "(name, scope = nil, **options)",                                          # method's parameters
+            #  "<p>Specifies a one-to-one association with another class..."]             # document preview
+            #
+            info.each do |ary|
```
paracycle, line 95: Oh, one small thing that's bugging me about this: I think we can make this code much more readable by expanding the array values here in the block params:
```suggestion
            info.each do |method_or_class, method_owner, doc_path, _, doc_preview|
```
which will allow us to get rid of the adhoc assignments with indices below inside the block.

st0012: Great catch 👍

vinistock (approved):

PR #318, opened 2022-10-03
Add release.yml

st0012:
### Motivation

To make GH smarter at generating release notes.

### Implementation

Replicate [`Tapioca`'s `release.yml`](https://github.com/Shopify/tapioca/blob/main/.github/release.yml#L2) with slightly different label names.

### Automated Tests

N/A

### Manual Tests

Need to wait for a couple of new PRs to be merged before we can test it.

### DIFF
--- a/.github/release.yml
+++ b/.github/release.yml
@@ -1,0 +1,22 @@
+# release.yml
+
+changelog:
+  exclude:
+    labels:
+      - dependencies
+      - chore
+    authors:
+      - dependabot
+  categories:
+    - title: 🚧 Breaking Changes
+      labels:
+        - breaking-change
+    - title: ✨ Enhancements
+      labels:
+        - enhancement
+    - title: 🐛 Bug Fixes
+      labels:
+        - bug
+    - title: 🛠 Other Changes
+      labels:
+        - "*"


.github/release.yml
```diff
@@ -0,0 +1,22 @@
...
+  exclude:
+    labels:
+      - dependencies
+      - chore
+    authors:
+      - dependabot
+  categories:
+    - title: 🚧 Breaking Changes
+      labels:
+        - breaking-change
```
vinistock, line 13: I think we use the unstable [label](https://github.com/Shopify/ruby-lsp/labels) for this one.

vinistock (approved):

.github/release.yml
st0012: I think `unstable` and `breaking-change` means different things? According to the new release strategy, `unstable` means the feature will be guarded behind feature flags and are hidden by default. So it should actually be safe to release.

Breaking changes, on the other hand, are expected to break users usages or expectations. This may be rare for this project though and I can remove this part for now to avoid confusion.

.github/release.yml
```diff
@@ -0,0 +1,22 @@
...
+    - title: ✨ Enhancements
+      labels:
+        - enhancement
+    - title: 🐛 Bug Fixes
+      labels:
+        - bug
```
paracycle, line 19: Nitpick: This should be `bugfix` and not `bug`. I think `bug` is appropriate for issues and `bugfix` is appropriate for PRs.

PR #323, opened 2022-10-11
Make parsing lazy

vinistock:
### Motivation

Currently, the server eagerly parses files on every `textDocument/didChange` notification. Because we receive those on every keypress, we're parsing the file an excessive amount of times. This is super expensive, especially for large files, where parsing can take up to a few hundred milliseconds.

Ideally, we would only parse the file after VS Code's internal debounce expires, so that we parse an entire group of edits at once.

### Implementation

We only receive feature requests (like folding ranges) when VS Code's debounce expires. By lazily parsing documents, we can make sure we do it only once for a sequence of edits, rather than eagerly parsing on each one.

The outcome would be something like this
```ruby
# If we were to type
Rails.logger.info("Some information")

# Before
# We would parse the file 37 times (the number of characters/key presses to type this line)
# In the middle of typing, the debounce could also expire and send feature requests in the middle of typing, further
# backlogging the server

# After
# It's impossible to tell exactly how many times we will parse, since it depends on how fast the user is typing.
# However, it's definitely not going to be 37 times, since it will only happen when the debounce expires and the first
# feature request is sent
```

- Instead of parsing the document on every edit, moved the parsing logic into a `Document#parse` method
- Started parsing documents when requests are instantiated at `BaseRequest`
- Added early returns to requests to prevent trying to execute them on unparsed documents (when there is a syntax error)

### Tradeoff

The tradeoff made here is that requests can no longer always assume they will have a parsed document and must perform an early return if they can't operate on files with syntax errors.

### Automated Tests

The existing tests should continue to pass.


KaanOzkan (approved):

lib/ruby_lsp/document.rb
```diff
@@ -81,7 +90,6 @@ def update_parsable_source(edits)
       # If the new edits caused a syntax error, make all edits blank spaces and line breaks to adjust the line and
       # column numbers. This is attempt to make the document parsable while partial edits are being applied
       edits.each do |edit|
-        @syntax_error_edits << edit
         next if edit[:text].empty? # skip deletions, since they may have caused the syntax error
 
         apply_edit(@parsable_source, edit[:range], edit[:text].gsub(/[^\r\n]/, " "))
```
adisonlampert, line 95: What is the difference between `@source` and `@parsable_source`/What is `@parsable_source` being used for here?

adisonlampert (approved):

PR #328, opened 2022-10-17
Test and handle the situation where RailsDocumentClient#search_index being nil

### LINKED ISSUE #327: Hover breaking
The hover request is breaking with `undefined method `[]' for nil:NilClass` at [this part of the code](https://github.com/Shopify/ruby-lsp/blob/3e806a4c453fe91036b1ae8d44424378f872f19f/lib/ruby_lsp/requests/support/rails_document_client.rb#L31).

We declared the `search_index` correctly as possibly being `nil`, but we're not handling that possibility in the code and incorrectly masking the problem with a `T.must`.

st0012:
### Motivation

Fixes #327

### Implementation

Test and handle the case where `search_Index` being `nil`.

### Automated Tests

Added

### Manual Tests

- Comment out `gem "railties" in `Gemfile`
- Run `Reload Window` to restart the LSP
- Open the `OUTPUT` tab 
- Hover some DSL calls, like `expectations_tests` of `expectations_tests RubyLsp::Requests::Hover, "hover"`
    - On `main`, this should produce 
        ```
        Request failed: #<NoMethodError: undefined method `[]' for nil:NilClass
        
                    docs = T.must(search_index)[name]
                                               ^^^^^^> (-32603).
        ```
    - On this branch, it should not cause any error

### DIFF
--- a/lib/ruby_lsp/requests/support/rails_document_client.rb
+++ b/lib/ruby_lsp/requests/support/rails_document_client.rb
@@ -28,7 +28,7 @@ module RubyLsp
             params(name: String).returns(T::Array[String])
           end
           def generate_rails_document_urls(name)
-            docs = T.must(search_index)[name]
+            docs = search_index&.fetch(name, nil)
 
             return [] unless docs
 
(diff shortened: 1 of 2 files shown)


vinistock:
The [build](https://github.com/Shopify/ruby-lsp/actions/runs/3268256714/jobs/5374448037) seems to be flaky after the new stub. Can you please make sure the class is loaded for the test?

PR #331, opened 2022-10-18
Autoload RailsDocumentClient

st0012:
The build became flaky after https://github.com/Shopify/ruby-lsp/pull/328, which is likely caused by `RailsDocumentClient` now being autoloadable.


vinistock (commented):
I think we can remove the [require in hover](https://github.com/Shopify/ruby-lsp/blob/b531bbfb6b32eb6b65cc2889ec48f839818e3d5f/lib/ruby_lsp/requests/hover.rb#L4) since it's now autoloaded.

PR #333, opened 2022-10-18
Add extra test cases

vinistock:
### Motivation

Add some new fixtures with some more uncommonly found syntax to make sure our requests can handle them.

### DIFF
--- a/test/fixtures/nested_block_arguments.rb
+++ b/test/fixtures/nested_block_arguments.rb
@@ -1,0 +1,2 @@
+[[1, [2, 3]]].each do |one, (two, (three, four))|
+end
--- a/test/fixtures/aref_call_aref_assign.rb
+++ b/test/fixtures/aref_call_aref_assign.rb
@@ -1,0 +1,1 @@
+object["attributes"].find { |a| a["field"] == "expected" }["value"] = "changed"
--- a/test/fixtures/aref_on_method_call.rb
+++ b/test/fixtures/aref_on_method_call.rb
@@ -1,0 +1,1 @@
+object.attributes["name"] = "something"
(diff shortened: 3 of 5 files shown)


test/fixtures/nested_block_arguments.rb
```diff
@@ -0,0 +1,2 @@
+[[1, [2, 3]]].each do |one, (two, three)|
```
st0012, line 1: Isn't this covered by `test/fixtures/double_nested_arguments.rb` as well?

st0012 (commented):
Not saying we **should** do that, but if we want to find some edge cases to cover, [Ruby tricks snippets](https://github.com/tric) feels like a good source 😂  

[Example](https://github.com/tric/trick2022/blob/master/03-mame/entry.rb):

```rb
2022;"#

.chars} {puts'TRICK+2022'
        \  {         ;
#';$><< b 
   ?!
};#{s=' ' #           0 
! s[0]? (   b=$<.read  ;'
}            ub(      ,''}
';a=            ''<<32
b.lines      {puts( ?.. *(
b.size)   .gsub(/./) {
b.sub!(        /^#$`\K(\S)
             /x,a)?$1:a }
               .rstrip)}):
      ( [ 12,1,12,11].
cycle { | i | t = ( s *
10<<
10)*
10+            %(\e[A)*
10
10.             times{t[i*
    _1  ] =  'TRICK+2022'[
    _1  ]     };$><<t
            sleep 1})
               }"
```

test/fixtures/nested_block_arguments.rb
vinistock: You're right, it is. I was capturing a few interesting cases and added them sequentially as I found them.

vinistock:
@st0012 that is some unholy piece of code. I'm hoping no one is adding code like that to their codebases 😂. That said, those trick snippets might contain some interesting examples that we can test.

Morriar (approved):
Very nice! ❤️

PR #336, opened 2022-10-19
Use DefaultReporter to get clearer failure reports

st0012:
### Motivation

The `SpecReporter` prints failures as they happen instead of all at once at the bottom. This means we need to scroll up to see test failures very often, which is annoying.

### Implementation

Just use a `DefaultReporter` instead. And now failures will be shown at the bottom :-)

DefaultReporter

### DIFF
--- a/test/test_helper.rb
+++ b/test/test_helper.rb
@@ -9,7 +9,7 @@ require "minitest/reporters"
 require "tempfile"
 require "debug"
 
-Minitest::Reporters.use!(Minitest::Reporters::SpecReporter.new(color: true))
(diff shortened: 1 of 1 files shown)


test/test_helper.rb
```diff
@@ -9,7 +9,7 @@
 require "tempfile"
 require "debug"
 
-Minitest::Reporters.use!(Minitest::Reporters::SpecReporter.new(color: true))
+Minitest::Reporters.use!(Minitest::Reporters::DefaultReporter.new(color: true))
```
vinistock, line 12: Do we even need the minitest-reporters gem to use the default one? Or is it because minitest comes without colour by default?

vinistock (approved):

PR #352, opened 2022-11-07
Ignore errors when trying to parse our parsable source

vinistock:
### Motivation

Most of the errors we are getting in telemetry are related to parsing the fake source we keep around while there are syntax errors. However, maintaining that parsable source is just a best effort to keep features working while developers are typing. If we fail to do so, features come back as soon as the syntax is valid again. Therefore, we should prevent any errors from trying to parse the fake source from bubbling up to our telemetry (or worse breaking the LSP).

### Implementation

Wrapped the parse invocation in a standard error rescue.


KaanOzkan (approved):

paracycle (approved):
I think I kept getting bit by this a lot, so hopefully this makes it better for me! Thanks

PR #358, opened 2022-11-11
Add range semantic highlighting

vinistock:
### Motivation

There are two semantic highlighting requests: full and range. When both are available, VS Code will use the range request on large files to compute semantic tokens only for the visible portion of the document and only uses the full request for highlighting the minimap.

Having the range request available is good from a performance standpoint because it returns fewer tokens to VS Code, which means they can be processed faster. The specification recommends the request to reduce flickering.

### Implementation

- Added an optional range parameter to semantic highlighting
- Added the "new" request to server.rb
- Extract the inlay hint logic of verifying if a node is visible to `BaseRequest`
- Started checking if nodes are visible on all semantic highlighting visits, so that we skip our functionality if they are not visible

### Automated Tests

Added new tests.


lib/ruby_lsp/server.rb
```diff
@@ -38,10 +38,8 @@ module RubyLsp
             token_types: Requests::SemanticHighlighting::TOKEN_TYPES.keys,
             token_modifiers: Requests::SemanticHighlighting::TOKEN_MODIFIERS.keys,
           ),
-          range: false,
-          full: {
-            delta: true,
-          },
+          range: true,
+          full: { delta: false },
```
vinistock, line 42: Note: we never actually supported delta requests, so I'm not sure why this was `true`.

lib/ruby_lsp/requests/semantic_highlighting.rb
```diff
@@ -105,6 +112,8 @@ def run
 
       sig { override.params(node: SyntaxTree::Call).void }
       def visit_call(node)
+        return super unless visible?(node, @range)
```
paracycle, line 115: This is a lot of repetition and ends up being brittle. If someone were to add a new method to visit and they forget to add this check, this would break.

I think we can do this for each node by overriding the `visit_child_nodes` which each visitor method is already aliased to.

I guess:
```ruby
def visit_child_nodes(node)
  return unless visible?(node, @range)

  super
end
```
should do the same thing?

paracycle (approved):

lib/ruby_lsp/requests/semantic_highlighting.rb
paracycle: Sorry, I'd meant to comment, not approve 😅

paracycle (approved):
TBH, I still don't like the manual checks, but let's merge this and then decide on how to do it better.

lib/ruby_lsp/requests/semantic_highlighting.rb
```diff
@@ -115,31 +124,41 @@ def visit_call(node)
 
       sig { override.params(node: SyntaxTree::Command).void }
       def visit_command(node)
+        return super unless visible?(node, @range)
```
st0012, line 127: I feel `visible?` should just read `@range` directly and only take `node`, like `visible?(node)`. 

Passing an instance variable to a method that's accessible to it seems to be unnecessary, at least in this case.

lib/ruby_lsp/requests/base_request.rb
```diff
@@ -79,6 +79,15 @@ def locate_node_and_parent(parent, target_nodes, position)
...
         end
       end
+
+      sig { params(node: T.nilable(SyntaxTree::Node), range: T.nilable(T::Range[Integer])).returns(T::Boolean) }
+      def visible?(node, range)
+        return true if range.nil?
+        return false if node.nil?
+
+        loc = node.location
+        range.cover?(loc.start_line - 1) && range.cover?(loc.end_line - 1)
```
st0012, line 89: I think something like `SyntaxTree::Location#in?(range)` for this would be handy 🤔

test/expectations/code_actions/def_bad_formatting.exp
```diff
@@ -4,126 +4,6 @@
         "end": 4
     },
     "result": [
-        {
```
st0012, line 7: Why are these results removed?

lib/ruby_lsp/requests/semantic_highlighting.rb
vinistock: I agree, but unfortunately if we do that we'll be forced to declare `@range` in `BaseRequest` because Sorbet complains that it's undeclared. Since not every request depends on a range, I ended up just passing it via arguments.

st0012: I see 👍

st0012 (approved):

PR #360, opened 2022-11-11
Take client encoding into consideration when applying text edits

vinistock:
### Motivation

Closes https://github.com/Shopify/ruby-lsp/issues/1494

VS Code (and possibly other clients) [use UTF-16 to represent internal strings because of NodeJS](https://github.com/microsoft/vscode/issues/20170). In UTF-16, 4 byte unicode characters (like emojis) have length 2, whereas in UTF-8 they have length 1.

When editing a line that includes an emoji, this means that VS Code will send us the `character` part of the position offset by +1 for each 4 bytes unicode character, which messes up our internal document state.

After doing some research, it appears that LSP clients broadcast their encodings to the servers so that they can adjust `character` positions for when there are 4 bytes characters. In fact, Rust Analyzer does exactly this, by checking UTF-16 lengths when the client is using that encoding.

### Implementation

I split the fix by commit.
1. Started saving the encoding coming from the client capabilities in the global store
2. Started passing the encoding to `Document::Scanner` so that we have that information accessible
3. Started making `character` adjustments when the encoding is UTF-16. The adjustment is simply to subtract 1 for each 4 byte character we find

### Automated Tests

Added a test including emojis.

### Manual Tests

**Check the problem in main first**
1. In the main branch, start the LSP
2. Open any Ruby file with emojis
3. Make edits on the same line and around the emoji
4. Verify that the Ruby LSP's internal state gets completely messes up (you should see incorrectly reported syntax errors, non functioning undo...)

**On this branch**
1. Verify that doing the same as above does not error and all LSP functionality continues to work with emojis present

### DIFF
--- a/lib/ruby_lsp/document.rb
+++ b/lib/ruby_lsp/document.rb
@@ -18,10 +18,11 @@ module RubyLsp
     sig { returns(T::Array[EditShape]) }
     attr_reader :syntax_error_edits
 
-    sig { params(source: String).void }
-    def initialize(source)
+    sig { params(source: String, encoding: String).void }
+    def initialize(source, encoding = "utf-8")
       @cache = T.let({}, T::Hash[Symbol, T.untyped])
       @syntax_error_edits = T.let([], T::Array[EditShape])
+      @encoding = T.let(encoding, String)
       @source = T.let(source, String)
       @parsable_source = T.let(source.dup, String)
       @unparsed_edits = T.let([], T::Array[EditShape])
@@ -83,6 +84,11 @@ module RubyLsp
       !@tree.nil?
     end
 
+    sig { returns(Scanner) }
+    def create_scanner
+      Scanner.new(@source, @encoding)
+    end
+
     private
 
     sig { params(edits: T::Array[EditShape]).void }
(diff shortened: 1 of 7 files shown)


lib/ruby_lsp/document.rb
```diff
@@ -113,11 +119,12 @@ def apply_edit(source, range, text)
...
-      def initialize(source)
+      sig { params(source: String, encoding: String).void }
+      def initialize(source, encoding)
         @current_line = T.let(0, Integer)
         @pos = T.let(0, Integer)
         @source = source
+        @encoding = encoding
       end
 
       sig { params(position: PositionShape).returns(Integer) }
```
st0012, line 130: It's not added in this PR, but I was confused by `find_position` taking a `position` argument and returning something that's also a position but with a different type (and perhaps meaning as well). Is it possible to give it a more clear name, like `find_char_position`?

lib/ruby_lsp/document.rb
```diff
@@ -83,6 +84,11 @@ def parsed?
       !@tree.nil?
     end
 
+    sig { returns(Scanner) }
+    def scanner
```
st0012, line 88: Nitpick: Since it initialises a new `Scanner` every time it's called, may be add a verb prefix, like `create_scanner`?
Otherwise, based on its usages I'd assume it's an attribute of the document.

lib/ruby_lsp/document.rb
vinistock: Renamed and added a comment for the method.

egiurleo (approved):

PR #368, opened 2022-11-25
Completely dettach IO from parallel request execution

### LINKED ISSUE #378: Refactor Queue, Handler and Server to decouple IO from request processing
Currently, responsibilities are leaking between `Queue`, `Handler` and `Server` and the structure is growing in complexity. We should completely decoupled running a request from IO and centralize responsibilities, so that introducing parallelism is easier.

vinistock:
### Motivation

Closes #378

Our request `Handler` class has leaked some responsibility over to the queue and server files. Both request definition and IO are now spread and that makes it difficult to guarantee that no IO will occur in the main request path.

This refactor tries to centralize all request running logic into an `Executor` class while leaving all queue and IO managing to the `Handler`. The only part that leaks into the `Handler` is defining which requests must be run sequentially inside the main process (typically client notifications and life cycle requests).

#### Before

- All requests are implemented in `server.rb`
- `Handler` and `Queue` work together to decide how to execute and finalize requests, as well as managing the request queue
- A bit difficult to display send arbitrary notifications to the client. We currently use the `on_error` configuration block in `server.rb` to show error message notifications, but we are unable to send more than one notification to the client (e.g.: we wouldn't be able to clear diagnostics for a file and display a dialogue in the same request)

#### After

- The `Executor` class knows how to run all requests (except for requests that need access to the main process - either the queue or the thread worker). It is basically a big case statement that matches the feature to a method containing the implementation
- In the `Executor`, every request returns a response and all notifications (which should not return a response) return `Void`
- In addition to request responses, we may want to display notifications to the user. For that, requests can set the `@notifications` variable, which is an array of notifications to be returned to the client. For example, this allows us to return a response, clear diagnostics and show a dialogue to the user all within the same request
- The `Handler` class manages the queue and decides which requests are run sequentially or in parallel. All IO happens here, which should make parallelism a bit easier. We can run requests in separate processes, get the response back and then write to IO

### What should reviewers focus on

- Do you think this new structure is cleaner?
- Is it easier to understand?
- How can we better document it in the source code?
- Do you have a suggestion on how to structure it differently?

### Manual tests

Please, take the time to play around with the LSP on this branch to see if you can make it break. I tested on a few projects and things seem to be fine.


lib/ruby_lsp/executor.rb
```diff
@@ -0,0 +1,343 @@
+# typed: strict
+# frozen_string_literal: true
+
+module RubyLsp
+  # This class dispatches a request execution to the right request class. No IO should happen anywhere here!
+  class Executor
```
vinistock, line 6: Should this be named `Server` instead?

lib/ruby_lsp/handler.rb
```diff
@@ -1,118 +1,157 @@
...
-      @handlers = T.let({}, T::Hash[String, RequestHandler])
       @store = T.let(Store.new, Store)
-      @queue = T.let(Queue.new(@writer, @handlers), Queue)
+
+      # The job queue is the actual list of requests we have to process
+      @job_queue = T.let(Thread::Queue.new, Thread::Queue)
+      # The jobs hash is just a way of keeping a handle to jobs based on the request ID, so we can cancel them
+      @jobs = T.let({}, T::Hash[T.any(String, Integer), Job])
+      @mutex = T.let(Mutex.new, Mutex)
+      @worker = T.let(new_worker, Thread)
```
shopify-aiden, line 23: Does the worker get utilized at all?

lib/ruby_lsp/executor.rb
shopify-aiden: From my domain experience `Executor` makes sense for this class. With the functionality change of `Handler` I could see that being updated to `Server` as it feels like that's what it semantically represents now.

shopify-aiden (commented):
From someone who is unfamiliar with the history of this project, this change seems to make a lot of sense. 

One thing that I do want to ask, is if you'd thought at all about this ticket during the refactor: https://github.com/Shopify/ruby-lsp/issues/443

It feels like they align quite well and at the very least would be worth documenting initial thoughts on how plugins could be added with the new code layout. 

Otherwise it's looking good!

lib/ruby_lsp/handler.rb
vinistock: It's only used during `shutdown`. We close the `Thread::Queue`, which allows the worker thread to finish its infinite loop and then we do `@worker.join` just to wait until it's finalized.

vinistock:
This refactor doesn't take much of the plugin mechanisms into consideration, but I don't think it makes it harder either. The way I'm currently thinking about it, we'll need to allow plugins to define middleware and overrides for requests and those will be loaded at boot.

The plugins shouldn't interact with global store, queues and things like that, they should basically just have access to what the request implementation already has (generally the document and some extra parameters like position).

shopify-aiden (approved):
Awesome work, this is really easy to follow now

lib/ruby_lsp/executor.rb
```diff
@@ -0,0 +1,343 @@
...
+    extend T::Sig
+
+    sig { params(store: Store).void }
+    def initialize(store)
+      # Requests that mutate the store must be run sequentially! Parallel requests only receive a temporary copy of the
+      # store
+      @store = store
+      @notifications = T.let(nil, T.nilable(T::Array[Notification]))
```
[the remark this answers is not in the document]
st0012: Given it's one empty array per request and there's no indication array allocation is causing issues to the performance here, I feel it's an good trade-off to make the code more straightforward.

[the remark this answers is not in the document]
vinistock: Changed it to an array 👍

lib/ruby_lsp/executor.rb
vinistock: Renamed it to `Server` 👍.

lib/ruby_lsp/server.rb
```diff
@@ -1,250 +1,158 @@
...
+    # Finalize a Queue::Result. All IO operations should happen here to avoid any issues with cancelling requests
+    sig { params(result: Result, request: T::Hash[Symbol, T.untyped]).void }
+    def finalize_request(result, request)
+      @mutex.synchronize do
+        error = result.error
+        response = result.response
+
+        # If the response include any notifications, go through them and publish each one
+        notifications = result.notifications
+        notifications.each { |n| @writer.write(method: n.message, params: n.params) } if notifications.any?
```
st0012, line 106: I feel the `any?` check is not necessary? If the `notifications` is empty, simply nothing would happen.


```suggestion
        result.notifications.each { |n| @writer.write(method: n.message, params: n.params) }
```

lib/ruby_lsp/server.rb
```diff
@@ -1,250 +1,158 @@
...
-      position = request.dig(:params, :position)
-      character = request.dig(:params, :ch)
+      if error
+        params[:errorClass] = error.class.name
+        params[:errorMessage] = error.message
 
-      Requests::OnTypeFormatting.new(store.get(uri), position, character).run
-    end
+        log_params = request[:params]&.reject { |k, _| k == :textDocument }
+        params[:params] = log_params.to_json if log_params&.any?
```
st0012, line 148: nitpick: I feel it's clearer if we can avoid lonely operators in 2 consecutive lines with an if statement.

We can keep an if statement but eliminate the lonely operators. WDYT?

```suggestion
        
        log_params = request[:params]
        if log_params
          params[:params] = log_params.reject { |k, _| k == :textDocument }.to_json
        end
```

st0012 (approved):
LGTM. Just 2 nitpicks 🙂

PR #371, opened 2022-12-01
Add semantic highlighting for sorbet methods

TobiasBales:
### Motivation

This adds semantic highlighting to "top level" sorbet methods (basically everything that is not used in sigs like `params`/`returns`/`void`).
This allows us to distinctly highlight type/sorbet related pieces of code.

### Implementation

We ended up moving the decision logic into a separate class `RubyLsp::Requests::Support::Sorbet` with one public method `annotation?` that takes either a `SyntaxTree::Call`, `SyntaxTree::FCall` or `SyntaxTree::VCall` and an argument indicating if a block was passed (since that is determined on a parent node of those which would have made the handling a lot more complicated).
This is used in all relevant calls to check if it is a sorbet annotations and if so we add the correct annotations.

### Automated Tests

There are positive and negative cases for the relevant positive and negative cases.

### Manual Tests

You can put
```json
  "editor.semanticTokenColorCustomizations": {
    "[Default Dark+]": {
      "rules": {
        "type": {
          "foreground": "#efec08",
          "fontStyle": "italic"
        }
      }
    }
  }
```

to your `settings.json` and sorbet signatures should be highlighted in yellow and be italic

### DIFF
--- a/lib/ruby_lsp/requests/support/annotation.rb
+++ b/lib/ruby_lsp/requests/support/annotation.rb
@@ -1,0 +1,46 @@
+# typed: strict
+# frozen_string_literal: true
+
+module RubyLsp
+  module Requests
+    module Support
+      class Annotation
+        extend T::Sig
+        sig do
+          params(
+            arity: T.any(Integer, T::Range[Integer]),
+            receiver: T::Boolean,
+          ).void
+        end
+        def initialize(arity:, receiver: false)
+          @arity = arity
+          @receiver = receiver
+        end
+
+        sig { returns(T.any(Integer, T::Range[Integer])) }
+        attr_reader :arity
+
+        sig { returns(T::Boolean) }
+        attr_reader :receiver
+
+        sig { params(arity: T.any(T::Range[Integer], Integer)).returns(T::Boolean) }
+        def supports_arity?(arity)
+          if @arity.is_a?(Integer)
+            @arity == arity
+          elsif @arity.is_a?(Range)
+            @arity.cover?(arity)
+          else
+            T.absurd(@arity)
+          end
+        end
+
+        sig { params(receiver: T.nilable(String)).returns(T::Boolean) }
+        def supports_receiver?(receiver)
+          return receiver.nil? || receiver.empty? if @receiver == false
+
+          receiver == "T"
+        end
+      end
+    end
+  end
+end
--- a/lib/ruby_lsp/requests/support/sorbet.rb
+++ b/lib/ruby_lsp/requests/support/sorbet.rb
@@ -1,0 +1,120 @@
+# typed: strict
+# frozen_string_literal: true
+
+module RubyLsp
+  module Requests
+    module Support
+      class Sorbet
+        class << self
+          extend T::Sig
+
+          ANNOTATIONS = T.let(
+            {
+              "abstract!" => Annotation.new(arity: 0),
+              "absurd" => Annotation.new(arity: 1, receiver: true),
+              "all" => Annotation.new(arity: (2..), receiver: true),
+              "any" => Annotation.new(arity: (2..), receiver: true),
+              "assert_type!" => Annotation.new(arity: 2, receiver: true),
+              "attached_class" => Annotation.new(arity: 0, receiver: true),
+              "bind" => Annotation.new(arity: 2, receiver: true),
+              "cast" => Annotation.new(arity: 2, receiver: true),
+              "class_of" => Annotation.new(arity: 1, receiver: true),
+              "enums" => Annotation.new(arity: 0),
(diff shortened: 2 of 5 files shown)


lib/ruby_lsp/requests/support/annotation.rb
```diff
@@ -0,0 +1,76 @@
...
+            block: Block,
+            receiver: T::Boolean,
+          ).void
+        end
+        def initialize(arity:, block:, receiver: false)
+          @arity = arity
+          @block = block
+          @receiver = receiver
+        end
+        sig { returns(T.any(Integer, T::Range[Integer])) }
```
andyw8, line 21: (nit) Normally I'd expect to see any `attr_*`s before the methods in class.

test/requests/support/sorbet_test.rb
```diff
@@ -0,0 +1,834 @@
...
+    refute(annotation?("untyped(x, y) {}"))
+
+    refute(annotation?("C.untyped"))
+    refute(annotation?("C.untyped()"))
+    refute(annotation?("C.untyped(x)"))
+    refute(annotation?("C.untyped(x, y)"))
+    refute(annotation?("C.untyped {}"))
+    refute(annotation?("C.untyped() {}"))
+    refute(annotation?("C.untyped(x) {}"))
+    refute(annotation?("C.untyped(x, y) {}"))
```
andyw8, line 792: This seems like a lot of tests due to the combinations, I'm sure they're very fast but wonder if it's necessary to be as exhaustive?

lib/ruby_lsp/requests/support/sorbet.rb
```diff
@@ -0,0 +1,158 @@
...
+          sig do
+            params(node: T.any(SyntaxTree::VCall, SyntaxTree::CallNode)).returns(Integer)
+          end
+          def node_arity(node)
+            case node
+            when SyntaxTree::VCall
+              0
+            when SyntaxTree::CallNode
+              argument_arity(node.arguments)
+            end
```
andyw8, line 135: Should this have `T.absurd` as an `else`?

lib/ruby_lsp/requests/support/sorbet.rb
```diff
@@ -0,0 +1,158 @@
...
+module RubyLsp
+  module Requests
+    module Support
+      class Sorbet
+        class << self
+          extend T::Sig
+
+          ANNOTATIONS = T.let(
+            {
+              "abstract!" => Annotation.new(arity: 0, block: Annotation::Block::Unsupported),
```
andyw8, line 13: How was this list generated? Is there a way to ensure that we didn't miss anything?

lib/ruby_lsp/requests/support/sorbet.rb
```diff
@@ -0,0 +1,158 @@
...
+            T::Hash[String, Annotation],
+          )
+
+          sig do
+            params(
+              node: T.any(SyntaxTree::CallNode, SyntaxTree::VCall),
+              block: T::Boolean,
+            ).returns(T::Boolean)
+          end
+          def annotation?(node, block = false)
```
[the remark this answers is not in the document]
TobiasBales: looking at this it seems like the code path that calls it with true seems to be missing - which is weird given the tests pass.
I'll have another look

lib/ruby_lsp/requests/support/sorbet.rb
TobiasBales: We did skip some things intentionally (mostly around T.proc.bind etc) and went with just the "top level" things.

Finding an automated way (and then deciding which we want and which we don't want) might be a good thing but I have no idea where to start and how much it really is needed (from a time spent vs benefit gained perspective)

lib/ruby_lsp/requests/support/annotation.rb
andyw8: It may not be universal, but it seems to be the convention we use in this repo.

It also what the Ruby Style Guide [suggests](https://github.com/rubocop/ruby-style-guide#consistent-classes), although the Shopify one doesn't mention it.

lib/ruby_lsp/requests/support/sorbet.rb
```diff
@@ -0,0 +1,126 @@
...
+            when SyntaxTree::VCall
+              node_name(node.value)
+            when SyntaxTree::Ident
+              node.value
+            when SyntaxTree::Backtick
+              node.value
+            when SyntaxTree::Const
+              node.value
+            when SyntaxTree::Op
+              node.value
```
vinistock, line 115: NIT: can we merge these into the same `when` branch since they use the same method?

vinistock (approved):

Issue #356, opened 2022-11-09
Upgrade requirement to Syntax Tree 5.0.0 and fix breaking changes


vinistock:
Syntax Tree v5.0.0 is out with many node simplifications, like merging `Def`, `Defs` and `DefEndless` into a single `DefNode`. This means we can simplify our code as well and not require so many different visits.

Upgrade our requirement to `>= 5.0.0` and adjust it to fit the breaking changes and node simplifications.

Changelog: https://github.com/ruby-syntax-tree/syntax_tree/blob/main/CHANGELOG.md#500---2022-11-09

PR #376, opened 2022-12-07
Upgrade to syntax-tree 5

andyw8:
Paired with @vinistock.

[View diff with RBIs ignored](https://github.com/Shopify/ruby-lsp/pull/376/files?file-filters%5B%5D=.gemspec&file-filters%5B%5D=.lock&file-filters%5B%5D=.rb&show-deleted-files=true&show-viewed-files=true)

### Motivation

Closes https://github.com/Shopify/ruby-lsp/issues/356

### Automated Tests

Most things were already well covered, but I added one new expectation file.

### Manual Tests

I spent a fair amount of time manually verifying the behaviour in VS Code against the fixtures, but there is a chance I missed something.

I focused 🎩 on these area, which are touched by the PR:

- [x] document_symbol
- [x] folding_ranges
- [x] hover
- [x] selection_ranges
- [x] semantic_highlighting
- [x] support - highlight_target

I also checked against each point from the syntax_tree changelog, copied below:

- [x] ~Nodes no longer have a `comments:` keyword on their initializers. By default, they initialize to an empty array. If you were previously passing comments into the initializer, you should now create the node first, then call `node.comments.concat` to add your comments.~ We have no calls to `.comments`.
- [x] A lot of nodes have been folded into other nodes to make it easier to interact with the AST. This means that a lot of visit methods have been removed from the visitor and a lot of class definitions are no longer present. This also means that the nodes that received more function now have additional methods or fields to be able to differentiate them. Note that none of these changes have resulted in different formatting. The changes are listed below:
  - [x] `IfMod`, `UnlessMod`, `WhileMod`, `UntilMod` have been folded into `IfNode`, `UnlessNode`, `WhileNode`, and `UntilNode`. Each of the nodes now have a `modifier?` method to tell if it was originally in the modifier form. Consequently, the `visit_if_mod`, `visit_unless_mod`, `visit_while_mod`, and `visit_until_mod` methods have been removed from the visitor.
  - [x] ~`VarAlias` is no longer a node, and the `Alias` node has been renamed. They have been folded into the `AliasNode` node. The `AliasNode` node now has a `var_alias?` method to tell you if it is aliasing a global variable. Consequently, the `visit_var_alias` method has been removed from the visitor interface. If you were previously using this method, you should now use `visit_alias` instead.~ No mentions of `alias`.
  - [x] ~`Yield0` is no longer a node, and the `Yield` node has been renamed. They has been folded into the `YieldNode` node. The `YieldNode` node can now have its `arguments` field be `nil`. Consequently, the `visit_yield0` method has been removed from the visitor interface. If you were previously using this method, you should now use `visit_yield` instead.~ No mentions of `yield`.
  - [x] `FCall` is no longer a node, and the `Call` node has been renamed. They have been folded into the `CallNode` node. The `CallNode` node can now have its `receiver` and `operator` fields be `nil`. Consequently, the `visit_fcall` method has been removed from the visitor interface. If you were previously using this method, you should now use `visit_call` instead.
  - [x] ~`Dot2` and `Dot3` are no longer nodes. Instead they have become a single new `RangeNode` node. This node looks the same as `Dot2` and `Dot3`, except that it additionally has an `operator` field that contains the operator that created the node. Consequently, the `visit_dot2` and `visit_dot3` methods have been removed from the visitor interface. If you were previously using these methods, you should now use `visit_range` instead.~ No mentions of Dot2 or Dot3.
  - [x] `Def`, `DefEndless`, and `Defs` have been folded into the `DefNode` node. The `DefNode` node now has the `target` and `operator` fields which originally came from `Defs` which can both be `nil`. It also now has an `endless?` method on it to tell if the original node was found in the endless form. Finally the `bodystmt` field can now either be a `BodyStmt` as it was or any other kind of node since that was the body of the `DefEndless` node. The `visit_defs` and `visit_def_endless` methods on the visitor have therefore been removed.
  - [x] `DoBlock` and `BraceBlock` have now been folded into a `BlockNode` node. The `BlockNode` node now has a `keywords?` method on it that returns true if the block was constructed with the `do`..`end` keywords. The `visit_do_block` and `visit_brace_block` methods on the visitor have therefore been removed and replaced with the `visit_block` method.
  - [x] ~`Return0` is no longer a node, and the `Return` node has been renamed. They have been folded into the `ReturnNode` node. The `ReturnNode` node can now have its `arguments` field be `nil`. Consequently, the `visit_return0` method has been removed from the visitor interface. If you were previously using this method, you should now use `visit_return` instead.~ No mentions.
- [x] The `ArgsForward`, `Redo`, `Retry`, and `ZSuper` nodes no longer have `value` fields associated with them (which were always string literals corresponding to the keyword being used).
- [x] The `Command` and `CommandCall` nodes now has `block` attributes on them. These attributes are used in the place where you would previously have had a `MethodAddBlock` structure. Where before the `MethodAddBlock` would have the command and block as its two children, you now just have one command node with the `block` attribute set to the `Block` node.
- [x] ~Previously the formatting options were defined on an unfrozen hash called `SyntaxTree::Formatter::OPTIONS`. It was globally mutable, which made it impossible to reference from within a Ractor. As such, it has now been replaced with `SyntaxTree::Formatter::Options.new` which creates a new options object instance that can be modified without impacting global state. As a part of this change, formatting can now be performed from within a non-main Ractor. In order to check if the `plugin/single_quotes` plugin has been loaded, check if `SyntaxTree::Formatter::SINGLE_QUOTES` is defined. In order to check if the `plugin/trailing_comma` plugin has been loaded, check if `SyntaxTree::Formatter::TRAILING_COMMA` is defined.~ No use of this.


lib/ruby_lsp/requests/support/highlight_target.rb
```diff
@@ -67,15 +67,17 @@ def value(node)
           case node
           when SyntaxTree::ConstPathRef, SyntaxTree::ConstPathField, SyntaxTree::TopConstField
             node.constant.value
-          when SyntaxTree::GVar, SyntaxTree::IVar, SyntaxTree::Const, SyntaxTree::CVar, SyntaxTree::Ident,
-               SyntaxTree::ArgsForward
+          when SyntaxTree::ArgsForward
+            raise "no test is hitting this"
```
andyw8, line 71: I need to look into this.

lib/ruby_lsp/requests/hover.rb
```diff
@@ -32,16 +32,19 @@ def run
...
         case target
         when SyntaxTree::Command
           message = target.message
           generate_rails_document_link_hover(message.value, message)
-        when SyntaxTree::FCall
-          message = target.value
-          generate_rails_document_link_hover(message.value, message)
+        when SyntaxTree::CallNode
+          if target.message == :call
+            generate_rails_document_link_hover(target.message.to_s, target)
```
andyw8, line 44: Alternatively:
`generate_rails_document_link_hover("call", target)`
🤷‍♂️

lib/ruby_lsp/requests/folding_ranges.rb
```diff
@@ -192,33 +197,37 @@ def emit_partial_range
...
-        add_lines_range(receiver.location.start_line, node.location.end_line - 1)
+        add_lines_range(receiver.location.start_line, node.location.end_line - 1) if receiver
 
         visit(node.arguments)
       end
 
-      sig { params(node: T.any(SyntaxTree::Def, SyntaxTree::Defs)).void }
+      sig { params(node: SyntaxTree::DefNode).void }
       def add_def_range(node)
+        # For an endless method with no arguments, `node.params` returns `nil` for Ruby 3.0, but a `Syntax::Params`
```
andyw8, line 227: @kddnewton is this an expected behaviour? I couldn't see any mention of it in the docs or changelog.

kddnewton: Dang yeah I missed that in the changelog, sorry. 

The type in the documentation is right though https://github.com/ruby-syntax-tree/syntax_tree/blame/7e1b2a8176c37786323e334e477da8bd216f6ad6/lib/syntax_tree/node.rb#L4052

andyw8:
@st0012 Thanks, both suggestions applied.

lib/ruby_lsp/requests/folding_ranges.rb
```diff
@@ -192,33 +197,37 @@ def emit_partial_range
...
+        # for Ruby 3.1
+        return unless node.params
```
vinistock, line 229: Is this guaranteed to work for regular method definitions? Will `params` never be `nil`?

lib/ruby_lsp/requests/hover.rb
vinistock: I think we should ignore `:call` methods instead. I don't believe those will ever be a Rails DSL.

lib/ruby_lsp/requests/semantic_highlighting.rb
```diff
@@ -110,14 +110,16 @@ def run
...
-      sig { override.params(node: SyntaxTree::Call).void }
+      sig { override.params(node: SyntaxTree::CallNode).void }
       def visit_call(node)
         return super unless visible?(node, @range)
 
         visit(node.receiver)
 
         message = node.message
-        add_token(message.location, :method) if message != :call
+        if message != :call && message&.value != "lambda"
```
vinistock, line 120: Why do we need to check for lambda here? Should we be checking for special methods instead?

lib/ruby_lsp/requests/hover.rb
andyw8: I think you're right, I've updated.

lib/ruby_lsp/requests/folding_ranges.rb
andyw8: Ruby 3.0:

```
ruby-lsp andyw8/vini/upgrade-to-syntax-tree-5 % chruby 3.0
ruby-lsp andyw8/vini/upgrade-to-syntax-tree-5 % irb -r syntax_tree
irb(main):001:0> SyntaxTree.parse("def foo = 1").child_nodes.first.child_nodes.first.params.class
=> NilClass
irb(main):002:0> SyntaxTree.parse("def foo; end").child_nodes.first.child_nodes.first.params.class
=> SyntaxTree::Params                                                                              
irb(main):005:0> SyntaxTree.parse("def foo; end").child_nodes.first.child_nodes.first.params.child_nodes
=> [nil, nil, nil]
irb(main):006:0>
```
Ruby 3.1:

```
ruby-lsp andyw8/vini/upgrade-to-syntax-tree-5 % chruby 3.1
ruby-lsp andyw8/vini/upgrade-to-syntax-tree-5 % irb -r syntax_tree
irb(main):001:0> SyntaxTree.parse("def foo = 1").child_nodes.first.child_nodes.first.params.class
=> SyntaxTree::Params
irb(main):002:0> SyntaxTree.parse("def foo; end").child_nodes.first.child_nodes.first.params.class
=> SyntaxTree::Params
irb(main):003:0> SyntaxTree.parse("def foo; end").child_nodes.first.child_nodes.first.params.child_nodes
=> [nil, nil, nil]
```

lib/ruby_lsp/requests/semantic_highlighting.rb
andyw8: Without the `lambda` check, we get a failure on `test/expectations/semantic_highlighting/lambdas.exp`.

I've updated to use the `special_methods?` check.

andyw8:
@vinistock Thanks for the feedback, everything should be addressed.

PR #381, opened 2022-12-08
Refactor to support codeAction/Resolve

alkhatib-shopify:
Previously we were using the textDocument/codeAction only for Rubocop Where the edits are available and sent back in the response.

To be able to support refactorings, we would need to use the codeAction/resolve

This refactor adds code action resolve, and uses it for the rubocop edits.

Co-authored-by: Vinicius Stock <vinicius.stock@shopify.com>
Co-authored-by: Jenny Shih <jenny.shih@shopify.com>

### DIFF
--- a/lib/ruby_lsp/requests/code_action_resolve.rb
+++ b/lib/ruby_lsp/requests/code_action_resolve.rb
@@ -1,0 +1,50 @@
+# typed: strict
+# frozen_string_literal: true
+
+module RubyLsp
+  module Requests
+    # ![Code actions demo](../../misc/code_actions.gif)
+    #
+    # The [code actions](https://microsoft.github.io/language-server-protocol/specification#textDocument_codeAction)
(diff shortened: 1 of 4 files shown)


lib/ruby_lsp/requests/code_action_resolve.rb
```diff
@@ -0,0 +1,50 @@
+# typed: strict
+# frozen_string_literal: true
+
+module RubyLsp
+  module Requests
+    # ![Code actions demo](../../misc/code_actions.gif)
+    #
+    # The [code actions](https://microsoft.github.io/language-server-protocol/specification#textDocument_codeAction)
+    # request informs the editor of RuboCop quick fixes that can be applied. These are accesible by hovering over a
```
andyw8, line 9:
```suggestion
    # request informs the editor of RuboCop quick fixes that can be applied. These are accessible by hovering over a
```

PR #382, opened 2022-12-09
Fix the code action cache level

vinistock:
### Motivation

While pairing on #381, I noticed that we had our `cache_fetch` invocation at the wrong place for code actions. We want to cache the diagnostics response, so that we don't have to run RuboCop more than once, but we do not want to cache the code actions response since it may change depending on which range is visible in the editor.

### Implementation

Removed the cache from the server part, which depends on the range, and wrapped the diagnostics call in the cache.

### DIFF
--- a/lib/ruby_lsp/requests/code_actions.rb
+++ b/lib/ruby_lsp/requests/code_actions.rb
@@ -35,7 +35,7 @@ module RubyLsp
 
       sig { override.returns(T.all(T::Array[LanguageServer::Protocol::Interface::CodeAction], Object)) }
       def run
-        diagnostics = Diagnostics.new(@uri, @document).run
+        diagnostics = @document.cache_fetch(:diagnostics) { Diagnostics.new(@uri, @document).run }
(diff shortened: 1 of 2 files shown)


egiurleo (approved):

vinistock:
Good call. Added a test for it in the latest commit.

test/requests/code_actions_test.rb
```diff
@@ -0,0 +1,37 @@
...
+    uri = "file://#{__FILE__}"
+    @counter = 0
+    increment_counter = ->(_uri, _document) {
+      @counter += 1
+      []
+    }
+
+    # If diagnostics is being properly cached, then we should only see the increment_counter block invoked once for the
+    # same URI, no matter how many times we invoke it and with which range
+    RubyLsp::Requests::Support::RuboCopDiagnosticsRunner.instance.stub(:run, increment_counter) do
```
st0012, line 28: TIL we can implement an increment counter like this.

st0012 (approved):

PR #392, opened 2022-12-14
WIP: try fixing Ruby 3.2 build

egiurleo:


.github/workflows/ci.yml
```diff
@@ -16,7 +16,7 @@ jobs:
         uses: ruby/setup-ruby@v1
         with:
           ruby-version: ${{ matrix.ruby }}
-          bundler-cache: true
+          # bundler-cache: true
```
andyw8, line 19: You need to explicitly call `bundle` if this isn't enabled.

Issue #393, opened 2022-12-14
Missing test case for `ArgsForward` for `HighlightTarget`


andyw8:
It seems no test is touching `ArgsForward` for `HighlightTarget`:

https://github.com/Shopify/ruby-lsp/compare/Shopify:10dc26f...Shopify:77fa974

I need to change this behaviour in #376 so want to ensure it's covered first.

andyw8:
Discussed with @vinistock on Slack: there's not really much value in having document highlighting on `...` since it has no name, so we'll remove that node types from the match list. I'll open a new PR for that. This will simplify the syntax_tree v5 upgrade.

PR #395, opened 2022-12-14
Treat .exp files as JSON

andyw8:
Treat `.exp` as JSON so that we benefit from syntax highlighting and auto-formatting.

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -3,5 +3,8 @@
   "ruby lsp.trace.server": "messages",
   "[ruby]": {
     "editor.defaultFormatter": "Shopify.ruby-lsp",
-  }
+  },
+  "files.associations": {
+    "*.exp": "json"
+  },
 }


andyw8:
Note: This will cause some existing `.exp` files to change when saved. Should we disable format-on-save for that type? Or should we format all the existing ones as part of this PR?

Morriar (commented):
Note that all `exp` files aren't always JSON. For example: https://github.com/Shopify/ruby-lsp/blob/main/test/expectations/formatting/def_bad_formatting.exp

Is there any reason you're trying to read these files? The idea was also to provide the necessary tooling to use/update them rather than reading/writing them manually.

andyw8:
```
Note that all exp files aren't always JSON. For example: https://github.com/Shopify/ruby-lsp/blob/main/test/expectations/formatting/def_bad_formatting.exp
```
Ah, maybe we can split them into `.json.exp` and `.rb.exp`?

> Is there any reason you're trying to read these files? The idea was also to provide the necessary tooling to use/update them rather than reading/writing them manually.

It was for https://github.com/Shopify/ruby-lsp/issues/393. I did copy the output from the test run (and verified it), but I had to first give it some `params` as input, i.e.

```
{
    "params": [
        {
            "line": 0,
            "character": 5
        }
    ]
}
```

(I copied the structure from another `exp` but then adjusted the numbers).

Happy to hear if there's a better way!

Morriar:
iirc when you run the test without an exp file, it prints the exp in the terminal and you can copy it verbatim in the file

vinistock:
I think it makes sense to split it into `json.exp` and `rb.exp`, even if we have to reformat the files, but you will need to refactor our expectations runner a bit.

PR #396, opened 2022-12-14
Add test for argument forwarding (`...`)

### LINKED ISSUE #393: Missing test case for `ArgsForward` for `HighlightTarget`

andyw8:
### Motivation

Closes #393

When working on https://github.com/Shopify/ruby-lsp/pull/376 I discovered there was no highlighting test covering `SyntaxTree::ArgsForward`.

It seems that although we had `test/fixtures/argument_forwarding.rb`, there was no corresponding `.exp` - does that mean the fixture was essentially unused? If so, might there be other 'gaps' like this, and could we catch those programatically? 🤔

### DIFF
--- a/test/expectations/document_highlight/argument_forwarding.exp
+++ b/test/expectations/document_highlight/argument_forwarding.exp
@@ -1,0 +1,23 @@
+{
+    "params": [
+        {
+            "line": 0,
+            "character": 5
+        }
+    ],
+    "result": [
+        {
+            "range": {
+                "start": {
+                    "line": 0,
+                    "character": 4
+                },
+                "end": {
+                    "line": 0,
+                    "character": 7
+                }
+            },
+            "kind": 3
+        }
+    ]
+}


test/expectations/document_highlight/argument_forwarding.exp
```diff
@@ -0,0 +1,23 @@
+{
```
andyw8, line 1: It seems that though we had `test/fixtures/argument_forwarding.rb`, there was no corresponding `.exp` - does that mean it was having no effect? If so, might there be other 'gaps' like this, and could we catch those programatically? 🤔

test/expectations/document_highlight/argument_forwarding.exp
```diff
@@ -0,0 +1,23 @@
...
+                },
+                "end": {
+                    "line": 0,
+                    "character": 7
+                }
+            },
+            "kind": 3
+        }
+    ]
+}
```
andyw8, line 23: This may need to be expanded.

vinistock:
Did you want to test the highlight on the argument forwarding itself? Because line 0 character 5 is still on top of the method `foo` definition.

andyw8:
> Did you want to test the highlight on the argument forwarding itself? Because line 0 character 5 is still on top of the method foo definition.

Yeah, there a few cases that I want to cover, but my initial goal was just to match `ArgsForward` in https://github.com/Shopify/ruby-lsp/blob/10dc26f3b582601583966730a85c9a684d49f594/lib/ruby_lsp/requests/support/highlight_target.rb#L71, which this example does.

I'll try add some more now.

vinistock (approved):

andyw8:
This is somewhat redundant now due to https://github.com/Shopify/ruby-lsp/pull/399 but I think it's worth keeping as at least ensures we don't raise when encountering `def(...)`.

PR #398, opened 2022-12-15
Add note about `.exp` files to README

andyw8:


vinistock (commented):
Thanks for adding this 🙏

PR #402, opened 2022-12-16
Add `.json` or `.rb` extension to `.exp` files

andyw8:
### Motivation

* Illustrates that there are two variants of expression files.
* Easier reading for if we are trying to debug a failure and need to inspect the expression file in detail.
* Encourages consistent formatting since the editor can auto-format them.

### Implementation

I added a `.json` extension to all `.exp` files, except for those in `formatting/`, since they are Ruby.

I then re-saved each file, to ensure consistency across all the files. Most had no issues, but around 12 needed formatting changes.

~I chose not to add an `.rb` extension to the Ruby files since the may contain intentionally incorrect formatting, which we would not wish to auto-format when saved.~

### Automated Tests

I verified that the number of assertions run is the same as on `main` (855).

### Manual Tests

n/a


.rubocop.yml
```diff
@@ -29,6 +29,7 @@ Sorbet/TrueSigil:
   Exclude:
     - "**/*.rake"
     - "lib/**/*.rb"
+    - "test/expectations/formatting/*.rb"
```
vinistock, line 32: Let's exclude the entire expectations folder. There's nothing there that we need to run RuboCop over.

vinistock (approved):

PR #404, opened 2022-12-16
Add note about VSCode Neovim and selection ranges

andyw8:


lib/ruby_lsp/requests/selection_ranges.rb
```diff
@@ -11,6 +11,8 @@ module Requests
     #
     # Trigger this request with: Ctrl + Shift + -> or Ctrl + Shift + <-
     #
+    # Note that if using VSCode Neovim, you will need to be in Insert mode for this to work correctly.
```
andyw8, line 14: I suspect there are other features with similar problems, I'll keep a look out.

PR #410, opened 2022-12-22
Adjust Gemfile.lock platforms

vinistock:
### Motivation

With the recent changes to fix CI, we now can't bundle install in our dev laptops. I'm hoping adding arm64-darwin, x86-darwin and x86-linux will fix the issue.

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -48,6 +48,7 @@ GEM
       crass (~> 1.0.2)
       nokogiri (>= 1.5.9)
     method_source (1.0.0)
+    mini_portile2 (2.8.0)
     minitest (5.16.3)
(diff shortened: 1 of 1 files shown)


Gemfile.lock
```diff
@@ -151,8 +158,8 @@ GEM
     zeitwerk (2.6.1)
 
 PLATFORMS
-  arm64-darwin-21
-  arm64-darwin-22
+  arm64-darwin
+  x86_64-darwin
   x86_64-linux
```
paracycle, line 163: Shouldn't we also add `ruby` as a platform?

st0012 (approved):

st0012:
I need this PR to test changes without patching Gemfile.lock locally, so merging now.

PR #411, opened 2022-12-22
Fix comment continuation incorrectly matching keywords

### LINKED ISSUE #407: Unexpected behaviour when hitting newline in a comment
# Reproduce 

When you have a comment that contains a comma, for example: 

```ruby
# For gems that have a normal default require and that load all of their constants through that, everything should work
```

If you put your cursor before the comma and hit enter, what I expect to see is a comment on the new line. But instead I see:

```ruby
# For gems that have a normal default require and that load
#
endal$0l of their constants through that, everything should work
```

It seems like it's dependent on where you hit enter in the comment  - if it's before the comma, then we get the buggy behaviour, if it's after the comma, then we get the expected behaviour.

vinistock:
### Motivation

Closes #407

The issue has more context, but basically we were matching two parts of on type formatting at once. We were trying to continue comments on the next line, but also matching `def`, `while` or other keywords that we complete with `end` tokens at the same time. This resulted in weird behaviour, since those words inside comments don't actually require us to complete the `end`.

### Implementation

I restructured our `case` statement to make sure that we don't match both completions at the same time.

### Automated Tests

Added an example that reproduces the problem.

### Manual Tests

2. Add a comment that includes a keyword that we complete `end` tokens for like `while`
3. Break the line
4. Verify that the new line is a comment, but that no `end` tokens were added

### DIFF
--- a/lib/ruby_lsp/requests/on_type_formatting.rb
+++ b/lib/ruby_lsp/requests/on_type_formatting.rb
@@ -44,17 +44,17 @@ module RubyLsp
 
       sig { override.returns(T.nilable(T.all(T::Array[Interface::TextEdit], Object))) }
       def run
-        handle_comment_line
-
-        return @edits unless @document.syntax_errors?
-
         case @trigger_character
         when "{"
-          handle_curly_brace
+          handle_curly_brace if @document.syntax_errors?
         when "|"
-          handle_pipe
+          handle_pipe if @document.syntax_errors?
         when "\n"
-          handle_statement_end
+          if (comment_match = @previous_line.match(/^#(\s*)/))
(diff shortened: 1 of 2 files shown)


andyw8:
🤔 This doesn't seem to fix #407.

If I take the original comment:
```ruby
        # For gems that have a normal default require and that load all of their constants through that, everything should work
```

and insert a line break after `and`, then it results in:

```ruby
        # For gems that have a normal default require and 
         
        endthat load all of their constants through that, everything should work
```

vinistock:
Indeed the `if` and `elsif` clauses were inverted. I fixed it and added a test case that catches the scenario.

Issue #389, opened 2022-12-12
Allow log level to be configurable


andyw8:
I was discussing with @st0012 how we might prevent `$stderr` output appearing when running the tests. He suggested:

>. A better way may be introducing logging level to LSP output. And in testing we just increase the logging level to `warn`. (this cases’s message is likely `info` or `debug`)

> Once we have that, we can add some detailed messages at debug level, which will make debugging easier

vinistock:
I love this idea. We should be relatively low effort if we just use Ruby's `logger` library and set that as a singleton somewhere.

PR #412, opened 2022-12-23
Reduce noise in test output

andyw8:
### Motivation

This is related to https://github.com/Shopify/ruby-lsp/issues/389 but taking a different approach.

The aim to reduce noise in the output when running tests.

### Implementation

If we use `warn` instead of `$stderr.puts` then we can selectively silence its output.

There is possibly a chance the hiding warnings could mask an underlying issue. To partly mitigate this, I'm only silencing the warnings in non-CI environments.

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -39,3 +39,6 @@ Sorbet/StrictSigil:
     - "**/*.rake"
     - "test/**/*.rb"
     - "lib/ruby-lsp.rb"
+
+Style/StderrPuts:
+  Enabled: true
--- a/README.md
+++ b/README.md
@@ -36,7 +36,7 @@ For more visibility into which tests are running, use the `SpecReporter`:
 
 `SPEC_REPORTER=1 bin/test`
 
-By default the tests run with warnings disabled to reduce. To enable warnings, pass `VERBOSE=1`.
+By default the tests run with warnings disabled to reduce noise. To enable warnings, pass `VERBOSE=1`.
 Warnings are always shown when running in CI.
 
 ### Expectation testing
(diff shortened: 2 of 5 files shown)


.rubocop.yml
```diff
@@ -39,3 +39,6 @@ Sorbet/StrictSigil:
     - "**/*.rake"
     - "test/**/*.rb"
     - "lib/ruby-lsp.rb"
+
+Style/StderrPuts:
```
andyw8, line 43: https://www.rubydoc.info/gems/rubocop/RuboCop/Cop/Style/StderrPuts

test/test_helper.rb
```diff
@@ -2,6 +2,7 @@
 # frozen_string_literal: true
 
 $LOAD_PATH.unshift(File.expand_path("../lib", __dir__))
+$VERBOSE = nil unless ENV["CI"]
```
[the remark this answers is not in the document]
andyw8: Thanks, updated with these suggestions.

andyw8:
Need to update rubocop-shopify first: https://github.com/Shopify/ruby-lsp/pull/425

README.md
```diff
@@ -36,6 +36,9 @@ For more visibility into which tests are running, use the `SpecReporter`:
 
 `SPEC_REPORTER=1 bin/test`
 
+By default the tests run with warnings disabled to reduce. To enable warnings, pass `VERBOSE=1`.
```
egiurleo, line 39: I think you didn't finish that first sentence 😆

egiurleo (approved):
Nice! Small comment about a missing word.

README.md
andyw8: Oops, updated.

st0012 (approved):

PR #425, opened 2023-01-03
Update rubocop-shopify to 2.11.1 and apply corrections

andyw8:
### Motivation

It's blocking https://github.com/Shopify/ruby-lsp/pull/412.

### Implementation

Use `rubocop -a`, then made a manual edit (435c3da770c3ebec26ccd9b0047c987ccbce47d1) for the remaining one issue.


test/requests/support/semantic_token_encoder_test.rb
```diff
@@ -13,14 +13,32 @@ def test_tokens_encoded_to_relative_positioning
     ]
 
     expected_encoding = [
-      0, 2, 1, 0, 1,
-      0, 2, 2, 9, 1,
-      1, 2, 3, 0, 64,
-      3, 6, 10, 4, 16,
+      0,
```
andyw8, line 16: Should we add an ignore directive for this?

Issue #338, opened 2022-10-21
Explore implementing a semantic search custom request


vinistock:
Explore adding the possibility for the user to insert a pattern expression, which then sends a request to the server, the server finds all matches and allows the user to cycle through them.

Think something like
- Some hotkey trigger a command that opens an input box, the user adds an expression like `var_ref (ident ("my_var"))`
- Upon submit, a request is sent and the Ruby LSP finds all occurrences matching the expression, returning a list of locations
- The extension then allows the user to cycle through all results

Morriar:
> Ruby LSP finds all occurrences matching the expression

Between this and #61 we may want to investigate a way to index the code in the project to provide fast lookups.

kddnewton:
Just want to put this here in case y'all didn't see it: https://github.com/ruby-syntax-tree/syntax_tree#search.

PR #429, opened 2023-01-05
[RFC] Constant/method indexing and go to definition

paracycle:
### Motivation

This is a quick-and-dirty, first-cut implementation of constant/method definition indexing based on the prototype code supplied by @kddnewton. An example "Go to definition" request implementation that only works for fully-qualified constants is also included.

### Implementation

The indexer is based on scanning the compiled instruction sequences that correspond to all project Ruby files (all files under the folder that contains the `Gemfile`). This technique was suggested by @kddnewton and can compile and process all files in Core in about 30s on my M1 Pro machine.

In its current form, the indexer runs serially before the Ruby LSP starts, which is obviously not the best way to do this, but is fast enough for this RFC.

### Automated Tests

None

### Manual Tests

None

### DIFF
--- a/lib/ruby_lsp/indexer.rb
+++ b/lib/ruby_lsp/indexer.rb
@@ -1,0 +1,68 @@
+# typed: true
+# frozen_string_literal: true
+
+class Indexer
+  VM_DEFINECLASS_TYPE_CLASS           = 0x00
+  VM_DEFINECLASS_TYPE_SINGLETON_CLASS = 0x01
+  VM_DEFINECLASS_TYPE_MODULE          = 0x02
+  VM_DEFINECLASS_FLAG_SCOPED          = 0x08
+  VM_DEFINECLASS_FLAG_HAS_SUPERCLASS  = 0x10
+
+  def initialize(root)
+    @root = root
+    @index = {
+      constant: (Hash.new { |h,k| h[k] = [] }),
+      method: (Hash.new { |h,k| h[k] = [] })
+    }
+  end
+
+  def run
+    Dir["#{@root}/**/*.rb"].each do |file|
+      index_file(file)
+    end
+  end
+
+  def locs_for_symbol(symbol, kind)
+    $stderr.puts "--- Looking for #{symbol} of kind #{kind}"
+    index.dig(kind, symbol)
+  end
+
+  private
+
+  attr_reader :root, :index
+
+  def index_file(file)
+    iseq = RubyVM::InstructionSequence.compile(File.read(file)).to_a
+    index_iseq(iseq) do |line, event, message|
+      index[event][message] << [File.absolute_path(file), line]
+    end
+  end
+
+  def index_iseq(iseq, nesting = [], &blk)
+    line = iseq[9]
+    iseq[13].each_with_index do |insn, index|
+      case insn
+      in Integer
+        line = insn
+      in [:defineclass, name, class_iseq, ^(VM_DEFINECLASS_TYPE_SINGLETON_CLASS)]
+        # if iseq[13][index - 2] == [:putself]
+        #   index_iseq(iseq, nesting + [name], &blk)
+        # else
+        #   raise NotImplementedError, "singleton class with non-self receiver"
+        # end
+      in [:defineclass, name, class_iseq, flags] if flags & VM_DEFINECLASS_TYPE_MODULE > 0
+        blk.call(line, :constant, (nesting + [name]).join("::"))
+        index_iseq(class_iseq, nesting + [name], &blk)
+      in [:defineclass, name, class_iseq, flags]
+        blk.call(line, :constant, (nesting + [name]).join("::"))
+        index_iseq(class_iseq, nesting + [name], &blk)
+      in [:definemethod, name, method_iseq]
+        blk.call(line, :method, "#{name}")
+      in [:definesmethod, name, method_iseq]
+        blk.call(line, :method, "#{name}")
+      else
+        # skip other instructions
+      end
+    end
+  end
+end
--- a/lib/ruby_lsp/handler.rb
+++ b/lib/ruby_lsp/handler.rb
@@ -4,6 +4,7 @@
 require "ruby_lsp/requests"
 require "ruby_lsp/store"
 require "ruby_lsp/queue"
+require "ruby_lsp/indexer"
 
 module RubyLsp
   Interface = LanguageServer::Protocol::Interface
@@ -48,18 +49,26 @@ module RubyLsp
     sig { returns(Store) }
     attr_reader :store
 
+    sig { returns(Indexer) }
+    attr_reader :indexer
+
     sig { void }
     def initialize
       @writer = T.let(Transport::Stdio::Writer.new, Transport::Stdio::Writer)
       @reader = T.let(Transport::Stdio::Reader.new, Transport::Stdio::Reader)
       @handlers = T.let({}, T::Hash[String, RequestHandler])
       @store = T.let(Store.new, Store)
+      @indexer = T.let(Indexer.new(Bundler.root), Indexer)
       @queue = T.let(Queue.new(@writer, @handlers), Queue)
     end
 
     sig { void }
     def start
       $stderr.puts "Starting Ruby LSP..."
+
+      $stderr.puts "Indexing..."
+      indexer.run
+      $stderr.puts "Indexing done!"
 
       @reader.read do |request|
         handler = @handlers[request[:method]]
--- a/lib/ruby_lsp/requests.rb
+++ b/lib/ruby_lsp/requests.rb
@@ -31,7 +31,7 @@ module RubyLsp
     autoload :CodeActions, "ruby_lsp/requests/code_actions"
     autoload :DocumentHighlight, "ruby_lsp/requests/document_highlight"
     autoload :InlayHints, "ruby_lsp/requests/inlay_hints"
-
+    autoload :GoToDefinition, "ruby_lsp/requests/go_to_definition"
     # :nodoc:
     module Support
       autoload :RuboCopDiagnostic, "ruby_lsp/requests/support/rubocop_diagnostic"
--- a/lib/ruby_lsp/requests/go_to_definition.rb
+++ b/lib/ruby_lsp/requests/go_to_definition.rb
@@ -1,0 +1,88 @@
+# typed: strict
+# frozen_string_literal: true
+
+module RubyLsp
+  module Requests
+    class GoToDefinition < BaseRequest
+      extend T::Sig
+
+      sig { params(indexer: Indexer, uri: String, document: Document, position: Document::PositionShape).void }
+      def initialize(indexer, uri, document, position)
+        super(document)
+        @uri = uri
+        @indexer = T.let(indexer, Indexer)
+        @position = T.let(document.create_scanner.find_char_position(position), Integer)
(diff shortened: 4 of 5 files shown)


lib/ruby_lsp/indexer.rb
```diff
@@ -0,0 +1,68 @@
+# typed: true
+# frozen_string_literal: true
+
+class Indexer
```
kddnewton, line 4: This should be nested under the module.

lib/ruby_lsp/indexer.rb
```diff
@@ -0,0 +1,68 @@
...
+    $stderr.puts "--- Looking for #{symbol} of kind #{kind}"
+    index.dig(kind, symbol)
+  end
+
+  private
+
+  attr_reader :root, :index
+
+  def index_file(file)
+    iseq = RubyVM::InstructionSequence.compile(File.read(file)).to_a
```
kddnewton, line 35: We should use `compile_file` here as it'll store the information a little more accurately inside the ISEQ.

lib/ruby_lsp/indexer.rb
```diff
@@ -0,0 +1,68 @@
...
+    index_iseq(iseq) do |line, event, message|
+      index[event][message] << [File.absolute_path(file), line]
```
kddnewton, line 37: Removing this block and instead having `index_iseq` either push directly into the hashes or call methods like `index_constant` and `index_method` would be a lot more performant.

lib/ruby_lsp/indexer.rb
```diff
@@ -0,0 +1,68 @@
...
+    end
+  end
+
+  def index_iseq(iseq, nesting = [], &blk)
+    line = iseq[9]
+    iseq[13].each_with_index do |insn, index|
+      case insn
+      in Integer
```
kddnewton, line 45: We should remove the pattern matching here. If it were instead:

```ruby
case insn
when Array
  # this is the common case, and should be checked first

  case insn[0]
  when :defineclass
  when :definemethod, :definesmethod
  end

  # the above case statement can take advantage of the
  # opt_case_dispatch instruction, which will make it a
  # lot faster to execute
when Integer
  line = insn
end
```

kddnewton (commented):
A couple of suggestions that should help speed this up quite a bit

lib/ruby_lsp/indexer.rb
vinistock, line 35: Does compiling the instruction sequences include comments? Indexing will also unlock properly implementing hover, but we'll need to keep track of the comments associated to each method/constant.

vinistock (commented):
This is super exciting. As a comment so that we do not forget: we should use the specifications [file watching notifications](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#workspace_didChangeWatchedFiles) to maintain the index in sync.

The server registers which files it is interested in and the client will notify when files are created, changed or deleted. We can update the index when processing these notifications and avoid having to implement any sort of file watching ourselves.

lib/ruby_lsp/indexer.rb
vinistock: In this case, we'd be compiling the file into instruction sequences and then parsing it with Ripper for the comments. Should we merge both things and just create a Ripper subclass that extracts only constant, methods and their comments?

kddnewton: You could if you wanted, but I wouldn't. 

The issue is, if you're using ripper to get all of the constants and methods, then you have to deal with every variation in syntax. You need to define handlers for every child node that _could_ be in the tree. And you need to allocate a ton of objects. You basically have to manually build your syntax tree and maintain it. At that point you may as well use Syntax Tree.

On the other hand, if you're just lexing the file to get out the comments, it's going to be much faster. And it won't allocate nearly as many objects.

vinistock: We could also subclass `SyntaxTree::Parser` to intercept only the methods we're interested in. Not sure which would be faster (compiling inseq + parsing comments vs doing it in one go with a subclass). Whatever the fastest option is, since indexing is expensive.

kddnewton:
Note that one limitation of this approach is that it will only run on CRuby at the moment. (Unless you use Syntax Tree to get the instruction sequences through the syntax tree on other VMs.)

Morriar (commented):
I wonder if using Ctags for indexing would be easier and faster rather that parsing iseqs ourselves?

kddnewton:
@morriar this is not dissimilar to doing ctags manually. I'd be fine with standardizing on that format however for the indexing operation. Do you have a tool that generates good ctags for Ruby at the moment?

Morriar:
I've been experimenting with https://github.com/universal-ctags/ctags lately. Not sure if this is the best one or how it compares to parsing iseqs but it's pretty fast and reliable.

vinistock:
Would ctags give us enough flexibility to implement hover, signature help, workspace symbols and other features we'd like to explore?

numbcoder:
> @Morriar this is not dissimilar to doing ctags manually. I'd be fine with standardizing on that format however for the indexing operation. Do you have a tool that generates good ctags for Ruby at the moment?

[ ripper-tags](https://github.com/tmm1/ripper-tags) generates ctags using the built-in Ruby parser Ripper, which is more accurate than universal-ctags.

kddnewton:
If we're going to generate with ripper we may as well generate with Syntax Tree because that's going to get us closer to YARP. I took at look at the CTags generated by universal ctags and it's not at all dissimilar from what Syntax Tree is currently generating. I think we could make the jump to just generating ctags if that makes it easier to work with.

paracycle:
I am not sure why we are considering alternatives here. The iseq based indexing is now encapsulated inside SyntaxTree with a fallback to doing AST based indexing, we could just start building on that. We know it is fast, it is now encapsulated and stable, so I don't see a reason to not use it. Moreover, since the encapsulation is at the ST layer, it will be easier to plug it out to use YARP behind the scenes if we find that to be a better approach later.

As for what the indexing format is going to be, let's discuss that, but I think we're in a pretty good place with how we are going to do the indexing.

kddnewton:
Okay @morriar - I spent the weekend looking at CTags mostly for this ticket. Long story short, Syntax Tree can now generate CTags exactly as `ripper-tags` can once I merge this PR: https://github.com/ruby-syntax-tree/syntax_tree/pull/334.

Here are the things that I learned:

* CTags don't inherently have any support for delineating between identifiers. Extended CTags provides a way to "tag" identifiers with different information, but it wouldn't be a very good format for accessing comments, for instance.
* Even with a CTags database, you would still have to provide the functionality in the language server. It doesn't just give you that functionality for free.
* `universal-ctags` seems to miss a couple of things that we can generate more accurately with Syntax Tree. We're also very slightly faster than both it and `ripper-tags`.

If you want to roll with CTags, that's fine, you can use that format. I don't really see any particular benefit to using it over just accessing the index directly, but if you feel strongly there's now nothing stopping you.

paracycle:
> The PR title is prefixed with `[RFC]` and it's been in draft for the last two months.

Totally fair.

> As for Ctags, regardless of the indexing format, it seems to be an approach already used by many, in Ruby as well as other languages. It allows a good control over which files should be considered or not, even outside of Ruby and which contents from these files should be indexed. This would also allow people to plug in what they already know and use.
> 
> I'm asking if we considered the pros and cons of the proposed approach compared to alternatives.

I think what I was trying to settle in my previous message was the difference between how we will be doing the indexing vs what the indexing format was going to be. I think we should continue discussing what the indexing format might be, and I would lean on a format more under our control, since it might give us more options. On the other hand, I think we can start by doing the indexing itself using the new ST indexer. I think we are all on the same page on that aspect, right?

vinistock:
I agree that indexing with Syntax Tree is the way to go. It will gives us more flexibility and control which should allow us to provide more complex features.

Concerning the format, since performance is critical for language servers, the answer should be whatever is fastest to search/access. Doesn't Syntax Tree already come with a structure like that? Pretty sure we're covered in both fronts.

Morriar:
Thanks @kddnewton for this run down. I think there is a misunderstanding about why I brought Ctags though.

I'm not promoting the format, I actually think Ctags is a horrible format for what we're trying to do here. What I'm promoting here is the flexibility of the tools that produce the Ctags.

Ctags generators come with options that let you include / exclude paths, files and or even constructs (let's say private things) and this kind of features seems really important in this context. For example, in codebases containing a lot of generated code may it be real Ruby, RBIs or something else. Imagine if the first result you always jump is the wrong one and brings you in the middle of generated code? Horrible experience.

Parsing generated Ctags means the user can generate them with the options they want. Maybe even having their own custom tools to generate Ctags that support their DSLs etc.

If we're going with SyntaxTree for the indexing, the Ruby LSP is then responsible to provide the options to pass or not certain files to the SyntaxTree, then select or not what SyntaxTree returns to be indexed.

So have we considered what features are necessary to provide a good experience here? Have we looked at what other tools that provide the same scope of features do?

I hope this clears things up.

Abson:
哥们，实在不行搞个 CTag 用着先，没有提示很难受

gwillcox-r7:
> Thanks @kddnewton for this run down. I think there is a misunderstanding about why I brought Ctags though.
> 
> I'm not promoting the format, I actually think Ctags is a horrible format for what we're trying to do here. What I'm promoting here is the flexibility of the tools that produce the Ctags.
> 
> Ctags generators come with options that let you include / exclude paths, files and or even constructs (let's say private things) and this kind of features seems really important in this context. For example, in codebases containing a lot of generated code may it be real Ruby, RBIs or something else. Imagine if the first result you always jump is the wrong one and brings you in the middle of generated code? Horrible experience.
> 
> Parsing generated Ctags means the user can generate them with the options they want. Maybe even having their own custom tools to generate Ctags that support their DSLs etc.
> 
> If we're going with SyntaxTree for the indexing, the Ruby LSP is then responsible to provide the options to pass or not certain files to the SyntaxTree, then select or not what SyntaxTree returns to be indexed.
> 
> So have we considered what features are necessary to provide a good experience here? Have we looked at what other tools that provide the same scope of features do?
> 
> I hope this clears things up.

Were you thinking something like Bust-A-Gem from https://github.com/gurgeous/bust-a-gem in terms of implementation here for comparison's sake? That also uses a CTags based approach but uses `ripper-tags` to implement its features. I have used it quite a bit and haven't personally found any issues though I deal more with pure Ruby than Rails related code so that may need some testing.

vinistock:
I'm super happy with everybody's excitement over go to definition and related functionality. Let me chime in here to clarify a few things:
- We have plans to work on code indexing, which will unblock go to definition, autocomplete, hover, signature help, workspace symbols and other features
- We're planning on working it soonish. It's important to get the implementation right for a good experience - especially in very large projects. Therefore, we will take the time necessary to prototype, explore alternatives and make sure we're delivering the best experience we can
- This PR here is not a final implementation, it's one of these prototypes exploring indexing through instruction sequences as opposed to indexing via parsing and visiting (the usual method for analyzing files in the Ruby LSP)
- While we might explore CTags as part of the prototypes, it's **very unusual** for an LSP to use that as its indexing backend. We're most likely going to be indexing files ourselves to give us enough flexibility to implement features - not only the current ones in the specification, but future additions and custom requests we may desire to add. For example, semantic search #338 is not a part of the specification, but it could be implemented as a custom request as long as we have enough information in our index

PR #430, opened 2023-01-05
Add Sorbet shims for debug gem

andyw8:
To prevent distracting typecheck errors during development.


KaanOzkan (approved):

st0012:
@andyw8 the `debugger` method is actually [added to `Kernel`](https://github.com/ruby/debug/blob/0fcfc28acae33ec1c08068fb7c33703cfa681fa7/lib/debug/session.rb#L2570-L2590), not `Binding`.

PR #431, opened 2023-01-05
WIP: Set up `ruby/debug`

andyw8:


andyw8:
Note: Calling `exit!` (which bypasses `at_exit`) does not seem to hang, like `exit` does.

andyw8:
Re. @st0012's suggestion of "adding an Ruby LSP: Restart with ruby/debug action on the extension side". I tried that [here](https://github.com/Shopify/vscode-ruby-lsp/compare/andyw8/support-ruby-debug).

The problem is that once you run Start Debuggin, Ruby LSP complains "The gems in the bundle are not installed". I think this is the same kind of issue we see when we load a project through VS Code rather than using `code .` in the terminal.

Is there some known workaround to use ruby-lsp when being run within the Extension Development Host?

vinistock:
What if you change the command from `exe/ruby-lsp` to `bundle exec ruby-lsp`?

.vscode/launch.json
```diff
@@ -0,0 +1,15 @@
...
+    // Hover to view descriptions of existing attributes.
+    // For more information, visit: https://go.microsoft.com/fwlink/?linkid=830387
+    "version": "0.2.0",
+    "configurations": [
+        {
+            "type": "rdbg",
+            "name": "Minitest - current file",
+            "request": "launch",
+            "script": "-Itest ${file}",
+            "askParameters": false
```
vinistock, line 12: Should we ask for the specific test the person wants to run (the fixture name)? If we were trying to debug tests that use fixtures, we'd always end up running every single one, which is a very high number of tests.

README.md
```diff
@@ -93,6 +93,28 @@ Possible values are:
...
+1. Add a breakpoint using the VS Code UI.
+1. Open the relevant test file.
+1. Open the **Run and Debug** panel on the sidebar.
+1. Ensure `Minitest - current file` is selected in the top dropdown.
+1. Press `F5` OR click the green triangle next to the top dropdown. VS Code will then run the test file with debugger activated.
+1. When the breakpoint is triggered, the process will pause and VS Code will connect to the debugger and activate the debugger UI.
+1. Open the Debug Console view to use the debugger's REPL.
+
+```
+rm `which rdbg`
```
st0012, line 109: Is this a leftover from previous changes?

BTW, Vini and I were investigating [this issue](https://github.com/Shopify/ruby-dev-exp-issues/issues/736) and we made some good progress.

README.md
```diff
@@ -93,6 +93,28 @@ Possible values are:
...
+```
+
+### Debugging using the command line
+
+You can attach to the ruby-lsp process from the command line with:
+
+`bundle exec rdbg --attach`
```
st0012, line 116: This section is also not needed I think?
FWIW, once VS Code is connected to the debugger, the `DEBUG CONSOLE` tab can be used as a console:
Screenshot 2023-01-11 at 18 44 25

vinistock (approved):

PR #434, opened 2023-01-06
Add comparison with Solargraph

### LINKED ISSUE #419: Comparison with Solargraph
I'm pretty new to language servers, and I've been hearing only about Solargraph so far. Since this seems to be a replacement, it would be useful if the README included some advantages over Solargraph, to help developers decide 🙂

andyw8:
Closes #419

My exposure to Solargraph is limited so some aspects here could be incorrect or incomplete.

## Motivation

Both Solargraph and ruby-lsp have similar goals, so it's not clear why a developer should pick one over the other. This is an attempt to give some useful information for an informed and objective decision.

### DIFF
--- a/README.md
+++ b/README.md
@@ -97,8 +97,7 @@ Possible values are:
 
 **Background**
 
-In July 2021, Shopify quietly released [rubocop-lsp](https://github.com/Shopify/rubocop-lsp). In March 2022, that was succeeded by [ruby-lsp](https://github.com/Shopify/ruby-lsp).
-As the name suggests, the newer library is intended to have a wider set of capabilities.
+[ruby-lsp](https://github.com/Shopify/ruby-lsp) was first released in March 2022.
 It has a corresponding [VS Code extension](https://github.com/Shopify/vscode-ruby-lsp).
 ruby-lsp is widely used at Shopify, and is installed by default as part of most Ruby or Rails codebases.
 
@@ -114,10 +113,7 @@ This is used to provide capabilities such as intellisense (context-aware auto-co
 Solargraph relies on [YARD](https://yardoc.org/), a popular tool for building documentation.
 Developers run the `yard gems` command on their local machine to generate this information
 
-ruby-lsp does not provide direct support for these features.
-The intention is that they are handled by Sorbet’s LSP server, running alongside ruby-lsp.
-This means some functionality is only available if the files are [typed at ‘true’ or higher](https://sorbet.org/docs/static#file-level-granularity-strictness-levels).
-However there is some experimental work [in progress](https://github.com/Shopify/ruby-lsp/pull/429) to improve support for untyped code.
+While the Ruby LSP does not yet implement these features, the goal is to have it index Ruby codebases to provide them without requiring any form of annotations.
 
 **Similarities**
 
@@ -136,7 +132,7 @@ Features only in ruby-lsp:
 
 Features only in Solargraph:
 
-* [Plugin support](https://solargraph.org/guides/plugins)
+* [Plugin support](https://solargraph.org/guides/plugins) (but it is [planned](https://github.com/Shopify/ruby-lsp/issues/443) for ruby-lsp)
 
 **Underlying Technologies**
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -93,6 +93,71 @@ Possible values are:
 * `messages`: display requests and responses notifications
 * `verbose`: display each request and response as JSON
 
+## How does ruby-lsp differ from Solargraph?
```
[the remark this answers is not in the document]
shopify-aiden: I'm personally fine with the length of it, but could see benefit of adding a comparison page to leave room for future edits.

shopify-aiden (approved):
This is an awesome write up, thanks for giving some great context on the difference of the two. I think this would be really helpful for anyone assessing which they would want to use. 

Are there any other extensions in this space that we would want to add to this in the future? I think that might be a good indicator on if we want to add this to a different page or keep in int the README.

st0012 (commented):
I like the idea 👍 
But is it possible to utilise tables here? Having stuff visually side-by-side via tables will make it even easier to compare them.

README.md
```diff
@@ -93,6 +93,70 @@ Possible values are:
...
+
+**Background**
+
+In July 2021, Shopify quietly released [rubocop-lsp](https://github.com/Shopify/rubocop-lsp). In March 2022, that was succeeded by [ruby-lsp](https://github.com/Shopify/ruby-lsp).
```
vinistock, line 100: Is it worth mentioning the `rubocop-lsp` here? It was basically an experiment that is already archived.

README.md
```diff
@@ -93,6 +93,70 @@ Possible values are:
...
+
+**Major Differences**
+
+The biggest difference between the two libraries is their approach to understanding the structure your code, and its dependencies.
+This is used to provide capabilities such as intellisense (context-aware auto-completion) and navigation (e.g. Go To Definition).
+
+Solargraph relies on [YARD](https://yardoc.org/), a popular tool for building documentation.
+Developers run `yard gems` command on their local machine to generate this information
+
+ruby-lsp does not provide direct support for these features.
```
vinistock, line 117: The Ruby LSP will provide those features eventually. We choose to use it with Sorbet for more accurate results, but it's not a requirement.

I would say something along the lines of `While the Ruby LSP does not yet implement these features, the goal is to have it index Ruby codebases to provide them without requiring any form of annotations`.

README.md
```diff
@@ -93,6 +93,70 @@ Possible values are:
...
+The intention is that they are handled by Sorbet’s LSP server, running alongside ruby-lsp.
+This means some functionality is only available if the files are [typed at ‘true’ or higher](https://sorbet.org/docs/static#file-level-granularity-strictness-levels).
+However there is some experimental work [in progress](https://github.com/Shopify/ruby-lsp/pull/429) to improve support for untyped code.
```
vinistock, line 120: Sorbet is not required to use the Ruby LSP, even though you can use both at the same time. I wouldn't mention this.

README.md
```diff
@@ -93,6 +93,70 @@ Possible values are:
...
+* Both libraries integrate with RuboCop for linting and auto-formatting. (Only ruby-lsp has a Quick Fix feature, allowing individual corrections.).
+* Both support Code Folding
+
+**Other differences**
+
+Features only in ruby-lsp:
+
+* Selection ranges (Expand Selection / Shrink Selection)
+* Inlay hints
+* Snippets
```
vinistock, line 135: If folks decide to implement these in Solargraph, we will need to maintain our own documentation in sync. I'm not a big fan of that...maybe we should focus on bigger, architectural and principle differences?

README.md
```diff
@@ -93,6 +93,70 @@ Possible values are:
...
+
+Features only in Solargraph:
+
+* [Plugin support](https://solargraph.org/guides/plugins)
```
vinistock, line 139: This is planned for the Ruby LSP https://github.com/Shopify/ruby-lsp/issues/443.

vinistock (commented):
I agree with Stan that a side by side table would be easier to digest

andyw8:
Closing in favor of https://github.com/Shopify/ruby-lsp/pull/447

PR #447, opened 2023-01-10
Add overview

andyw8:
### Motivation

Explain the goals of ruby-lsp and provide an overview of its major features. Outline the differences with Solargraph to allow a reader to make an informed decision on whether to choose it.

Addresses https://github.com/Shopify/ruby-lsp/issues/419

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,6 +4,33 @@
 
 This gem is an implementation of the [language server protocol specification](https://microsoft.github.io/language-server-protocol/) for Ruby, used to improve editor features.
 
+# Overview
+
+The intention of Ruby LSP is to provide a fast, robust and feature-rich coding environment for Ruby developers.
+
+It's part of a [wider Shopify goal](https://github.com/Shopify/vscode-shopify-ruby) to provide a state-of-the-art experience to Ruby developers using modern standards for cross-editor features, documentation and debugging.
+
+It provides many features, including:
+
+* Syntax highlighting
+* Linting and formatting
+* Code folding
+* Selection ranges
+
+It does not perform typechecking, so its features are implemented on a best-effort basis, aiming to be as accurate as possible.
+
+Planned future features include:
+
+* Auto-completion and navigation ("Go To Definition") ([prototype](https://github.com/Shopify/ruby-lsp/pull/429))
+* Support for plug-ins to extend behavior
+
+The Ruby LSP does not perform any type-checking or provide any type-related assistance, but it can be used alongside [Sorbet](https://github.com/sorbet/sorbet)'s LSP server.
+
+At the time of writing, these are the major differences between Ruby LSP and [Solargraph](https://solargraph.org/):
+
+* Solargraph [uses](https://solargraph.org/guides/yard) YARD documentation to gather information about your project and its gem dependencies. This provides functionality such as context-aware auto-completion and navigation ("Go To Definition")
+* Solargraph can be used as a globally installed gem, but Ruby LSP must be added to the Gemfile or gemspec if using RuboCop. (There are pros and cons to each approach)
+
 ## Usage
 
 Install the gem. There's no need to require it, since the server is used as a standalone executable.
@@ -17,7 +44,7 @@ end
 If using VS Code, install the [Ruby LSP extension](https://github.com/Shopify/vscode-ruby-lsp) to get the extra features in
 the editor.
 
-See the [documentation](https://shopify.github.io/ruby-lsp) for
+See the [documentation](https://shopify.github.io/ruby-lsp) for more in-depth details about the
 [supported features](https://shopify.github.io/ruby-lsp/RubyLsp/Requests.html).
 
 ## Contributing


README.md
```diff
@@ -17,9 +48,11 @@ end
 If using VS Code, install the [Ruby LSP plugin](https://github.com/Shopify/vscode-ruby-lsp) to get the extra features in
 the editor.
 
-See the [documentation](https://shopify.github.io/ruby-lsp) for
+See the [documentation](https://shopify.github.io/ruby-lsp) for more in-depth details about the
 [supported features](https://shopify.github.io/ruby-lsp/RubyLsp/Requests.html).
 
+Currently is not possible to use a 'globally installed' version of the gem - it must be part of the `Gemfile`.
```
vinistock, line 54: You can actually use it from a global installation for editors other than VS Code (e.g.: NeoVim) as long as you're not using RuboCop.

README.md
```diff
@@ -4,6 +4,37 @@
...
+It's part of a [wider Shopify goal](https://github.com/Shopify/vscode-shopify-ruby) to provide a state-of-the-art experience to Ruby developers using modern standards for cross-editor features, documentation and debugging.
+
+It provides many features, including:
+
+* Syntax highlighting
+* Linting and formatting
+* Code folding
+* Selection ranges
+
+When used on typed Sorbet code, additional features are available:
```
vinistock, line 20: The Ruby LSP and Sorbet work completely independently - with no interaction between the two whatsoever. I don't think we should include this section, we can simply say "The Ruby LSP is designed to work by itself or paired with Sorbet".

README.md
```diff
@@ -4,6 +4,37 @@
 
 This gem is an implementation of the [language server protocol specification](https://microsoft.github.io/language-server-protocol/) for Ruby, used to improve editor features.
 
+# Overview
+
+The intention of Ruby LSP is to provide a fast, accurate, and robust coding environment for Ruby developers, especially for large codebases.
```
vinistock, line 9: Because we don't perform typechecking, our features are actually never 100% accurate, but rather are a best effort. For example, when we implement go to definition, we'll match methods based on name and arity, which is guaranteed to include incorrect results.

```suggestion
The intention of Ruby LSP is to provide a fast, robust and feature rich coding environment for Ruby developers. It does not perform typechecking, so its features are implemented in a best effort basis, attempting to be as accurate as possible.
```

README.md
```diff
@@ -4,6 +4,37 @@
...
+* Selection ranges
+
+When used on typed Sorbet code, additional features are available:
+
+* Quick documentation lookup when hovering over a Ruby or Rails method.
+* Support for Go To Definition.
+
+Planned future features include:
+
+* Context-aware auto-completion and navigation ("Go To Definition") for non-typed code ([prototype](https://github.com/Shopify/ruby-lsp/pull/429))
```
vinistock, line 27: What does context-aware mean in this case? Can it be confused with proper typechecking?

Also, let's avoid the usage of "non-typed code". It feels like we're explaining the Ruby LSP from a Sorbet standpoint, when we should just explain the Ruby LSP. It provides features for Ruby code.

README.md
```diff
@@ -4,6 +4,37 @@
...
+* Support for plug-ins to extend behavior
+
+ruby-lsp does not perform any type-checking or provide any type-related assistance, but it can be used alongside [Sorbet](https://github.com/sorbet/sorbet)'s LSP server.
+
+At the time of writing, these are the major differences between Ruby LSP and [Solargraph](https://solargraph.org/):
+
+* Solargraph [uses](https://solargraph.org/guides/yard) YARD documentation to gather information about your project and its gem dependencies. This provides functionality such as context-aware auto-completion and navigation ("Go To Definition")
+* Solargraph can be used as a globally installed gem, but Ruby LSP must be added to the Gemfile or gemspec. (There are pros and cons to each approach)
```
vinistock, line 35: Similar to my other comment, I'd point out that this is only true for projects using RuboCop.

README.md
```diff
@@ -4,6 +4,37 @@
...
+* To parse Ruby code, Solargraph uses [Parser](https://github.com/whitequark/parser) and Ruby uses [Syntax Tree](https://github.com/ruby-syntax-tree/syntax_tree).
```
vinistock, line 36: When we chatted about describing technical details, the feedback was that it's not relevant for users deciding which one to use. Does it matter which parser is used underneath?

andyw8: You're right, I was trying to convey that it's built on top of a set of modern technologies, but that's probably too specific.

README.md
andyw8: Ah, hover didn't seem to be working without Sorbet when I tried previously so I thought there was some connection, but it's working now so something separate must have been happening.

README.md
andyw8: > What does context-aware mean in this case? Can it be confused with proper typechecking?

I suppose it depends how far https://github.com/Shopify/ruby-lsp/pull/429 goes, but I think at the minimum we'd would be able to auto-complete constant names? Context-aware is maybe a bit vague though, so I'll take that out.

README.md
```diff
@@ -4,6 +4,34 @@
...
+* Selection ranges
+
+It does not perform typechecking, so its features are implemented on a best-effort basis, aiming to be as accurate as possible.
+
+Planned future features include:
+
+* Auto-completion and navigation ("Go To Definition") ([prototype](https://github.com/Shopify/ruby-lsp/pull/429))
+* Support for plug-ins to extend behavior
+
+ruby-lsp does not perform any type-checking or provide any type-related assistance, but it can be used alongside [Sorbet](https://github.com/sorbet/sorbet)'s LSP server.
```
andyw8, line 28: This is now a somewhat duplicating what was said previously, but it's such key point that I think it's worth repeating.

README.md
```diff
@@ -4,6 +4,34 @@
 
 This gem is an implementation of the [language server protocol specification](https://microsoft.github.io/language-server-protocol/) for Ruby, used to improve editor features.
 
+# Overview
+
+The intention of Ruby LSP is to provide a fast, accurate, and robust coding environment for Ruby developers, especially for large codebases.
```
vinistock, line 9: Is this duplicated?

README.md
vinistock, line 28:
```suggestion
The Ruby LSP does not perform any type-checking or provide any type-related assistance, but it can be used alongside [Sorbet](https://github.com/sorbet/sorbet)'s LSP server.
```

README.md
andyw8: Indeed, fixed.

PR #451, opened 2023-01-11
Refactor locating nodes based on positions

### LINKED ISSUE #444: Hover targets are not specific enough
If a method invocation includes a block and we hover over the block's contents, we will still consider as if the user is hovering over the surrounding method invocation. This is not ideal since we would also like to provide hover for method invocations inside the block's body.

Take the following example: if we hover anywhere inside the block passed to `scope`, the Ruby LSP considers that we're hover over the scope invocation. However, we are hovering over the `all` invocation, for which we should also be able to provide hover functionality.

Screenshot 2023-01-09 at 3 05 49 PM

The issue probably comes from [locate_node_and_parent](https://github.com/Shopify/ruby-lsp/blob/b3bc40845d8093e769b7951339b8d4a2a3e4744e/lib/ruby_lsp/requests/base_request.rb#L68). We need to make sure we always match the most specific node possible (i.e.: the node for which its location's boundaries are the closest as possible to the position provided by the client).

vinistock:
### Motivation

Closes #444 

When locating specific nodes on the AST, we were always returning the first node that matched a list of classes and covered the requested position. However, that is sometimes not precise enough.

For example, when hovering inside a block of a method call, we shouldn't consider it as the original invocation - but that was happening because the first node found `MethodAddBlock` already covered the position.
```ruby
scope :something do
  # hovering over this where should display information about `where`, but because the MethodAddBlock node
  # for scope was hit first, we always displayed the hover information for `scope`
  where(...)
end
```

### Implementation

I split it by commit to make it easier to review.
1. Create a new locate method and add a few tests for it. It keeps trying to find a node that has the narrowest range that still covers the requested position. I also removed the logic to match against specific node types, I think that belongs in the requests that are trying to find nodes
2. Migrate document highlight to the new locate method
3. Migrate hover to the new locate method. Now that we're finding more specific nodes, hovering over the ActiveRecord namespace doesn't trigger the request. It needs to be over the Base class, so I updated the test
4. Remove the old implementation of locate

### Automated Tests

Added new automated tests for base request.

### Manual Tests

2. Open our fixture files for Rails apps
3. Play around with hovering over DSL methods, like `validates`
4. Verify hover information is displayed properly

### DIFF
--- a/lib/ruby_lsp/requests/base_request.rb
+++ b/lib/ruby_lsp/requests/base_request.rb
@@ -84,6 +84,43 @@ module RubyLsp
         end
       end
 
+      sig do
+        params(
+          node: SyntaxTree::Node,
+          position: Integer,
+        ).returns([T.nilable(SyntaxTree::Node), T.nilable(SyntaxTree::Node)])
+      end
+      def locate(node, position)
+        queue = T.let(node.child_nodes.compact, T::Array[T.nilable(SyntaxTree::Node)])
+        closest = node
+
+        until queue.empty?
+          candidate = queue.shift
+
+          # Skip nil child nodes
+          next if candidate.nil?
+
+          # Add the next child_nodes to the queue to be processed
+          queue.concat(candidate.child_nodes)
+
+          # Skip if the current node doesn't cover the desired position
+          loc = candidate.location
+          next unless (loc.start_char...loc.end_char).cover?(position)
+
+          # If the node's start character is already past the position, then we should've found the closest node already
+          break if position < loc.start_char
+
+          # If the current node is narrower or equal than the previous closest node, than it is more precise
+          closest_loc = closest.location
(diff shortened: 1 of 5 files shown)


lib/ruby_lsp/requests/base_request.rb
```diff
@@ -60,28 +60,39 @@ def full_constant_name(node)
...
+          queue.concat(candidate.child_nodes)
+
+          # Skip if the current node doesn't cover the desired position
+          loc = candidate.location
+          next unless (loc.start_char...loc.end_char).cover?(position)
+
+          # If the node's start character is already past the position, then we should've found the closest node already
+          break if position < loc.start_char
+
+          # If the current node is narrower or equal than the previous closest node, than it is more precise
```
andyw8, line 87:
```suggestion
          # If the current node is narrower than or equal to the previous closest node, then it is more precise
```

Issue #250, opened 2022-08-15
Inform users of failing cops


vinistock:
When running RuboCop on files for both diagnostics and formatting, a bug may cause a cop to fail. The default behaviour is for RuboCop to print to `stderr` that a specific check failed and continue running without that particular cop.

While both diagnostics and formatting will still work (just ignoring the failing cop), we may want to display a message to users to let them know that the cop is failing to process the file they're currently working on.

vinistock:
Let's put this on pause and see if they're open to raising a new custom error for this. The current options are brittle or hacky and if we can get first class support for it, it'll be better. Thanks for looking into it!

wildmaples:
With the hopes of moving this issue forward, I've put up a PR on Rubocop for the proposed option: https://github.com/rubocop/rubocop/issues/10977

wildmaples:
Update: The [latest release of Rubocop v1.38](https://github.com/rubocop/rubocop/releases/tag/v1.38.0) contains the fix we need to resolve this issue: https://github.com/rubocop/rubocop/pull/11001

andyw8:
I can take a look at this now that the RuboCop fix is in (since v1.38.0).

PR #454, opened 2023-01-13
Inform users of failing cops

andyw8:
### Motivation

Closes https://github.com/Shopify/ruby-lsp/issues/250

### Implementation

See my GitHub comments for notes.

### Automated Tests

There doesn't seem to be any existing tests for `rubocop_runner.rb` so I was unsure how to approach this... opinions welcome on whether we need tests for this.

### Manual Tests

* Check out the branch and point your repo's Gemfile to it. Add a corrupt cop to the root, e.g.:

```
module CustomCops
  class Duffy < RuboCop::Cop::Base
    def on_send(node)
       raise "fake failure!"
    end
  end
end
```
* Require the cop from your `.rubocop.yml`:

```yml
require:
  - ./duffy.rb
```

* Run `bundle exec rubocop .` to ensure you see RuboCop complain.

* Restart Ruby LSP within VS Code and open a Ruby file.

* Ensure you see an error pop up which matches the one on the command-line.

### DIFF
--- a/lib/ruby_lsp/requests/support/rubocop_runner.rb
+++ b/lib/ruby_lsp/requests/support/rubocop_runner.rb
@@ -16,6 +16,22 @@ end
 module RubyLsp
   module Requests
     module Support
+      class InternalRuboCopError < StandardError
+        extend T::Sig
+
+        MESSAGE = <<~EOS
+          An internal error occurred for the %s cop.
+          Updating to a newer version of RuboCop may solve this.
+          For more details, run RuboCop on the command line.
+        EOS
+
+        sig { params(rubocop_error: RuboCop::ErrorWithAnalyzedFileLocation).void }
+        def initialize(rubocop_error)
+          message = format(MESSAGE, rubocop_error.cop.name)
+          super(message)
+        end
+      end
+
       # :nodoc:
       class RuboCopRunner < RuboCop::Runner
         extend T::Sig
@@ -31,9 +47,17 @@ module RubyLsp
             "--force-exclusion",
             "--format",
             "RuboCop::Formatter::BaseFormatter", # Suppress any output by using the base formatter
-          ].freeze,
+          ],
           T::Array[String],
         )
+
+        begin
+          RuboCop::Options.new.parse(["--raise-cop-error"])
+          DEFAULT_ARGS << "--raise-cop-error"
+        rescue OptionParser::InvalidOption
+          # older versions of RuboCop don't support this flag
+        end
+        DEFAULT_ARGS.freeze
 
         sig { params(args: String).void }
         def initialize(*args)
@@ -63,6 +87,8 @@ module RubyLsp
           raise Formatting::Error, error.message
         rescue RuboCop::ValidationError => error
           raise ConfigurationError, error.message
+        rescue RuboCop::ErrorWithAnalyzedFileLocation => error
+          raise InternalRuboCopError, error
         end
 
         sig { returns(String) }


lib/ruby_lsp/requests/support/rubocop_runner.rb
```diff
@@ -61,6 +62,11 @@ def run(path, contents)
           super([path])
         rescue RuboCop::Runner::InfiniteCorrectionLoop => error
           raise Formatting::Error, error.message
+        rescue RuboCop::ErrorWithAnalyzedFileLocation => error
+          line = ":#{error.line}" if error.line
+          column = ":#{error.column}" if error.column
+          location = "#{path}#{line}#{column}"
+          raise Formatting::Error, "An error occurred while #{error.cop.name} cop was inspecting #{location}"
```
andyw8, line 69: This matches the message generated by RuboCop on the command line:
https://github.com/rubocop/rubocop/blob/5be2df73ec57e8a7615461b0006a3ab70cb16999/lib/rubocop/cop/team.rb#L247

lib/ruby_lsp/requests/support/rubocop_runner.rb
```diff
@@ -61,6 +62,11 @@ def run(path, contents)
           super([path])
         rescue RuboCop::Runner::InfiniteCorrectionLoop => error
           raise Formatting::Error, error.message
```
andyw8, line 64: `Formatting::Error` may not be appropriate here, since the error could occur either during linting _or_ formatting...

lib/ruby_lsp/requests/support/rubocop_runner.rb
```diff
@@ -61,6 +62,11 @@ def run(path, contents)
...
+        rescue RuboCop::ErrorWithAnalyzedFileLocation => error
+          line = ":#{error.line}" if error.line
```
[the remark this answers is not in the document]
shopify-aiden: nit: Should we check for line && column so we don't have `<path>:<column>`. I am guessing they will never be separate, but might be worth doing explicitly?

lib/ruby_lsp/requests/support/rubocop_runner.rb
```diff
@@ -29,6 +29,7 @@ class ConfigurationError < StandardError; end
           [
             "--stderr", # Print any output to stderr so that our stdout does not get polluted
             "--force-exclusion",
+            "--raise-cop-error",
```
vinistock, line 32: We'll need to add a RuboCop version check and conditionally include this flag since it only exists after a certain version. Otherwise, it'll raise an error and complain about a non-existing flag.

lib/ruby_lsp/requests/support/rubocop_runner.rb
vinistock: Are you talking about `InfiniteCorrectionLoop`? That error can only occur when running with `--auto-correct` which we only do in formatting.

lib/ruby_lsp/requests/support/rubocop_runner.rb
vinistock: UX question: since this error will indicate an issue in a cop's implementation, which
- Not every developer will know how to fix
- Will require in depth debugging of RuboCop separately
Would it be enough to just mention which cop failed without fully reconstructing the location?

lib/ruby_lsp/requests/support/rubocop_runner.rb
andyw8: Sorry, commented on the wrong line, I was referring to the new `raise` I added below.

vinistock: Oh, I see. We can create a parent class for the formatting and diagnostics error and then rescue from that parent class.

lib/ruby_lsp/requests/support/rubocop_runner.rb
andyw8: I thought about that, but might make the message overly long. It's mentioned in the default bug report template: https://github.com/rubocop/rubocop/issues/new?assignees=&labels=&template=bug_report.md

lib/ruby_lsp/server.rb
```diff
@@ -4,6 +4,22 @@
 require "ruby_lsp/internal"
 
 module RubyLsp
+  class InternalRubocopError < StandardError
```
vinistock, line 7: Can we move this to RuboCop runner?

lib/custom_cops/duffy.rb
```diff
@@ -0,0 +1,10 @@
+# typed: false
+# frozen_string_literal: true
+
+module CustomCops
```
paracycle, line 4: I don't think this file should be in the `lib` folder, since it is not relevant to the operation of the gem. IMO, it shouldn't even be packaged into the gem bundle.

lib/ruby_lsp/requests/support/rubocop_runner.rb
```diff
@@ -76,6 +78,11 @@ def formatted_source
         def file_finished(_file, offenses)
           @offenses = offenses
         end
+
+        sig { returns(T::Boolean) }
+        def rubocop_version_supports_raise_cop_error?
+          Gem::Version.new(@rubocop_version) >= Gem::Version.new("1.38.0")
```
paracycle, line 84: I don't think we should be doing version checks. Can't we do feature detection? We have access to the whole `RuboCop` namespace, so we can just check for the existence of `RuboCop::ErrorWithAnalyzedFileLocation`

lib/ruby_lsp/server.rb
```diff
@@ -231,6 +247,8 @@ module RubyLsp
       uri = request.dig(:params, :textDocument, :uri)
       response = store.cache_fetch(uri, :diagnostics) do |document|
         Requests::Diagnostics.new(uri, document).run
+      rescue RuboCop::ErrorWithAnalyzedFileLocation => error
```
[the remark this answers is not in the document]
paracycle: There is a bigger problem here: `RuboCop::ErrorWithAnalyzedFileLocation` only exists past the PR that introduces `--raise-cop-error` in the first place. For older versions of RuboCop, this line will fail with `NameError`.

.rubocop.yml
```diff
@@ -5,6 +5,8 @@ require:
  - rubocop-sorbet
  - rubocop-minitest
  - rubocop-rake
+ # enable this to test handling of failing cops
+ # - ./test/fixtures/custom_cops/duffy
```
vinistock, line 9: This is helpful for verifying the behaviour on this PR, but I doubt we'll ever need this again. We should remove these associated configs and the duffy cop before shipping.

lib/ruby_lsp/requests/support/rubocop_runner.rb
andyw8: `RuboCop::ErrorWithAnalyzedFileLocation` already existed prior to the flag, but I'll see if there's another way.

paracycle: Ah right, sorry I read the PR wrong.

This seems to work for doing feature detection, btw:
```ruby
irb> RuboCop::Options.new.parse(["--foo", "bar"])
OptionParser::InvalidOption: invalid option: --foo
```

lib/ruby_lsp/server.rb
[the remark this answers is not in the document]
paracycle: Ok, strike that comment since the exception predates the PR that added the `raise-cop-error` option.

lib/ruby_lsp/requests/support/rubocop_runner.rb
paracycle: Confirmed that this works:
```ruby
[1] pry> gem("rubocop", "1.36")
=> true
[2] pry(main)> require "rubocop"
=> true
[3] pry(main)> RuboCop::Options.new.parse(["--raise-cop-error"])
OptionParser::InvalidOption: invalid option: --raise-cop-error
from /Users/ufuk/.gem/ruby/3.1.2/gems/rubocop-1.36.0/lib/rubocop/options.rb:28:in `parse'
```
vs
```ruby
[1] pry(main)> gem("rubocop", "1.39")
=> true
[2] pry(main)> require "rubocop"
=> true
[3] pry(main)> RuboCop::Options.new.parse(["--raise-cop-error"])
=> [{:raise_cop_error=>true}, []]
```

andyw8:
Thanks for all the feedback, everything should now be addressed.

I have remaining question about testing to resolve (see the comment in the PR).

test/requests/support/rubocop_runner_test.rb
```diff
@@ -0,0 +1,26 @@
...
+      class RuboCopRunnerTest < Minitest::Test
+        def test_adds_raise_cop_error_flag_if_rubocop_version_supports_it
+          runner = RuboCopRunner.new("--lint")
+
+          assert(runner.instance_variable_get(:@options).fetch(:raise_cop_error))
+        end
+
+        def test_does_not_add_raise_cop_error_flag_if_rubocop_version_does_not_support_it
+          skip("fix this")
+          runner = RuboCopRunner.new("--lint")
```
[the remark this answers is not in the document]
paracycle: IMO, I think that's fine as long we test the full behaviour that depends on that, i.e. setup an environment that raises on that flag, and then test a real RuboCop failure and sure that we just process the files and not show any errors.

In order to test a real RuboCop failure, we can also programmatically stub other RuboCop methods so that our custom cop is included in the run as well, for example.

The alternative way to test it is to test against the real old version of RuboCop and that means forking or shelling out to a separate process, which is what we do on Tapioca. However, that is more time consuming and harder to get right, and seen as our implementation is tightly coupled with how RuboCop behaves anyway, I think stubbing RuboCop behaviour is fine.

lib/ruby_lsp/requests/support/rubocop_formatting_runner.rb
```diff
@@ -29,6 +29,8 @@ def run(uri, document)
           @runner.run(filename, document.source)
 
           @runner.formatted_source
+        rescue RuboCop::ErrorWithAnalyzedFileLocation => error
```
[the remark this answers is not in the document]
paracycle: Can't we move this `rescue` and the one in the diagnostics runner and move them both inside the `RuboCopRunner#run` method?

lib/ruby_lsp/requests/support/rubocop_runner.rb
```diff
@@ -76,6 +93,14 @@ def formatted_source
...
           @offenses = offenses
         end
+
+        sig { returns(T::Boolean) }
+        def rubocop_version_supports_raise_cop_error?
+          RuboCop::Options.new.parse(["--raise-cop-error"])
+          true
+        rescue OptionParser::InvalidOption
+          false
+        end
```
paracycle, line 103: This implementation means that we will likely keep calling `parse` multiple times to detect the same unchanging feature. We should either memoize this value or do this operation at top level and assign the result to a constant to make it always run once. With memoization it would be like:
```suggestion
        sig { returns(T::Boolean) }
        def rubocop_version_supports_raise_cop_error?
          return @raises_cop_error if defined?(@raises_cop_error)
          
          @raises_cop_error = begin
            RuboCop::Options.new.parse(["--raise-cop-error"])
            true
          rescue OptionParser::InvalidOption
            false
          end
        end
```

test/requests/support/rubocop_runner_test.rb
```diff
@@ -0,0 +1,26 @@
...
+
+          refute(runner.instance_variable_get(:@options).key?(:raise_cop_error))
```
paracycle, line 21: Actually, I think this assertion is assuming more about the internals of our code than stubbing RuboCop internals.

lib/ruby_lsp/requests/support/rubocop_runner.rb
```diff
@@ -57,6 +57,7 @@ def initialize(*args)
           @offenses = T.let([], T::Array[RuboCop::Cop::Offense])
           @errors = T.let([], T::Array[String])
           @warnings = T.let([], T::Array[String])
+          @raises_cop_error = T.let(nil, T.nilable(T::Boolean))
```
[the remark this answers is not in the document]
andyw8: Maybe...? I've pushed an update to show how that would look.

PR #455, opened 2023-01-13
Stop trying to maintain a parsable source

vinistock:
### Motivation

After #323, trying to maintain features working on a virtual parsable source has no longer been working properly. The reality is that it leads to more complex code, errors that are difficult to understand and worse performance (since we'll try to parse edits twice while there are syntax errors).

I vote in favour of removing this. We'll just need to wait for YARP for actual error tolerance and proper diagnostics.

### Implementation

Just removed everything related to parsable source and inlined the `apply_edits` method.

### Automated Tests

We didn't test parsable source, so they should just continue to pass.

### DIFF
--- a/lib/ruby_lsp/document.rb
+++ b/lib/ruby_lsp/document.rb
@@ -20,7 +20,6 @@ module RubyLsp
       @cache = T.let({}, T::Hash[Symbol, T.untyped])
       @encoding = T.let(encoding, String)
       @source = T.let(source, String)
-      @parsable_source = T.let(source.dup, String)
       @unparsed_edits = T.let([], T::Array[EditShape])
       @tree = T.let(SyntaxTree.parse(@source), T.nilable(SyntaxTree::Node))
     rescue SyntaxTree::Parser::ParseError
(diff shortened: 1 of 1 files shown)


lib/ruby_lsp/document.rb
```diff
@@ -63,9 +69,8 @@ def parse
 
       @tree = SyntaxTree.parse(@source)
       @unparsed_edits.clear
-      @parsable_source = @source.dup
-    rescue SyntaxTree::Parser::ParseError
-      update_parsable_source(@unparsed_edits)
+    rescue StandardError
```
shopify-aiden, line 72: Do we want to suppress other errors?

shopify-aiden (approved):
This makes sense, especially since we aren't reporting diagnostics the same either.

lib/ruby_lsp/document.rb
st0012: I'm also a bit worry about this biting us later in the future. Can we be more specific on the error class?

lib/ruby_lsp/document.rb
```diff
@@ -50,8 +49,15 @@ def cache_fetch(request_name, &block)
 
     sig { params(edits: T::Array[EditShape]).void }
     def push_edits(edits)
-      # Apply the edits on the real source
-      edits.each { |edit| apply_edit(@source, edit[:range], edit[:text]) }
+      edits.each do |edit|
+        range = edit[:range]
+        scanner = Scanner.new(source, @encoding)
```
st0012, line 54:
```suggestion
        scanner = create_scanner
```

lib/ruby_lsp/document.rb
vinistock: I switched back to parser errors. The reason I had changed to `StandardError` is because we've been getting some `Comment location overlaps node location` errors, which are `RuntimeError` and notoriously difficult to reproduce. But indeed this may mask other problems, I agree.

PR #462, opened 2023-01-20
Properly place `end` when there's extra content after keyword

vinistock:
### Motivation

The original functionality for completing the `end` tokens on keywords didn't take into account breaking the line in the middle, which resulted in weird behaviour. For example
```ruby
if something
# ^ If you break the line here it results in

if
endsomething
```
which is of course not the experience we want. We should make sure the extra content after the line break is properly placed in the middle like this
```ruby
if
  something
end
```

### Implementation

- If the `@previous_line` includes a line break, that means there's content **after** the line break that triggered this on type formatting completion
- In that case, we want to place the `end` token one line below the extra content
- The only corner case is that, if the document does not have enough lines (e.g.: a single line document with an if statement), then we need to insert an extra line break ourselves

### Automated Tests

Added a new example.

### Manual Tests

2. Open any file or create a new one
3. Type `if something`
4. Break the line right before `something`
5. Verify that the `end` token is inserted in the right place
6. Verify that the cursor is moved back into the conditional of the if statement

Play around with other keywords and scenarios.

### DIFF
--- a/lib/ruby_lsp/requests/on_type_formatting.rb
+++ b/lib/ruby_lsp/requests/on_type_formatting.rb
@@ -32,8 +32,8 @@ module RubyLsp
 
         scanner = document.create_scanner
         line_begin = position[:line] == 0 ? 0 : scanner.find_char_position({ line: position[:line] - 1, character: 0 })
-        line_end = scanner.find_char_position(position)
-        line = T.must(@document.source[line_begin..line_end])
+        @line_end = T.let(scanner.find_char_position(position), Integer)
+        line = T.must(@document.source[line_begin..@line_end])
 
         @indentation = T.let(find_indentation(line), Integer)
         @previous_line = T.let(line.strip.chomp, String)
@@ -84,8 +84,30 @@ module RubyLsp
 
         indents = " " * @indentation
 
-        add_edit_with_text(" \n#{indents}end")
-        move_cursor_to(@position[:line], @indentation + 2)
+        if @previous_line.include?("\n")
+          # If the previous line has a line break, then it means there's content after the line break that triggered
+          # this completion. For these cases, we want to add the `end` after the content and move the cursor back to the
+          # keyword that triggered the completion
+
+          line = @position[:line]
+
+          # If there are enough lines in the document, we want to add the `end` token on the line below the extra
+          # content. Otherwise, we want to insert and extra line break ourselves
+          correction = if T.must(@document.source[@line_end..-1]).count("\n") >= 2
+            line -= 1
+            "#{indents}end"
+          else
+            "#{indents}\nend"
+          end
+
+          add_edit_with_text(correction, { line: @position[:line] + 1, character: @position[:character] })
+          move_cursor_to(line, @indentation + 3)
+        else
+          # If there's nothing after the new line break that triggered the completion, then we want to add the `end` and
+          # move the cursor to the body of the statement
+          add_edit_with_text(" \n#{indents}end")
+          move_cursor_to(@position[:line], @indentation + 2)
+        end
       end
 
       sig { params(spaces: String).void }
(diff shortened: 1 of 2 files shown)


lib/ruby_lsp/requests/on_type_formatting.rb
```diff
@@ -84,8 +84,30 @@ def handle_statement_end
...
+        if @previous_line.include?("\n")
+          # If the previous line has a line break, then it means there's content after the line break that triggered
+          # this completion. For these cases, we want to add the `end` after the content and move the cursor back to the
+          # keyword that triggered the completion
+
+          line = @position[:line]
+
+          # If there are enough lines in the document, we want to add the `end` token on the line below the extra
+          # content. Otherwise, we want to insert and extra line break ourselves
+          correction = if T.must(@document.source[@line_end..-1]).count("\n") >= 2
```
bitwise-aiden, line 96: How does this play out in the case of: 

```ruby
if something
immediate_content
```

Will it result in: 
``` ruby 
if 
  something 
endimmediate_content
```

lib/ruby_lsp/requests/on_type_formatting.rb
```diff
@@ -84,8 +84,30 @@ def handle_statement_end
...
+            line -= 1
+            "#{indents}end"
+          else
+            "#{indents}\nend"
+          end
+
+          add_edit_with_text(correction, { line: @position[:line] + 1, character: @position[:character] })
+          move_cursor_to(line, @indentation + 3)
```
bitwise-aiden, line 104: Do we want to restore the initial cursor? I would expect the cursor to be: 
``` ruby 
if 
   something
# ^ cursor
```

This reads as behind `end`, so correct me if I'm wrong.

lib/ruby_lsp/requests/on_type_formatting.rb
vinistock: Good catch. Yeah, indeed we end up in a scenario like that. I'll think about how we can avoid it.

lib/ruby_lsp/requests/on_type_formatting.rb
vinistock: The idea of moving the cursor would kinda be to "guess" where the user will type next. If you're breaking the line there, then you'd need to fill in the conditional next. Not sure if it matches everybody's expectation though. What do you think?

bitwise-aiden: Oh right! That makes more sense. I was wondering how line was updated to the line for end, but makes more sense that it's at the end of `if `. 

Personally I'd expect to be where I pointed out in the above snippet. My litmus test would be: If I am typing while looking away from the screen, what would me mental cursor be. For me that's after the new line. So I'd expect to hit enter then have to press up to get back to it. 

That said, that could be a bias of not having a tool that is trying to allow for auto complete like that. 

Another case that could be interesting is this one: 
``` ruby 
if (something
```

If the user were to hit enter after the `(` but before the `something`, it would indicate that they are handling a multi-line conditional. 

Though is that handled differently because it's inside `()`?

test/requests/on_type_formatting_test.rb
```diff
@@ -152,4 +152,27 @@ def test_comment_continuation_when_inserting_new_line_in_the_middle
...
+        range: { start: { line: 1, character: 2 }, end: { line: 1, character: 2 } },
+        newText: " \nend",
+      },
+      {
+        range: { start: { line: 1, character: 2 }, end: { line: 1, character: 2 } },
+        newText: "$0",
+      },
+    ]
+
+    assert_equal(expected_edits.to_json, T.must(edits).to_json)
```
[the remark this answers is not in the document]
andyw8: Yes, fixed, thanks.

st0012 (approved):

lib/ruby_lsp/requests/on_type_formatting.rb
vinistock: I frankly haven't found an easy way to handle this. How do you feel about merging this one and trying to address this other case later?

To fix this, we would need to perform some lookahead operations to figure out what exists in the next line.

bitwise-aiden: I'm happy with that

bitwise-aiden (approved):

PR #469, opened 2023-01-26
Avoid running RuboCop outside of project folder

### LINKED ISSUE #401: RuboCop errors when opening files outside of the current project
If a developer has a project open in the editor and then decides to open a file from a different project, RuboCop functionality breaks all over the place.

I believe what happens is that the LSP is running with the bundle from the current project, but tries to read the configuration from from some other folder when opening those external files. Not sure if it's looking for `.rubocop.yml` in the other file's project root folder or in the same folder the file is in.

This results in all sorts of errors triggering, like not being able to require RuboCop extensions or invalid custom cops that don't exist in that external's file project.

We should reflect on how to make the experience of opening external files better. I believe disabling diagnostics and formatting makes sense, since most of the time external files are probably only going to be opened as a reference and not actually for editing.

vinistock:
### Motivation

Closes #401 

If a user has the LSP open a project and decides to open an external file that belongs to a different project, RuboCop will try to load the `.rubocop.yml` configuration from the folder where that external file exists.

This will always fail if there are RuboCop plugins in that configuration since Bundler configured the load paths using the current project's `Gemfile` and not the one where the external file is. For example:
- Bundler sets up load paths for the `Gemfile` in the main project (includes rubocop and rubocop-performance)
- User opens external file in different folder. This other folder's configuration requires rubocop-rails
- Because rubocop-rails is not present in the main project's bundle, it is not in the load paths and requiring it fails

This is not a major issue in the experience, but it fills our telemetry with error reports that aren't very useful.

### Implementation

We wouldn't be able to re-configure load paths with Bundler just for the external file, so I don't believe there's any practical way of making RuboCop functionality simply work in them.

The approach here is to just return `nil` if the file URI doesn't belong to the same folder where the LSP process is running (`Dir.pwd`).

**Note**

This is a RuboCop specific issue, because it tries to load configurations found in any folder. Other requests can continue to work normally in external files, such as semantic highlighting.

### Automated Tests

Added one for diagnostics and one for formatting.

### Manual Tests

1. Start the LSP for this branch on another project `gem "ruby-lsp", require: false, path: "ruby-lsp/folder/here"`
2. Verify diagnostics and formatting work on any file that belongs to the project
3. Open a file that belongs to a different project that also uses RuboCop
4. Verify diagnostics and formatting do not work

### DIFF
--- a/lib/ruby_lsp/requests/diagnostics.rb
+++ b/lib/ruby_lsp/requests/diagnostics.rb
@@ -30,15 +30,20 @@ module RubyLsp
 
       sig do
         override.returns(
-          T.any(
-            T.all(T::Array[Support::RuboCopDiagnostic], Object),
-            T.all(T::Array[Support::SyntaxErrorDiagnostic], Object),
+          T.nilable(
+            T.any(
+              T.all(T::Array[Support::RuboCopDiagnostic], Object),
+              T.all(T::Array[Support::SyntaxErrorDiagnostic], Object),
+            ),
           ),
         )
       end
       def run
-        return [] if @document.syntax_error?
-        return [] unless defined?(Support::RuboCopDiagnosticsRunner)
+        return if @document.syntax_error?
+        return unless defined?(Support::RuboCopDiagnosticsRunner)
+
+        # Don't try to run RuboCop diagnostics for files outside the current working directory
+        return unless @uri.sub("file://", "").start_with?(Dir.pwd)
 
         Support::RuboCopDiagnosticsRunner.instance.run(@uri, @document)
       end
--- a/lib/ruby_lsp/executor.rb
+++ b/lib/ruby_lsp/executor.rb
@@ -223,7 +223,7 @@ module RubyLsp
       Requests::InlayHints.new(document, start_line..end_line).run
     end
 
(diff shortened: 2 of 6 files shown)


lib/ruby_lsp/requests/diagnostics.rb
```diff
@@ -30,15 +30,20 @@ def initialize(uri, document)
...
         )
       end
       def run
-        return [] if @document.syntax_error?
-        return [] unless defined?(Support::RuboCopDiagnosticsRunner)
+        return if @document.syntax_error?
+        return unless defined?(Support::RuboCopDiagnosticsRunner)
+
+        # Don't try to run RuboCop diagnostics for files outside the current working directory
+        return unless @uri.sub("file://", "").start_with?(Dir.pwd)
```
andyw8, line 46: What was your thinking around returning nils instead of empty arrays as we did previously?

vinistock: This is just my preference, but I think returning `nil` better conveys that there are no diagnostics for this file rather than an empty array.

bitwise-aiden: I support this decision

Issue #335, opened 2022-10-19
Create a benchmark suite for the Ruby LSP


vinistock:
Performance is critical when it comes to language servers. Slowness will often make developers frustrated and has the potential to make their editor feel clunky or unstable.

We should have some sort of CI step that benchmarks our requests against fixtures of a reasonable size (e.g.: a fake Rails model with 800+ lines).

Defining what passing or failing this check means could be tricky, but it would be nice to at least get a report commented on PRs saying whether it made things slower or faster by a significant amount.

### Idea

Maybe we could have an automated step that
- runs benchmark-ips on main (using the `hold!` functionality to benchmark through different executions)
- switches to the PR branch and re-runs the benchmarks
- captures the report
- comments the results on the PR (if possible, re-using the same comment for different commits so that PRs don't become cluttered with reports)

PR #471, opened 2023-01-27
Add benchmark script

vinistock:
### Motivation

Step towards #335 

Performance is critical in language servers and we need some quick way to validate that we're not introducing major slowness into our requests.

### Implementation

This PR creates a benchmark script that prints a list of results. All requests are executed through the `Executor` class by sending it a method name and the parameters.

The next step is to use GitHub actions cache to save results and then compare results from `main` vs branches used in PRs, informing of performance differences.

### DIFF
--- a/bin/benchmark
+++ b/bin/benchmark
@@ -1,0 +1,180 @@
+#!/usr/bin/env ruby
+# typed: true
+# frozen_string_literal: true
+
+$LOAD_PATH.unshift(File.expand_path("../lib", __dir__))
+$VERBOSE = nil
+
+require "ruby_lsp/internal"
+require "benchmark"
+
+ITERATIONS = 50
+
+def avg_bench(method, params)
+  (0...ITERATIONS).map do
+    # Create a new store every time to prevent caching
+    store = RubyLsp::Store.new
+    store.set(FILE_URI, FIXTURE)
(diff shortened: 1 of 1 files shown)


bitwise-aiden (approved):

bin/benchmark
```diff
@@ -0,0 +1,178 @@
...
+# frozen_string_literal: true
+
+$LOAD_PATH.unshift(File.expand_path("../lib", __dir__))
+$VERBOSE = nil
+
+require "ruby_lsp/internal"
+require "benchmark"
+
+def avg_bench(method, params)
+  (0...50).map do
```
andyw8, line 12: Maybe pull the `50` out to a constant?

bin/benchmark
```diff
@@ -0,0 +1,178 @@
...
+    # Parse ahead of time or else one of the requests will do it
+    store.get(FILE_URI).parse
+
+    Benchmark.realtime do
+      RubyLsp::Executor.new(store).execute({
+        method: method,
+        params: params,
+      })
+    end
+  end.reduce(&:+).to_f / 50
```
andyw8, line 25:
```suggestion
  end.sum.to_f / 50
```

andyw8 (approved):

PR #478, opened 2023-02-01
Run benchmarks on CI

### LINKED ISSUE #335: Create a benchmark suite for the Ruby LSP

vinistock:
### Motivation

Closes #335 

Having the benchmark script added in #471 is nice, but we need to run the script regularly to get feedback on what's happening in terms of performance when we make changes. The ideal scenario is to get performance feedback directly on PRs.

### Implementation

Added a new workflow to run benchmarks. The steps are
- Restore the cached benchmark results from the `main` branch, which will be our point of comparison
- Run the benchmark script. This will benchmark the code in the current PR and compare it with the results restore from cache. It will also always overwrite the results file, but that won't be saved in the cache for pull requests
- The final step is to save the results file in the cache only once the commits are merged into main to update our baseline

### Concerns

Not sure if this is the best way to cache results and compare. Any other ideas?


bin/benchmark
```diff
@@ -5,10 +5,12 @@
 $LOAD_PATH.unshift(File.expand_path("../lib", __dir__))
 $VERBOSE = nil
 
+require "bundler/setup"
 require "ruby_lsp/internal"
 require "benchmark"
 
-ITERATIONS = 50
+ITERATIONS = 500
```
vinistock, line 12: Had to bump the number of iterations to get a more reliable average. Using 50 iterations made the script always fail because there was also some variance.

bin/benchmark
```diff
@@ -178,3 +180,33 @@ longest_key_length = requests.keys.max_by(&:length).length
...
+    if absolute_ratio < 0.15
+      puts "#{adjusted_method} unchanged"
+    elsif ratio.negative?
+      puts "#{adjusted_method} slower by #{percentage} %"
+    else
+      puts "#{adjusted_method} faster by #{percentage} %"
+    end
+  end
+
+  unless success
```
KaanOzkan, line 206: This is always `true` since we don't write to it when we're slower. I'd think we'd want to fail the CI step so that we're notified when a particular PR introduces slowdown?

st0012 (approved):

KaanOzkan (approved):
I'd love to have comments as your issue pointed out so that we get notified when a PR improves performance but this is a great first step.

TIL: Cache action

vinistock:
@KaanOzkan I'll look into how we can do that next.

PR #481, opened 2023-02-03
Post results from benchmark as a comment on PRs

vinistock:
### Motivation

Final part of #335 

Post the results of running benchmarks as a comment on the PR.

### Implementation

1. Push the STDOUT results of running `bin/benchmark` into the GitHub actions output variable
2. Use github-script to take that output and post it as a comment

### DIFF
--- a/.github/workflows/benchmark.yml
+++ b/.github/workflows/benchmark.yml
@@ -7,7 +7,11 @@ jobs:
     runs-on: ubuntu-latest
     name: Benchmark
     steps:
-      - uses: actions/checkout@v3
+      # Setup the main branch to run benchmarks there as a baseline for comparison
+      - name: Checkout to main
+        uses: actions/checkout@v3
+        with:
+          ref: main
 
       - name: Set up Ruby
         uses: ruby/setup-ruby@v1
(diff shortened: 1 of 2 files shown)


KaanOzkan (approved):

andyw8:
I see in `bin/benchmark` we're using `Benchmark.realtime`, but wouldn't `Benchmark.measure` be more accurate since it provides the CPU time?

vinistock:
I've made some changes to try to make the benchmark more stable. It now
- Considers any different smaller than a standard deviation as unchanged
- Uses only user cpu time (time spent in Ruby)
- Bumped the number of iterations
- Did some preloading of things that could impact results
- Changed the position and added some `source://` comments, so that `DocumentLink` and `Hover` have actual results to process

vinistock:
@dirceu gave a great suggestion to run both the main and the branch benchmark on the same CI run to avoid some variance. This also allows us to get rid of using cache.

Currently, I believe the benchmark will always fail because I added some extra things to the fixture, but once this is merged it should be considerably more stable.

dirceu (approved):

PR #488, opened 2023-02-07
Refactor code actions to use context instead of our own cache

### LINKED ISSUE #487: Refactor code actions to use context instead of cached diagnostics
According to the [specification](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#textDocument_codeAction), the code action request actually receives all previously reported diagnostics in the `context` parameter.

We can probably refactor our implementation to use that parameter instead of reading from the [cached diagnostics](https://github.com/Shopify/ruby-lsp/blob/7d7c9d5874dafcdc9a7368b7916c7221970a34c5/lib/ruby_lsp/requests/code_actions.rb#L38).

vinistock:
### Motivation

Closes #487 

Instead of caching diagnostics ourselves, we should leverage the specification and use what's already available. Every code action request receives a `context` object, which contains the current diagnostics for the file, so there's no point in reading from our cache.

**Note**: the reason this is coming during this week is because it helps my hackdays idea 😂.

### Implementation

- Removed a ton of unused code, especially related to syntax errror diagnostics which we no longer produce
- Started returning the code actions straight from the diagnostics request inside the `data` attribute which is always preserved between a diagnostic -> code action request
- Simplified code actions to use the `context` parameter rather than trying to read from our own cache

### Automated Tests

Had to change a bunch of tests. Some didn't make sense anymore, others needed to have expectations updated since the response changed for diagnostics and code actions.

### Manual Tests

Verify code actions still work. For example
1. Open the LSP on this branch
2. Make a violation (e.g.: wrong indentation)
3. Hover over the violation
4. Use the quickfix
5. Verify that the violation has been fixed


lib/ruby_lsp/requests/code_actions.rb
```diff
@@ -24,26 +24,30 @@ class CodeActions < BaseRequest
...
-        end
-        return [] if corrections.empty?
+        diagnostics.filter_map do |diagnostic|
+          code_action = diagnostic.dig(:data, :code_action)
+          range = code_action.dig(:edit, :documentChanges, 0, :edits, 0, :range)
 
-        T.cast(corrections, T::Array[Support::RuboCopDiagnostic]).map!(&:to_lsp_code_action)
+          if diagnostic.dig(:data, :correctable) && @range.cover?(range.dig(:start, :line)..range.dig(:end, :line))
+            code_action
+          end
```
omegahm, line 49: If the `code_action` key in the `diagnostic` hash is `nil` this will fail.

```suggestion
          code_action = diagnostic.dig(:data, :code_action)
          
          if code_action
            range = code_action.dig(:edit, :documentChanges, 0, :edits, 0, :range)

            if diagnostic.dig(:data, :correctable) && @range.cover?(range.dig(:start, :line)..range.dig(:end, :line))
              code_action
            end
          end
```

st0012 (commented):
Looks good 👍 
Just some thoughts:

- It'd help reviewing if the `LanguageServer::Protocol::` removal has its own commit
- Since it changes the server's behaviour, should this be categorised as a fix instead of refactoring?

vinistock:
@st0012 That's a good point, I split the changes to use aliased constants into a separate commit.

I would still consider this a refactor since it's not fixing any bugs or changing behaviour in any way. It's just an internal change on how we gather the code action quickfixes.

PR #491, opened 2023-02-09
Add codeAction/resolve and variable extraction refactoring

jenny-codes:
### Motivation

Two changes are included: 
1. Implement codeAction/resolve specification that populates the "edits" field for a given code action.
2. Use the new codeAction/resolve to implement a use case of refactoring: extracting variables.

### Implementation

Limitation: Currently the use case of extracting variable does not allow customizing the new variable name, and will arbitrarily place the extracted text to the immediate line above, disregarding if it is a valid syntax.

### Automated Tests

✅

### Manual Tests
code_action_resolve

1. Select a chunk of length in the editor. 
2. Right click to see the menu, choose `Refactor...` -> `Refactor: Extract variable`
3.  Expected: The selected text will be extracted away into a `new_variable` variable defined in the line above the selected text.


lib/ruby_lsp/requests/code_action_resolve.rb
```diff
@@ -0,0 +1,105 @@
...
+  module Requests
+    # ![Code action resolve demo](../../misc/code_action_resolve.gif)
+    #
+    # The [code action resolve](https://microsoft.github.io/language-server-protocol/specification#codeAction_resolve)
+    # request is used to to resolve the edit field for a given code action, if it is not already provided in the
+    # textDocument/codeAction response. We can use it for scenarios that require more computation such as refactoring.
+    #
+    # Currently one use case is implemented: Extract to variable.
+    #
+    # # Example
```
vinistock, line 14: If we say that we only support one use case, there's a very high chance we'll forget to update this phrase once more are supported 😅.
```suggestion
    # # Extract to variable example
```

lib/ruby_lsp/requests/code_action_resolve.rb
```diff
@@ -0,0 +1,105 @@
...
+      sig { params(document: Document, code_action: T::Hash[Symbol, T.untyped]).void }
+      def initialize(document, code_action)
+        super(document)
+
+        @code_action = code_action
+      end
+
+      sig { override.returns(T.any(Interface::CodeAction, Error)) }
+      def run
+        return Error::UnsupportedAction unless @code_action[:kind] == Constant::CodeActionKind::REFACTOR_EXTRACT
```
vinistock, line 48: The server only receives resolve requests for the code actions we return ourselves. Is there a scenario in which we would receive an unsupported resolve?

test/requests/code_action_resolve_expectations_test.rb
```diff
@@ -0,0 +1,90 @@
...
+      ).run
+    end
+
+    assert_empty(stdout)
+    result
+  end
+
+  def assert_expectations(source, expected)
+    actual = run_expectations(source)
+    # assert_equal(expected, JSON.parse(actual.to_json))
```
vinistock, line 36: ✂️

jenny-codes: Oops

lib/ruby_lsp/executor.rb
```diff
@@ -236,11 +238,33 @@ def inlay_hint(uri, range)
...
-      end_line = range.dig(:end, :line)
       document = @store.get(uri)
 
-      Requests::CodeActions.new(uri, document, start_line..end_line, context).run
+      Requests::CodeActions.new(uri, document, range, context).run
+    end
+
+    sig do
+      params(params: T::Hash[Symbol, T.untyped])
+        .returns(T.any(Interface::CodeAction, Requests::CodeActionResolve::Error))
```
vinistock, line 248: I think this method only returns an `Interface::CodeAction`. The raised error is not considered a return of the method.

vinistock (approved):
Some minor comments, but it looks good to me! 🎩

test/requests/code_action_resolve_expectations_test.rb
```diff
@@ -0,0 +1,89 @@
...
+
+  def run_expectations(source)
+    params = @__params&.any? ? @__params : default_args
+    document = RubyLsp::Document.new(source)
+    result = T.let(
+      nil,
+      T.nilable(
+        T.any(
+          RubyLsp::Interface::CodeAction,
+          RubyLsp::Requests::CodeActionResolve::Error,
```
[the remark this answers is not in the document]
vinistock: Oh, right. Sorry, yes it's necessary here.

andyw8:
This is great! One small optional suggestion: Once `new_variable` is added, can we have it selected that you can easily type a new name?

andyw8:
One limitation I noticed is that it doesn't yet work on single line blocks, e.g.

```
    [0..10].each { |i| 5 * 2 * i }
```

becomes
```
    new_variable = 5 * 2 * i
    [0..10].each { |i| new_variable }
```
but the `i` would not be in scope.

Maybe there's a more general design decision to be made: Do we prefer to generate potentially incorrect code, or to show an error if we can't be sure an operation is safe?

andyw8:
Disregard my comments above, I should have read the description first!

jenny-codes:
@andyw8 
> One limitation I noticed is that it doesn't yet work on single line blocks

Yes, and among others, it also does not work in cases where an expression spans more than one line. Example:
```ruby
a = {
   x: 1  # Use Extract Variable on 1
}
```
becomes
```ruby
a = {
  new_variable = 1
  x: new_variable
}
```
In other words, this refactor functionality is currently quite limited. This is aimed as more of a quick showcase of what is possible with the implementation of codeAction/resolve than a full-fledge feature. And, to make it more robust, the complexity deserves a pr of its own I think :)

PR #493, opened 2023-02-10
Substitute all T::Struct usage for plain Objects

vinistock:
### Motivation

I wanted to get a more complete understanding of the LSP's performance on a long running session, so I added the following to `server.rb` to generate profile data.
```ruby
def start
  # ...
  when "initialized"
    StackProf.start(mode: :cpu, raw: true)

  when "shutdown"
    StackProf.stop
    data = StackProf.results
    File.write("profile.json", JSON.pretty_generate(data))
  # ...
end
```

I edited and saved files for 2 minutes to triggered as many requests as possible. To my surprise, of the 120 seconds we spent 30 of them just initializing `T::Struct` objects (specifically, running the construct with defaults thing and performing runtime typechecks with `is_a?`).

Other than the desire to have very compact class definitions, we really have no other requirement for using `T::Struct` and it doesn't seem like a good trade off for the performance penalty.

Using benchmark IPS to check the difference on a single request, not counting objects created for queueing, there's a 7% performance difference.
```
Warming up --------------------------------------
                 new    15.350k i/100ms
Calculating -------------------------------------
                 new    154.727k (± 1.1%) i/s -    782.850k in   5.060214s

Comparison:
                 new:   154727.1 i/s
                 old:   144077.7 i/s - 1.07x  (± 0.00) slower
```

This also benefits YJIT, which in my local machine goes from 1.24 -> 1.32 interp / yjit ratio.

### Implementation

Switched all `T::Struct` to regular classes.

### DIFF
--- a/lib/ruby_lsp/requests/semantic_highlighting.rb
+++ b/lib/ruby_lsp/requests/semantic_highlighting.rb
@@ -78,11 +78,28 @@ module RubyLsp
         T::Array[String],
       )
 
-      class SemanticToken < T::Struct
-        const :location, SyntaxTree::Location
-        const :length, Integer
-        const :type, Integer
-        const :modifier, T::Array[Integer]
+      class SemanticToken
+        extend T::Sig
+
+        sig { returns(SyntaxTree::Location) }
+        attr_reader :location
+
+        sig { returns(Integer) }
+        attr_reader :length
+
+        sig { returns(Integer) }
+        attr_reader :type
+
+        sig { returns(T::Array[Integer]) }
+        attr_reader :modifier
+
+        sig { params(location: SyntaxTree::Location, length: Integer, type: Integer, modifier: T::Array[Integer]).void }
+        def initialize(location:, length:, type:, modifier:)
+          @location = location
+          @length = length
+          @type = type
+          @modifier = modifier
+        end
       end
 
       sig do
--- a/lib/ruby_lsp/utils.rb
+++ b/lib/ruby_lsp/utils.rb
@@ -6,29 +6,73 @@ module RubyLsp
   VOID = T.let(Object.new.freeze, Object)
 
   # A notification to be sent to the client
-  class Notification < T::Struct
-    const :message, String
-    const :params, Object
+  class Notification
+    extend T::Sig
+
+    sig { returns(String) }
+    attr_reader :message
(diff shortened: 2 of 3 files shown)


lib/ruby_lsp/requests/semantic_highlighting.rb
```diff
@@ -78,11 +78,28 @@ class SemanticHighlighting < BaseRequest
...
+        attr_reader :length
+
+        sig { returns(Integer) }
+        attr_reader :type
+
+        sig { returns(T::Array[Integer]) }
+        attr_reader :modifier
+
+        sig { params(location: SyntaxTree::Location, length: Integer, type: Integer, modifier: T::Array[Integer]).void }
+        def initialize(location, length, type, modifier)
```
paracycle, line 97: Can we continue using kwargs please? I find them to be more flexible to use and it will result in less code changes as well.

paracycle (commented):
Can we figure out how to improve `T::Struct` as well? Maybe we should always be shipping our own simplified version of `T::Struct` at runtime, so that we never have to do this work, nor worry about performance.

lib/ruby_lsp/utils.rb
```diff
@@ -6,29 +6,73 @@ module RubyLsp
...
+    extend T::Sig
+
+    sig { returns(String) }
+    attr_reader :message
+
+    sig { returns(Object) }
+    attr_reader :params
+
+    sig { params(message: String, params: Object).void }
+    def initialize(message, params)
```
Morriar, line 19: Why not using named parameters so the change is transparent to the rest of the codebase?

Morriar (commented):
It seems there are some improvements we could be shipping Sorbet's upstream either on `T::Struct` itself or its runtime configuration.

bitwise-aiden:
> Can we figure out how to improve T::Struct as well? Maybe we should always be shipping our own simplified version of T::Struct at runtime, so that we never have to do this work, nor worry about performance.

+1 to this. Off the back of the [investigation into the constructor](https://github.com/Shopify/shopify-types/pull/313), I think that it might be worth replacing `T::Struct` for PORO at runtime. I will have a bit of a look into that this week before going on break.

paracycle (approved):

PR #498, opened 2023-02-13
Upgrade to syntax tree 6

### LINKED ISSUE #497: Upgrade `syntax_tree` to 6.0
[Several changes](https://github.com/ruby-syntax-tree/syntax_tree/blob/main/CHANGELOG.md#changed) are required for this upgrade.

andyw8:
### Motivation

Closes #497

### DIFF
--- a/test/expectations/selection_ranges/heredoc.exp.json
+++ b/test/expectations/selection_ranges/heredoc.exp.json
@@ -25,7 +25,7 @@
                     },
                     "end": {
                         "line": 2,
-                        "character": 0
+                        "character": 8
                     }
                 }
             }


test/expectations/selection_ranges/heredoc.exp.json
```diff
@@ -25,7 +25,7 @@
                     },
                     "end": {
                         "line": 2,
-                        "character": 0
+                        "character": 8
```
st0012, line 28: Do we know which change leads to this difference?

andyw8: I can't actually find a commit/PR about this, perhaps it's a side-effect of some other change. The new position is the correct though, since we want it to include to the end of `HEREDOC` in https://github.com/Shopify/ruby-lsp/blob/main/test/fixtures/heredoc.rb

Issue #499, opened 2023-02-13
Intermittent failing test on Ruby 2.7


andyw8:
After merging https://github.com/Shopify/ruby-lsp/pull/498 we saw a failing test on main. It passed on the re-run, but it seems order dependent, I was able to reproduce it locally with:

```
git checkout 5733d790f251db87e9960a7e5d81d2b09dc425c9
TESTOPTS="--seed=4107" bin/test
```

vinistock:
If you run that specific example and put a breakpoint in the exact point where it fails, are you able to see the exception?

PR #501, opened 2023-02-14
Fix failing test on Ruby 2.7

andyw8:
### Motivation

Resolves https://github.com/Shopify/ruby-lsp/issues/499

### DIFF
--- a/test/requests/base_request_test.rb
+++ b/test/requests/base_request_test.rb
@@ -42,4 +42,15 @@ class BaseRequestTest < Minitest::Test
 
     assert_instance_of(SyntaxTree::CallNode, parent)
   end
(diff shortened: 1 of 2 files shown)


test/requests/base_request_test.rb
```diff
@@ -42,4 +42,15 @@ def test_locate
 
     assert_instance_of(SyntaxTree::CallNode, parent)
   end
+
+  # We can remove this once we drop support for Ruby 2.7
```
andyw8, line 46: @vinistock I moved this into `base_request` since it's not really specific to `SemanticHighlighting`.

vinistock (approved):

PR #502, opened 2023-02-16
Fix completion on syntax error

vinistock:
### Motivation

This PR fixes two issues that I noticed thanks to our telemetry. I recommend reviewing by commit. The issues were
1. We were using `@unparsed_edits` as the indication of whether a file had syntax errors or not, which is brittle. If a file is opened with syntax errors, `@unparsed_edits` is empty, but we still have a syntax error
2. We were trying to run completion when a syntax error existed, which resulted to passing `nil` to `locate` and then failed when invoking `child_nodes` on `nil`

### Implementation

1. Started using an instance variable to indicate the presence of syntax errors
2. Returned early from completion if the file is not properly parsed
3. Used the aliased language server constants to avoid very long constant names

### Automated Tests

Added a test for both fixes and adapted an existing test assumption.

### DIFF
--- a/lib/ruby_lsp/document.rb
+++ b/lib/ruby_lsp/document.rb
@@ -21,9 +21,10 @@ module RubyLsp
       @encoding = T.let(encoding, String)
       @source = T.let(source, String)
       @unparsed_edits = T.let([], T::Array[EditShape])
+      @syntax_error = T.let(false, T::Boolean)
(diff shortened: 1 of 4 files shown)


bitwise-aiden (approved):
Nice find!

paracycle (approved):
Nice, thank you!

PR #505, opened 2023-02-16
Warn that the LSP is ready on initialized

vinistock:
### Motivation

When we're initializing the LSP, we print `Starting Ruby LSP...`, but then we never print that it's ready, which is a minor source of confusion.

Also, after the refactor in #368, we automatically respond with `nil` even for notifications (which do not require a response). The only notification we receive without explicitly requesting in our capabilities is `initialized`, but it does print an error to the console saying `received response without id`.

### Implementation

Implement the `initialized` notification to avoid seeing the errors in the console and to let users know the LSP finished booting.

### DIFF
--- a/lib/ruby_lsp/executor.rb
+++ b/lib/ruby_lsp/executor.rb
@@ -37,6 +37,9 @@ module RubyLsp
       case request[:method]
       when "initialize"
         initialize_request(request.dig(:params))
+      when "initialized"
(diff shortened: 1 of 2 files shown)


paracycle (approved):

andyw8:
Similarly, when we stop the extension we display "Shutting down Ruby LSP..." but never indicate that it completes. It is worth addressing that too?

PR #509, opened 2023-02-17
Fix path completions when partial path is typed

paracycle:
### Motivation

With the current implementation when a completion request is received with the require string like `"ruby_lsp/reque"`, the completion suggestions are returned correctly, but when the user selects one of the suggestions, for example, `ruby_lsp/requests/base_request`, then their require string becomes `"ruby_lsp/sts/base_request"`.

Apparently, this happens because [the contents of `insert_text` is open to interpretation by the client](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#completionItem):

> The `insertText` is subject to interpretation by the client side.
> Some tools might not take the string literally. For example
> VS Code when code complete is requested in this example
> `con<cursor position>` and a completion item with an `insertText` of
> `console` is provided it will only insert `sole`. Therefore it is
> recommended to use `textEdit` instead since it avoids additional client
> side interpretation.

### Implementation

This PR changes the implementation to use `text_edit` instead of `insert_text` for completion items.

### Automated Tests

Updates existing tests to be more resilient and added a new test case for partial path completion.

### DIFF
--- a/test/requests/path_completion_test.rb
+++ b/test/requests/path_completion_test.rb
@@ -88,6 +88,31 @@ class PathCompletionTest < Minitest::Test
     assert_equal(expected.to_json, result.to_json)
   end
 
+  def test_completion_with_partial_path
+    prefix = "foo/suppo"
+
+    document = RubyLsp::Document.new(+<<~RUBY)
+      require "#{prefix}"
+    RUBY
+
+    position = {
+      line: 0,
+      character: document.source.rindex('"') || 0,
+    }
+
+    result = with_file_structure do
+      RubyLsp::Requests::PathCompletion.new(document, position).run
+    end
+
+    expected = [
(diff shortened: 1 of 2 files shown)


test/requests/path_completion_test.rb
```diff
@@ -5,45 +5,114 @@
...
+      path_completion("foo/support/bar", prefix, position),
+      path_completion("foo/support/baz", prefix, position),
+      path_completion("foo/support/quux", prefix, position),
+    ]
+
+    assert_equal(expected.to_json, result.to_json)
   end
 
+  # require("ruby_lsp/requ")
+
```
vinistock, line 64: ✂️ 
```suggestion
```

vinistock (approved):

PR #516, opened 2023-02-23
Use `visit_methods` syntax from Syntax Tree

andyw8:
View with whitespace changes ignored: https://github.com/Shopify/ruby-lsp/pull/516/files?w=1

### Motivation

Closes https://github.com/Shopify/ruby-lsp/issues/476


vinistock:
We'd still need some sort of rake task to make sure people invoked `visit_methods` on every class, so I don't think there'd be a lot of difference in terms of maintenance cost.

Regarding using the private API, it's a CI/development thing and not something that actually runs when using the Ruby LSP, so I don't think that's a big deal. It's would be a big issue if different versions of the `sorbet-runtime` could result in different behaviour of the server itself, but that's not the case.

My opinion is that we can continue to use Sorbet's override functionality, get instant feedback on the editor about incorrect visits and avoid another level of nesting in our request classes.

PR #518, opened 2023-02-23
Avoid unnecessary String#sub in diagnostic

vinistock:
### Motivation

Instead of doing `sub` and then `start_with?`, we can do a single operation to check if we're inside the current workspace.

### Implementation

Moved the workspace URI into a constant since it might be useful for other requests. Removed the `sub`.

### DIFF
--- a/lib/ruby_lsp/utils.rb
+++ b/lib/ruby_lsp/utils.rb
@@ -5,6 +5,9 @@ module RubyLsp
   # Used to indicate that a request shouldn't return a response
   VOID = T.let(Object.new.freeze, Object)
 
+  # This freeze is not redundant since the interpolated string is mutable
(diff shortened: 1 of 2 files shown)


st0012 (approved):

lib/ruby_lsp/utils.rb
```diff
@@ -4,6 +4,7 @@
 module RubyLsp
   # Used to indicate that a request shouldn't return a response
   VOID = T.let(Object.new.freeze, Object)
+  WORKSPACE_URI = T.let("file://#{Dir.pwd}", String)
```
paracycle, line 7: Should we `.freeze` this string?

paracycle (approved):

lib/ruby_lsp/utils.rb
vinistock: Oh, this was actually fixed already. The reason it fails for us is because we aim RuboCop for 2.7 (the earliest version we support).

Not actually a bug, just our configs.

PR #519, opened 2023-02-26
Treat test name strings as document symbols

andyw8:
### Motivation

When tests are defined as normal methods using the `def test_something` syntax, they are treated as document symbols, as any other method would be. But when using the `test "some string"` syntax provided by ActiveSupport, they are not detected. By adding support for those,  VS Code can show them in the Outline View, and list them you press `@` in the command palette.

Screenshot 2023-02-26 at 9 46 30 AM

This uses some of the earlier exploratory work by @DeeChau in https://github.com/Shopify/ruby-lsp/pull/490

### Implementation

Currently this does not support other syntax such as RSpec, but I hope to add that in future.

### Automated Tests

I've added a new fixture and expectation.

### Manual Tests

Choose a large project which uses ActiveSupport tests, e.g. a Rails project. Ensure that each tests is listed in the Outline view.

### DIFF
--- a/lib/ruby_lsp/requests/document_symbol.rb
+++ b/lib/ruby_lsp/requests/document_symbol.rb
@@ -112,7 +112,14 @@ module RubyLsp
 
         return unless node.call.message.value == "test" && node.call.arguments.is_a?(SyntaxTree::ArgParen)
 
-        full_name = node.call.arguments.arguments.parts.first.parts.map do |part|
+        arg_paren = node.call.arguments
+        args = arg_paren.arguments
+        return unless args
+
+        arg = args.parts.first
+        return unless arg.is_a?(SyntaxTree::StringLiteral)
+
+        full_name = arg.parts.map do |part|
           case part
           when SyntaxTree::TStringContent
             part.value
--- a/lib/ruby_lsp/requests/document_symbol.rb
+++ b/lib/ruby_lsp/requests/document_symbol.rb
@@ -108,32 +108,17 @@ module RubyLsp
 
       sig { override.params(node: SyntaxTree::MethodAddBlock).void }
       def visit_method_add_block(node)
-        # TODO: extract commonality with parse_test
-
         return unless node.call.message.value == "test" && node.call.arguments.is_a?(SyntaxTree::ArgParen)
 
         arg_paren = node.call.arguments
         args = arg_paren.arguments
         return unless args
 
-        arg = args.parts.first
-        return unless arg.is_a?(SyntaxTree::StringLiteral)
+        name_node = args.parts.first
+        return unless name_node.is_a?(SyntaxTree::StringLiteral)
 
-        full_name = arg.parts.map do |part|
-          case part
-          when SyntaxTree::TStringContent
-            part.value
-          when SyntaxTree::StringEmbExpr
-            "{ }"
-          end
-        end.join("")
-
-        create_document_symbol(
-          name: full_name,
-          kind: :method,
-          range_node: node, # TODO: check
-          selection_range_node: node, # TODO: check
-        )
+        # T.reveal_type(name_node)
+        add_test(name_node, node)
       end
 
       sig { override.params(node: SyntaxTree::Command).void }
@@ -256,17 +241,24 @@ module RubyLsp
         symbol
       end
 
-      sig { params(node: SyntaxTree::Command).void }
-      def parse_test(node)
-        return unless node.block
+      sig { params(command_node: SyntaxTree::Command).void }
+      def parse_test(command_node)
+        return unless command_node.block && command_node.arguments.parts.one?
 
-        argument = node.arguments.parts.first
-        return unless argument.is_a?(SyntaxTree::StringLiteral) && node.arguments.parts.size == 1
+        name_node = command_node.arguments.parts.first
+        return unless name_node.is_a?(SyntaxTree::StringLiteral) && name_node.parts.first
 
-        name_node = argument.parts.first
-        return unless name_node
+        add_test(name_node, command_node)
+      end
 
-        full_name = argument.parts.map do |part|
+      sig do
+        params(
+          name_node: SyntaxTree::StringLiteral,
+          test_node: SyntaxTree::Node,
+        ).returns(LanguageServer::Protocol::Interface::DocumentSymbol)
+      end
+      def add_test(name_node, test_node)
+        full_name = name_node.parts.map do |part|
           case part
           when SyntaxTree::TStringContent
             part.value
@@ -278,8 +270,8 @@ module RubyLsp
         create_document_symbol(
           name: full_name,
           kind: :method,
-          range_node: node, # TODO: check
-          selection_range_node: argument.parts.first, # TODO: check
+          range_node: test_node,
+          selection_range_node: name_node,
         )
       end
     end
--- a/lib/ruby_lsp/requests/document_symbol.rb
+++ b/lib/ruby_lsp/requests/document_symbol.rb
@@ -106,8 +106,36 @@ module RubyLsp
         @stack.pop
       end
 
+      sig { override.params(node: SyntaxTree::MethodAddBlock).void }
+      def visit_method_add_block(node)
+        # TODO: extract commonality with parse_test
+
+        return unless node.call.message.value == "test" && node.call.arguments.is_a?(SyntaxTree::ArgParen)
+
+        full_name = node.call.arguments.arguments.parts.first.parts.map do |part|
+          case part
+          when SyntaxTree::TStringContent
+            part.value
+          when SyntaxTree::StringEmbExpr
+            "{ }"
+          end
+        end.join("")
+
+        create_document_symbol(
+          name: full_name,
+          kind: :method,
+          range_node: node, # TODO: check
+          selection_range_node: node, # TODO: check
+        )
+      end
+
       sig { override.params(node: SyntaxTree::Command).void }
       def visit_command(node)
+        if node.message.value == "test"
+          parse_test(node)
+          return
+        end
+
         return unless ATTR_ACCESSORS.include?(node.message.value)
 
         node.arguments.parts.each do |argument|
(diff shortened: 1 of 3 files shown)


lib/ruby_lsp/requests/document_symbol.rb
```diff
@@ -199,6 +216,15 @@ def visit_var_field(node)
 
       private
 
+      sig { params(node: SyntaxTree::Node).returns(T::Boolean) }
+      def test_class?(node)
+        class_node = node.child_nodes[1]
+        return false unless class_node.is_a?(SyntaxTree::ConstPathRef)
+
+        class_name = class_node.constant.value
+        class_name.end_with?("TestCase") || class_name.end_with?("IntegrationTest")
```
andyw8, line 225: This should cover all the types provided by Rails:
https://guides.rubyonrails.org/testing.html#a-brief-note-about-test-cases

However if someone defines another kind of test class to inherit from then it could be missed, if named another way. We could possibly make this more generic by matching on anything ending or beginning with `Test`.

lib/ruby_lsp/requests/document_symbol.rb
```diff
@@ -108,6 +111,17 @@ def visit_class(node)
 
       sig { override.params(node: SyntaxTree::Command).void }
       def visit_command(node)
+        if @test_class && node.message.value == "test"
+          argument = node.arguments.parts.first
+          test_name = argument.child_nodes.first.value
+          create_document_symbol(
+            name: test_name,
+            kind: :method,
```
andyw8, line 119: It's not _actually_ a method but this is a good enough approximation, and keeps it consistent with tests which use the traditional syntax.

lib/ruby_lsp/requests/document_symbol.rb
```diff
@@ -108,6 +111,17 @@ def visit_class(node)
 
       sig { override.params(node: SyntaxTree::Command).void }
       def visit_command(node)
+        if @test_class && node.message.value == "test"
+          argument = node.arguments.parts.first
+          test_name = argument.child_nodes.first.value
+          create_document_symbol(
+            name: test_name,
```
andyw8, line 118: I'm wondering whether to prefix this with `test `.

andyw8:
Thanks @paracycle, I've updated with your suggestion.

My thinking behind using `@test_class` was that we could set it then skip later checks for `node.message.value == "test"` in non-test classes, but I didn't measure the difference and I expect it's negligible.

> Start hooking into the runtime and collect the names of all the tests that the application has loaded from this file. I think this will be a bigger thing to pull off, but the resulting knowledge is exactly what we are after, including the actual names of the tests as they appear at runtime. This will also include any tests that might have been defined dynamically, which the other method will miss.

This is interesting, and I'll keep in mind for future enhancements.

lib/ruby_lsp/requests/document_symbol.rb
```diff
@@ -108,6 +108,17 @@ def visit_class(node)
 
       sig { override.params(node: SyntaxTree::Command).void }
       def visit_command(node)
+        if node.message.value == "test"
```
st0012, line 111: Since we're making this change, how about including other common names as well?:

- describe
- context
- it

They can still appear in `minitest`-based tests I believe.

andyw8: I do want to add support for that soon, but it gets a little more complex since they can be nested.

lib/ruby_lsp/requests/document_symbol.rb
```diff
@@ -108,6 +108,17 @@ def visit_class(node)
...
+          argument = node.arguments.parts.first
```
vinistock, line 112: We should add a test case for things that might break this. For example, what happens if an application has a `test` method that takes no arguments?

lib/ruby_lsp/requests/document_symbol.rb
```diff
@@ -108,6 +108,17 @@ def visit_class(node)
...
+          test_name = argument.child_nodes.first.value
+          create_document_symbol(
+            name: test_name,
+            kind: :method,
+            range_node: argument,
+            selection_range_node: argument,
```
vinistock, line 118: The selection range should encapsulate the entire test, which matches the expected behaviour from the spec. For example, method definitions always include the body in the selection.

lib/ruby_lsp/requests/document_symbol.rb
andyw8: Tests without a block are actually [valid](https://github.com/rails/rails/blob/1b10f5c04cf1c7bdf823c77780586e8f09e73c3f/activesupport/lib/active_support/testing/declarative.rb#L20-L22), it outputs as:

`No implementation provided for <test name>`

I think the intention is that you can use it as a reminder to write the test.

paracycle: Sure, but it doesn't mean that we should surface those as valid test methods either. Regardless, I don't see us doing a strict check on the count and type.of arguments being passed to the "test" call.

andyw8:
I've pushed an update which has some additional rules for what we treat as a test, and expanded the number of variations in the fixture.

I've also tried adding support for cases where the name contains interpolations. We can't catch every possible variation in the ways that people might construct test names, but this seems like a common one that is fairly easy to handle.

lib/ruby_lsp/requests/document_symbol.rb
```diff
@@ -106,8 +106,36 @@ def visit_class(node)
         @stack.pop
       end
 
+      sig { override.params(node: SyntaxTree::MethodAddBlock).void }
+      def visit_method_add_block(node)
+        # TODO: extract commonality with parse_test
+
+        return unless node.call.message.value == "test" && node.call.arguments.is_a?(SyntaxTree::ArgParen)
+
+        full_name = node.call.arguments.arguments.parts.first.parts.map do |part|
```
vinistock, line 115: Are no elements in this entire chain nilable? It looks like a lot of potential for breaking.

lib/ruby_lsp/requests/document_symbol.rb
```diff
@@ -220,6 +248,33 @@ def create_document_symbol(name:, kind:, range_node:, selection_range_node:)
...
+      def parse_test(node)
+        return unless node.block
+
+        argument = node.arguments.parts.first
+        return unless argument.is_a?(SyntaxTree::StringLiteral) && node.arguments.parts.size == 1
+
+        name_node = argument.parts.first
+        return unless name_node
+
+        full_name = argument.parts.map do |part|
```
vinistock, line 262: What are the parts in this case? Would we have a single string argument?

lib/ruby_lsp/requests/document_symbol.rb
```diff
@@ -106,8 +106,36 @@ def visit_class(node)
         @stack.pop
       end
 
+      sig { override.params(node: SyntaxTree::MethodAddBlock).void }
+      def visit_method_add_block(node)
```
paracycle, line 110: I am not sure what this is matching, can you clarify why we need this in addition to what `visit_commands` does?

If we still do need this, I think we should share a lot of this logic by calling out to `parse_test` anyway.

lib/ruby_lsp/requests/document_symbol.rb
```diff
@@ -220,6 +248,33 @@ def create_document_symbol(name:, kind:, range_node:, selection_range_node:)
...
+          case part
+          when SyntaxTree::TStringContent
+            part.value
+          when SyntaxTree::StringEmbExpr
+            "{ }"
+          end
+        end.join("")
+
+        create_document_symbol(
+          name: full_name,
```
paracycle, line 272: If we are doing special treatment for AS::TestCase `test` method calls, then I think we should also do the name transformation it does as well: https://github.com/rails/rails/blob/7c70791470fc517deb7c640bead9f1b47efb5539/activesupport/lib/active_support/testing/declarative.rb#L14

lib/ruby_lsp/requests/document_symbol.rb
paracycle: Agree, and I think `parse_test` is already doing this the right way.

andyw8: Yes I found there were some problems with this, I've added more tests cases and broke it down a bit. I'll aim to unify it with `parse_test`.

lib/ruby_lsp/requests/document_symbol.rb
vinistock: Then do we need to map? Can't we just access the first item?

andyw8: I added the `map` so it could handle strings with interpolation, e.g. we could show `"valid for #{category} something"` as `"valid for category { } something"`. Without the map, it would only show `"valid for category"`.

 I'm on the fence about whether it's worthwhile though.

vinistock: Oh, I understand now what the curly braces are. In that case, I agree with the `map`, but instead of using curly braces it might look better if we use a fake object notation. Something like
```ruby
"valid for category #<Object ...> something"
```

lib/ruby_lsp/requests/document_symbol.rb
andyw8: Yep, I had those in mind for a follow-up.

lib/ruby_lsp/requests/document_symbol.rb
paracycle: Ok, so this was the point I was trying to make initially about runtime method generation and how we are skirting on the border of something that doesn't cleanly belong in the current implementation.

The way I see it is that this feature is implementing support for syntax that does metaprogramming to create methods at runtime. All we are doing is to expose those methods as they would be represented at runtime.

If we are doing that, then we should be doing it faithfully, i.e. show the method with the name it would have at runtime.

The concern about describe, etc blocks is not important right now, since each testing framework will need special handling for this anyway.

vinistock:
Is it difficult to support test methods with parenthesis? I'd say it makes sense to do it since we're already looking at it.

Issue #515, opened 2023-02-23
Spec blocks in RSpec files are not foldable


amiratak88:
```
RSpec.describe SomeObject do
	describe "#some_method" do
		...
    end

	describe "#some_other_method" do
		...
    end
end
```

In the above code, only the main `RSpec.describe` block is foldable. The two nested `describe` blocks are not.

andyw8:
Ah, I see the issue now - it happens only if you omit the parenthesis for `SomeObject`.

andyw8:
The upgrade of syntax_tree in https://github.com/Shopify/ruby-lsp/pull/538 by @vinistock will unblock this.

PR #528, opened 2023-03-01
Prevent duplicate ranges for code folding

andyw8:


lib/ruby_lsp/requests/folding_ranges.rb
```diff
@@ -110,6 +112,17 @@ def visit(node)
         super
       end
 
+      # This is to prevent duplicate ranges
+      sig { params(node: SyntaxTree::Command).returns(T::Boolean) }
+      def same_lines_for_command_and_block?(node)
```
andyw8, line 117: @vinistock I changed this from `same_location_for_command_and_block`

test/fixtures/block_with_class.rb
```diff
@@ -0,0 +1,5 @@
+RSpec.describe(SomeObject)do
```
andyw8, line 1: The situation where there are no parens around `SomeObject` needs a SyntaxTree update - see https://github.com/Shopify/ruby-lsp/issues/515.

PR #538, opened 2023-03-03
Add highlighting support for more local variables

### LINKED ISSUE #344: Add support for block locals
Block locals are currently not captured by [Syntax Tree's WithEnvironment](https://github.com/ruby-syntax-tree/syntax_tree/blob/main/lib/syntax_tree/visitor/with_environment.rb) and the Ruby LSP is also not processing the node to get correct semantic highlighting. Currently, block locals are completely ignored and get no semantic highlighting.

The syntax is a bit uncommon, but we should support it for consistency and completion sake.
```ruby
var = 1

# adding this var after a semicolon declares it in the context of the block
# this means that when re-assigning it to 2, it does not impact the outer scope
foo.each do |some_other_param; var|
  var = 2 
end

var # => 1
```

vinistock:
### Motivation

Closes #344

There were still a few difficult to support local variable highlighting we didn't handle. Starting on SyntaxTree v6.0.2, we can finally complete local variable support.

### Implementation

I recommend reviewing by commit.
- Upgrade our dependency on SyntaxTree to `>= 6.0.2`
- Add semantic highlighting to block and lambda locals (very uncommon, but existing feature of Ruby)
- Add semantic highlighting to regexp named captures (which are in fact local variables)
- Add a test for destructured splat block arguments (there was a bug that was fixed in SyntaxTree v6.0.2)
- Add semantic highlighting for pinned variables (made possible by another SyntaxTree improvement)

### Automated Tests

Added tests for all of those.

### DIFF
--- a/lib/ruby_lsp/requests/semantic_highlighting.rb
+++ b/lib/ruby_lsp/requests/semantic_highlighting.rb
@@ -238,13 +238,12 @@ module RubyLsp
 
         value = node.value
 
-        case value
-        when SyntaxTree::Ident
+        if value.is_a?(SyntaxTree::Ident)
           type = type_for_local(value)
           add_token(value.location, type)
-        else
-          visit(value)
         end
+
+        super
       end
 
       sig { override.params(node: SyntaxTree::VarRef).void }
--- a/lib/ruby_lsp/requests/semantic_highlighting.rb
+++ b/lib/ruby_lsp/requests/semantic_highlighting.rb
@@ -280,10 +280,45 @@ module RubyLsp
       def visit_vcall(node)
         return super unless visible?(node, @range)
 
-        return if special_method?(node.value.value)
+        # A VCall may exist as a local in the current_scope. This happens when used named capture groups in a regexp
+        ident = node.value
+        value = ident.value
+        local = current_scope.find_local(value)
+        return if local.nil? && special_method?(value)
 
-        type = Support::Sorbet.annotation?(node) ? :type : :method
+        type = if local
+          :variable
+        elsif Support::Sorbet.annotation?(node)
+          :type
+        else
+          :method
+        end
+
         add_token(node.value.location, type)
+      end
+
+      sig { override.params(node: SyntaxTree::Binary).void }
+      def visit_binary(node)
+        # It's important to visit the regexp first in the WithScope module
+        super
+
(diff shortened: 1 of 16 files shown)


lib/ruby_lsp/requests/semantic_highlighting.rb
```diff
@@ -262,16 +261,58 @@ def visit_var_ref(node)
...
 
-        return if special_method?(node.value.value)
+        # A VCall may exist as a local in the current_scope. This happens when used named capture groups in a regexp
+        value = node.value
+        local = current_scope.find_local(value.value)
+        return if local.nil? && special_method?(value.value)
 
-        type = Support::Sorbet.annotation?(node) ? :type : :method
+        type = :variable if local
+        type ||= Support::Sorbet.annotation?(node) ? :type : :method
```
andyw8, line 288: I think this is a little more readable as:
```
type = if local
  :variables
elsif Support::Sorbet.annotation?(node)
  :type
else
  :method
end
```

lib/ruby_lsp/requests/semantic_highlighting.rb
```diff
@@ -262,16 +261,58 @@ def visit_var_ref(node)
...
         add_token(node.value.location, type)
       end
 
+      sig { override.params(node: SyntaxTree::Binary).void }
+      def visit_binary(node)
+        # It's important to visit the regexp first in the WithScope module
+        super
+
+        # You can only capture local variables with regexp by using the =~ operator
+        if node.operator == :=~
```
andyw8, line 298: Maybe use a guard here to avoid a level of nesting?

lib/ruby_lsp/requests/semantic_highlighting.rb
```diff
@@ -262,16 +261,58 @@ def visit_var_ref(node)
...
+          left = node.left
+          parts = left.parts
+
+          if left.is_a?(SyntaxTree::RegexpLiteral) && parts.length == 1 && parts.first.is_a?(SyntaxTree::TStringContent)
```
andyw8, line 302:
```suggestion
          if left.is_a?(SyntaxTree::RegexpLiteral) && parts.one? && parts.first.is_a?(SyntaxTree::TStringContent)
```

st0012 (approved):
TIL we can directly assign captured regexp groups into local variables.

Issue #503, opened 2023-02-16
Add custom cop for language server constant aliases


vinistock:
Anything inside the `RubyLsp` module has direct access to the `Interface`, `Transport` and `Constant` modules, which saves us from typing `LanguageServer::Protocol` in front of them every time.

Let's add a custom cop inside this repo to autocorrect usages if they happen inside the `RubyLsp` module.

caalberts:
Hi @vinistock, I'm interested to take this one up.

Is the intent to replace lines such as the following:

```rb
      sig { override.returns(T.all(T::Array[LanguageServer::Protocol::Interface::FoldingRange], Object)) }
```

into this?

```rb
      sig { override.returns(T.all(T::Array[Interface::FoldingRange], Object)) }
```

I'm thinking of creating the cop first, then depending on how many violations it finds, decide whether to fix them or create todos.

PR #543, opened 2023-03-06
Create Rubocop Cop to detect LanguageServer constants

caalberts:
### Motivation

#503 - Add custom cop to detect references to `LanguageServer::Protocol` constants and autocorrect to use its aliases.

### Implementation

Custom Rubocop Cop.

### Automated Tests

Added RSpec spec for the custom cop. RSpec is needed as Rubocop test uses matchers defined in `rubocop/rspec/support`.

### DIFF
--- a/lib/ruby_lsp/requests/base_request.rb
+++ b/lib/ruby_lsp/requests/base_request.rb
@@ -36,16 +36,16 @@ module RubyLsp
         nodes.each { |node| visit(node) }
       end
 
-      sig { params(node: SyntaxTree::Node).returns(LanguageServer::Protocol::Interface::Range) }
+      sig { params(node: SyntaxTree::Node).returns(Interface::Range) }
       def range_from_syntax_tree_node(node)
         loc = node.location
 
-        LanguageServer::Protocol::Interface::Range.new(
-          start: LanguageServer::Protocol::Interface::Position.new(
+        Interface::Range.new(
+          start: Interface::Position.new(
             line: loc.start_line - 1,
             character: loc.start_column,
           ),
-          end: LanguageServer::Protocol::Interface::Position.new(line: loc.end_line - 1, character: loc.end_column),
+          end: Interface::Position.new(line: loc.end_line - 1, character: loc.end_column),
         )
       end
 
--- a/lib/ruby_lsp/requests/document_highlight.rb
+++ b/lib/ruby_lsp/requests/document_highlight.rb
@@ -29,14 +29,14 @@ module RubyLsp
       def initialize(document, position)
         super(document)
 
-        @highlights = T.let([], T::Array[LanguageServer::Protocol::Interface::DocumentHighlight])
+        @highlights = T.let([], T::Array[Interface::DocumentHighlight])
         return unless document.parsed?
 
         position = document.create_scanner.find_char_position(position)
         @target = T.let(find(T.must(document.tree), position), T.nilable(Support::HighlightTarget))
       end
 
-      sig { override.returns(T.all(T::Array[LanguageServer::Protocol::Interface::DocumentHighlight], Object)) }
+      sig { override.returns(T.all(T::Array[Interface::DocumentHighlight], Object)) }
       def run
         # no @target means the target is not highlightable
         visit(@document.tree) if @document.parsed? && @target
@@ -90,7 +90,7 @@ module RubyLsp
       sig { params(match: Support::HighlightTarget::HighlightMatch).void }
       def add_highlight(match)
         range = range_from_syntax_tree_node(match.node)
-        @highlights << LanguageServer::Protocol::Interface::DocumentHighlight.new(range: range, kind: match.type)
+        @highlights << Interface::DocumentHighlight.new(range: range, kind: match.type)
       end
     end
   end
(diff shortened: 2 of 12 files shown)


Gemfile
```diff
@@ -21,3 +21,5 @@ gem "yard", "~> 0.9", require: false
 
 # The Rails documentation link only activates when railties is detected.
 gem "railties", "~> 7.0", require: false
+
+gem "rspec", "~> 3.11", require: false
```
caalberts, line 25: `rspec` is added because Rubocop tests are written using RSpec. I'm open to discussing about this. As far as I'm aware, Rubocop doesn't provide test helpers in minitest. I'm not sure how much effort it would take to write the Rubocop tests from scratch in minitest.

If this is acceptable, it needs to be added to CI.

caalberts (commented):
@vinistock could you take a look please?

Gemfile
andyw8: https://github.com/rubocop/rubocop-minitest doesn't use RSpec for its tests - you could take a look how that's done.

vinistock: Yeah, let's please avoid adding rspec as a dependency.

rubocop/ruby_lsp/use_language_server_aliases.rb
```diff
@@ -0,0 +1,59 @@
+# typed: strict
```
vinistock, line 1: Can we put this file under `lib/rubocop` please?

vinistock (commented):
This is great! Thank you so much for looking into it. If we can just avoid using `rspec`, this looks ready to be merged

test/rubocop_helper.rb
```diff
@@ -0,0 +1,23 @@
...
+module InjectRubyLspCop
+  def setup
+    super
+    return unless @cop.nil?
+
+    cop_name = self.class.to_s.delete_suffix("Test")
+    return unless ::RuboCop::Cop::RubyLsp.const_defined?(cop_name)
+
+    # rubocop:disable Sorbet/ConstantsFromStrings
+    @cop = ::RuboCop::Cop::RubyLsp.const_get(cop_name).new
```
caalberts, line 18: This is to replace https://github.com/rubocop/rubocop-minitest/blob/v0.29.0/lib/rubocop/minitest/assert_offense.rb#L82

.rubocop.yml
```diff
@@ -20,6 +21,11 @@ Naming/FileName:
   Exclude:
   - "lib/ruby-lsp.rb"
 
+RubyLsp/UseLanguageServerAliases:
+  Exclude:
+    - "test/**/*.rb"
+    - "lib/ruby_lsp/server.rb"
```
vinistock, line 27: Why is server excluded?

test/rubocop/cop/ruby_lsp/use_language_server_aliases_test.rb
```diff
@@ -0,0 +1,121 @@
...
+# frozen_string_literal: true
+
+require "test_helper"
+
+class UseLanguageServerAliasesTest < Minitest::Test
+  def test_registers_offense_when_using_interface_within_ruby_lsp
+    assert_offense(<<~RUBY)
+      module RubyLsp
+        sig { override.returns(T.all(T::Array[LanguageServer::Protocol::Interface::FoldingRange], Object)) }
+                                              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ Use constant alias `Interface`.
```
vinistock, line 11: I noticed the error seems to only include from `::Protocol` forward. Would it be possible to start the from `LanguageServer`?

test/rubocop_helper.rb
vinistock: Instead of using this module, can't we set `@cop` in `UseLanguageServerAliasesTest` manually? For example

```ruby
class UseLanguageServerAliasesTest < Minitest::Test
  def setup
    @cop = ::RuboCop::Cop::RubyLsp::UseLanguageServerAliases.new
  end
end
```

Gemfile
vinistock: We unload RuboCop to test that Syntax Tree formatting also works. Requiring rubocop in setup will make sure it passes regardless of the order in which they run.

```ruby
def setup
  require "rubocop"
end
```

.rubocop.yml
caalberts: It is where the aliases are defined. I've changed to disable the cop in-line on that file.

test/rubocop/cop/ruby_lsp/use_language_server_aliases_test.rb
caalberts: @vinistock I don't think I understand your question. What I see is that the error starts from `LanguageServer`. Do you want it to be something else?

Screenshot 2023-03-12 at 10 07 26 PM

caalberts:
@vinistock I've fixed the loading issues. Could you review again please?

test/rubocop/cop/ruby_lsp/use_language_server_aliases_test.rb
vinistock: Oh, interesting. It appears distorted to me, in my GitHub view it shows the error starting from `::Protocol` all the way to `FoldingRange` - which I couldn't understand since the node is supposed to be the namespace. Just a visual thing then, the code uses the right location.

vinistock (approved):

caalberts:
@vinistock I've updated the PR with the sorbet changes.

PR #549, opened 2023-03-08
Test against TruffleRuby

vinistock:
### Motivation

We should make sure the Ruby LSP also works with TruffleRuby.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -8,10 +8,12 @@ jobs:
       fail-fast: false
       matrix:
         os: [ubuntu-latest, macos-latest, windows-latest]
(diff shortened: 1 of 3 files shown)


.github/workflows/ci.yml
```diff
@@ -8,7 +8,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        ruby: ["2.7", "3.0", "3.1", "3.2"]
+        ruby: ["2.7", "3.0", "3.1", "3.2", "jruby", "truffleruby"]
```
st0012, line 11: I recommend testing against their `head` versions instead so we can raise the issues we found to them, or receive addressed patches in time.

PR #556, opened 2023-03-14
Allow specifying a formatter

vinistock:
### Motivation

Part of https://github.com/Shopify/vscode-ruby-lsp/issues/430

We want to allow users to be able to configure which formatter they want to use. In order to do so for every editor, the server must accept the formatter option when being initialized and then decide what to do based on it.

### Implementation

- Save the formatter option in store
- Always pass it along to the formatting request
- Decide what to do based on the option

### Automated Tests

Added a test where we select Syntax Tree despite RuboCop being present.

### DIFF
--- a/lib/ruby_lsp/requests/formatting.rb
+++ b/lib/ruby_lsp/requests/formatting.rb
@@ -22,14 +22,23 @@ module RubyLsp
     # ```
     class Formatting < BaseRequest
       class Error < StandardError; end
+      class InvalidFormatter < StandardError; end
 
       extend T::Sig
 
-      sig { params(uri: String, document: Document).void }
-      def initialize(uri, document)
+      sig { params(uri: String, document: Document, formatter: String).void }
+      def initialize(uri, document, formatter: "auto")
         super(document)
 
         @uri = uri
+        @formatter = T.let(
+          if formatter == "auto"
+            defined?(Support::RuboCopFormattingRunner) ? "rubocop" : "syntax_tree"
+          else
+            formatter
+          end,
+          String,
+        )
       end
 
       sig { override.returns(T.nilable(T.all(T::Array[Interface::TextEdit], Object))) }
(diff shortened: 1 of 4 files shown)


st0012 (approved):
LGTM 👍 
Just a question: Do you think we can validate the formatter option passed in? Like raise if the value is not known by us. I know we don't support non-`vscode-ruby-lsp` clients but it's something that can make other editor's integration with `ruby-lsp` easier.

lib/ruby_lsp/requests/formatting.rb
```diff
@@ -25,11 +25,19 @@ class Error < StandardError; end
...
-      def initialize(uri, document)
+      sig { params(uri: String, document: Document, formatter: String).void }
+      def initialize(uri, document, formatter: "auto")
         super(document)
 
         @uri = uri
+        @formatter = T.let(
+          if formatter == "auto"
+            defined?(Support::RuboCopFormattingRunner) ? "rubocop" : "syntax_tree"
+          else
```
bitwise-aiden, line 36: Should we check if the formatter is supported / error if it isn't? 

In the case of VSCode we will be able to set that the variable must be `auto|rubocop|syntax_tree`, but for cases outside of that, we don't explicitly support passing the wrong type.

vinistock:
I agree with the suggestions. Started showing a message if the formatter is invalid.

lib/ruby_lsp/requests/formatting.rb
```diff
@@ -60,10 +69,13 @@ def run
...
       sig { returns(T.nilable(String)) }
       def formatted_file
-        if defined?(Support::RuboCopFormattingRunner)
+        case @formatter
+        when "rubocop"
           Support::RuboCopFormattingRunner.instance.run(@uri, @document)
-        else
+        when "syntax_tree"
           SyntaxTree.format(@document.source)
+        else
```
bitwise-aiden, line 77: Looking at the VS code implementation `none` is also a valid response here. I think given the feature won't be enabled on `none` it won't matter, but wanted to comment just incase.

vinistock: Yeah, just avoid broadcasting the capability on `none`, which makes the editor not send any requests.

Issue #559, opened 2023-03-16
uninitialized constant RubyLsp::Requests::SemanticHighlighting::Bundler


icnagy:
Hey,

Really great project! After seeing the conference talk I wanted to try it in SublimeText. Interesting that not all features are documented/enabled by default but it's not hard to figure out how to enable them. 
Here's the SublimeLSP setting I ended up with:
```
# ...
    "ruby-lsp": {
      "enabled": true,
      "command": [
        "ruby-lsp"
      ],
      "initializationOptions": {
        "enabledFeatures": [
          "completion",
          "hover",
          "documentHighlights",
          "documentSymbols",
          "codeActions",
          "documentLink",
          "formatting",
          "onTypeFormatting",
          "foldingRanges",
          "selectionRanges",
          "semanticHighlighting",
          "inlayHint",
          "diagnostics"
        ],
        "telemetryEnabled": true
      },
      "selector": "source.ruby | text.html.ruby"
    },
# ...
```
When `semanticHighlighting` is enabled `ruby-lsp` throws a `#<NameError: uninitialized constant RubyLsp::Requests::SemanticHighlighting::Bundler>` on initialize: (sorry for the wall of text, formatted for readability)

```
:: [11:00:03.967] <?  ruby-lsp telemetry/event: {
"lspVersion": "0.4.1", 
"errorMessage": "uninitialized constant RubyLsp::Requests::SemanticHighlighting::Bundler", 
"backtrace": "/usr/local/lib/ruby/gems/3.2.0/gems/ruby-lsp-0.4.1/lib/ruby_lsp/requests/semantic_highlighting.rb:75:in `<class:SemanticHighlighting>
/usr/local/lib/ruby/gems/3.2.0/gems/ruby-lsp-0.4.1/lib/ruby_lsp/requests/semantic_highlighting.rb:21:in `<module:Requests>`
/usr/local/lib/ruby/gems/3.2.0/gems/ruby-lsp-0.4.1/lib/ruby_lsp/requests/semantic_highlighting.rb:5:in `<module:RubyLsp>`
/usr/local/lib/ruby/gems/3.2.0/gems/ruby-lsp-0.4.1/lib/ruby_lsp/requests/semantic_highlighting.rb:4:in `<top (required)>`
<internal:/usr/local/lib/ruby/site_ruby/3.2.0/rubygems/core_ext/kernel_require.rb>:88:in `require`
[...]

PR #560, opened 2023-03-16
Require bundler

icnagy:
### Motivation

https://github.com/Shopify/ruby-lsp/issues/559

Bundler is referenced in SemanticHighlighting but when running `ruby-lsp` without bundle, it is not defined and throws an exception.

### Implementation

Explicitly require `bundler`

### DIFF
--- a/lib/ruby_lsp/requests/semantic_highlighting.rb
+++ b/lib/ruby_lsp/requests/semantic_highlighting.rb
@@ -72,7 +72,7 @@ module RubyLsp
           Module.instance_methods(false),
           Kernel.instance_methods(false),
           Kernel.methods(false),
-          *(Bundler::Dsl.instance_methods(false) if defined?(Bundler)),
+          Bundler::Dsl.instance_methods(false),
           Module.private_instance_methods(false),
         ].flatten.map(&:to_s),
         T::Array[String],


icnagy:
Updated the commit message to re-start the checks. Couldn't find the button re-run the failed CLA step after accepting it.

paracycle:
I wonder if we could (also) make the original method safer by doing:
```ruby
        [
          Module.instance_methods(false),
          Kernel.instance_methods(false),
          Kernel.methods(false),
          *(Bundler::Dsl.instance_methods(false) if defined?(Bundler)),
          Module.private_instance_methods(false),
        ].flatten.map(&:to_s)
```

vinistock:
If we require bundler, is there any scenario where it wouldn't be defined?

andyw8:
Possibly if someone is using a very old version of Rubygems since ([before March 2020](https://github.com/rubygems/rubygems/pull/3166#issuecomment-597583663)) and also still relying on pre-Bundler tooling such as those mentioned in https://cloudcity.io/blog/2015/07/10/how-bundler-works-a-history-of-ruby-dependency-management/.

icnagy:
@vinistock anything I can do to make he benchmark step not fail with `403 Unhandled error: HttpError: Resource not accessible by integration`?

vinistock:
Ignore that. Thanks for the contribution!

Issue #535, opened 2023-03-02
Stale responses should not be returned to the editor


vinistock:
### Reproduction steps

1. Open a very large file (4k+ lines) with the Ruby LSP
2. Edit the file quickly. Make insertions and deletions and make sure to save the file in between edits
3. Eventually, you will start getting errors. Mainly the infamous `Comment location overlaps with node location`

I believe the reason this occurs is due to this flow
- Make some text edits
- Save the file (runs formatting)
- Make more edits before formatting finishes running
- The document is now in a different state than when we started running formatting
- After a while, the edits from formatting are returned and are applied, but those edits don't correspond to the current version of the document

### Ideas on how to fix this

I believe the only way to fix this is to
- Start keeping track of the document version sent by editors in our `Document` class https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#versionedTextDocumentIdentifier
- After each request, check if the current version of the `Document` is newer than the version we executed the request for
- If that's the case, return a `nil` result as if the request had been cancelled

vinistock:
Note: this was fixed by making any requests that return text edits run in the main thread (without the requests being placed in the queue), which guarantees the order of operations.

However, I think the more adequate fix is something we tried in the past and didn't quite work as we hoped, which is cancelling requests in the middle of execution. If you save a file and type again before formatting finishes, I believe editors send a cancel request for formatting but, because the Ruby LSP already started processing, we just continue letting it run.

We may want to revisit in the future and try to get requests to halt execution midway. This was very difficult to get right without accidentally killing the worker thread completely so we ended up reverting our initial attempt at this.

PR #561, opened 2023-03-16
Include version and uri in document

vinistock:
### Motivation

Step towards #535

We currently don't include the URI or the version of the document in our `Document` class. This results in
- Having to pass the document _and_ the URI around, instead of just passing the document
- Not being able to verify the current version of documents

In order to fix the bug described in #535, we need the version to be present. Additionally, after this refactor we can stop passing the URI to requests and just read it straight from the document.

### Implementation

I split in two commits to make it easier to review
- Add URI and version to Document, which also requires updating the Store and Executor to provide those values
- Update all references in tests


st0012 (approved):
👍

PR #573, opened 2023-03-23
Streerc support

### LINKED ISSUE #570: Support SyntaxTree options from .streerc file
👋🏼 First off, this project is really exciting, especially it's SyntaxTree integration!

Starting with [`v3.3.0`](https://github.com/ruby-syntax-tree/syntax_tree/blob/main/CHANGELOG.md#330---2022-08-02), SyntaxTree has supported storing CLI options in an `.streerc` file.  Right now, running the format command with this LSP only uses the SyntaxTree defaults.

Are there any plans to support that `.streerc` file?  

If there's an appetite for this from the maintainers' end, I'd be happy to take a swing at a PR.

molawson:
### Motivation

Fixes #570

SyntaxTree supports a config file that it's CLI commands (including its LSP) will pick up and use for formatting. Supporting that file makes formatting with SyntaxTree via this LSP consistent with other uses of SyntaxTree in a given repo.

### Implementation

I largely sought to mirror the custom runner setup that already exists for RuboCop formatting, first extracting a new runner class, then adding the new functionality. 

### Automated Tests

I added a new test for the `Formatting` class to exercise this new setup, testing the two types of options (print width & plugins) that SyntaxTree supports.

### Manual Tests

I used this branch for an existing project that uses an `.streerc` file and confirmed that it uses those config values and plugins.


lib/ruby_lsp/requests/support/syntax_tree_formatting_runner.rb
```diff
@@ -0,0 +1,39 @@
...
+                opts = SyntaxTree::CLI::Options.new
+                opts.parse(SyntaxTree::CLI::ConfigFile.new.arguments)
+                opts
+              end,
+              SyntaxTree::CLI::Options,
+            )
+        end
+
+        sig { params(_uri: String, document: Document).returns(T.nilable(String)) }
+        def run(_uri, document)
```
vinistock, line 29: Can we remove `uri` here? It's not used and these runners don't necessarily need to comply to the same exact interface.

vinistock:
This is looking great!

lib/ruby_lsp/requests/support/syntax_tree_formatting_runner.rb
molawson: Sure thing!

I called that out in the first commit. The thinking was that establishing an interface would be handy for other formatters, but I'm game for keeping this more focused the bits this new runner needs at this point.

vinistock: Oh, I didn't see the commit descriptions. Apologies!

That's a fair point. I guess it could be worth having an interface and it doesn't hurt to keep it.

vinistock (approved):

kddnewton (approved):
Thanks!!

vinistock:

PR #574, opened 2023-03-24
Avoid analyzing named regexp captures if left side is not a literal

vinistock:
### Motivation

When the left side is not a regex literal, then we can't analyze and identify local variables for named capture groups.

### Implementation

Return early if the left side of the comparison is anything but a regexp literal.

### Automated Tests

Added a test that catches the problem.

### DIFF
--- a/lib/ruby_lsp/requests/semantic_highlighting.rb
+++ b/lib/ruby_lsp/requests/semantic_highlighting.rb
@@ -305,6 +305,7 @@ module RubyLsp
         return unless node.operator == :=~
 
         left = node.left
+        # The regexp needs to be on the left hand side of the =~ for local variable capture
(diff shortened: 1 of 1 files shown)


lib/ruby_lsp/requests/semantic_highlighting.rb
```diff
@@ -305,6 +305,8 @@ def visit_binary(node)
         return unless node.operator == :=~
 
         left = node.left
+        return unless left.is_a?(SyntaxTree::RegexpLiteral)
```
paracycle, line 308: Can we add a comment here similar to the one in the previous early return?
```suggestion
        # The regexp needs to be on the left hand side of the =~ for local variable capture
        return unless left.is_a?(SyntaxTree::RegexpLiteral)
```

paracycle (approved):

PR #580, opened 2023-03-27
Restructure README

vinistock:
### Motivation

Our README has become quite bulky and it includes information that quickly goes out of date (e.g.: future plans that we're already working on or a link to a closed prototype PR).

We need to make sure relevant information is easy to find and not hidden away in the middle of too much text.

### Implementation

- Added a link to our public Ruby DX Slack workspace
- Took the relevant parts in the `Overview` section and put it in our initial `Ruby LSP` section. Removed everything else that was either outdated or soon to be outdated
- Moved learn more below usage. Being able to quickly understand how to use is more important
- Moved the debugging and spell checking sections under contributing since they are development related


README.md
```diff
@@ -2,56 +2,42 @@
...
-At the time of writing, these are the major differences between Ruby LSP and [Solargraph](https://solargraph.org/):
-
-* Solargraph [uses](https://solargraph.org/guides/yard) YARD documentation to gather information about your project and its gem dependencies. This provides functionality such as context-aware auto-completion and navigation ("Go To Definition")
-* Solargraph can be used as a globally installed gem, but Ruby LSP must be added to the Gemfile or gemspec if using RuboCop. (There are pros and cons to each approach)
-
-## Learn More
+## Usage
 
-* [RubyConf 2022: Improving the development experience with language servers](https://www.youtube.com/watch?v=kEfXPTm1aCI) ([Vinicius Stock](https://github.com/vinistock))
+If using VS Code, install the [Ruby LSP extension](https://github.com/Shopify/vscode-ruby-lsp) to get the extra features
```
st0012, line 14: I think we need to be extra explicit about NOT to add it to Gemfile while using VS Code. I missed 
`No other configuration is required and you can skip this section.` part on the first pass so I assume some readers may do that too.

Example:

```
## Usage

### With VS Code

Just install the Ruby LSP extension and there's no need to install this gem manually.

### With other editors

See the editors doc....
```

st0012 (approved):
Love the change 👍

PR #593, opened 2023-03-28
Allow registering hover listeners

### LINKED ISSUE #586: Add hover after_request middleware
As our first server extension point, let's add an after request middleware for hover where extensions can enhance the content displayed at the editor.

This involves
- Activating extensions
- Creating a mechanism to register the middleware
- Creating a well defined interface that extensions can use

vinistock:
### Motivation

Closes #586 

Add the ability for other gems to register hover listeners in the Ruby LSP.

### Implementation

- Add extension activation point. Notice that we can't use `ruby_lsp/extension.rb` as the path or else it conflicts with our own extension class. So extensions must be declared under `ruby_lsp/name_of_gem/extension.rb`
- Add ability to register listeners for hover

### Automated Tests

Added tests for the new functionality.


lib/ruby_lsp/extensions/base.rb
```diff
@@ -0,0 +1,55 @@
+# typed: strict
+# frozen_string_literal: true
+
+module RubyLsp
+  module Extensions
+    class Base
```
paracycle, line 6: I don't think I like the `RubyLsp::Extensions::Base` name.

A better name structure would be `RubyLsp::Extension` class that extension writers can subclass to define their `activate` method, and then `RubyLsp::Provider::Hover` as the class to inherit for a hover provider. Since the hover provider "provides" hover content, but is not the "extension" itself.

lib/ruby_lsp/extensions/hover.rb
```diff
@@ -0,0 +1,30 @@
...
+        end
+
+        sig { params(child_class: T.class_of(Extensions::Hover)).void }
+        def inherited(child_class)
+          hooks << child_class
+          super
+        end
+
+        sig { abstract.params(target: SyntaxTree::Node).returns(T.nilable(String)) }
+        def run(target); end
```
paracycle, line 26: What if the provider needs to reference other nodes in the document, or even other documents to get its job done?

Also, why a class method? Instance methods are easier to work with and the implementation might want to hold some kind of state in its implementation. If `run` was an instance method, the provider could accept a document, etc, state in its constructor, for example.

This was my biggest regret of the initial Tapioca compiler design, that I'd made the `decorate` method a class method.

lib/ruby_lsp/requests/hover.rb
```diff
@@ -54,27 +55,36 @@ def run
...
+          create_hover(target, "")
         end
+
+        return response if Extensions::Hover.hooks.empty?
+
+        # Run after request hooks coming from extensions and append the returned content
+        Extensions::Hover.hooks.each do |hook|
+          hook_response = hook.run(target)
+          response.contents.value << hook_response << "\n\n" unless hook_response.nil?
+        end
```
paracycle, line 68: This is not essential for this PR review, but I think this should all happen one layer above this one, i.e. at the layer where `Hover.new(document, position).run` is being called. I know that that creates a problem with how `target` is passed along, but I am sure we can find a way to handle that. For example, maybe finding the target is not the responsibility of this class in the first place.

lib/ruby_lsp/extensions/base.rb
```diff
@@ -0,0 +1,55 @@
...
+        sig { returns(T::Array[StandardError]) }
+        def load_extensions
+          # Require all extensions entry points, which should be placed under `some_gem/lib/ruby_lsp/extension.rb`
+          errors = $LOAD_PATH.filter_map do |path|
+            extension_path = File.join(path, "ruby_lsp", "extension.rb")
+            require(extension_path) if File.exist?(extension_path)
+            nil
+          rescue => e
+            e
+          end
```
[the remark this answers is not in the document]
vinistock: Nice! Didn't know about this one.

lib/ruby_lsp/requests/hover.rb
vinistock: What's the concern with having the hooks at this level? In hover, it will be a small inconvenience because of `target`, but for other requests it may become a pretty big issue and could require us to leak a lot of request state to `Executor`.

vinistock: Again, for hover, this might be straight forward because the only requirement is `target` and there are no visits. But for requests that involve visits and require mutating a lot of state, I think it could get messy very quickly. We also want to have a different degree of control and flexibility in our own requests when compared to extensions.

Besides being a purist view, what's the benefit in following that approach?

paracycle: I think I am not sure if we are setting up any of the other hover providers for success by not treating them on the same level as we are treating our own hover request handler. But then, I think we can also ship this and see if that's true or not.

lib/ruby_lsp/extensions/hover.rb
paracycle: > About the class method, there's no big reason for it. Even if you wanted state, you could still do
> 
> ```ruby
> def self.run(target)
>   new(target).do_something_else
> end
> ```
> 
> But I'm okay with changing to an instance method.

You can do that, but if we are providing a base class for the hover provider, then we might as well make the base provider have an `initialize` method that accepts a `target` parameter and stores it in a `target` attr reader, so that the provider just needs to subclass and implement a `run` method without any parameters. I am suggesting that that would make it simpler to implement a provider than implementing a stateless class method.

vinistock:
I've repurposed the PR to use the new listeners approach.

lib/ruby_lsp/extension.rb
```diff
@@ -0,0 +1,53 @@
...
+
+        # Activate each one of the discovered extensions. If any problems occur in the extensions, we don't want to
+        # fail to boot the server
+        extensions.each do |extension|
+          extension.activate
+        rescue => e
+          errors << e
+        end
+
+        errors
```
st0012, line 42: I wonder if it'll be better to not just return errors but also other info. For example, `errors` is actually an array of hashes like: 

```
{
  extension_file: # for error during require
  extension_class: # for error during activation
  error: e
}
```

Otherwise, if there are multiple extensions and an error like `NoMethodError` on `nil` is being raised, just seeing `Exception#message` won't be too helpful. At least we should let the user know which extension caused that?

andyw8: Along the same lines of what @st0012 mentioned, returning an array of `StandardError` instance feels not quite right here. Maybe if they could be encapsulated in something like an array of `ExtensionLoadResult`.

andyw8 (approved):
I expect we'll iterate on this some more, but I think it's a good enough state to merge in to start developing against internally.

lib/ruby_lsp/extension.rb
```diff
@@ -0,0 +1,53 @@
...
+          nil
+        rescue => e
+          e
+        end
+
+        # Activate each one of the discovered extensions. If any problems occur in the extensions, we don't want to
+        # fail to boot the server
+        extensions.each do |extension|
+          extension.activate
+        rescue => e
```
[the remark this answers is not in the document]
vinistock: Good idea.

lib/ruby_lsp/extension.rb
vinistock: Okay, I changed the way extensions are registered in the last commit to allow keeping track of names and better encapsulate errors. Let me know what you think.

lib/ruby_lsp/executor.rb
```diff
@@ -38,6 +38,22 @@ def run(request)
       when "initialize"
         initialize_request(request.dig(:params))
       when "initialized"
+        Extension.load_extensions
+
+        error_extensions = Extension.extensions.select(&:error?)
```
andyw8, line 43:
```suggestion
        errored_extensions = Extension.extensions.select(&:error?)
```

lib/ruby_lsp/extension.rb
```diff
@@ -0,0 +1,104 @@
+# typed: strict
+# frozen_string_literal: true
+
+module RubyLsp
+  # To register an extension, inherit from this class and implement both `name` and `activate`
```
andyw8, line 5: Should this comment be within `class Extension`?

andyw8 (approved):

lib/ruby_lsp/extension.rb
vinistock: I think the comment needs to be on top for it to be associated with the `Extension` class.

st0012 (approved):

Issue #584, opened 2023-03-28
Add code lens for running tests


vinistock:
Discover test examples using the code lens request and return `Run` entries for each one plus one for running the entire file at the class level.

We should
- Support only Minitest / test-unit for now
- Allow running specific examples or the entire test class
- Use the `data` attribute (which allows for any information we want) to
    - Specify that the code lens is a test (e.g.: `type: "test"`). This will make it easier to create a test controller on VS Code
    - Specify the command that the client should use to execute that test. It'll most likely be `TESTOPTS="-n='/NAME_OF_THE_TEST/'" ruby -Itest PATH_TO_TEST_FILE`. Do not use `rake` since we want to be able to run tests even in projects that may not be using it

**Note**: this also involves creating a VS Code command that is used to execute the test. The first version of this should only grab whatever command was received in the `data` attribute and run that in the terminal.

Documentation: https://microsoft.github.io/language-server-protocol/specification#textDocument_codeLens

adisonlampert:
@matteeyah 
> The main things that I think are missing as of right is starting a debugging session for a test

This is one of our goals. We've been working on setting up `ruby/debug` (for example, #431). Adding test running capability is just the first step 😃

PR #598, opened 2023-03-29
Add code lens for running entire test class

adisonlampert:
### Motivation

The first step of #584 

Using [code lens](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#textDocument_codeLens), identify test classes and (in combination https://github.com/Shopify/vscode-ruby-lsp/pull/506) allow users to run all of the tests in that class by clicking "Run" 
- See gif below for demo

The tests are run in the terminal with the command `bundle exec ruby -Itest  <test_path>` (or just `ruby` if there is no `Gemfile`)

### Implementation

- Create new request `CodeLens`
- Visit every class in the document and check if the name of the class ends with `Test`
  - If so, adds a new `Interface::CodeLens` to the results array
  - `Interface::CodeLens` contains the range, command information, and data (`type: "test"`)
  - The `command` is a `Interface::Command` that runs the VSCode command `rubyLsp.runTest` (defined here) with the terminal command passed in as an argument
 - The `command` will be visible to users when they are on a file with a test class (see gif demo below)
 - When "Run" is clicked, [`rubyLsp.runTest`](https://github.com/Shopify/vscode-ruby-lsp/pull/506/files#diff-25d66d74617fe2e23d7946bd6e3ba95640ab1b9bc8947445d604fc271c7c1f12R243) is called and the entire test class is run from the terminal

### Automated Tests

Added tests for the new functionality.

### Manual Tests

See manual test instructions in https://github.com/Shopify/vscode-ruby-lsp/pull/506.


lib/ruby_lsp/requests/code_lens.rb
```diff
@@ -0,0 +1,61 @@
...
+      def run
+        visit(@document.tree) if @document.parsed?
+        @results
+      end
+
+      sig { override.params(node: SyntaxTree::ClassDeclaration).void }
+      def visit_class(node)
+        class_name = node.constant.constant.value
+        if class_name.end_with?("Test")
+          add_code_lens(node, [@path, class_name, "ruby -Itest " + @path])
```
andyw8, line 45: Should this be prefixed with `bundle exec`?

vinistock: As far as I know, it shouldn't be necessary. This is how we run things with `debug` as well. @st0012 any thoughts?

andyw8: _Most_ of the time it won't matter, but there's a chance that someone could have a newer, incompatible version of a gem compared to what's in `Gemfile.lock`.

vinistock: Our current workflow with VS Code creates a custom bundle even for projects that don't have a `Gemfile`. I agree with Andy, we can do something like this
```ruby
class Codelens < BaseRequest
  BASE_COMMAND = T.let(File.exist?("Gemfile.lock") ? "bundle exec ruby" : "ruby", String)
end
```

vinistock (approved):

PR #607, opened 2023-04-03
Refactor hover with event emitter

vinistock:
### Motivation

Step towards #586 

Another proposed implementation for the extension system.

If we create a level of abstraction above `SyntaxTree::Visitor` using an event emitter and listener pattern, we are able to provide extensions with a consistent interface for implementing both positional and nonpositional requests. Additionally, our own requests can make use of it to also match the same interface as extensions.

Benefits:
- Consistency between all requests including our own implementations
- More resiliency against SyntaxTree changes. Since we listen to our own events, changes in SyntaxTree are less impactful
- Allows us to implement request merging for better performance. This is a more sophisticated step than this, but the general idea is that we can visit the AST only once for all nonpositional requests and gather information about everything at once. For example
```ruby
# On every nonpositional requests, we can always use all listeners to collect all of the information in a single
# visit run. Then we cache the results use the Document#version and just returned them straight away if it's
# already cached, saving us several passes on the document
listeners = [
  Requests::FoldingRange.new,
  Requests::DocumentSymbol.new,
  Requests::SemanticHighlighting.new,
]
EventEmitter(*listeners).visit(document.tree)
document.cache_results(listeners)
```

For extensions, this will allow us to implement something like
```ruby
listeners = [
  Requests::Hover.new,
  *Requests::Hover.extensions,
]
EventEmitter(*listeners).visit(document.tree)
# merge responses
```

### Implementation

- Moved `locate` to `Document`. It's necessary to make this PR work, but I also feel like it makes sense given that it's locating nodes inside of the tree
- Migrate existing requests to use `Document#locate` instead of `BaseRequest#locate`
- Create the `EventEmitter` and `Listener`
- Extract common helpers into a module. After all our requests are migrated, this module will not longer be necessary and we can instead move it all inside `Listener`
- Use the listener pattern to implement hover


lib/ruby_lsp/event_emitter.rb
```diff
@@ -0,0 +1,29 @@
...
+      @listeners = listeners
+
+      super()
+    end
+
+    # Emit events for a specific node. This is similar to the regular `visit` method, but avoids going deeper into the
+    # tree for performance
+    sig { params(node: T.nilable(SyntaxTree::Node)).void }
+    def emit_for_position(node)
+      case node
```
Morriar, line 31: I wonder if we should also trigger a `on_node` so the listener can subscribe to the one they want depending on the level of precision they need.

For example a listener that does the same thing for `Command` and `CallNode` could just listen to `on_node` and replicate the switch statement you have here rather than duplicating code across `on_command` and `on_call`.

This may be something we can add later if needed.

lib/ruby_lsp/requests/base_request.rb
```diff
@@ -31,94 +32,10 @@ def run; end
       # Syntax Tree implements `visit_all` using `map` instead of `each` for users who want to use the pattern
       # `result = visitor.visit(tree)`. However, we don't use that pattern and should avoid producing a new array for
       # every single node visited
-      sig { params(nodes: T::Array[SyntaxTree::Node]).void }
+      sig { params(nodes: T::Array[T.nilable(SyntaxTree::Node)]).void }
```
Morriar, line 35: Why do we need this to be nilable?

Authorizing nilables on `visit(node)` makes sense to make it easier to call `visit(node.something)` where `something` may be nil and we don't want to repeat the guard `if node.something`. But in `visit_all` shouldn't we always pass around collections of non-nilable nodes, even if we need to compact them?

If this change coming from the upstream signature maybe?

lib/ruby_lsp/event_emitter.rb
```diff
@@ -2,6 +2,18 @@
 # frozen_string_literal: true
 
 module RubyLsp
+  # EventEmitter is an intermediary between our requests and Syntax Tree visitors. It's used to visit the document's AST
```
Morriar, line 5: Should it be `SyntaxTree` in one word? Not sure.

lib/ruby_lsp/event_emitter.rb
```diff
@@ -2,6 +2,18 @@
...
+  # and emit events that the requests can listen to for providing functionality. Usages:
```
Morriar, line 6: Let's separate the usages list with a new line before and after, it will make the comment easier to read.

lib/ruby_lsp/listener.rb
```diff
@@ -2,13 +2,18 @@
...
+  # Listener is an abstract class to be used by requests for listening to events emitted when visiting an AST using the
+  # EventEmitter.
   class Listener
     extend T::Sig
     extend T::Helpers
     include Requests::Support::Common
 
     abstract!
 
+    sig { abstract.returns(Object) }
```
Morriar, line 14: It may be useful to document at least the abstract methods so implementers know how they should override them. An example would be nice too.

Morriar: Is there any kind of common interface we could use to type the return value? If not should we be introducing one? Or how is the system going to handle every kind of response here?

Morriar (commented):
This is very nice!

lib/ruby_lsp/event_emitter.rb
vinistock: Done!

Morriar (approved):

PR #609, opened 2023-04-04
Add code lens for test methods

### LINKED ISSUE #584: Add code lens for running tests

adisonlampert:
### Motivation

Closes #584 
A continuation of #598 

See updated gif for demo.

### Implementation

- Extended the functionality of the `CodeLens` request
- Added an instance variable `@modifier`
    - Visitor visits every `SyntaxTree::VCall` to check if it is an access modifier, if so updates `@modifier`
    - Visits every `SyntaxTree::Command` to check if it is a `public` access modifier, if so temporarily updates `@modifer` for that one method, then reverts the modifier back to what it was previously. (ex. `public def test_method(); end`)
- Visits every method definition to check if it is a public method to starts with "test"
   - If so, creates a new code lens for the method (setting the `path` to the specific line number the test begins and setting the `command` to the appropriate terminal command to run the test.
 - Refactored `add_code_lens` a bit to make the parameters easier to read and understand.

### Automated Tests

Updated automated testing.

### Manual Tests

See manual test instructions in https://github.com/Shopify/vscode-ruby-lsp/pull/506. 
- Run individual tests instead of running all class tests.
- Look at demo for example


lib/ruby_lsp/requests/code_lens.rb
```diff
@@ -44,17 +45,52 @@ def run
...
+      sig { override.params(node: SyntaxTree::DefNode).void }
+      def visit_def(node)
+        if @modifier == "public"
+          method_name = node.name.value
+          if method_name.start_with?("test")
+            add_code_lens(
+              node,
+              path: @path + ":" + node.location.start_line.to_s,
+              name: method_name,
+              command: BASE_COMMAND + " -Itest " + @path + " --name " + method_name,
```
andyw8, line 62: We should probably treat `method_name` as a regexp something like `%r{^{method_name}}` otherwise this may be run the wrong test if you have something like `test_foo` and `test_foo_another` where one is a substring of the other.

lib/ruby_lsp/requests/code_lens.rb
```diff
@@ -44,17 +45,52 @@ def run
...
+          add_code_lens(node, path: @path, name: class_name, command: BASE_COMMAND + " -Itest " + @path)
+        end
+        visit(node.bodystmt)
+      end
+
+      sig { override.params(node: SyntaxTree::DefNode).void }
+      def visit_def(node)
+        if @modifier == "public"
+          method_name = node.name.value
+          if method_name.start_with?("test")
```
andyw8, line 57:
```suggestion
          if method_name.start_with?("test_")
```

lib/ruby_lsp/requests/code_lens.rb
```diff
@@ -44,17 +45,52 @@ def run
...
+            add_code_lens(
+              node,
+              path: @path + ":" + node.location.start_line.to_s,
```
vinistock, line 60: The `:` technique won't work without Rails. We need to use `TESTOPTS="-n='/NAME_OF_TEST/'"` to select which one to run unfortunately.

Also, it seems like we have both `--name` and the line number. We should probably only use one. I think `path:` here is just `@path` and command is
```ruby
TESTOPTS="-n='/NAME_OF_TEST$/'" + BASE_COMMAND ...
```

test/fixtures/minitest_tests.rb
```diff
@@ -1,11 +1,15 @@
 class Test < Minitest::Test
-  def test_method
-  end
+  private def test_private_command; end
```
vinistock, line 2: Let's also add the following scenarios (which believe it or not are valid Ruby).
```ruby
private(def test_another_private
end)

public(def test_another_public
end)
```

lib/ruby_lsp/requests/code_lens.rb
adisonlampert: When I run something like:
```
TESTOPTS="-n='/test_initialize_enabled_features_with_array$/'" bundle exec ruby -Itest <PATH_TO_TEST>/test/executor_test.rb
```
It runs the entire test suite instead of the individual test.

vinistock: Sorry, it should actually just be the `--name` parameter. It accepts a complete string, which is what I believe we should use to avoid matching other tests.
```shell
bundle exec ruby -Itest <PATH_TO_TEST>/test/executor_test.rb --name test_initialize_enabled_features_with_array
```

lib/ruby_lsp/requests/code_lens.rb
andyw8: correction: `%r{^#{method_name}$}`

vinistock: Yeah, that one is just so that we don't have to type the entire name when running manually.

adisonlampert:
I started working on https://github.com/Shopify/vscode-ruby-lsp/issues/498 and discovered we will need to send `start_line`, `start_char`, `end_line`, and `end_char` for the test controller so I went ahead and added it in this PR.

lib/ruby_lsp/requests/code_lens.rb
```diff
@@ -86,11 +99,25 @@ def visit_vcall(node)
...
+            arguments: [
+              @path,
+              name,
+              command,
+              {
+                start_line: node.location.start_line,
+                start_char: node.location.start_char,
+                end_line: node.location.end_line,
+                end_char: node.location.end_char,
+              },
```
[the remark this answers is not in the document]
adisonlampert: To create a [`TestItem`](https://code.visualstudio.com/api/references/vscode-api#TestItem) in the Test Explorer, you need to supply a [`Range`](https://code.visualstudio.com/api/references/vscode-api#Range). Ranges need a `start_line`, `start_char`, `end_line`, and `end_char` so this info is necessary.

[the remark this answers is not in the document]
vinistock: In this case, we probably need to adjust lines by 1, since Syntax Tree is not zero based. And we probably want the column instead of the absolute character.

```suggestion
              {
                start_line: node.location.start_line - 1,
                start_column: node.location.start_column,
                end_line: node.location.end_line - 1,
                end_column: node.location.end_column,
              },
```

lib/ruby_lsp/requests/code_lens.rb
```diff
@@ -44,17 +45,79 @@ def run
...
+            @modifier = modifier
+          end
+        end
+      end
+
+      sig { override.params(node: SyntaxTree::VCall).void }
+      def visit_vcall(node)
+        vcall_value = node.value.value
+
+        if ["private", "protected", "public"].include?(vcall_value)
```
vinistock, line 95: Let's put this array in a constant to avoid allocating it for every test we find.

vinistock (approved):

Issue #532, opened 2023-03-02
Option to skip benchmark


andyw8:
While a PR is still at an early stage, the benchmarking comments can get a bit noisy. It would be nice to have an option to temporarily skip these, e.g. by putting `[skip bm]` in the commit message.

PR #613, opened 2023-04-05
Add option to skip benchmark in CI

andyw8:
### Motivation

Closes https://github.com/Shopify/ruby-lsp/issues/532

### Implementation

Based on https://github.com/veggiemonk/skip-commit/issues/5

(the repo is an action but the comments discuss how it can be done without).


andyw8:
🎉 https://github.com/Shopify/ruby-lsp/actions/runs/4620850296/jobs/8171527613

Issue #546, opened 2023-03-08
Only fallback to Syntax Tree if part of the project's `Gemfile` or `gemspec`


andyw8:
Currently, if RuboCop isn't present, we [fall back](https://github.com/Shopify/ruby-lsp/blob/e145cfb74928ab43e2da37ed13a2ae787633f463/lib/ruby_lsp/requests/formatting.rb#L62-L67) to using Syntax Tree for formatting.

With the [recent change](https://github.com/Shopify/vscode-ruby-lsp/pull/419) of allowing Ruby LSP to be used even if not included in the Gemfile, this may result in undesired formatting on projects which use neither RuboCop nor Syntax Tree.

We will change this so that we only fall back to Syntax Tree if it is listed in the `Gemfile` or `gemspec`, since that can be taken as a strong signal that formatting should be applied.

vinistock:
Sorry, I must disagree. I think this needs to be configured on the server or else it won't work for other editors.

To work with our long term plans of having server extensions, I think the right way to go is to
- Solve https://github.com/Shopify/vscode-ruby-lsp/issues/430 and https://github.com/Shopify/vscode-ruby-lsp/issues/399#issuecomment-1460465843. We need to receive another [initialization option](https://github.com/Shopify/ruby-lsp/blob/ac7d7132ee0337d5be9aff2646107d7d172644f3/lib/ruby_lsp/executor.rb#L312) to know which formatter was selected (and the default can be `nil`).
- Then the server will only broadcast the [formatting capability](https://github.com/Shopify/ruby-lsp/blob/ac7d7132ee0337d5be9aff2646107d7d172644f3/lib/ruby_lsp/executor.rb#L388) if:
    - Formatting is enabled
    - RuboCop or Syntax Tree are present in the original bundle if no formatter was selected in the initialization options
    - If a formatter was selected, then we use that one

st0012:
How can the server know about the original bundler 🤔 

BTW, I seriously think this should be resolved before we make an official `0.2.0` release on the extension because I have to disable `formatOnSave` for all projects now (can't do that just for irb or debug without committing).

andyw8:
Completed in https://github.com/Shopify/ruby-lsp/pull/654

Issue #384, opened 2022-12-11
Can we NOT enable formatting when the project doesn't have RuboCop?


st0012:
Currently, `ruby-lsp` uses `SyntaxTree`'s formatting when the project doesn't have RuboCop installed. I know the intention was good, but imo it provides bad experience to some new users.

Assumptions:

- If the project doesn't have RuboCop installed, it "likely" means there's no strong coding style preference in it. For example, when it's a personal project or it has few collaborators.
- When users want to try `ruby-lsp`, many will try it on existing projects.

If we combine both, some users (like me) would experience these:

1. Install `ruby-lsp` to see what it provides
2. Features like highlighting look good 👍 
3. But now it aggressively format my code and I need to manually pick changes when I commit them. This is annoying 😕
4. Go to setting and set `formatOnSave` to `false` (I actually think many users wouldn't know it and just uninstall the extension)
5. Lost formatting in other projects that do have RuboCop 🤦‍♂️

I understand that enforcing consistent coding style is a good practice. But I don't want `ruby-lsp` pushing me doing so and I feel most users would feel the same.

So I hope `ruby-lsp` won't enable `SyntaxTree` formatting by default.

PR #614, opened 2023-04-05
Only fallback to Syntax Tree if part of the project's `Gemfile` or `gemspec`

andyw8:
### Motivation

Closes https://github.com/Shopify/ruby-lsp/issues/546 and https://github.com/Shopify/ruby-lsp/issues/384

### Implementation

There is a chance that auto-detection could be wrong in particular situations: If a gem has a runtime dependency on RuboCop, but the project using that gem uses `syntax_tree`, it will incorrectly default to using RuboCop. We could choose to mitigate this by:
- displaying an error if the correct formatter cannot be reliably determined.
- not applying any formatting unless a specific formatter is set

### Automated Tests

Included.

### Manual Tests

(in each case below, checkout this branch and add `ruby-lsp` to the Gemfile, with the appropriate `path`).

* Choose a project which has neither RuboCop or syntax_tree
* Attempt to format a Ruby file
* Ensure no formatting is performance
* Choose a project which uses RuboCop
* Attempt to format a Ruby file
* Ensure it's formatted using RuboCop (confirm via the logging in the LSP output panel)
* Choose a project which uses Syntax Tree
* Attempt to format a Ruby file
* Ensure it's formatted using Syntax Tree (confirm via the logging in the LSP output panel)

### DIFF
--- a/lib/ruby_lsp/requests/formatting.rb
+++ b/lib/ruby_lsp/requests/formatting.rb
@@ -25,6 +25,10 @@ module RubyLsp
     # puts "Hello" # --> formatting: fixes the indentation on save
     # end
     # ```
+
+    SYNTAX_TREE_AS_DIRECT_DEPENDENCY = Bundler::LockfileParser.new(Bundler.read_file(Bundler.default_lockfile))
+      .dependencies.key?("syntax_tree")
+
     class Formatting < BaseRequest
       class Error < StandardError; end
       class InvalidFormatter < StandardError; end
@@ -40,7 +44,7 @@ module RubyLsp
           if formatter == "auto"
             if defined?(Support::RuboCopFormattingRunner)
               "rubocop"
-            elsif syntax_tree_as_direct_dependency?
+            elsif SYNTAX_TREE_AS_DIRECT_DEPENDENCY
               "syntax_tree"
             else
               "none"
@@ -90,11 +94,6 @@ module RubyLsp
         else
           raise InvalidFormatter, "Unknown formatter: #{@formatter}"
         end
-      end
-
-      sig { returns(T::Boolean) }
-      def syntax_tree_as_direct_dependency?
-        Bundler::LockfileParser.new(Bundler.read_file(Bundler.default_lockfile)).dependencies.key?("syntax_tree")
       end
     end
   end
--- a/lib/ruby_lsp/requests/formatting.rb
+++ b/lib/ruby_lsp/requests/formatting.rb
@@ -26,10 +26,11 @@ module RubyLsp
     # end
     # ```
 
-    SYNTAX_TREE_AS_DIRECT_DEPENDENCY = Bundler::LockfileParser.new(Bundler.read_file(Bundler.default_lockfile))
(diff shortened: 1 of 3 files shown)


test/requests/formatting_test.rb
```diff
@@ -167,4 +193,11 @@ def clear_syntax_tree_runner_singleton_instance
 
     T.unsafe(Singleton).__init__(RubyLsp::Requests::Support::SyntaxTreeFormattingRunner)
   end
+
+  def stub_syntax_tree(present:)
```
andyw8, line 197: Rather than trying to manipulate the loaded gems to mirror Bundler's behaviour, this seemed more straightforward.

lib/ruby_lsp/requests/formatting.rb
```diff
@@ -34,7 +38,15 @@ def initialize(document, formatter: "auto")
         @uri = T.let(document.uri, String)
         @formatter = T.let(
           if formatter == "auto"
-            defined?(Support::RuboCopFormattingRunner) ? "rubocop" : "syntax_tree"
+            if defined?(Support::RuboCopFormattingRunner)
+              warn("Using RuboCop for formatting")
```
vinistock, line 42: Let's remove these warns please.

lib/ruby_lsp/requests/formatting.rb
```diff
@@ -71,6 +85,7 @@ def run
       sig { returns(T.nilable(String)) }
       def formatted_file
         case @formatter
+        when "none"
```
vinistock, line 88: I don't think we need to add a branch for `none` if it just returns `nil`.

lib/ruby_lsp/requests/formatting.rb
```diff
@@ -79,6 +94,11 @@ def formatted_file
           raise InvalidFormatter, "Unknown formatter: #{@formatter}"
         end
       end
+
+      sig { returns(T::Boolean) }
+      def syntax_tree_as_direct_dependency?
+        Bundler::LockfileParser.new(Bundler.read_file(Bundler.default_lockfile)).dependencies.key?("syntax_tree")
```
vinistock, line 100: We should perform IO in the middle of requests or it creates the risk of switching threads in the middle of formatting. Also, it's wasteful to check this every time, we can just check this once while loading the class.

vinistock (commented):
I'm still hesitant about this change a bit. It feels to me that we should be able to offer some formatting without requiring people to declare extra dependencies. With this change, even if the formatter is configured to be syntax tree, it won't work until the gem is added as a dependency.

lib/ruby_lsp/requests/formatting.rb
andyw8: Removed, it was a remnant from a previous approach.

st0012:
> My main concern is to prevent surprising the user with unwanted formatting.

I agree with Andy. Surprising aside, it's also inconvenient despite the good intention. I still have formatting disabled "globally" because I don't want to format code in projects like irb or debug. But I also don't want to ship that config per-project.

PR #615, opened 2023-04-05
Improve robustness of extract to variable

### LINKED ISSUE #602: `Extract Variable` refactor won't prompt for variable name and misindents
On the heels of #594, I thought I should post this.

recording

You can see above that the refactored line did not indent correctly, and we were not prompted for a variable name.

For reference, [a good implementation](https://code.visualstudio.com/docs/editor/refactoring#_extract-variable).

vinistock:
### Motivation

Closes #602

This PR improves how we were finding the spot to place a variable extraction at. The code is unfortunately more complex than I wished for it to be, but there are some corner cases we have to handle.

### Implementation

The idea of finding the right spot is
1. If the expression selected is inside a single line block, then we have to extract it as `new_variable = 123; `
2. If the closest node to the selected expression covers the selected range, then it means we're extracting a nested expression. In these cases, we need to place the extraction one line above the closest node line to generate valid Ruby
3. Otherwise, we try to place it immediately under the closest expression

### Automated Tests

Added some tests to make sure these scenarios are working properly.


test/requests/code_action_resolve_expectations_test.rb
```diff
@@ -10,25 +10,8 @@ class CodeActionResolveExpectationsTest < ExpectationsTestRunner
...
-    stdout, _ = capture_io do
-      result = RubyLsp::Requests::CodeActionResolve.new(
-        document,
-        params,
-      ).run
-    end
 
-    assert_empty(stdout)
-    result
+    RubyLsp::Requests::CodeActionResolve.new(document, params).run
```
vinistock, line 14: I don't know why we were swallowing stdout here, but it's definitely not necessary.

PR #617, opened 2023-04-06
Avoid trying to parse document if there are no new edits

vinistock:
### Motivation

This is a tiny change, but that may provide considerable improvement. Before, we only cleared `@unparsed_edits` if we successfully parsed the document. What we actually want is to only try to parse the document if there are new edits we haven't tried parsing yet.

In the current approach, if someone types in a syntax error, we receive a bunch of requests from the editor (e.g.: folding range, document symbol, semantic highlighting...). Because trying to parse will fail and we won't clear `@unparsed_edits`, we'll actually try to parse again for each single one of these requests even though there's no possibility of successfully parsing. Parsing can only succeed after a new edit has been made.

### Implementation

The `@unparsed_edits` variable must indicate that there are edits waiting for a parse attempt. We should always clear the array even if parsing fails to prevent unnecessary parses.

### Automated Tests

Added a test that verifies the array is cleared even when there are syntax errors.

### DIFF
--- a/lib/ruby_lsp/document.rb
+++ b/lib/ruby_lsp/document.rb
@@ -77,9 +77,9 @@ module RubyLsp
     def parse
       return if @unparsed_edits.empty?
 
+      @unparsed_edits.clear
       @tree = SyntaxTree.parse(@source)
       @syntax_error = false
-      @unparsed_edits.clear
     rescue SyntaxTree::Parser::ParseError
       @syntax_error = true
     end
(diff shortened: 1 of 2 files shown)


vinistock:
Yes, I ended up catching this by chance. I noticed on type formatting was abnormally slow, especially on large files. This seemed odd since on type formatting does almost nothing other than checking a regex.

Turns out the issue was that, when autocompleting `end` tokens, the file is always in a syntax error state and so we were actually trying to parse the file before autocompleting. This makes it MUCH faster.

PR #618, opened 2023-04-06
Fix server encoding registration and prefer UTF-8

vinistock:
### Motivation

Our client encoding reading was wrong. The `positionEncodings` field is actually an array of encodings that the editor supports and not a single one. This means that our UTF-16 corrections weren't being properly applied.

The server must look into the list and select out of the possible ones its preference. Because Ripper doesn't support anything but UTF-8, we should always prefer that.

### Implementation

- If UTF-8 is a possible encoding, we pick that
- Otherwise, we try to use whatever encoding we received

### Automated Tests

Added positional encodings to the expected initialization capabilities.

### DIFF
--- a/bin/console
+++ b/bin/console
@@ -6,14 +6,4 @@ require "bundler/setup"
 require_relative "../lib/ruby_lsp/internal"
 require "irb"
 
-extend T::Sig
-
-sig { params(source: String).returns(RubyLsp::Document) }
-def new_doc(source)
-  RubyLsp::Document.new(source)
-end
-
-@source = T.let(File.read(File.expand_path("../lib/ruby_lsp/server.rb", __dir__)), String)
-@document = T.let(new_doc(@source), RubyLsp::Document)
-
 IRB.start(__FILE__)
--- a/lib/ruby_lsp/executor.rb
+++ b/lib/ruby_lsp/executor.rb
@@ -348,7 +348,16 @@ module RubyLsp
     sig { params(options: T::Hash[Symbol, T.untyped]).returns(Interface::InitializeResult) }
     def initialize_request(options)
       @store.clear
-      @store.encoding = options.dig(:capabilities, :general, :positionEncodings)
+
+      encodings = options.dig(:capabilities, :general, :positionEncodings)
+      @store.encoding = if encodings.nil? || encodings.empty?
+        Constant::PositionEncodingKind::UTF16
+      elsif encodings.include?(Constant::PositionEncodingKind::UTF8)
+        Constant::PositionEncodingKind::UTF8
+      else
+        encodings.first
+      end
+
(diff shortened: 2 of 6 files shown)


paracycle (approved):

lib/ruby_lsp/executor.rb
```diff
@@ -348,7 +348,9 @@ def completion(uri, position)
     sig { params(options: T::Hash[Symbol, T.untyped]).returns(Interface::InitializeResult) }
     def initialize_request(options)
       @store.clear
-      @store.encoding = options.dig(:capabilities, :general, :positionEncodings)
+
+      encodings = options.dig(:capabilities, :general, :positionEncodings)
+      @store.encoding = encodings.include?("utf-8") ? "utf-8" : encodings.first if encodings
```
paracycle, line 353: Can we use `Constant::PositionEncodingKind::UTF8` instead of `"utf-8"`?

lib/ruby_lsp/store.rb
```diff
@@ -9,8 +9,8 @@ module RubyLsp
   class Store
     extend T::Sig
 
-    sig { params(encoding: String).void }
-    attr_writer :encoding
+    sig { returns(String) }
+    attr_accessor :encoding
```
paracycle, line 13: I think we need to default this to `Constant::PositionEncodingKind::UTF16` since the spec states:
```
        /**
		 * The position encodings supported by the client. Client and server
		 * have to agree on the same position encoding to ensure that offsets
		 * (e.g. character position in a line) are interpreted the same on both
		 * side.
		 *
		 * To keep the protocol backwards compatible the following applies: if
		 * the value 'utf-16' is missing from the array of position encodings
		 * servers can assume that the client supports UTF-16. UTF-16 is
		 * therefore a mandatory encoding.
		 *
		 * If omitted it defaults to ['utf-16'].
		 *
		 * Implementation considerations: since the conversion from one encoding
		 * into another requires the content of the file / line the conversion
		 * is best done where the file is read which is usually on the server
		 * side.
		 *
		 * @since 3.17.0
		 */
		positionEncodings?: PositionEncodingKind[];
```

specifically because of the part that says: "If omitted it defaults to ['utf-16']."

vinistock: You are right. I did this in the initialize result instead just because we'd need to update many tests if we switched it directly in the store.

bin/console
```diff
@@ -6,14 +6,4 @@ require "bundler/setup"
 require_relative "../lib/ruby_lsp/internal"
 require "irb"
 
-extend T::Sig
```
vinistock, line 9: I tried to test something and noticed this was broken.

st0012 (approved):

Issue #504, opened 2023-02-16
Add rake task to check if all requests are benchmarked


vinistock:
With the addition of the new completion request, our benchmarks are already outdated - and that's bound to happen as we continue to add requests.

Let's add a rake task similar to [check_docs](https://github.com/Shopify/ruby-lsp/blob/main/rakelib/check_docs.rake) that verifies that every request is benchmarked at least once.

vinistock:
Hello! Thank you very much for the assistance. I assigned you to the issue.

What we want to do here is ensure that all requests that exist are benchmarked in `bin/benchmark`. I'm actually not 100% sure on what the best approach would be.

Part of the challenge is that in the benchmark, we run everything using the `Executor` class, which receives request methods (e.g.: `textDocument/foldingRange`), but we don't have a connection between that and the request class that handles the logic.

I see two options: using static analysis or changing the benchmark to check this at runtime. Static analysis will be difficult because of the point I made above. You can give it a try, but I think it won't be easy to achieve.

The other option would be, instead of creating a new rake task, simply changing the benchmark to ensure that every request class got executed. Since requests are autoloaded, you might be able to do this at the end of the benchmark by
1. Map which classes inherit from `BaseRequest` and `Listener` (we're in the process of doing some refactors, you'll need to check both). This map will only include classes that were autoloaded (requests that were executed)
2. Then eager load all requests by requiring everything under the `requests` folder
3. Map the classes again and check to see if there are any extra entries between the two lists. If there are, then it means some request didn't get executed because it wasn't autoloaded

PR #619, opened 2023-04-07
feat(benchmark): display missing requests

snutij:
### Motivation

Closes https://github.com/Shopify/ruby-lsp/issues/504

Display in benchmarks results if there is any missing request benchmark. 

### Implementation

The second option proposed here: https://github.com/Shopify/ruby-lsp/issues/504#issuecomment-1499068788

* Map all classes which inherit from `BaseRequest` or `Listener`
* Require everything under `ruby_lsp/requests`
* Re-map all classes which inherit from `BaseRequest` or `Listener`
* If a delta is present, display it. 

Two open questions: 

* Is the output OK this way? Should I format it like removing `RubyLsp::Requests` (or any other way)?

```bash
================================================================================
Missing benchmarks:

RubyLsp::Requests::PathCompletion
RubyLsp::Requests::CodeLens
RubyLsp::Requests::CodeActionResolve
```

* Should I edit `github/workflows/benchmark.yml` to make it fail if there are any missing requests?

### Automated Tests

I think there is no test on the task, but I may be wrong. Feel free to point it out if necessary.

### Manual Tests

There are already missing benchmarks so: 

* run `bin/benchmark` 
* output `PathCompletion`, `CodeLens` & `CodeActionResolve`
* comment an item in `requests` hash, re-run `bin/benchmark` it should be added on missing benchmarks

### DIFF
--- a/bin/benchmark
+++ b/bin/benchmark
@@ -314,6 +314,7 @@ if missing_benchmarks.any?
   puts "=" * 80
   puts "Missing benchmarks:\n\n"
   puts missing_benchmarks.map(&:to_s)
+  exit(1)
 end
 
 File.write(CACHE_FILE_PATH, results.to_json)
--- a/.github/workflows/benchmark.yml
+++ b/.github/workflows/benchmark.yml
@@ -58,7 +58,11 @@ jobs:
       # Fail the job based on the benchmark output
       - name: Finalize job
         run: |
-          if [[ "${{ steps.benchmark.outputs.REPORT }}" == *"At least one benchmark is slower than the main branch"* ]]; then
+          REPORT="${{ steps.benchmark.outputs.REPORT }}"
+          IS_SLOWER=$(grep "At least one benchmark is slower than the main branch" <<< "$REPORT")
+          IS_MISSING=$(grep "Missing benchmarks" <<< "$REPORT")
+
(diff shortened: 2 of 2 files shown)


snutij:
I've signed the CLA!

vinistock:
The implementation looks great! The only issue is we now need to add the missing benchmarks for CI to pass 😂. Would you like to take a stab at that too? Otherwise, we'll need to add the missing ones before we can merge this PR.

snutij:
I tried to add the missing benchmarks, but I'm not sure I did it right. 😆 Feel free to let me know if I should do something different!

bin/benchmark
```diff
@@ -298,4 +301,19 @@ if File.exist?(CACHE_FILE_PATH)
...
+
+all_requests = RubyLsp::Requests::BaseRequest.subclasses.concat(RubyLsp::Listener.subclasses)
+
+missing_benchmarks = all_requests - requests_benchmarked
+
+if missing_benchmarks.any?
+  puts "\n\n"
+  puts "=" * 80
+  puts "Missing benchmarks:\n\n"
+  puts missing_benchmarks.map(&:to_s)
```
vinistock, line 316: Here we need to `exit(1)`

vinistock (approved):

vinistock:

Issue #199, opened 2022-07-15
Explore adding an indexer to provide go to definition


vinistock:
Despite not being able to provide a fully correct go to definition implementation (which requires type checking), the Ruby LSP could provide a simpler version of it.

One possible idea is to index all existing classes and methods in the code base and then providing go to definition results for methods that match the same name. Despite being naive, it's better than no support at all.

### Ideas for exploration

- The indexer can run in a separate process (maybe a Ractor?) to not block the LSP start up
- The indexer can cache indices from files that haven't been modified and re-use those to speed up the initial start up
- The indexer can try to be a bit smart about argument count. In this case, it would find method definitions by searching not only the name, but argument count. Need to be careful about optionals and splats

Note: the Ruby LSP should not provide any go to definition at all if Sorbet is present, since Sorbet does proper type checking and returns correct go to definition results.

vinistock:
The [LSIF spec](https://microsoft.github.io/language-server-protocol/specifications/lsif/0.6.0/specification/) may be useful for what we're trying to do.

pheen:
👋 I wanted to drop-in and say Hi! I have tried to implement a solution for Go To Definition both [in Ruby](https://github.com/pheen/elastic_ruby_server/) which I couldn't get it to be reliable, so I moved to a Ruby parser written in Rust that I implemented an LSP on-top of here: https://github.com/pheen/fuzzy_ruby_vscode_client

I mentioned it because it's an example of what Go To Definition can look like using only heuristics, without any type annotations. It works reasonably well, but could be much better if more was added like lookbehind, argument counts, argument keyword names, etc. Unfortunately, I have more pressing side-projects and likely won't have time to work on this anymore.

Regarding LSIF, I looked at it and my interpretation is it will require two index passes. First, one to index all of the various tokens, then a second pass to populate all of the connections. That's not a problem if you're exporting that data from a CI build or something, but if you need to build an index quickly when a developer opens up their editor then that sounds rough. 

I'm looking forward to tracking this work and how it develops!

matteeyah:
The reason why I think `lsif` is the best approach is simply because a lot of other tools use it and build their functionality on top of it. Other approaches might be more efficient or easier to implement, but using lsif has the potential to unlock a bunch of tools that are inaccessible to Ruby codebases right now.

jmgarnier:
[...]

PR #632, opened 2023-04-12
Add basic index

vinistock:
### Motivation

Currently, we're rebuilding the prefix tree in path completion every time we execute it, which is super wasteful. We can create a simple version of what the #199 index is going to be just to keep track of files and manage the prefix tree.

(I also needed to test this out for my talk 😂)

### Implementation

1. Required `singleton` only in internal since we were doing it in multiple places
2. Created an Index to track files and the prefix tree
3. Started building the Index on `initialized` and watching Ruby files to synchronize it
4. Started using the index on path completion

### Automated Tests

Added tests for the Index and changed the ones for path completion.

### Manual Tests

There aren't any behaviour changes for path completion, it's basically just a refactor. I guess you could verify the require paths are still being completed.


lib/ruby_lsp/executor.rb
```diff
@@ -444,6 +455,34 @@ def initialize_request(options)
         )
       end
 
+      # # # Dynamically registered capabilities
+      file_watching_caps = options.dig(:capabilities, :workspace, :didChangeWatchedFiles)
+
+      # # Not every client supports dynamic registartion or file watching
```
paracycle, line 461:
```suggestion
      # # Not every client supports dynamic registration or file watching
```

lib/ruby_lsp/index.rb
```diff
@@ -0,0 +1,63 @@
...
+      @prefix_tree = T.let(Requests::Support::PrefixTree.new(@files), Requests::Support::PrefixTree)
+    end
+
+    sig { params(changes: T::Array[{ uri: String, type: Integer }]).void }
+    def synchronize(changes)
+      has_addition_or_removals = T.let(false, T::Boolean)
+
+      changes.each do |change|
+        # File change events include folders, but we're only interested in files
+        file_path = change[:uri].delete_prefix("file://")
```
paracycle, line 27: This will not always work properly. We should not be treating URIs as if they were plain strings, or we will get bitten by URL encoding problems.

For example `file:///some%20path/that%20includes%20spaces.txt` would fail this operation.

test/requests/path_completion_test.rb
```diff
@@ -160,17 +160,33 @@ def with_file_structure(&block)
...
-      ])
+      ]
+
+      files.each do |file|
+        FileUtils.touch(file)
+        RubyLsp::Index.instance.synchronize([{
+          uri: "file://#{file}",
+          type: RubyLsp::Constant::FileChangeType::CREATED,
+        }])
+      end
```
paracycle, line 178: Would this be better as:
```suggestion
      changes = files.map do |file|
        FileUtils.touch(file)

        {
          uri: "file://#{file}",
          type: RubyLsp::Constant::FileChangeType::CREATED,
        }
      end
      RubyLsp::Index.instance.synchronize(changes)
```

test/requests/path_completion_test.rb
```diff
@@ -160,17 +160,33 @@ def with_file_structure(&block)
...
       return block.call
     ensure
+      T.unsafe(files).each do |file|
+        FileUtils.rm(file)
+
+        RubyLsp::Index.instance.synchronize([{
+          uri: "file://#{file}",
+          type: RubyLsp::Constant::FileChangeType::DELETED,
+        }])
+      end
```
paracycle, line 189: Similarly:
```suggestion
      changes = T.unsafe(files).map do |file|
        FileUtils.rm(file)

        {
          uri: "file://#{file}",
          type: RubyLsp::Constant::FileChangeType::DELETED,
        }
      end
      RubyLsp::Index.instance.synchronize(changes)
```

lib/ruby_lsp/index.rb
```diff
@@ -0,0 +1,63 @@
...
+        next if File.directory?(file_path)
+
+        # Get the relative path based on the LOAD_PATH
+        base_load_path = $LOAD_PATH.find { |path| file_path.start_with?(path) }
+        next if base_load_path.nil?
+
+        file_path.delete_prefix!("#{base_load_path}/")
```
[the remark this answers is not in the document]
vinistock: Done.

lib/ruby_lsp/index.rb
```diff
@@ -0,0 +1,63 @@
...
+
+    sig { returns(T::Array[String]) }
+    attr_reader :files
+
+    sig { returns(Requests::Support::PrefixTree) }
+    attr_reader :prefix_tree
+
+    sig { void }
+    def initialize
+      @files = T.let(collect_files, T::Array[String])
```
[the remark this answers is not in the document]
vinistock: This makes it a bit harder to type since we'll need to make the prefix tree nilable. Given that, outside of tests, we will never really want an empty index, are you okay with us moving forward like this?

lib/ruby_lsp/index.rb
paracycle: I don't think you should need `Shellwords` at all, since we are not running anything in the shell. Ruby's `File` and `Pathname` classes can handle paths with spaces, etc inside them perfectly fine:
```ruby
[1] pry(main)> require "pathname"
=> false
[2] pry(main)> p = Pathname.new("some path/that includes spaces and an @ sign.txt")
=> #<Pathname:some path/that includes spaces and an @ sign.txt>
[3] pry(main)> p.parent.mkpath
=> nil
[4] pry(main)> p.write("foo bar baz")
=> 11
[5] pry(main)> p.read
=> "foo bar baz"
[6] pry(main)> .cat "some path/that includes spaces and an @ sign.txt"
foo bar baz
```

lib/ruby_lsp/index.rb
```diff
@@ -0,0 +1,63 @@
+# typed: strict
+# frozen_string_literal: true
+
+module RubyLsp
+  class Index
+    extend T::Sig
+    include Singleton
+
+    sig { returns(T::Array[String]) }
```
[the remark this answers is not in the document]
vinistock: Oh, good point. I got it now.

lib/ruby_lsp/index.rb
```diff
@@ -0,0 +1,73 @@
...
+          @files << require_path
+        when Constant::FileChangeType::CHANGED
+          # Do nothing for now
+        when Constant::FileChangeType::DELETED
+          has_addition_or_removals = true
+          @files.delete(require_path)
+        end
+      end
+
+      @prefix_tree = Requests::Support::PrefixTree.new(@files) if has_addition_or_removals
```
Morriar, line 59: I think `PrefixTree#insert` should be made public, and we could add `insert_all` so you don't have to initialize a new one here.

vinistock:
I don't know why Sorbet keeps failing complaining that the prefix tree is the same type but a different object id. Any clues?

PR #635, opened 2023-04-13
Windows compatibility

andyw8:
### Motivation

See https://github.com/Shopify/ruby-lsp/pull/620


bin/rdbg
```diff
@@ -1,18 +1,13 @@
 #!/usr/bin/env ruby
 # frozen_string_literal: true
 
-#
```
andyw8, line 4: Removed since we have already made alterations to the generated binstub.

st0012:
@andyw8 Does it build on windows at all? I think adding `ruby` to the bundler platform may make it compile locally?

Gemfile.lock
```diff
@@ -162,6 +160,8 @@ GEM
 
 PLATFORMS
   arm64-darwin
+  x64-mingw-ucrt
```
andyw8, line 163: "UCRT64: The Universal C Runtime (UCRT) is used by recent versions of Microsoft Visual Studio. It ships by default with Windows 10."

bin/benchmark
```diff
@@ -31,7 +32,7 @@ RubyLsp::Requests::Support::RailsDocumentClient.send(:search_index)
 RubyLsp::Requests::DocumentLink.gem_paths
 
 ITERATIONS = 1000
-CACHE_FILE_PATH = "/tmp/ruby_lsp_benchmark_results.json"
+CACHE_FILE_PATH = "#{Etc.systmpdir}/ruby_lsp_benchmark_results.json"
```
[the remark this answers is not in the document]
andyw8: They are the same for me, except that one has a trailing slash:

```
irb(main):005:0> Dir.tmpdir
=> "/var/folders/h7/3_mvpj6x3bv1cst7rcj1nkw00000gn/T"
irb(main):006:0> Etc.systmpdir
=> "/var/folders/h7/3_mvpj6x3bv1cst7rcj1nkw00000gn/T/"
```

But I'm not sure about Windows. In the [previous PR](https://github.com/Shopify/ruby-lsp/pull/620#issuecomment-1502759221) the author wrote:
> Now using Etc.systmpdir which sould be equal to node os.tmpdir() on all platforms where using vscode actually makes sense

andyw8:
@vinistock I had to update the branch protection rules to match the new names, so in-flight PRs will need to be rebased against main.

Screenshot 2023-04-14 at 12 49 15 PM

PR #636, opened 2023-04-13
Fix benchmark exit status

vinistock:
### Motivation

Not exactly sure why, but the code for the exit status was failing even when the benchmark passed.

### DIFF
--- a/.github/workflows/benchmark.yml
+++ b/.github/workflows/benchmark.yml
@@ -59,7 +59,7 @@ jobs:
       - name: Finalize job
         run: |
           report="${{ steps.benchmark.outputs.REPORT }}"
-          if [[ "$report" == *"At least one benchmark is slower than the main branch"* || "$report" == *"Missing benchmarks"*]]; then
+          if [[ "$report" == *"At least one benchmark is slower than the main branch"* ]] || [[ "$report" == *"Missing benchmarks"* ]]; then
             exit 1
           else
             exit 0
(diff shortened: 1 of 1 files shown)


.github/workflows/benchmark.yml
```diff
@@ -58,11 +58,8 @@ jobs:
       # Fail the job based on the benchmark output
       - name: Finalize job
         run: |
-          REPORT="${{ steps.benchmark.outputs.REPORT }}"
-          IS_SLOWER=$(grep "At least one benchmark is slower than the main branch" <<< "$REPORT")
-          IS_MISSING=$(grep "Missing benchmarks" <<< "$REPORT")
-
-          if [[ $IS_SLOWER || $IS_MISSING ]]; then
+          report="${{ steps.benchmark.outputs.REPORT }}"
+          if [[ "$report" == *"At least one benchmark is slower than the main branch"* || "$report" == *"Missing benchmarks"*]]; then
```
snutij, line 62: Sorry for the red CI. 😞 

I think here you should try:

```suggestion
          if [[ ("$report" == *"At least one benchmark is slower than the main branch"*) || ("$report" == *"Missing benchmarks"*) ]]; then
```

These are the two command result that is tested in the if condition.

vinistock: Your solution was correct, but the blank spaces were unicode characters instead of regular spaces. Thanks for the help!

PR #638, opened 2023-04-14
Fix benchmark script

andyw8:
Missed in https://github.com/Shopify/ruby-lsp/pull/635

### DIFF
--- a/bin/benchmark
+++ b/bin/benchmark
@@ -7,6 +7,7 @@ $VERBOSE = nil
 
 require "bundler/setup"
 require "sorbet-runtime"
+require "tmpdir"
 
 # Disable any Sorbet checks
 begin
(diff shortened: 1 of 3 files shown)


bin/benchmark
```diff
@@ -7,6 +7,7 @@ $VERBOSE = nil
 
 require "bundler/setup"
 require "sorbet-runtime"
+require "tmpdir"
```
andyw8, line 10: I'm confused why this is failing in CI, since this succeeds:

```
ruby -e "require 'tmpdir'; puts Dir.tmpdir"
```

andyw8: Ah, it's because that's from the `main` branch:
https://github.com/Shopify/ruby-lsp/blob/main/.github/workflows/benchmark.yml#L15

So this should be fine to merge.

Issue #611, opened 2023-04-04
Execute all nonpositional requests in one round of visits


vinistock:
Depends on #610 

Now that all requests are listeners, we can execute all the nonpositional requests in one round of visits to avoid having to go through the AST more than once.

For all of the different features involved, we will execute the same thing, always checking if there's a cached result already. For example, the executor implementation would look something like this
```ruby
case request[:method]
when "textDocument/foldingRange", "textDocument/documentSymbol", ...
  document = @store.get(uri)
  
  # If a cached result exists for the requested method and current document version, just return that straight away
  cached = document.cache_fetch(request[:method])
  return cached if cached

  # Otherwise, build the cache for all requests
  listeners = [
    Requests::FoldingRange.new,
    Requests::DocumentSymbol.new,
    Requests::SemanticHighlighting.new,
  ]
  EventEmitter(*listeners).visit(document.tree)
  document.save_cache(listeners)
  # Return the result for the right listener according to the requested method
end
```

andyw8:
Already being done step by step as we migrate to the new Emitter pattern.

PR #640, opened 2023-04-14
Make sending client messages async

vinistock:
### Motivation

Currently, we can only send messages to the client in association with a request. We basically push items to the `@messages` array either in the beginning or end of a request.

This is super limiting and kind of defeats the whole asynchronous approach of language servers, where messages can be sent at any time to the client.

For example, in the current design, it is impossible for the server to send notifications to the client to display progress for long running tasks, such as indexing, loading extensions or even just a slow request.

I believe it's safe to assume we won't be investing in actual parallelism for a while (maybe never?) given that
- We have been able to improve performance by a lot so far
- #611 will further improve performance
- Adopting YARP will greatly improve parsing performance

Based on this, I believe it is safe to assume that we can allow requests to push messages into a queue as they please to provide more flexibility in our implementations and provide a nicer experience.

### Implementation

The whole idea here is to have a `Thread::Queue` where we can push messages that have to be sent to the client. A Thread will watch that queue and write the messages back to the client. I believe the best place to put this is in our `Store` - since it is our global state object.

- The first commit makes Store a singleton. There's no reason why we'd have more than one global state and this makes it easier to use it without having to pass the store around
- The second commit adds the new `Thread::Queue` and `message_dispatcher` thread

### Automated Tests

This approach also makes it easier to test certain things, like the RuboCop failure window messages or that we clear diagnostics when closing files. I moved those tests to `executor_test.rb`, which is faster and less brittle.

### Manual Tests

This is a refactor, there shouldn't be any differences in behaviour.

### DIFF
--- a/lib/ruby_lsp/listener.rb
+++ b/lib/ruby_lsp/listener.rb
@@ -14,6 +14,11 @@ module RubyLsp
 
     abstract!
 
+    sig { params(message_queue: Thread::Queue).void }
+    def initialize(message_queue)
+      @message_queue = message_queue
+    end
+
     class << self
       extend T::Sig
 
--- a/bin/benchmark
+++ b/bin/benchmark
@@ -34,6 +34,7 @@ ITERATIONS = 1000
 CACHE_FILE_PATH = "/tmp/ruby_lsp_benchmark_results.json"
 
 def avg_bench(method, params)
+  message_queue = Thread::Queue.new
   results = (0...ITERATIONS).map do
     # Create a new store every time to prevent caching
     store = RubyLsp::Store.new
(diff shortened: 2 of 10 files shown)


st0012 (commented):
It feels a bit weird to put the message queue inside the `Store` class and invoke it when enqueueing messages because:

1. For the most part, `Store` still looks like the storage of documents, which shouldn't be caring about messages?
1. It doesn't look like it depends on `Store` much?

I feel a better way could be:

1. Put the message queue in server: `@message_queue = Thread::Queue.new`
2. We pass it to `Executor.new(@message_queue)`
3. In `executor.rb`, we just change `@messages <<` to `@message_queue <<`
4. In `server.rb`, we call `@message_queue.pop` instead

WDYT?

vinistock:
Based on Stan's suggestions, I have changed this to let the server own the message queue and pass it down as arguments to requests.

lib/ruby_lsp/listener.rb
```diff
@@ -14,6 +14,11 @@ class Listener
 
     abstract!
 
+    sig { params(message_queue: Thread::Queue).void }
+    def initialize(message_queue)
+      @message_queue = message_queue
```
st0012, line 19: Just a note: this is not covered by any tests yet.

PR #647, opened 2023-04-19
Make DocumentSymbol a Listener

### LINKED ISSUE #646: Migrate document symbol
Migrate document symbol to the new event emitter and listener strategy.

vinistock:
### Motivation

Closes #646

This PR migrates document symbol to the new listener strategy.

### Implementation

1. Started emitting visit events from the `EventEmitter`. Also removed the instance variable `@listeners` which isn't used for anything. For emitting events, I created `on_*` and `after_*` for nodes that create new scopes. This is necessary because certain requests (like document symbol) care about when we exit a certain scope (e.g.: when a class declaration ends)
2. Migrated executor to use the listener strategy
3. Migrated document symbol to use the listener strategy

### Automated Tests

I adapted the expectations test runner to properly run tests when a request is a listener. Other than that, existing tests should just continue to pass.


lib/ruby_lsp/executor.rb
```diff
@@ -82,7 +82,19 @@ def run(request)
...
+
+        # If the response has already been cached by another request, return it
+        cached_response = document.cache_get(request[:method])
+        return cached_response if cached_response
+
+        # Run listeners for the document
+        listener = Requests::DocumentSymbol.new(@message_queue)
+        EventEmitter.new(listener).visit(document.tree) if document.parsed?
+
+        # Store all responses retrieve in this round of visits in the cache. The last one we save must always be the one
```
andyw8, line 95:
```suggestion
        # Store all responses retrieved in this round of visits in the cache. The last one we save must always be the one
```

adisonlampert (approved):

st0012 (approved):

lib/ruby_lsp/listener.rb
```diff
@@ -19,11 +19,18 @@ def initialize(message_queue)
       @message_queue = message_queue
     end
 
+    @event_to_listener_map = T.let(Hash.new { |h, k| h[k] = [] }, T::Hash[Symbol, T::Array[T.class_of(Listener)]])
+
+    sig { params(event: Symbol).returns(T.nilable(T::Boolean)) }
```
andyw8, line 24: (nit) Worth removing the possibility of a nil here? e.g.

```
    sig { params(event: Symbol).returns(T::Boolean) }
    def registered_for_event?(event)
      Listener.event_to_listener_map[event]&.include?(self.class) || false
    end
```

lib/ruby_lsp/listener.rb
```diff
@@ -42,7 +49,10 @@ def add_listener(listener)
       def listener_events(&block)
         current_methods = instance_methods
         block.call
-        @events = T.let(instance_methods - current_methods, T.nilable(T::Array[Symbol]))
+
+        (instance_methods - current_methods).each do |event|
+          T.must(Listener.event_to_listener_map[event]) << self
```
andyw8, line 54:
```suggestion
          Listener.event_to_listener_map.fetch(event) << self
```

andyw8 (approved):

lib/ruby_lsp/listener.rb
vinistock: Not sure there's a lot of benefit other than the small simplification in the signature.

lib/ruby_lsp/listener.rb
vinistock: I'm not really a fan of this `fetch` pattern to be honest. Unless there's consensus from the team that we should be using it, I prefer hash access.

PR #650, opened 2023-04-21
Remove server telemetry

vinistock:
### Motivation

The client side is implemented in https://github.com/Shopify/vscode-ruby-lsp/pull/571.

Doing benchmarks in the server itself is not reliable and doesn't include the entire roundtrip that requests make until they are processed in the editor.

Additionally, benchmarking stuff on the server means it will spend time sending telemetry events to the client for every single request and notification.

### Implementation

1. Stopped sending `telemetry/event` notifications completely
2. Started returning error information as part of the request response, which we can read in the client

### Automated Tests

Adjusted the related tests for the removed functionality.


andyw8:
Moving to backlog as discussed at team meeting on 2023-06-12.

PR #654, opened 2023-04-24
Determine formatter for 'auto' (no longer fallback to `syntax_tree`).

andyw8:
### Motivation

Ruby LSP supports two formatters, RuboCop and syntax_tree. When formatting is set to `auto`, we check for RuboCop, and use that if available, otherwise we fallback to syntax tree.

This means for the repos which use neither, we were applying syntax tree formatting. This led to unexpected changes in formatting.

### Implementation

A better way is to check for the presence of RuboCop or syntax_tree in `Gemfile.lock`. But we want to avoid false positives through transitive dependencies, so we only check for **direct** dependencies.

We also want to support RuboCop-based gems such as `rubocop-shopify`, so we pattern match on gems beginning named `^rubocop`.

### Automated Tests

Included

### Manual Tests

With `rubyLsp.formatting` set to `auto`:

* Open a repo which doesn't use any formatter, e.g. `ruby/debug`. Add `ruby-lsp` to the Gemfile, pointing to this branch. Ensure that no formatting is applied when saving a file.
* Open a repo which uses RuboCop, e.g. `Shopify/tapioca`. Add `ruby-lsp` to the Gemfile, pointing to this branch. Ensure that RuboCop formatting is applied when saving a file.
* Open a repo which uses Syntax Tree, e.g. https://github.com/andyw8/syntax_tree_example. Add `ruby-lsp` to the Gemfile, pointing to this branch. Ensure that Syntax Tree formatting is applied when saving a file.

With `rubyLsp.formatting` set to `none`:

* Open a repo which uses RuboCop, e.g. Shopify/tapioca. Add ruby-lsp to the Gemfile, pointing to this branch. Ensure that formatting is **not** applied when saving a file.

With `rubyLsp.formatting` set to `rubocop`:

* Open a repo which doesn't use any formatter, e.g. `ruby/debug`. Add `ruby-lsp` to the Gemfile, pointing to this branch. Try format the file and verify you see an error "Formatter is set to 'rubocop' but RuboCop was not found".

With `rubyLsp.formatting` set to `syntax_tree`:

* Open a repo which uses RuboCop, e.g. Shopify/tapioca. Add ruby-lsp to the Gemfile, pointing to this branch. Ensure that **syntax_tree** formatting is applied when saving a file.

(syntax_tree is always available since it ships with Ruby LSP)

**Known Limitations**

On a repo which has syntax_tree but also has a gem named `rubocop-*`, e.g. https://github.com/discourse/discourse it will wrongly use RuboCop formatting.

### DIFF
--- a/lib/ruby_lsp/executor.rb
+++ b/lib/ruby_lsp/executor.rb
@@ -54,6 +54,8 @@ module RubyLsp
           warn(errored_extensions.map(&:backtraces).join("\n\n"))
         end
 
+        check_formatter_is_available
+
         warn("Ruby LSP is ready")
         VOID
       when "textDocument/didOpen"
@@ -259,6 +261,9 @@ module RubyLsp
 
     sig { params(uri: String).returns(T.nilable(T::Array[Interface::TextEdit])) }
     def formatting(uri)
+      # If formatter is set to `auto` but no supported formatting gem is found, don't attempt to format
+      return if @store.formatter == "none"
+
       Requests::Formatting.new(@store.get(uri), formatter: @store.formatter).run
     end
 
@@ -381,7 +386,11 @@ module RubyLsp
       end
 
       formatter = options.dig(:initializationOptions, :formatter)
-      @store.formatter = formatter unless formatter.nil?
+      @store.formatter = if formatter == "auto"
+        detected_formatter
+      else
+        formatter
+      end
 
       configured_features = options.dig(:initializationOptions, :enabledFeatures)
       experimental_features = options.dig(:initializationOptions, :experimentalFeaturesEnabled)
@@ -489,6 +498,40 @@ module RubyLsp
           code_lens_provider: code_lens_provider,
         ),
       )
+    end
+
+    sig { returns(String) }
+    def detected_formatter
+      # NOTE: Intentionally no $ at end, since we want to match rubocop-shopify, etc.
+      if direct_dependency?(/^rubocop/)
+        "rubocop"
+      elsif direct_dependency?(/^syntax_tree$/)
+        "syntax_tree"
+      else
+        "none"
+      end
+    end
+
+    sig { params(gem_pattern: Regexp).returns(T::Boolean) }
+    def direct_dependency?(gem_pattern)
+      Bundler.locked_gems.dependencies.keys.grep(gem_pattern).any?
+    end
+
+    sig { void }
+    def check_formatter_is_available
+      # Warn of an unavailable `formatter` setting, e.g. `rubocop` on a project which doesn't have RuboCop.
+      # Syntax Tree will always be available via Ruby LSP so we don't need to check for it.
+      return unless @store.formatter == "rubocop"
+
+      unless defined?(Support::RuboCopFormattingRunner)
+        @message_queue << Notification.new(
+          message: "window/showMessage",
+          params: Interface::ShowMessageParams.new(
+            type: Constant::MessageType::ERROR,
+            message: "Ruby LSP formatter is set to `rubocop` but RuboCop was not found in the bundle.",
+          ),
+        )
+      end
     end
   end
 end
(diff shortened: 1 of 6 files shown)


lib/ruby_lsp/requests/formatting.rb
```diff
@@ -34,14 +40,30 @@ def initialize(document, formatter: "auto")
         @uri = T.let(document.uri, String)
         @formatter = T.let(
           if formatter == "auto"
-            defined?(Support::RuboCopFormattingRunner) ? "rubocop" : "syntax_tree"
+            if rubocop_as_direct_dependency
```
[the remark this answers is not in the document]
andyw8: @vinistock would it make sense to persist this on `@store`?

andyw8:
> What do you think about moving this logic to `Executor` instead and just saving the result in the original `@formatter` option?

I've moved it into there. I have a slight concern that `Executor` is growing quite large (> 500 lines), so we might want to starting thinking about if there's anything that could be extracted out of there.

lib/ruby_lsp/executor.rb
```diff
@@ -12,6 +12,7 @@ def initialize(store, message_queue)
       # store
       @store = store
       @message_queue = message_queue
+      @store.formatter = detected_formatter
```
vinistock, line 15: When formatter is `auto`, this is going to perform the detection on every single request - including when it's not formatting.

We need to move this to the initialize request, so that we only set it once. [Here](https://github.com/Shopify/ruby-lsp/blob/2b608f98501dd65f6561c33885ed172bbc407ac0/lib/ruby_lsp/executor.rb#L385) where we set the formatter, you can just invoke the logic and set the right one there.

Then only the formatting request will read `@store.formatter`.

test/integration_test.rb
```diff
@@ -385,6 +385,7 @@ def initialize_lsp(enabled_features, experimental_features_enabled: false)
         initializationOptions: {
           enabledFeatures: enabled_features,
           experimentalFeaturesEnabled: experimental_features_enabled,
+          formatter: "rubocop",
```
andyw8, line 388: (I'm confirming if this is valid).

andyw8:
Discourse is an interesting edge case for this: It uses syntax_tree, but also has a few custom cops in a `rubocop-discourse` gem.

https://github.com/discourse/discourse/blob/552697e194c840b687b00c6125504734121205ef/Gemfile#L164-L171

One thing we could do to help improve the accuracy is to check for a `.streerc` file.

vinistock:
In ambiguous cases, I would say the best option is to configure the formatter instead of using `auto`.

test/executor_test.rb
```diff
@@ -141,4 +141,36 @@ def test_did_close_clears_diagnostics
...
+  end
+
+  def test_gives_rubocop_precedence_if_syntax_tree_also_present
+    stub_dependencies(rubocop: true, syntax_tree: true)
+    RubyLsp::Executor.new(@store, @message_queue).send(
+      :initialize_request,
+      { initializationOptions: { formatter: "auto" } },
+    )
+    assert_equal("rubocop", @store.formatter)
+  end
```
st0012, line 166: Don't we also want to test if it sets `none` when rubocop and syntax_tree both aren't present?

andyw8:
@st0012 I have verified on Core as suggested.

st0012 (approved):
Amazing, I've been waiting for this for so long 😂

lib/ruby_lsp/executor.rb
```diff
@@ -260,6 +260,9 @@ def semantic_tokens_full(uri)
 
     sig { params(uri: String).returns(T.nilable(T::Array[Interface::TextEdit])) }
     def formatting(uri)
+      # If formatter is set to `auto` but no supported formatting gem is found, don't attempt to format
+      return [] if @store.formatter == "none"
```
vinistock, line 264: We can just avoid the array allocation here.
```suggestion
      return if @store.formatter == "none"
```

vinistock (approved):

PR #660, opened 2023-04-26
Migrate document link to listener

### LINKED ISSUE #659: Migrate document link

vinistock:
### Motivation

Closes #659

Migrate our document link request to use the listener pattern. This PR also demonstrates how we would cache multiple responses and then just return the cached values on subsequent requests.

### Implementation

1. Migrated document link to use `Listener`
2. Started executing document link together with document symbol
3. We cache both responses and return only the response relevant to the current request. This way we can retrieve information about multiple requests in a single AST visit

### Automated Tests

Adjusted some tests, but this is mostly a refactor and current tests should cover the functionality.

### Manual Tests

If you wish to see the caching idea in practice, you can try this out:
1. In `.vscode/settings.json`, change the trace server configuration to `off` to make it easier to see output
2. Add `warn "Returning cached response for #{request[:method]} (#{request[:id]})"` when returning a cached response
3. Add `warn "Running request #{request[:method]} (#{request[:id]})"` when running the event emitter
4. In the output tab, verify that we never execute both document link and document symbol. One of them should be executed, but then the next should always returned a cached response

### DIFF
--- a/lib/ruby_lsp/executor.rb
+++ b/lib/ruby_lsp/executor.rb
@@ -79,11 +79,9 @@ module RubyLsp
         )
       when "textDocument/foldingRange"
         folding_range(uri)
-      when "textDocument/documentLink"
-        document_link(uri)
       when "textDocument/selectionRange"
         selection_range(uri, request.dig(:params, :positions))
-      when "textDocument/documentSymbol"
+      when "textDocument/documentSymbol", "textDocument/documentLink"
         document = @store.get(uri)
 
         # If the response has already been cached by another request, return it
@@ -91,12 +89,15 @@ module RubyLsp
         return cached_response if cached_response
 
         # Run listeners for the document
-        listener = Requests::DocumentSymbol.new(@message_queue)
-        EventEmitter.new(listener).visit(document.tree) if document.parsed?
+        document_symbol = Requests::DocumentSymbol.new(uri, @message_queue)
+        document_link = Requests::DocumentLink.new(uri, @message_queue)
+        EventEmitter.new(document_symbol, document_link).visit(document.tree) if document.parsed?
 
-        # Store all responses retrieved in this round of visits in the cache. The last one we save must always be the
-        # one related to the request we're receiving since that's the response we want to return to the editor
-        document.cache_set(request[:method], listener.response)
+        # Store all responses retrieve in this round of visits in the cache and then return the response for the request
+        # we actually received
+        document.cache_set("textDocument/documentSymbol", document_symbol.response)
+        document.cache_set("textDocument/documentLink", document_link.response)
+        document.cache_get(request[:method])
       when "textDocument/semanticTokens/full"
         semantic_tokens_full(uri)
       when "textDocument/semanticTokens/range"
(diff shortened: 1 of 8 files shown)


vinistock:
Yes! The idea is to execute all nonpositional requests together in a single round of AST visits #611.

That way, if we have 8 nonpositional requests, instead of visiting the AST 8 times, we just visit once when receiving any of the 8 requests, cache all responses and then we can just return the appropriate response from the cache.

lib/ruby_lsp/executor.rb
```diff
@@ -77,24 +77,25 @@ def run(request)
...
-        document.cache_set(request[:method], listener.response)
+        document_symbol = Requests::DocumentSymbol.new(uri, @message_queue)
+        document_link = Requests::DocumentLink.new(uri, @message_queue)
+        EventEmitter.new(document_symbol, document_link).visit(document.tree) if document.parsed?
+
+        # Store all responses retrieve in this round of visits in the cache and then return the response for the request
+        # we actually received
+        document.cache_set("textDocument/documentSymbol", document_symbol.response)
+        document.cache_set("textDocument/documentLink", document_link.response)
+        document.cache_get(request[:method])
```
bitwise-aiden, line 98: Is this supposed to be `cache_get`?

lib/ruby_lsp/executor.rb
```diff
@@ -77,24 +77,25 @@ def run(request)
...
-        # related to the request we're receiving since that's the response we want to return to the editor
-        document.cache_set(request[:method], listener.response)
+        document_symbol = Requests::DocumentSymbol.new(uri, @message_queue)
+        document_link = Requests::DocumentLink.new(uri, @message_queue)
+        EventEmitter.new(document_symbol, document_link).visit(document.tree) if document.parsed?
+
+        # Store all responses retrieve in this round of visits in the cache and then return the response for the request
+        # we actually received
+        document.cache_set("textDocument/documentSymbol", document_symbol.response)
+        document.cache_set("textDocument/documentLink", document_link.response)
```
bitwise-aiden, line 97: I see that we are storing these, but should we be retrieving them for use at any point?

bitwise-aiden: Ignore this, I see now that we are setting both to then use the request method as the get, also getting the cached value at the top of the method.

bitwise-aiden (approved):

lib/ruby_lsp/executor.rb
```diff
@@ -77,24 +77,25 @@ def run(request)
...
-        EventEmitter.new(listener).visit(document.tree) if document.parsed?
-
-        # Store all responses retrieve in this round of visits in the cache. The last one we save must always be the one
-        # related to the request we're receiving since that's the response we want to return to the editor
-        document.cache_set(request[:method], listener.response)
+        document_symbol = Requests::DocumentSymbol.new(uri, @message_queue)
+        document_link = Requests::DocumentLink.new(uri, @message_queue)
+        EventEmitter.new(document_symbol, document_link).visit(document.tree) if document.parsed?
+
+        # Store all responses retrieve in this round of visits in the cache and then return the response for the request
```
andyw8, line 94:
```suggestion
        # Store all responses retrieved in this round of visits in the cache and then return the response for the request
```
(optional since was already present)

vinistock: This was fixed on the branch this was based on 👍

PR #667, opened 2023-04-28
Fix check for RuboCop availability

andyw8:
### Motivation

The unavailable RuboCop warning was appearing even when RuboCop was available.

### Implementation

The previous approach seems somewhat fragile since it depends on the order that other classes are loaded. I think trying to require the gem is a better approach.

### Automated Tests

n/a

### Manual Tests

* Open a repo which uses RuboCop, e.g. `Shopify/tapioca`. Add `ruby-lsp` to the Gemfile, pointing to this branch. Ensure that RuboCop formatting is applied when saving a file, and that no warning about a missing RuboCop is shown.


lib/ruby_lsp/executor.rb
```diff
@@ -523,7 +523,14 @@ def check_formatter_is_available
       # Syntax Tree will always be available via Ruby LSP so we don't need to check for it.
       return unless @store.formatter == "rubocop"
 
-      unless defined?(Support::RuboCopFormattingRunner)
+      rubocop_available = begin
+        require "rubocop"
+        true
```
andyw8, line 528: Needed because if something else has already called `require "rubocop"` then the second require will return false.

lib/ruby_lsp/executor.rb
```diff
@@ -523,7 +523,14 @@ def check_formatter_is_available
       # Syntax Tree will always be available via Ruby LSP so we don't need to check for it.
       return unless @store.formatter == "rubocop"
 
-      unless defined?(Support::RuboCopFormattingRunner)
+      rubocop_available = begin
+        require "rubocop"
```
vinistock, line 527: This is basically a copy of what we're doing in the formatting and diagnostic runner. Instead of repeating the require for rubocop, we can just move the require for the diagnostic and formatting runners to `internal`

vinistock (changes requested):
Let's add tests to executor to verify this functionality. You can check that the right notification is pushed to a message queue and verify that `@store` keeps the right formatter.

andyw8:
I'm having some trouble getting the tests stable for this. As agreed with @vinistock, to reduce the annoyances for those working on `main`, I've fixed the behaviour here without adding a test, and will aim to add a stable test in https://github.com/Shopify/ruby-lsp/pull/674

test/executor_test.rb
```diff
@@ -174,8 +174,43 @@ def test_sets_formatter_to_none_if_neither_rubocop_or_syntax_tree_are_present
...
+    end
+  end
+
   private
 
+  def with_uninstalled_rubocop(&block)
+    rubocop_paths = $LOAD_PATH.select { |path| path.include?("gems/rubocop") }
+    rubocop_paths.each { |path| $LOAD_PATH.delete(path) }
+    RubyLsp::Requests::Support.send(:remove_const, :RuboCopRunner)
+    require "ruby_lsp/requests/support/rubocop_runner"
```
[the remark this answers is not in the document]
andyw8: I've changed approach: https://github.com/Shopify/ruby-lsp/pull/674

But still having some trouble making it stable.

Issue #531, opened 2023-03-01
Use Ruby fixtures with expected final state for code action refactor tests


vinistock:
Instead of matching the result against and array of text edits, it would be much nicer if we could use a Ruby file as a fixture with the final state after the refactor.

For this, we may need to take the edits returned by running the request, apply them to the document instance and then compare the fixture with `document.source`.

andyw8:
It could possibly use the `diff` format.

snutij:
This is perfect! Thank you for the guidance, I'll go into it, and if needed I open a draft. Feel free to assign it to me 👍

vinistock:
Yes, you would need to put the source inside the result key. So the expectation files would look something like this
```json
{
  "params": [...],
  "result": "
    def method
      abc = "abc"
      foo = 123
      # ...
    end
  " 
}
```
I don't know if we can have multiline strings like this in a JSON. If we can't and need to squish the entire expectation into a single line using `\n`, then I think we need to reconsider since it won't really improve things.

snutij:
This is not possible to have multiline in JSON. For my previous example `extraction_with_empty_line_between` this will lead to this expected result:

```
"params": [foo],
"result":  "def method\n  abc = \"abc\"\n\n  foo = 123\n\n \n  new_variable = { some: \"hash\" } hash = {\n    info: new_variable,\n  }\nend\n"
``` 

In my opinion (someone unfamiliar with LSP specification) is still more readable than an array of text edits. But not sure it's worth implementing it and having a pattern not similar to other requests for this specific request.   

I let you decide, both are OK to me 👍

vinistock:
Yeah, I agree with you that it's not worth making this request different than others for now. We may want to re-think our test framework in a way that better handles parameters, but I think it's not worth the investment immediately.

Thank you for looking into it 🙏.

PR #673, opened 2023-05-01
Remove exit code benchmark

snutij:
### Motivation

- Since failing is not monitored anymore on CI (https://github.com/Shopify/ruby-lsp/pull/531), this exit should be removed too, to avoid dead code 👍 .

- Then I saw that PathCompletion wasn't loaded anymore, so the benchmark wasn't run on the request really. I fixed it but can be reverted.

### Implementation

- Simply remove the `exit(1)` which is not caught anymore. 
- Change the position param to be sure that `lib/ruby_lsp/executor.rb#completion` load the PathCompletion request.

### Automated Tests

Not relevant

### Manual Tests

Run the `bin/benchmark`, output is not a failing command if a request is slower.

### DIFF
--- a/bin/benchmark
+++ b/bin/benchmark
@@ -299,7 +299,6 @@ if File.exist?(CACHE_FILE_PATH)
     # If this phrase is changed, we have to update .github/workflows/benchmark.yml since this is used to determine
     # whether to fail the build or not
     puts "\n\nAt least one benchmark is slower than the main branch."
-    exit(1)
   end
 end
 
--- a/bin/benchmark
+++ b/bin/benchmark
@@ -296,8 +296,6 @@ if File.exist?(CACHE_FILE_PATH)
   end
 
   unless success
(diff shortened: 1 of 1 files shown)


bin/benchmark
```diff
@@ -250,7 +250,7 @@ requests = {
   "textDocument/codeAction" => code_action_params,
   "textDocument/onTypeFormatting" => base_params.merge(position: { line: 1, character: 31 }, ch: "\n"),
   "codeAction/resolve" => base_params.merge(data: { range: range, uri: FILE_URI }),
-  "textDocument/completion" => base_params.merge(position: position),
+  "textDocument/completion" => base_params.merge(position: { line: 1, character: 50 }),
```
vinistock, line 253: Line 1 character 50 in the fixture matches the magic comment `frozen_string_literal`. The position here needs to match a forward slash `/` inside of a require path.

vinistock (approved):

vinistock:
Looks good. Thank you for the contribution!

PR #675, opened 2023-05-01
Switch to RDoc and expose check_docs rake task to gems

### LINKED ISSUE #648: Define a consistent way to document features for extensions and the Ruby LSP itself

### Use case

When authors start writing new extensions, we'll want them to have tooling to make sure their features are documented. Having features documented is important for discovery and providing a standard mechanism from the Ruby LSP itself will ensure consistency.

### Description

It would be nice to have a standard way to document features inside of the Ruby LSP itself that extension authors can also use. We could also provide tooling to verify if all listeners are documented.

### Implementation

1. Define a good way to document features. Preferably something that includes GIFs similar to what we already have
2. Export rake tasks from the Ruby LSP that allow authors to check if their listeners are documented

vinistock:
### Motivation

Closes #648

Don't review yet.


st0012 (commented):
With rdoc I think we can still write docs in markdown and generate them to the same result.

For example, with: `bundle exec rdoc --markup=markdown` we don't need to change request files' comments.

And we can also add options to `.rdoc_options`, like:

```
---
markup: markdown
exclude: [sorbet, test, rakelib]
```

vinistock:
Yes, but I think RDoc works better on GitHub, doesn't it? If we add an RDoc link to the README, it works both on GitHub and on the static website, which is not the case for markdown.

KaanOzkan (approved):

PR #681, opened 2023-05-02
Refactor listener pattern to pass emitter around

vinistock:
### Motivation

I have been reflecting on how our listener pattern is structured and I was not really satisfied with two things:
1. The way we're registering for events doesn't quite feel like a proper implementation of the listener and emitter pattern
2. Requests that require extra arguments, like URI or position were "poisoning" the base listener and forcing it to require extra arguments that aren't relevant for all requests

This PR is basically a refactor that changes our event registration to work like this
```ruby
class Emitter
  def initialize
    @listeners = Hash.new { |h, k| h[k] = [] }
  end

  def register(listener, *events)
    events.each { |e| @listeners[e] << listener }
  end
end

class MyListener
  def initialize(emitter)
    emitter.register(self, :on_class, :on_command, :on_module)
  end
end
```

I would love feedback on the proposal. My opinion is that this makes the code more organized and straight forward. It is also about 43% faster than the previous implementation.

### Implementation

1. All listeners are initialized with the emitter and any relevant extra arguments for the request
2. The emitter instance allows registering for events
3. We now just invoke the events directly based on which listeners registered for it (rather than always performing a conditional check)
4. Migrated the existing listeners to the new pattern


andyw8:
Slightly easier reading: https://github.com/Shopify/ruby-lsp/pull/681/files?w=1

lib/ruby_lsp/executor.rb
```diff
@@ -89,9 +89,10 @@ def run(request)
         return cached_response if cached_response
 
         # Run listeners for the document
-        document_symbol = Requests::DocumentSymbol.new(uri, @message_queue)
-        document_link = Requests::DocumentLink.new(uri, @message_queue)
-        EventEmitter.new(document_symbol, document_link).visit(document.tree) if document.parsed?
+        emitter = EventEmitter.new
+        document_symbol = Requests::DocumentSymbol.new(emitter, @message_queue)
```
andyw8, line 93: For consistency, might it help to always pass the `uri` if not needed for some listeners?

vinistock: That's part of what I'm trying to avoid. Some requests will require position, others range or URI. At the end of migrating, we'll need 5 or 6 positional arguments for every listener even if they don't require these arguments.

I think it's best to not have consistency in this case and just let each listener receive the relevant arguments.

lib/ruby_lsp/executor.rb
```diff
@@ -202,11 +204,12 @@ def hover(uri, position)
       end
 
       # Instantiate all listeners
-      base_listener = Requests::Hover.new(document.uri, @message_queue)
-      listeners = Requests::Hover.listeners.map { |l| l.new(document.uri, @message_queue) }
+      emitter = EventEmitter.new
+      base_listener = Requests::Hover.new(emitter, @message_queue)
+      listeners = Requests::Hover.listeners.map { |l| l.new(emitter, @message_queue) }
```
st0012, line 209: I wonder if it's possible to encapsulate these `<Listener>.new`, `<Listener>.listeners.map` boilerplate into something like: `<Listener>.run`?

In this example, I imagine the final result would be:

```rb
emitter = EventEmitter.new
# returns a response
Requests::Hover.run(emitter, @message_queue) do |emitter|
  emitter.emit_for_target(target)
end

# listener.rb
class Listener
  # maybe with uri?
  def self.run(emitter, queue)
    base_listener = self.new(emitter, queue)
    listeners = listeners.map { |l| l.new(emitter, queue) }
    yield(emitter)
    listeners.each { |ext| base_listener.merge_response!(ext) }
    base_listener.response
  end
end
```

lib/ruby_lsp/event_emitter.rb
```diff
@@ -20,96 +20,99 @@ module RubyLsp
   class EventEmitter < SyntaxTree::Visitor
```
st0012, line 20: Just a reminder: we need to update the document if we do make this change.

lib/ruby_lsp/executor.rb
st0012: Which request could use both listener classes at the same time. For individual requests, the pattern I proposed should still help?

vinistock: We can follow up with another refactor. One possibility would be to include our own listener as the first entry in the list and then merge all responses into the first. Something like this
```ruby
# document_symbol
class DocumentSymbol < Listener
  add_listener(self)
end

# executor
emitter = EventEmitter.new
document_symbol_listeners = DocumentSymbol.listeners.each { |l| l.new(emitter, @message_queue) }
document_link_listeners = DocumentLink.listeners.each { |l| l.new(emitter, @message_queue) }
emitter.visit(document.tree)
# then use some loop to merge responses into the first entry in the listeners array
```
What do you think?

vinistock: That's nice. I'll create an issue to make this refactor after this one.

lib/ruby_lsp/executor.rb
vinistock: I'm not sure I agree. There will always be consistency for all listeners for the same request type. For example, all document symbol listeners are instantiated the same.

However, trying to enforce that different requests have the same listener arguments is very limiting and I can't see the benefit. We'd need to account for every possible extra argument in our base listener, which is very messy. It would include
- position
- range
- trigger character
- uri

And possibly more in the future. It would also be confusing for extension authors. If they wanted to override `initialize` to carry some state, they would need to always include these arguments even if it's not related to the request they are extending. E.g.: why would you need to accept trigger character for semantic highlighting?

Also each request has a different response, which is why we always override `initialize`. For example, `hover` returns a single object as the response, but `document symbol` returns an array. If we initialized the response in the parent class as an array for all requests, it would be inconvenient for requests like `hover`. If we initialized it as `nil`, it's inconvenient for requests that return arrays.

I believe keeping consistency only for same request listeners will make the code more organized. Of course, we can always revisit if we conclude otherwise.

st0012 (approved):

Issue #679, opened 2023-05-02
Make CodeLens a Listener (followup)


andyw8:
Originally posted by @vinistock in https://github.com/Shopify/ruby-lsp/pull/676#discussion_r1182941702

> Can we follow up with a PR to execute this request together with document symbol and document link? We can go through code lens on the same round of visits and use the caching mechanism to return the response.

PR #685, opened 2023-05-02
Run CodeLens together with DocumentSymbol and DocumentLink

andyw8:
### Motivation

Closes https://github.com/Shopify/ruby-lsp/issues/679

### Automated Tests

I didn't add anything new, I think we can trust that the caching is already working.


vinistock:
This will run the code lens listener with document symbol and document link, but it's missing a few things:
- You need to remove the old `when` branch for code lens and related `code_lens` method as we are merging that with the other requests
- You need to add `textDocument/codeLens` to the branch that is handling document symbol and document link
- You need to cache the results of the code lens listener or else the response will never be returned

PR #686, opened 2023-05-02
Add `Run In Terminal` code lens

st0012:
### Motivation

Because the test controller isn't able to handle dynamically generated tests, the `Run` code lens is not able to run them. This means those tests are not supported by any code lens anymore, while in prior versions they could be run in the terminal.

This commits adds a new code lens `Run In Terminal` that will run the tests in the terminal.

(This requires https://github.com/Shopify/vscode-ruby-lsp/pull/596)

### Implementation

1. Update the `create_code_lens` method so it handles more than 2 types of code lens.
2. Add a new type of code lens titled: `Run In Terminal`

Result: 

Screenshot 2023-05-02 at 23 18 11

### Automated Tests

The existing tests have covered it.

### DIFF
--- a/lib/ruby_lsp/requests/code_lens.rb
+++ b/lib/ruby_lsp/requests/code_lens.rb
@@ -119,6 +119,8 @@ module RubyLsp
       def add_code_lens(node, name:, command:)
         @response << create_code_lens(
           node,
+          title: "Run",
+          command_name: "rubyLsp.runTest",
           path: @path,
           name: name,
           test_command: command,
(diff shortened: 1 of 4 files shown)


andyw8:
I personally would enjoy a Run In Terminal option, but I think it would be worth looking into how other language extensions deal with this kind of thing. If clicking "Run" doesn't show failed tests then that seems like a regression.

(Perhaps we should mark the Test Explorer as experimental for now?)

st0012:
> If clicking "Run" doesn't show failed tests then that seems like a regression.

It's not that it doesn't show failed tests. It's because we can't create test controller items for generated tests, they're not being run at all. For example, if you click `Run` on `DocumentHighlightExpectationsTest`, it only runs 1 test that's explicitly defined in the class. But yeah, I feel it's a regression.

> (Perhaps we should mark the Test Explorer as experimental for now?)

I have a very mixed feeling about this:

- Given the technical limitation on detecting generated tests, it'll probably stay experimental for a long time.
- Generated tests should be relatively few in a project. Hiding the very convenient Test Explorer for them seems a bit of a shame.
- However, not seeing generated tests in Test Explorer is pretty confusing. I had to ping @vinistock to confirm what's causing it.

vinistock (approved):
Given that we can't easily discover tests defined via meta-programming, I think this is a good compromise.

We can always gather feedback and revisit if necessary

andyw8 (approved):

andyw8:
btw Code Lens _is_ still experimental, so we might want to mention that in the release notes.

PR #689, opened 2023-05-03
Add warning to `Support::Common`

andyw8:
### Motivation

Extension may rely on code in `Support::Common`. If we change things it could break extensions.

### Implementation

We could have something as part of CI to validate this, but for now I have just added a comment warning.


lib/ruby_lsp/requests/support/common.rb
```diff
@@ -1,6 +1,9 @@
 # typed: strict
 # frozen_string_literal: true
 
+# WARNING: Methods in this class may be used by Ruby LSP extensions such as https://github.com/Shopify/ruby-lsp-rails,
+# or extensions by created by developers outside of Shopify, so be cautious of changing anything.
```
st0012, line 5: Wouldn't this become the document of the top-level `RubyLsp`? Should it be directly above the `Common` class instead?

andyw8: Actually, this probably shouldn't appear anywhere in published docs, just the source, but I'm not sure if that's possible...

dirceu (approved):

PR #694, opened 2023-05-03
Add --help to ruby-lsp command, and detect invalid arguments

technicalpickles:
### Motivation

I ran `ruby-lsp --help` and was surprised to see the program start. I've added support it, and documented the --debug while at it.

I'm also detecting invalid options, which will fail and exit non-zero.

### Implementation

I used optparse, which is included in the stdlib. This avoids any extra dependencies.

I've tried to use exit codes meaningfully: if you ask for help explicitly, it is success (0) exit code, but an invalid argument is a failure (exit 1)

Similarly, I print usage to stdout when it's explicitly requested (ie you can grep or page or whatever), but invalid argument uses warn to send to stderr.

### Automated Tests

I didn't add any new tests, as there weren't any for the CLI already. I can follow up with a PR to add a test, and refactor to make it more testable

### Manual Tests

```
$ bundle exec exe/ruby-lsp --help
$ bundle exec exe/ruby-lsp -h
$ bundle exec exe/ruby-lsp --trololol
```


technicalpickles:
Sure, happy to elaborate. When working with CLI programs, it's pretty common and accepted to try to figure out what options are available from the system it's installed. Most commonly, I think of `man <command>` and `<command> --version`.

More specifically, I was reading this issue: https://github.com/Shopify/ruby-lsp/issues/384#issuecomment-1346608237

> Because it can be turned off both by editor.formatOnSave and by changing the Ruby LSP's enabled features configuration.

`editor.formatOnSave` is a VS Code option iirc, so I wanted to find out how to enable/disable features. That is when I tried `ruby-lsp --help`.


> The reason there is no support for --help is because there are no options. There's only one command: launching the server.

I found `--debug` while reading the code, which is the only way to find out about it or docs that refer to that.

> I noticed it recognizes a --version flag but currently outputs version unknown, it would be nice to support that too.

That is weird, I haven't seen this in optparse before, but it seems to be something it adds automatically.

It would be a good one to add though, so consumers can check the version if they want/need to. Over on https://github.com/neovim/nvim-lspconfig/pull/2498 for example, there's a workaround needed for vim for `ruby-lsp > 0.2.2`.

vinistock:
Thank you for elaborating, that's helpful to understand your mental model when using the LSP.

> editor.formatOnSave is a VS Code option iirc, so I wanted to find out how to enable/disable features. That is when I tried ruby-lsp --help.

Editor configurations, like picking which features are enabled, are all passed as part of the `initializationOptions` during the `initialize` request (while booting the server). In my opinion, any configuration for the LSP should always be passed through that request and never through CLI arguments.

The `--debug` flag is an odd one out. We use to allow debugging a live server while using the development version of `vscode-ruby-lsp`. The only reason it's a flag is because VS Code passes it as one, otherwise I would also prefer to make it part of the initialize request.

I think the overarching theme here is discoverability of our documentation. People need an easy way to figure out how to configure the LSP correctly and that's a super valid point. However, I'm not 100% sure adding the `--help` flag is necessarily the best option.

Maybe a manual entry linking to our [docs](https://shopify.github.io/ruby-lsp/) would be easier? I'd love to keep all documentation in the same place if possible. @andyw8 do you have opinions on this one?

technicalpickles:
Thanks for the consideration! I believe I've applied all the suggestions/feedback, so ready for a second review.

andyw8 (approved):

exe/ruby-lsp
```diff
@@ -20,7 +20,39 @@ end
...
+options = begin
+  parser.parse!
+rescue OptionParser::InvalidOption => e
+  warn(e)
+  warn("")
+  warn(parser.help)
+  exit(1)
+end
+
+if options[:debug]
```
vinistock, line 55: Is this `if` statement missing an `end`?

vinistock: Oh, yeah, indeed. The other `if` statement was removed. Why is the integration test failing then 🤔.

vinistock:
@technicalpickles did you try booting the LSP on this branch? Does it work properly?

Trying to understand what's breaking the integration tests. They spawn an actual server instance in the background, so maybe something broke how the server is booting.

Issue #585, opened 2023-03-28
Add gem actions in the Gemfile


vinistock:
We can use the code lens request (#584) to attach actions to gems in the Gemfile as well. For example, opening the remote on GitHub or connecting to the [Bundler view](https://github.com/Shopify/vscode-ruby-lsp/pull/464) exploration.

To do this in the code lens request:
- Handle the `on_command` event
- Check if the path in the URI is `Gemfile`
- Check if the command message is `gem` with a string argument
- If both match, then search the gem specifications like we do [here](https://github.com/Shopify/ruby-lsp/blob/cd7f548cc1232c9cff70927e84adf0c6baa3d529/lib/ruby_lsp/requests/document_link.rb#L40) to find the gem's homepage and add a new code lens to open the GitHub repository in the browser

In order for this to work, we will need to define a new command in `vscode-ruby-lsp` to handle these code lenses. Something like `rubyLsp.openGem` passing the remote as an argument. The implementation of the command will basically just open the external URL.

PR #710, opened 2023-05-09
Add gem actions in the Gemfile

snutij:
### Motivation

Closes: #585
Requires: https://github.com/Shopify/vscode-ruby-lsp/pull/605

### Implementation

Following indications in the issue: 

- on the `on_command` event check if the file is `Gemfile`, the command message includes `gem` 
- extract the gem name from the node 
- find the specification for this name
- extract its homepage property
- return a Codelens with this form:

```json
{
      "range": {
        "start": { "line": 1, "character": 0 },
        "end": { "line": 1, "character": 21 }
      },
      "command": {
        "title": "Open",
        "command": "rubyLsp.openGem",
        "arguments": ["https://github.com/ruby/rake"]
      },
      "data": { "type": "browser" }
    }
```

### Automated Tests

I added a fixture and expectation file for this specific Codelens. To make it work I had to:

- not faking URI in the document variable (I check it through the `@path` in the Codelens request)
- fix expectation (as it's not faked anymore) for the `minitest_test.exp.json`

To be honest, it's not a "hermetic" test. For example, I explicitly added `gem "rubocop-minitest"` in the fixture, because there is no homepage for this gem, so I test it by not providing it in `Gemfile.exp.json` but, if it's added, CI will fail. Not sure how to handle this differently. 

I'm also not sure that adding a Gemfile (even in `test/fixtures`) is without consequences.

### Manual Tests

- Check out this branch
- Checkout vscode-ruby-lsp branch
- Open a Gemfile
- Something like that is expected:

<details><summary>Record</summary>

Screen-Recording-2023-05-09-at-22 37 45
</details>

### DIFF
--- a/lib/ruby_lsp/requests/code_lens.rb
+++ b/lib/ruby_lsp/requests/code_lens.rb
@@ -73,12 +73,10 @@ module RubyLsp
           @prev_visibility = @visibility
           @visibility = node.message.value
         elsif @path.include?("Gemfile") && node.message.value.include?("gem") && node.arguments.parts.any?
-          gem_name = node.arguments.parts.map(&:child_nodes).flatten.first.value
-          spec = Gem::Specification.stubs.find { |gem| gem.name == gem_name }&.to_spec
+          homepage = resolve_gem_homepage(node)
+          return unless homepage
 
-          return if spec.nil? || spec.homepage.nil?
-
-          add_open_gem_code_lens(node, homepage: spec.homepage)
+          add_open_gem_homepage_code_lens(node, homepage: homepage)
         end
       end
 
@@ -156,15 +154,24 @@ module RubyLsp
         )
       end
 
+      sig { params(node: SyntaxTree::Command).returns(T.nilable(String)) }
+      def resolve_gem_homepage(node)
+        gem_name = node.arguments.parts.flat_map(&:child_nodes).first.value
+        spec = Gem::Specification.stubs.find { |gem| gem.name == gem_name }&.to_spec
+        return if spec.nil?
+
+        spec.homepage || spec.metadata.fetch("homepage_uri", nil)
+      end
+
       sig { params(node: SyntaxTree::Command, homepage: String).void }
-      def add_open_gem_code_lens(node, homepage:)
+      def add_open_gem_homepage_code_lens(node, homepage:)
         range = range_from_syntax_tree_node(node)
 
         @response << Interface::CodeLens.new(
           range: range,
           command: Interface::Command.new(
-            title: "Open",
-            command: "rubyLsp.openGem",
+            title: "Open Homepage",
+            command: "rubyLsp.openGemHomepage",
             arguments: [homepage],
           ),
           data: { type: "browser" },
--- a/lib/ruby_lsp/requests/code_lens.rb
+++ b/lib/ruby_lsp/requests/code_lens.rb
@@ -72,6 +72,13 @@ module RubyLsp
         if ACCESS_MODIFIERS.include?(node.message.value) && node.arguments.parts.any?
           @prev_visibility = @visibility
           @visibility = node.message.value
+        elsif @path.include?("Gemfile") && node.message.value.include?("gem") && node.arguments.parts.any?
+          gem_name = node.arguments.parts.map(&:child_nodes).flatten.first.value
+          spec = Gem::Specification.stubs.find { |gem| gem.name == gem_name }&.to_spec
+
+          return if spec.nil? || spec.homepage.nil?
+
+          add_open_gem_code_lens(node, homepage: spec.homepage)
         end
       end
 
@@ -146,6 +153,20 @@ module RubyLsp
           name: name,
           test_command: command,
           type: "debug",
+        )
+      end
+
+      def add_open_gem_code_lens(node, homepage:)
+        range = range_from_syntax_tree_node(node)
+
+        @response << Interface::CodeLens.new(
+          range: range,
+          command: Interface::Command.new(
+            title: "Open",
+            command: "rubyLsp.openGem",
+            arguments: [homepage],
+          ),
+          data: { type: "browser" },
         )
       end
     end
--- a/lib/ruby_lsp/requests/code_lens.rb
+++ b/lib/ruby_lsp/requests/code_lens.rb
@@ -156,6 +156,7 @@ module RubyLsp
         )
       end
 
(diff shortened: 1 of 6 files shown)


andyw8:
:wave: Thank youfor this! Some initial thoughts:

* For rubocop-minitest, there is actually a homepage in the [gemspec](https://github.com/rubocop/rubocop-minitest/blob/master/rubocop-minitest.gemspec), but it's specified differently, as part of the metadata. I'm not sure how common that it, but it maybe worth investigating if there are other gems that do this.
* Instead of 'Open' I think we should say 'Open Homepage', since we may later want add other options such as 'Open Repo' or 'Open on rubygems.org'.

snutij:
Hello, @andyw8 thank you for your feedback. 🙏 

Both ways are recommended in the gem specification reference: 

- https://guides.rubygems.org/specification-reference/#homepage
- https://guides.rubygems.org/specification-reference/#metadata

So to me, it is worth checking both. I added it.

And the title of the command has been fixed too.

lib/ruby_lsp/requests/code_lens.rb
```diff
@@ -70,6 +70,11 @@ def on_command(node)
         if ACCESS_MODIFIERS.include?(node.message.value) && node.arguments.parts.any?
           @prev_visibility = @visibility
           @visibility = node.message.value
+        elsif @path.include?("Gemfile") && node.message.value.include?("gem") && node.arguments.parts.any?
```
vinistock, line 73: Let's extract `node.message.value` into a local variable since it's used twice.

lib/ruby_lsp/requests/code_lens.rb
```diff
@@ -146,6 +151,30 @@ def add_code_lens(node, name:, command:)
...
+      end
+
+      sig { params(node: SyntaxTree::Command, homepage: String).void }
+      def add_open_gem_homepage_code_lens(node, homepage:)
+        range = range_from_syntax_tree_node(node)
+
+        @response << Interface::CodeLens.new(
+          range: range,
+          command: Interface::Command.new(
+            title: "Open Homepage",
```
vinistock, line 171:
```suggestion
            title: "Open remote",
```

lib/ruby_lsp/requests/code_lens.rb
```diff
@@ -146,6 +151,30 @@ def add_code_lens(node, name:, command:)
...
+            command: "rubyLsp.openGemHomepage",
+            arguments: [homepage],
+          ),
+          data: { type: "browser" },
```
vinistock, line 175: We may want to link to other things in the editor eventually.
```suggestion
          data: { type: "link" },
```

lib/ruby_lsp/requests/code_lens.rb
```diff
@@ -146,6 +151,30 @@ def add_code_lens(node, name:, command:)
...
         )
       end
+
+      sig { params(node: SyntaxTree::Command).returns(T.nilable(String)) }
+      def resolve_gem_homepage(node)
+        gem_name = node.arguments.parts.flat_map(&:child_nodes).first.value
+        spec = Gem::Specification.stubs.find { |gem| gem.name == gem_name }&.to_spec
+        return if spec.nil?
+
+        spec.homepage || spec.metadata.fetch("homepage_uri", nil)
```
vinistock, line 161: The idea is to open the source control for the gem, not necessarily the homepage. So I think what we want is
```ruby
spec.homepage || spec.metadata.fetch("source_code_uri", nil)
```
I think it's also worth checking if the URL is from GitHub or GitLab and reject any others. The majority of gems are probably hosted on these two and we shouldn't allow people to click on random URLs. So something like
```ruby
[spec.homepage, spec.metadata["source_code_uri"]].find { |page| page&.start_with?("https://github.com") || page&.start_with?("https://gitlab.com") }
```

lib/ruby_lsp/requests/code_lens.rb
```diff
@@ -146,6 +151,30 @@ def add_code_lens(node, name:, command:)
...
+      end
+
+      sig { params(node: SyntaxTree::Command, homepage: String).void }
+      def add_open_gem_homepage_code_lens(node, homepage:)
+        range = range_from_syntax_tree_node(node)
+
+        @response << Interface::CodeLens.new(
```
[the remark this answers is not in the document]
snutij: I can give it a try after this merge 👍

snutij:
Thanks for the feedback @vinistock 🙏 

Just a reminder, this PR requires https://github.com/Shopify/vscode-ruby-lsp/pull/605.

If you make a new release of this gem, without releasing before the VS Code extension, it will break the test code lenses feature. Is it OK for you?

test/fixtures/Gemfile.rb
```diff
@@ -0,0 +1,7 @@
+gem "rake", "~> 13.0"
```
andyw8, line 1: Could this be named `Gemfile`?

st0012 (approved):
Tested it locally and it worked great 👍 

I think for gems that don't have `source_code_uri` (like `rdoc`), we can fallback to `homepage` instead. But then we should also update the code lens' name to reflect that difference. So I think it could be a follow up improvement.
Another concern is that in large apps with hundreds of line of `Gemfile`, this feature will basically double the space to display the file, which some users may find annoying?

But anyway, none of those are blockers and I think we can merge this first 😄

vinistock:
@st0012 for `rdoc`, the problem is that the homepage is GitHub pages and there's actually no link to the repository anywhere in the gemspec.

As I had previously commented, I think we should limit showing these links only to GitHub and GitLab links for security reasons, given that we can't trust random URLs in the gemspec.

test/fixtures/Gemfile.rb
snutij: Tried before, it's not parsed by the `expectation_test_runner`: 

```ruby
TEST_FIXTURES_DIR = "test/fixtures"
TEST_FIXTURES_GLOB = File.join(TEST_FIXTURES_DIR, "**", "*.rb") # .rb expected
```

If you have any input to success it, I'm open to it 🙏

st0012:
> As I had previously commented, I think we should limit showing these links only to GitHub and GitLab links for security reasons, given that we can't trust random URLs in the gemspec.

That's fair. Sorry that I missed previous comments.

vinistock:
Merging this already since code lenses are still marked as experimental (so not having the extension command won't be a major issue for now).

Thank you so much for the contribution!

PR #712, opened 2023-05-09
Include test library type in Code Lens

andyw8:
### Motivation

https://github.com/Shopify/vscode-ruby-lsp/pull/604/files#r1188930063

### Implementation

Pass as additional `data`.

### Automated Tests

Fixtures updated.

### Manual Tests

n/a

### DIFF
--- a/lib/ruby_lsp/requests/support/common.rb
+++ b/lib/ruby_lsp/requests/support/common.rb
@@ -82,7 +82,7 @@ module RubyLsp
               command: command_name,
               arguments: arguments,
             ),
-            data: { type: type },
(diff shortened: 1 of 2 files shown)


vinistock:
I went through this PR and https://github.com/Shopify/vscode-ruby-lsp/pull/604.

The reason we were using the `test_` prefix in the test controller is to differentiate between test examples and the entire test suite for a file (i.e.: the test class vs the test methods). 

Instead of associating the code lens with the test framework, we should instead indicate whether the code lens is for a single example or for a group, which will be much more generic and work regardless of which framework is used.

PR #721, opened 2023-05-17
Escape test name in the commandline to run it in the terminal

soutaro:
### Motivation

The test name may end with `?` or `!`, and passing it without escape results in a failure like `zsh: no matches found: test_has?`.

### Implementation

Escaping the file name with `Shellwords.escape`.

Not sure if the same problem exists for rspec or other supported frameworks...

### DIFF
--- a/lib/ruby_lsp/requests/code_lens.rb
+++ b/lib/ruby_lsp/requests/code_lens.rb
@@ -1,6 +1,8 @@
 # typed: strict
 # frozen_string_literal: true
 
+require "shellwords"
+
 module RubyLsp
   module Requests
     # ![Code lens demo](../../code_lens.gif)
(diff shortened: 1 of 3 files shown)


lib/ruby_lsp/requests/code_lens.rb
```diff
@@ -59,7 +59,7 @@ def on_def(node)
             add_code_lens(
               node,
               name: method_name,
-              command: BASE_COMMAND + @path + " --name " + method_name,
+              command: BASE_COMMAND + @path + " --name " + Shellwords.escape(method_name),
```
andyw8, line 62: Should we `require "shellwords"` in this class to ensure it's loaded?

soutaro:
I have signed the CLA!

PR #727, opened 2023-05-26
Migrate semantic highlighting to the listener pattern

### LINKED ISSUE #678: Migrate semantic highlighting (full only)

vinistock:
### Motivation

Closes #678

This PR migrates semantic highlighting to the new listener pattern. For full semantic highlighting, we execute it as a part of all other nonpositional requests. Range semantic highlighting is still executed separately.

### Implementation

1. Added the necessary events to `EventEmitter`. Also included `WithScope` since it is required for semantic highlighting
2. Migrated `SemanticHighlighting` to be a listener and made the related changes to `Executor`

### Question

Now that the emitter controls the visits, one problem that surfaced for semantic highlighting is that we can't avoid visiting the `Const` node inside a `ClassDeclaration` or `ModuleDeclaration`.

Without any extra handling, this is what would happen
1. We find the class or module declaration, then push a `class.declaration` token
2. After, we find the const node inside of the declaration and then we push a `namespace` token for it, overriding the previously defined token

I couldn't come up with an elegant way of preventing this. The way I ended up circumventing it was by checking if any of the previous two tokens already found match the same location before trying to push const tokens.

Any better ideas than this?

### Automated Tests

Updated tests to match the new listener pattern. Also, the workaround of checking the two previous tokens actually reduced duplication in our responses and I had to update some expectations.


lib/ruby_lsp/requests/semantic_highlighting.rb
```diff
@@ -102,109 +104,101 @@ def initialize(location:, length:, type:, modifier:)
...
       end
 
-      sig { override.params(node: SyntaxTree::Const).void }
-      def visit_const(node)
-        return super unless visible?(node, @range)
+      sig { params(node: SyntaxTree::Const).void }
+      def on_const(node)
+        return unless visible?(node, @range)
+        # When finding a module or class definition, we will have already pushed a token related to this constant. We
+        # need to look at the previous two tokens and if they match this locatione exactly, avoid pushing another token
```
KaanOzkan, line 175:
```suggestion
        # need to look at the previous two tokens and if they match this location exactly, avoid pushing another token
```

KaanOzkan (commented):
I'm confused by the question because I don't have enough context on the traversal. Is `class.declaration` token being pushed by `SemanticHighlighting` or another listener? I'd need some detail before chiming in but rest of the code looks good to me.

vinistock:
It's pushed by `SemanticHighlighting` itself. If you have the following file
```ruby
class Foo
end
```
The AST looks like this
```
(program (statements ((class (const_ref (const "Foo")) (bodystmt (statements ((void_stmt))))))))
```
Notice how there's a node for `class` and inside of it there's a node for the constant `Foo`. When we're visiting the AST, we first find the `class` node and push a `class.declaration` token. Then we find the `const` node and push a `namespace` token for the same location, overriding the previous one.

I'm going to see if we can somehow stop visiting `const` to make this simpler.

lib/ruby_lsp/requests/semantic_highlighting.rb
```diff
@@ -102,109 +104,101 @@ def initialize(location:, length:, type:, modifier:)
...
 
-      sig do
-        override.returns(
-          T.any(
-            Interface::SemanticTokens,
-            T.all(T::Array[SemanticToken], Object),
-          ),
+        emitter.register(
+          self,
+          :on_call,
```
andyw8, line 126: Maybe alphabetize these?

andyw8 (approved):

PR #730, opened 2023-05-29
Exclude problematic Windows CI jobs

andyw8:


andyw8:
(I'll update the required checks for `main` once this is merged)

vinistock:
Is it possible to achieve the right behaviour with Gemfile platforms constraints rather than excluding these Ruby versions?

vinistock (approved):
Let's merge this for now to unblock CI and we can improve later

PR #736, opened 2023-06-02
Fix `Run In Terminal` code lens for test class

st0012:
### Motivation

Currently, the `Run in Terminal` code lens always run all tests in the file
when it's clicked on a class definition. This means if a file contains
multiple test classes, all tests will be run when click on any of the class.

I will add a similar support for `test-unit` in a follow up PR because it expects using `--testcase` for the class name instead.

### Implementation

To resolve this, we need to specify the class name in the test command.
But when a class is provided, minitest would try to match the full class
name, including namespaces. This is not what we want, so we need to
quote `--name`'s value to make minitest do partial matching.

### Automated Tests

I've updated existing test cases.

### Manual Tests

1. Visit `test/requests/code_lens_expectations_test.rb`
2. Click `Run In Terminal` on the test class to make sure it runs.
3. Click `Run In Terminal` on a test method to make sure it runs.
4. Add `module Foo` outside of `CodeLensExpectationsTest` and repeat 2 and 3


vinistock (approved):
What happens if there's a module involved too? Do we need to pass `Foo::Bar` in that case?

st0012:
I just found that I forgot to append `class_name` to the command by default 🤦‍♂️  Another PR coming.

Issue #432, opened 2023-01-06
Adding a Standard runner alongside Rubocop?


searls:
I'm new to this project but newly interested in LSPs and editor extensions since seeing how fast @will's [LSP implementation](https://github.com/testdouble/standard#language-server-protocol-support) for Standard Ruby is.

Are you open to a Standard Ruby runner alongside this [RuboCop one]?(https://github.com/Shopify/ruby-lsp/commits/07b8cdbc5ba0fa32c9a15dd65aed3d646bc854d8/lib/ruby_lsp/requests/support/rubocop_runner.rb)? Beyond that, I imagine there'd be a little more work to connect it to editor plugins like [vscode-ruby-lsp](https://github.com/Shopify/vscode-ruby-lsp)?

cc/@vinistock @camilopayan

vinistock:
We definitely want to support other formatters and have tool or framework specific functionality coming from the Ruby LSP. There's also a proposal to support rubyfmt #409.

We should first figure out how to introduce Ruby LSP plugins in #443 to be able to scale tool or framework specific features being provided from the server. Ideally, each tool that wants to provide editor features could export their own plugin and have the Ruby LSP surface that functionality - removing the need for the Ruby LSP to have tool specific implementations.

searls:
@j-manu - I appreciate this, but running Standard's configuration via a RuboCop config and the RuboCop CLI or Rake task directly isn't quite the same thing, since any other configuration or ruleset could override the intended-to-not-be-overridable rules in Standard. I [touched on this](https://blog.testdouble.com/posts/2023-01-19-super-standard-adding-gem-extensions-and-custom-rules/#how-it-works) briefly in a blog post last week when announcing Standard's new support for _importing RuboCop configurations_ in such a way as to prevent them the risk that any of them override Standard's (I realize this is a bit of an ouroboros at this point)

j-manu:
@searls I understand. The above config is meant as a stopgap solution. I didn't want to give up on ruby-lsp or standard and this helped me use both.

iamajvillalobos:
With the vs code extension, can we use them both now at the same time?

searls:
Thank you for clarifying!

drnic:
I was setting up a new vscode, and so I have a tiny [sample] `settings.json` that I can share that I think summarises @searls + @vinistock above (plus my own preference for rvm)

```json
{
    "workbench.colorTheme": "Spinel",
    "editor.inlineSuggest.enabled": true,
[...]

Issue #409, opened 2022-12-21
Question: What do folks think of adding rubyfmt as a formatter option?


jscharf:
I have been experimenting with the rubyfmt project recently https://github.com/fables-tales/rubyfmt and wanted to ask here what folks think of adding it as a formatter for this project?

I'm not sure if it even fits with this project's goals or architecture, but I want to ask anyway just in case. It's my understand that it already supports a built-in formatter, and Rubocop if the user has Rubocop configured. So the idea would be to add rubyfmt as another formatting option.

vinistock:
Hey! Please take a look at this comment https://github.com/Shopify/ruby-lsp/issues/432#issuecomment-1376019129. We're definitely open to having different formatting options, but we'd like to be able to do that through plugins and not have tool-specific features be all implemented inside of the Ruby LSP codebase.

vinistock:
Closing as https://github.com/Shopify/ruby-lsp/pull/739 allows for other formatters to be registered now.

jscharf:
Posting here for future readers, since I opened this over a year ago... I have created the https://github.com/jscharf/ruby-lsp-rubyfmt project to bridge the technologies together: `rubyfmt`, `ruby-lsp`, and `vscode-ruby-lsp`. It's a VS code addon that can be used to hook into the formatting feature of the ruby LSP.

PR #739, opened 2023-06-02
Allow extensions to register formatters

### LINKED ISSUE #649: Define mechanism to register formatters
Gems will need the ability to register themselves as formatters to be used in the Ruby LSP. Our formatting request will then select [which formatter to use](https://github.com/Shopify/ruby-lsp/blob/694ca5c4cd10190575da576c0d18961dc15cc79b/lib/ruby_lsp/requests/formatting.rb#L72) based on the user's configuration.

vinistock:
### Motivation

Closes #649 #409 #432

This PR is a proposal of how we can allow other gems to register formatters in the Ruby LSP.

### Implementation

I tried to make a generic formatter registration mechanism that would work for our bundled ones too.

1. Added a common interface that all formatter runners need to comply to. Basically, we want the formatters to receive the URI and the document and then return the formatted source
2. Added a way to register formatters in `Formatting` - which we are now using to register RuboCop and Syntax Tree
3. Started setting the formatter to `none` if we identify that RuboCop is not available. This is just so that we can avoid special handling for RuboCop

This is how the formatter gem would use this:
```ruby
class MyFormatterRubyLspExtension < RubyLsp::Extension
  def name
    "My Formatter"
  end

  def activate
    # The first argument here is what people will need to pass as the `rubyLsp.formatter`
    # configuration
    RubyLsp::Requests::Formatting.register_formatter("my_formatter", MyFormatterRunner.instance)
  end
end

class MyFormatterRunner
  include RubyLsp::Requests::Support::FormatterRunner

  def run(uri, document)
    source = document.source
    formatted_source = format_the_source_using_my_formatter(source)
    formatted_source
  end
end
```

### Automated Tests

Added a test with a dummy formatter.

### DIFF
--- a/lib/ruby_lsp/requests/formatting.rb
+++ b/lib/ruby_lsp/requests/formatting.rb
@@ -29,6 +29,29 @@ module RubyLsp
       class Error < StandardError; end
       class InvalidFormatter < StandardError; end
 
+      @formatters = T.let(
+        {
+          "syntax_tree" => Support::SyntaxTreeFormattingRunner.instance,
+        },
+        T::Hash[String, Support::FormatterRunner],
+      )
+
+      class << self
+        extend T::Sig
+
+        sig { returns(T::Hash[String, Support::FormatterRunner]) }
+        attr_reader :formatters
+
+        sig { params(identifier: String, instance: Support::FormatterRunner).void }
+        def register_formatter(identifier, instance)
+          @formatters[identifier] = instance
+        end
+      end
+
+      if defined?(Support::RuboCopFormattingRunner)
+        register_formatter("rubocop", Support::RuboCopFormattingRunner.instance)
+      end
+
       extend T::Sig
 
       sig { params(document: Document, formatter: String).void }
@@ -41,6 +64,8 @@ module RubyLsp
 
       sig { override.returns(T.nilable(T.all(T::Array[Interface::TextEdit], Object))) }
       def run
+        return if @formatter == "none"
+
         # Don't try to format files outside the current working directory
         return unless @uri.sub("file://", "").start_with?(Dir.pwd)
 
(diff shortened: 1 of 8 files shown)


lib/ruby_lsp/requests/formatting.rb
```diff
@@ -67,16 +92,10 @@ def run
...
-        when "rubocop"
-          if defined?(Support::RuboCopFormattingRunner)
-            Support::RuboCopFormattingRunner.instance.run(@uri, @document)
-          end
-        when "syntax_tree"
-          Support::SyntaxTreeFormattingRunner.instance.run(@uri, @document)
-        else
-          raise InvalidFormatter, "Unknown formatter: #{@formatter}"
-        end
+        formatter = Formatting.formatters[@formatter]
```
andyw8, line 95: (nit) I have a slight aversion to having a local variable with the same name as an instance variable... maybe this could be `formatter_instance`?

andyw8 (approved):

lib/ruby_lsp/requests/formatting.rb
```diff
@@ -67,16 +92,10 @@ def run
...
+        raise InvalidFormatter, "Formatter is not available: #{@formatter}" unless formatter
```
bitwise-aiden, line 96: Will this error with `Formatter is not available: none` if `@formatter = "none"`?

bitwise-aiden (approved):
I really like this pattern. Nicely done.

Could you fill me in on how a gem would go about doing this. I'm not quite sure about that connection wrt Ruby

lib/ruby_lsp/requests/formatting.rb
vinistock: It won't because we return early in the `run` method.

vinistock:
Adding the Ruby LSP extension into `standard` itself means that users will just get the extension out of the box without having to install a separate gem. All they'd need to do is select which formatter to use. IMO, it's a nice experience.

In terms of loading, you don't have to load anything in `standard`. As long as the extension activation class is in the right path structure, the Ruby LSP loads it automatically. The [ruby-lsp-rails extension](https://github.com/Shopify/ruby-lsp-rails/blob/main/lib/ruby_lsp/ruby_lsp_rails/extension.rb) has a good example of how to do it.

The Ruby LSP discovers those based on this glob `"ruby_lsp/**/extension.rb"` and then requires it. You basically need that `extension.rb` file and a [formatting runner implementation](https://github.com/Shopify/ruby-lsp/blob/main/lib/ruby_lsp/requests/support/rubocop_formatting_runner.rb).

Issue #738, opened 2023-06-02
Windows CI is mis-reporting successful runs


andyw8:
e.g. see https://github.com/Shopify/ruby-lsp/actions/runs/5155957143/jobs/9286294384

From looking at the output, it seems the Windows tests are not running at all, but they are being reported as successful.

andyw8:
https://github.com/Shopify/ruby-lsp/pull/747 shows the changes needed to the CI configuration, but there are some tests failing due to platform differences.

PR #747, opened 2023-06-06
Ensure CI runs on Windows

andyw8:
Closes https://github.com/Shopify/ruby-lsp/issues/738

Skips the tests that are not yet passing on Windows. The work to fix them is being tracked in https://github.com/Shopify/ruby-lsp/issues/751.

It makes sense that the exe path in the binstub would not be valid on Windows.

I don't yet know why it why it wasn't being marked as a failure though.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -35,7 +35,7 @@ jobs:
         run: bundle exec srb tc
 
       - name: Lint Ruby files
-        run: bin/rubocop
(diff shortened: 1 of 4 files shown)


andyw8:
(We're now seeing some test failures that were previously being missed)

rafaelfranca:
Can we merge this PR, skip the failing test on windows and then fix the tests slowly? That way new tests will not fail on windows anymore

PR #748, opened 2023-06-07
Migrate DocumentHighlight to Listener pattern

st0012:
Since DocumentHighlight focuses on the type of the visited node instead of the visiting event, I think it makes sense to introduce a new `on_visit` event for it. It saves us from expanding the nodes list in HighlightTarget into a list of fragmented `on_<node>` events.

It should also benefit other requests that currently define the `visit` method.


st0012:
I've rerun benchmark around 8 times and `documentHighlight` is consistently ~25% slower, so I think it's safe to say this approach causes performance regression. But I'm not sure why this is the case 🤔

lib/ruby_lsp/event_emitter.rb
```diff
@@ -53,6 +53,12 @@ def emit_for_target(node)
 
     # Visit dispatchers are below. Notice that for nodes that create a new scope (e.g.: classes, modules, method defs)
     # we need both an `on_*` and `after_*` event. This is because some requests must know when we exit the scope
+    sig { override.params(node: T.nilable(SyntaxTree::Node)).void }
+    def visit(node)
+      @listeners[:on_visit]&.each { |l| T.unsafe(l).on_visit(node) }
```
vinistock, line 58: Let's call this `on_node` since listeners only deal with node types and have no knowledge of the underlying visitor implementation.

lib/ruby_lsp/executor.rb
```diff
@@ -275,7 +275,12 @@ def on_type_formatting(uri, position, character)
       ).returns(T::Array[Interface::DocumentHighlight])
     end
     def document_highlight(uri, position)
-      Requests::DocumentHighlight.new(@store.get(uri), position).run
+      document = @store.get(uri)
```
vinistock, line 278: We can return early if the document has syntax errors

andyw8: Since `executor.rb` has `emitter.visit(document.tree) if document.parsed?`, do we need to be concerned with that here?

vinistock: Where? It seems we're visiting the tree a few lines down and we're not checking if the document is parsed.

andyw8: Ah, ignore, I was looking at the wrong section.

lib/ruby_lsp/event_emitter.rb
st0012: Good call, updated 👍

lib/ruby_lsp/executor.rb
st0012: Updated 👍

st0012:
Wrt the slowdown, I notice that it also happens to other migration PRs:

- https://github.com/Shopify/ruby-lsp/pull/727 sees all related requests slow down:
    - `textDocument/semanticTokens/full slower by 71.448 %`
    - `textDocument/documentSymbol slower by 78.778 %`
    - `textDocument/documentLink slower by 108.592 %`
    - `textDocument/codeLens slower by 112.232 %`
- https://github.com/Shopify/ruby-lsp/pull/703 sees `textDocument/inlayHint slower by 20.868 %`
- https://github.com/Shopify/ruby-lsp/pull/660 sees `textDocument/documentLink slower by 54.743 %`


So it's likely not specific to this change. We should look into what's causing the slowdown though OR if our benchmark is too sensitive.

lib/ruby_lsp/requests/document_highlight.rb
```diff
@@ -22,34 +22,49 @@ module Requests
...
-        visit(@document.tree) if @document.parsed? && @target
-        @highlights
+        highlight_target =
+          case target
+          when *DIRECT_HIGHLIGHTS
+            Support::HighlightTarget.new(target)
+          when SyntaxTree::Ident
+            relevant_node = parent.is_a?(SyntaxTree::Params) ? target : parent
+            Support::HighlightTarget.new(relevant_node)
+          end
```
vinistock, line 55: Not a blocker, but just a thought: what if we move this to `Executor`? Would that maybe simplify things a bit?

vinistock (approved):

lib/ruby_lsp/requests/document_highlight.rb
st0012: Consider the data type it uses (`HighlightTarget`) is still specific to this request, I prefer keeping it at the same place until we share it somewhere else.

Issue #684, opened 2023-05-02
Enhance our documentation rake task to check if the referenced GIF link exists


vinistock:
We're currently only checking if a request's documentation includes a link to a demo GIF in [check_docs](https://github.com/Shopify/ruby-lsp/blob/main/lib/ruby_lsp/check_docs.rb). However, if you add a link to a GIF file that doesn't exist, the task still passes.

We should check that the GIF file exists inside of the project. Maybe we can add a `gif_folder` option to our rake task so that we know where to look for the GIFs.

Issue #683, opened 2023-05-02
Create VS Code profile for recording our GIFs


vinistock:
It would be good to have a VS Code profile that we can use as the standard for recording all of our documentation GIFs.

I'm not super familiar with the profiles feature yet, but I assume you can export it and put it in the `.vscode` folder to make it accessible to others.

The profile should use
- A large font size so that it's easy to see in the GIFs (16?)
- The Spinel theme
- The default font bundled in VS Code
- Only the extensions that are part of [vscode-shopify-ruby](https://github.com/Shopify/vscode-shopify-ruby)

D-Forz:
Hi, I would like to help with this!

andyw8:
@D-Forz Thank you, feel free to open a PR.

PR #749, opened 2023-06-07
Documentation: make sure referenced GIF link exists

snutij:
### Motivation

Closes: #684 

Make sure GIFs documentation really exists in check_docs task.

### Implementation

- Pass to the rake task the GIF paths list
- From path request, extract related GIF ex: `lib/ruby_lsp/requests/hover.rb` search for `hover.gif` in GIF paths list

Notes: 
- This test requires matching name between request and GIF (`inlay_hints` for example was without `s`), but I think it's acceptable. As always, you can challenge this choice!
- `PathCompletion` was missing, I added it to avoid red CI.    

### Automated Tests

This task is not tested (or I missed it, feel free to correct me).

### Manual Tests

- checkout this branch
- remove one GIF
- run `bundle exec rake ruby_lsp:check_docs`
- rake task should fail


lib/ruby_lsp/check_docs.rb
```diff
@@ -34,6 +35,13 @@ def define_task
       task(@name) { run_task }
     end
 
+    sig { params(request_path: String).returns(T::Boolean) }
+    def gif_exists?(request_path)
+      request_gif = request_path.gsub(".rb", ".gif").split("/").pop
```
andyw8, line 40:
```suggestion
      request_gif = request_path.gsub(".rb", ".gif").split("/").last
```
(since there's no need to mutate `request_path`).

andyw8:
The new GIF looks slightly different from the others in style, but I think it's fine for now and we can re-record them after https://github.com/Shopify/ruby-lsp/issues/683.

andyw8 (approved):

snutij:
Yep I know there is a poor resolution for my recording, don't know why. Sorry for that 🙏

lib/ruby_lsp/check_docs.rb
```diff
@@ -34,6 +35,13 @@ def define_task
       task(@name) { run_task }
     end
 
+    sig { params(request_path: String).returns(T::Boolean) }
+    def gif_exists?(request_path)
+      request_gif = request_path.gsub(".rb", ".gif").split("/").last
+
+      @gif_list.any? { |gif_path| gif_path.include?(request_gif) }
```
vinistock, line 42:
```suggestion
      @gif_list.any? { |gif_path| gif_path.end_with?(request_gif) }
```

lib/ruby_lsp/check_docs.rb
```diff
@@ -86,7 +94,7 @@ def run_task
             # end
             # ```
           DOCS
-        elsif !/\[.* demo\]\(.*\.gif\)/.match?(documentation)
+        elsif !/\[.* demo\]\(.*\.gif\)/.match?(documentation) || !gif_exists?(file_path)
```
vinistock, line 97: I think the error message on this branch is more tailored to the comment missing the `[]()` gif link. Can we add a different branch for this case saying something like "The GIF referenced in the documentation does not exist"?

lib/ruby_lsp/check_docs.rb
```diff
@@ -94,13 +94,21 @@ def run_task
...
+        elsif !/\[.* demo\]\(.*\.gif\)/.match?(documentation)
           T.must(missing_docs[class_name]) << <<~DOCS
             Missing demonstration GIF. Each request and extension must be documented with a GIF that shows the feature
             working. For example:
 
             # [Inlay hint demo](../../inlay_hint.gif)
           DOCS
+        elsif !gif_exists?(file_path)
+          T.must(missing_docs[class_name]) << <<~DOCS
+            The GIF for the request documentation does not exist. Make sure to add it in the `misc/` folder,
```
vinistock, line 106: Can we use the full path here instead of the `misc` folder? We're exporting this rake task to allow Ruby LSP extensions to use it and we want to allow them to be flexible about where they store their GIFs.

snutij: Hello, I'm not sure to understand what you mean by `full path`? As we provide flexibility on where extension store their GIFs, maybe the best way is to not specify path at all? I did a try here: https://github.com/Shopify/ruby-lsp/pull/749/commits/31694e549fafa73dbd4bd9dbbd486aae2560d166

Tell me if I'm wrong, I can fix it. 👍

vinistock (approved):
Thank you for the contribution 🚀

Issue #754, opened 2023-06-13
Incorrect Folding for `else`


abhiroopdutta:
### Ruby version

3.1.2

### Description

Ruby LSP incorrectly folds if-else conditional block. The "else" condition is completely hidden. 
This can be reproduced by simply writing an if-else block and folding the if statement in VS Code. 
<img width="765" alt="Screenshot 2023-06-13 at 2 51 09 PM" src="https://github.com/Shopify/ruby-lsp/assets/19819930/976b685b-6e09-471a-90dc-daf6130283c9">

### Expected output

The else block should not be hidden when folding the if statement - 
<img width="765" alt="Screenshot 2023-06-13 at 2 53 18 PM" src="https://github.com/Shopify/ruby-lsp/assets/19819930/4f9ce466-f965-418d-a1f9-5d4963d5a97a">

andyw8:
Note: The 'expected' output matches how VS Code folds JavaScript.

abhiroopdutta:
I would like to work on this. I have no prior experience developing VS Code Extensions though, it would be helpful if you could point me to the right resources to get started.

andyw8:
Great! The implementation is in https://github.com/Shopify/ruby-lsp/blob/main/lib/ruby_lsp/requests/folding_ranges.rb

You will need to update the expectation files to match the new behaviour, https://github.com/Shopify/ruby-lsp/blob/main/CONTRIBUTING.md explains how those are used.

Feel free to ask other questions here.

andyw8:
I've added some details here: https://github.com/Shopify/ruby-lsp/pull/756

PR #759, opened 2023-06-15
Incorrect Folding for `else`

abhiroopdutta:
### Motivation

This fixes Issue #754 

### Implementation

`If` node was incorrectly listed in `SIMPLE_FOLDABLES` when it should rather be in `NODES_WITH_STATEMENTS`.

### Automated Tests

The bug originated because the expectation file was incorrect. I have corrected the same.

### Manual Tests

I have tested this against Ruby LSP's own local code in VS Code as mentioned in [CONTRIBUTING.md](https://github.com/Shopify/ruby-lsp/blob/main/CONTRIBUTING.md)


andyw8:
Oh, there are some test failures, do you want to try address those?

andyw8 (changes requested):
(test to be updated)

abhiroopdutta:
> Oh, there are some test failures, do you want to try address those?

My bad! It's fixed now.

vinistock (approved):

PR #772, opened 2023-06-20
Only handle gem statements if they have string literal arguments

vinistock:
### Motivation

We're still getting some errors while users are actively editing the `Gemfile`. I couldn't figure out exactly what triggers some of them, but I figured it would be worth to make the implementation more explicit that we only handle string literals.

### Implementation

Started returning early if the first argument is not a string literal. And then if inside that literal the first part is not a string content.

### Automated Tests

Added a new test.

### DIFF
--- a/lib/ruby_lsp/requests/code_lens.rb
+++ b/lib/ruby_lsp/requests/code_lens.rb
@@ -199,10 +199,13 @@ module RubyLsp
 
       sig { params(node: SyntaxTree::Command).returns(T.nilable(String)) }
       def resolve_gem_remote(node)
-        gem_statement = node.arguments.parts.flat_map(&:child_nodes).first
-        return unless gem_statement
+        gem_statement = node.arguments.parts.first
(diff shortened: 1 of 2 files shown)


st0012 (approved):

andyw8 (approved):

snutij:
Hi @vinistock, I'm sorry about this feature which needed 2 bug fixes. I was focused on "reading" a Gemfile, without thinking about the users who can edit it or edge-case like that!

I'm curious about the `We're still getting some errors` do you monitor errors/usages somewhere? 🤔 

Thanks for both fixes!

vinistock:
Hey @snutij! Don't worry, it's hard to predict what intermediate states files can be in.

The Ruby LSP extension allows connecting to a private metrics extension ([docs](https://github.com/Shopify/vscode-ruby-lsp#telemetry)). We collect metrics internally from our devs to find crashes.

PR #779, opened 2023-06-22
Make on type formatting `end` completions more robust

### LINKED ISSUE #500: Better handle on type formatting when there's content inside a block
Based on https://github.com/Shopify/ruby-lsp/pull/462#discussion_r1082638008, we should better handle what happens with on type formatting in case there is content in the next line.

For example
```ruby
if something
# ^ break here
immediate_content

# This results in this, which is definitely not ideal
if something
endimmediate_content
```

Fixing this means we have to perform some sort of lookahead to figure out what comes on the next lines.

vinistock:
### Motivation

Closes #500

This bug was bothering me too much. Instead of trying to use the document scanner, I switched the implementation of the on type formatting `end` completions to just deal with document lines. This simplifies the implementation and makes it easier to address the bug, so I think it's a win-win.

### Implementation

The way the lines work are as follows
```ruby
if something
#  ^ break here

# previous_line = "if "
# current_line = "  something"
# next_line = nil
```

So, the final logic is:
- If the current line is empty, then we broke the line at the end and so we can just put the `end` token normally
- If the current line is not empty and the next line is not empty either, then it's a bit hard to know what the intent of the user was. We just return early (this is the bug fix)
- Otherwise, we add the `end` token in a way that allows `something` to be in between

### Automated Tests

Added a scenario and fixed another one that was weird.

### Manual Tests

Basically fire up the LSP on this branch and play around breaking `if` statements. See if the `end` token ends up weird in any combination.

### DIFF
--- a/lib/ruby_lsp/requests/on_type_formatting.rb
+++ b/lib/ruby_lsp/requests/on_type_formatting.rb
@@ -30,13 +30,11 @@ module RubyLsp
       def initialize(document, position, trigger_character)
         super(document)
 
-        scanner = document.create_scanner
-        line_begin = position[:line] == 0 ? 0 : scanner.find_char_position({ line: position[:line] - 1, character: 0 })
-        @line_end = T.let(scanner.find_char_position(position), Integer)
-        line = T.must(@document.source[line_begin..@line_end])
+        @lines = T.let(@document.source.lines, T::Array[String])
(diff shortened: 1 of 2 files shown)


bitwise-aiden:
Actually, do we have a way to check if an end already exists for this code block? One thing I've noticed is additional ends in cases where I edit the opening line and hit enter.

Issue #709, opened 2023-05-09
Error running ruby-lsp in path with spaces


tonklon:
When running ruby-lsp in a project with a space somewhere in the absolute path it crashes with:

```
bundler: failed to load command: ruby-lsp (/Users/tono/.asdf/installs/ruby/3.1.3/bin/ruby-lsp)
/Users/tono/.asdf/installs/ruby/3.1.3/lib/ruby/gems/3.1.0/gems/uri-0.12.1/lib/uri/rfc3986_parser.rb:66:in `split': bad URI(is not URI?): "file:///Users/tono/some/path with spaces/code" (URI::InvalidURIError)
	from /Users/tono/.asdf/installs/ruby/3.1.3/lib/ruby/gems/3.1.0/gems/uri-0.12.1/lib/uri/rfc3986_parser.rb:71:in `parse'
	from /Users/tono/.asdf/installs/ruby/3.1.3/lib/ruby/gems/3.1.0/gems/uri-0.12.1/lib/uri/common.rb:193:in `parse'
	from /Users/tono/.asdf/installs/ruby/3.1.3/lib/ruby/gems/3.1.0/gems/uri-0.12.1/lib/uri/common.rb:722:in `URI'
	from /Users/tono/.asdf/installs/ruby/3.1.3/lib/ruby/gems/3.1.0/gems/ruby-lsp-0.5.1/lib/ruby_lsp/utils.rb:9:in `<module:RubyLsp>'
	from /Users/tono/.asdf/installs/ruby/3.1.3/lib/ruby/gems/3.1.0/gems/ruby-lsp-0.5.1/lib/ruby_lsp/utils.rb:4:in `<top (required)>'
	from /Users/tono/.asdf/installs/ruby/3.1.3/lib/ruby/gems/3.1.0/gems/ruby-lsp-0.5.1/lib/ruby_lsp/internal.rb:12:in `require'
	from /Users/tono/.asdf/installs/ruby/3.1.3/lib/ruby/gems/3.1.0/gems/ruby-lsp-0.5.1/lib/ruby_lsp/internal.rb:12:in `<top (required)>'
	from /Users/tono/.asdf/installs/ruby/3.1.3/lib/ruby/gems/3.1.0/gems/ruby-lsp-0.5.1/exe/ruby-lsp:21:in `require_relative'
[...]

PR #781, opened 2023-06-23
WIP: fix working in directories with spaces in path

tonklon:
This addresses #709 and allows ruby-lsp to work in directories with spaces in the path name.

The handling of paths in ruby-lsp is not really consistent, sometimes URIs are used sometimes, strings. Sometimes converting between the two is correct sometimes not.

This PR currently solves the problem I am facing in my setup, but I like to further improve this by:
- adding specs
- checking with other special chars (👍)

Feedback is already welcome, if there's something to improve.

### DIFF
--- a/lib/ruby_lsp/requests/code_lens.rb
+++ b/lib/ruby_lsp/requests/code_lens.rb
@@ -42,7 +42,7 @@ module RubyLsp
         @external_listeners = T.let([], T::Array[RubyLsp::Listener[ResponseType]])
         @test_library = T.let(test_library, String)
         @response = T.let([], ResponseType)
-        @path = T.let(T.must(URI(uri).path), String)
+        @path = T.let(T.must(URI(URI::Parser.new.escape(uri)).path), String)
         # visibility_stack is a stack of [current_visibility, previous_visibility]
         @visibility_stack = T.let([["public", "public"]], T::Array[T::Array[T.nilable(String)]])
         @class_stack = T.let([], T::Array[String])


tonklon:
I have signed the CLA!

lib/ruby_lsp/utils.rb
```diff
@@ -6,7 +6,7 @@ module RubyLsp
   VOID = T.let(Object.new.freeze, Object)
 
   # This freeze is not redundant since the interpolated string is mutable
-  WORKSPACE_URI = T.let(URI("file://#{Dir.pwd}".freeze), URI::Generic)
+  WORKSPACE_URI = T.let(URI(URI::Parser.new.escape("file://#{Dir.pwd}".freeze)), URI::Generic)
```
vinistock, line 9: I have a question regarding this one: does the editor send spaces encoded or unencoded? That is, do we receive `file:///something otherthing/blah.rb` or `file:///something%20otherthing/blah.rb`?

And does `URI()` not handle does already? If I'm not mistaken `URI()` invokes the parser, so I'm surprised it doesn't handle the spaces.

vinistock (changes requested):

lib/ruby_lsp/requests/formatting.rb
```diff
@@ -67,7 +67,7 @@ def run
         return if @formatter == "none"
 
         # Don't try to format files outside the current working directory
-        return unless @uri.sub("file://", "").start_with?(Dir.pwd)
+        return unless URI(@uri).path&.start_with?(T.must(WORKSPACE_URI.path))
```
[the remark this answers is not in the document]
tonklon: Yes, that might be a better way to go, but that is out of my perceived capability for now.
Aka: I don't think I could do this at this time. Especially because I am not familiar with the project structure and with sorbet.

tonklon:
I just open https://github.com/Shopify/vscode-ruby-lsp/pull/716 on the extensions itself, because that apparently broke  space-paths, too, since I last worked on this.

lib/ruby_lsp/utils.rb
tonklon: The URI gem does not have a better answer to this. I created https://github.com/ruby/uri/issues/87 over there.

vinistock:
Thank you very much for the PR, but I do think it's worth for us to start handling URIs properly and stop the mess of representing them as strings from spreading further.

I put up https://github.com/Shopify/ruby-lsp/pull/848 to fully change to URI objects and it should address the issues with spaces/emojis in paths.

vinistock:
@tonklon you are correct, I missed some spots. This PR should get us to a better state https://github.com/Shopify/ruby-lsp/pull/850.

PR #782, opened 2023-06-23
Move external listeners handling into extended requests

### LINKED ISSUE #690: Aggregate all listeners for each request type including our own
Currently, our listeners are separate from extension listeners. This is slightly inconvenient and not really necessary. Instead of always doing
```ruby
emitter = EventEmitter.new
base_listener = BaseListener.new(emitter)
extension_listeners = BaseListener.listeners.map { |l| l.new(emitter) }
emitter.visit(document.tree)
extension_listeners.each { |l| base_listener.merge_response!(l) }
```
We can include our listener in the list as the first element so that we can do
```ruby
emitter = EventEmitter.new
all_listeners = BaseListener.listeners.map { |l| l.new(emitter) }
emitter.visit(document.tree)
all_listeners[1..].each { |l| all_listeners[0].merge_response!(l) } # or something more elegant
```
Original discussion https://github.com/Shopify/ruby-lsp/pull/681#discussion_r1183061683.

st0012:
### Motivation

Closes #690 

### Implementation

Make extensable requests (`CodeLens` and `Hover` atm) handle external listeners' initialisation and collect responses internally, instead of exposing them in `Executor`.

Given that we currently only have 2 extendable requests and we're not sure what will come next, @vinistock and I decided to hold off further extraction (e.g. move those methods into `Listener` class) for now.

### Automated Tests

No new tests, all existing tests should pass.

### Manual Tests

Nope.


egiurleo (approved):
Nice! Excited to see how this evolves once we have more extensible requests.

lib/ruby_lsp/requests/code_lens.rb
```diff
@@ -56,6 +55,28 @@ def initialize(uri, emitter, message_queue, test_library)
...
+      sig { void }
+      def merge_external_listeners_responses!
+        @external_listeners.each do |l|
+          merge_response!(l)
+        end
+      end
+
+      sig { override.returns(ResponseType) }
+      def response
+        merge_external_listeners_responses!
```
vinistock, line 78: How about we invoke `merge_external_listeners_responses!` ourselves in the `Executor` and do not change `response`?

I find it a bit weird that invokeing `request.response` is no longer idempotent (since it always merges the responses).

st0012: Make sense. Updated 👍

lib/ruby_lsp/requests/hover.rb
```diff
@@ -39,10 +39,25 @@ class Hover < Listener
       def initialize(emitter, message_queue)
         super
 
+        @external_listeners = T.let(external_listeners, T::Array[RubyLsp::Listener[T.nilable(Interface::Hover)]])
```
vinistock, line 42: The type here should be
```suggestion
        @external_listeners = T.let(external_listeners, T::Array[RubyLsp::Listener[Hover]])
```
It's referring to a `Hover` the request and not `Interface::Hover`.

lib/ruby_lsp/requests/code_lens.rb
```diff
@@ -38,6 +38,8 @@ class CodeLens < Listener
       def initialize(uri, emitter, message_queue, test_library)
         super(emitter, message_queue)
 
+        @uri = T.let(uri, String)
+        @external_listeners = T.let(external_listeners, T::Array[RubyLsp::Listener[T::Array[Interface::CodeLens]]])
```
vinistock, line 42: I think there's an extra `T::Array` in this type. Also, the listener generic parameter is the request `CodeLens` not the `Interface::CodeLens`.
```suggestion
        @external_listeners = T.let(external_listeners, T::Array[RubyLsp::Listener[CodeLens]])
```

lib/ruby_lsp/requests/code_lens.rb
```diff
@@ -58,6 +60,21 @@ def initialize(uri, emitter, message_queue, test_library)
         )
       end
 
+      sig { returns(T::Array[RubyLsp::Listener[T::Array[Interface::CodeLens]]]) }
+      def external_listeners
+        self.class.listeners.map do |l|
+          # T.unsafe is needed because we expect CodeLens listeners to have different constructor arguments
+          T.unsafe(l).new(@uri, @emitter, @message_queue)
```
vinistock, line 67: Can we remove the `T.unsafe` with the updated type?

lib/ruby_lsp/requests/code_lens.rb
vinistock: Nevermind, the right type is `Listener[ResponseType]` indeed. We won't be able to escape the `T.unsafe`.

Issue #448, opened 2023-01-11
Use shared fixture repository for our tests


vinistock:
The https://github.com/ruby-syntax-tree/ruby-syntax-fixtures repository includes an extensive list of valid Ruby snippets that is being used to test Syntax Tree and YARP. The Ruby LSP should also use these fixtures in its tests, so that we can all contribute and benefit from a complete list of Ruby examples.

PR #788, opened 2023-06-27
Tests: Use shared ruby syntax tree fixtures

snutij:
### Motivation

Closes: https://github.com/Shopify/ruby-lsp/issues/448

### Implementation

- Add as a submodule https://github.com/ruby-syntax-tree/ruby-syntax-fixtures
- Edit CI Action to pull submodule dependency, see [documentation](https://github.com/actions/checkout#usage)
- Edit Dependabot to update it automatically every week as other gems dependencies, see [documentation](https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file#package-ecosystem)
- Edit test example name, because all fixtures provided by ruby-syntax-tree are structured like that: 
```
test/fixtures/ruby-syntax-fixtures/fixtures
├── CHAR
│   ├── 0.rb
│   ├── 1.rb
├── access_ctrl
│   ├── 0.rb
│   ├── 1.rb
```

So it was creating two time, for example `test_code_action_resolve_0__does_not_raise` which is making `minitest` crash.
I'm not sure this proposal is the best solution, feel free to challenge this part! 

- **Then, the main pain point**: Actually there are 23 failing tests, not sure is on `ruby-lsp` side. I'm opening this PR to have some feedback on how we can handle that 😃 

<details><summary>Summary failing tests</summary>

```
DiagnosticsExpectationsTest#test_diagnostics_do_block_4__does_not_raise:
RuntimeError: Neutered Exception RubyLsp::Requests::support::InternalRuboCopError: An internal error occurred for the Sorbet/KeywordArgumentOrdering cop.

DiagnosticsExpectationsTest#test_diagnostics_CHAR_3__does_not_raise:
RuntimeError: Neutered Exception RubyLsp::Requests::support::InternalRuboCopError: An internal error occurred for the Style/FrozenStringLiteralComment cop.

DiagnosticsExpectationsTest#test_diagnostics_mlhs_paren_1__does_not_raise:
RuntimeError: Neutered Exception RubyLsp::Requests::support::InternalRuboCopError: An internal error occurred for the VariableForce cop.

DiagnosticsExpectationsTest#test_diagnostics_CHAR_2__does_not_raise:
RuntimeError: Neutered Exception RubyLsp::Requests::support::InternalRuboCopError: An internal error occurred for the Style/FrozenStringLiteralComment cop.

DiagnosticsExpectationsTest#test_diagnostics_for_5__does_not_raise:
RuntimeError: Neutered Exception RubyLsp::Requests::support::InternalRuboCopError: An internal error occurred for the VariableForce cop.

DiagnosticsExpectationsTest#test_diagnostics_for_4__does_not_raise:
RuntimeError: Neutered Exception RubyLsp::Requests::support::InternalRuboCopError: An internal error occurred for the VariableForce cop.

DiagnosticsExpectationsTest#test_diagnostics_mlhs_paren_3__does_not_raise:
RuntimeError: Neutered Exception RubyLsp::Requests::support::InternalRuboCopError: An internal error occurred for the VariableForce cop.

DiagnosticsExpectationsTest#test_diagnostics_for_3__does_not_raise:
RuntimeError: Neutered Exception RubyLsp::Requests::support::InternalRuboCopError: An internal error occurred for the VariableForce cop.

DiagnosticsExpectationsTest#test_diagnostics_mlhs_paren_2__does_not_raise:
RuntimeError: Neutered Exception RubyLsp::Requests::support::InternalRuboCopError: An internal error occurred for the VariableForce cop.

FormattingExpectationsTest#test_formatting_do_block_4__does_not_raise:
RuntimeError: Neutered Exception RubyLsp::Requests::support::InternalRuboCopError: An internal error occurred for the Sorbet/EmptyLineAfterSig cop.

FormattingExpectationsTest#test_formatting_CHAR_3__does_not_raise:
RuntimeError: Neutered Exception RubyLsp::Requests::support::InternalRuboCopError: An internal error occurred for the Style/FrozenStringLiteralComment cop.

FormattingExpectationsTest#test_formatting_mlhs_paren_1__does_not_raise:
RuntimeError: Neutered Exception RubyLsp::Requests::support::InternalRuboCopError: An internal error occurred for the VariableForce cop.

FormattingExpectationsTest#test_formatting_CHAR_2__does_not_raise:
RuntimeError: Neutered Exception RubyLsp::Requests::support::InternalRuboCopError: An internal error occurred for the Style/FrozenStringLiteralComment cop.

FormattingExpectationsTest#test_formatting_for_5__does_not_raise:
RuntimeError: Neutered Exception RubyLsp::Requests::support::InternalRuboCopError: An internal error occurred for the VariableForce cop.

FormattingExpectationsTest#test_formatting_void_stmt_0__does_not_raise:
RubyLsp::Requests::Formatting::Error: Infinite loop detected in /Users/justinrabiller/Documents/www/ruby-lsp/test/requests/formatting_expectations_test.rb and caused by Layout/SpaceAfterSemicolon -> Layout/SpaceBeforeSemicolon

FormattingExpectationsTest#test_formatting_for_4__does_not_raise:
RuntimeError: Neutered Exception RubyLsp::Requests::support::InternalRuboCopError: An internal error occurred for the VariableForce cop.

FormattingExpectationsTest#test_formatting_mlhs_paren_3__does_not_raise:
RuntimeError: Neutered Exception RubyLsp::Requests::support::InternalRuboCopError: An internal error occurred for the VariableForce cop.

FormattingExpectationsTest#test_formatting_for_3__does_not_raise:
RuntimeError: Neutered Exception RubyLsp::Requests::support::InternalRuboCopError: An internal error occurred for the VariableForce cop.

FormattingExpectationsTest#test_formatting_when_7__does_not_raise:
NoMethodError: undefined method "type" for nil:NilClass

FormattingExpectationsTest#test_formatting_mlhs_paren_2__does_not_raise:
RuntimeError: Neutered Exception RubyLsp::Requests::support::InternalRuboCopError: An internal error occurred for the VariableForce cop.

FormattingExpectationsTest#test_formatting_when_8__does_not_raise:
NoMethodError: undefined method "type" for nil:NilClass

CodeActionResolveExpectationsTest#test_code_action_resolve_statements_0__does_not_raise:
TypeError: Passed `nil` into T.must

CodeActionResolveExpectationsTest#test_code_action_resolve_embdoc_0__does_not_raise:
TypeError: Passed `nil` into T.must
```
</details> 

### Automated Tests

Difficult to test Dependabot, but for CI workflow, this PR should trigger a lot more tests.

### Manual Tests

- checkout this branch
- `git submodule update --init --recursive`
- `bin/test`

Then to have a complete stack trace of `rubocop` fail: `bin/rubocop -d path/for/fixture`


andyw8:
This looks great! Some of the failures might take a while to address, so it could be useful to have a way that to mark those as known failures, which can be fixed incrementally in future.

snutij:
- I know that in `test/expectations/expectations_test_runner.rb` we already skip some tests (based on ruby version requirement). Maybe we can do something similar for tests files included in a `WAINTING_FOR_RUBOCOP_FIX` array, for example. 
- In the same time, I can open some issues on `rubocop` to report failing fixtures with the related stack trace, and keep an eye on it to update this array when fixes are released.    

This is not ideal, but allow us to go forward, and already provide a lot more fixture. wdyt?

andyw8:
> Maybe we can do something similar for tests files included in a WAINTING_FOR_RUBOCOP_FIX array, for example.

That seems reasonable. I haven't looked in detail at the failures yet, are they all from RuboCop? I would have thought we might see some caused by bugs in ruby-lsp itself.

Btw, ne thing I noticed when trying this locally: Initially everything was wrongly passing, because I hadn't run yet `git submodule update --init`, so `test/fixtures/ruby-syntax-fixtures/` was empty. Maybe we should have a check that this is non-empty?

vinistock (commented):
This is awesome and already showing value given that it caught many errors

andyw8:
:wave: After some more internal discussion:
* We would like to go with using https://github.com/ruby/yarp/tree/main/test/fixtures (still as a submodule) for testing against.
* Let's leave out the Dependabot configuration for now - the fixtures don't change too often, and we can manually pull in the latest when needed.
* For the failures due to `InternalRuboCopError` exceptions, let's report these upstream to RuboCop, but ignore them for our test suite, since they are outside of our control.
* For the failures that are due to ruby-lsp's code, let's add the breaking examples to our own fixtures. Feel free to try and fix those if you would like to, otherwise we can assist.

Thanks again for your contribution!

snutij:
Thanks for coming back @andyw8. 🙏 

For the first point, YARP fixtures are not a repository (but only a subfolder from a repository). If we want to use it as a submodule, we have to point to the full git project (and pull the full YARP repo). It's look not ideal to me, maybe I'm missing your point? Perhaps there is a way to do that, that I'm not aware of yet. 😄 

For the rest of your comment, I'll do it 👍

andyw8:
Yes we're ok with pulling the full YARP repo, it's fairly small. (@kddnewton said he may publish the fixtures to a separate repo at some point).

test/expectations/expectations_test_runner.rb
```diff
@@ -93,6 +93,10 @@ def test_#{expectation_suffix}__#{test_name}__does_not_raise
       end
     end
 
+    def uniq_name_from_path(path)
```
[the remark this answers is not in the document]
snutij: Done, I must keep all path context, because with YARP, we still have duplicate name, even with 3 parent folder context 😄

test/expectations/expectations_test_runner.rb
```diff
@@ -3,8 +3,44 @@
 
 class ExpectationsTestRunner < Minitest::Test
   TEST_EXP_DIR = "test/expectations"
-  TEST_FIXTURES_DIR = "test/fixtures"
-  TEST_FIXTURES_GLOB = File.join(TEST_FIXTURES_DIR, "**", "*.rb")
+  TEST_RUBY_LSP_FIXTURES_GLOB = File.join("test/fixtures", "*.rb")
+  TEST_RAILS_FIXTURES_GLOB = File.join("test/fixtures/rails_app", "**", "*.rb")
+  TEST_YARP_FIXTURES_GLOB = File.join("test/fixtures/yarp/test/fixtures", "**", "*.txt")
+  WAINTING_FOR_RUBOCOP_FIX = %w[
```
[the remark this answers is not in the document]
snutij: I replaced the list by a rescue of this specific error that skip the test. Without doing anything, skipped tests count should be less and less important, with Rubocop updates. Way better, tell me if this is what you expected! 

In the same time, I started (based on this previous array) to report them issues, and few are already fixed.

There are still some failures, especially some by nil value given to a `T.must()` sorbet syntax, which need maybe a fix on requests side. I'm not very confident with sorbet stuff, but I will dig into it.

Sorry I cannot pass this PR in draft but still some work!

Rakefile
```diff
@@ -8,7 +8,7 @@ require "ruby_lsp/check_docs"
 Rake::TestTask.new(:test) do |t|
   t.libs << "test"
   t.libs << "lib"
-  t.test_files = FileList["test/**/*_test.rb"]
+  t.test_files = FileList["test/**/*_test.rb"].exclude("test/fixtures/**/*.rb")
```
vinistock, line 11: I don't think this change is necessary since all fixtures are `txt` files.

test/expectations/expectations_test_runner.rb
```diff
@@ -3,8 +3,10 @@
 
 class ExpectationsTestRunner < Minitest::Test
   TEST_EXP_DIR = "test/expectations"
-  TEST_FIXTURES_DIR = "test/fixtures"
-  TEST_FIXTURES_GLOB = File.join(TEST_FIXTURES_DIR, "**", "*.rb")
+  TEST_RUBY_LSP_FIXTURES_GLOB = File.join("test/fixtures", "*.rb")
+  TEST_RAILS_FIXTURES_GLOB = File.join("test/fixtures/rails_app", "**", "*.rb")
```
vinistock, line 7: Since the fixtures have different extensions, I think we don't need to separate the `rails_app` here from the other ones.

test/requests/hover_expectations_test.rb
```diff
@@ -8,6 +8,8 @@
 class HoverExpectationsTest < ExpectationsTestRunner
   expectations_tests RubyLsp::Requests::Hover, "hover"
 
+  TEST_FIXTURES_DIR = "test/fixtures"
```
vinistock, line 11: This shouldn't be necessary.

vinistock (commented):
There seems to be at least two failures that are also related to infinite loops in RuboCop when running formatting.

I think we should just use the same technique and ignore any RuboCop related errors by rescuing `RubyLsp::Requests::Formatting::Error`.

Rakefile
vinistock: Oh, right. I didn't think about YARP's own tests. Do you know if there's some git submodule magic that will allow us to only pull the `txt` fixtures?

If not, that's fine, let's move forward with the exclude.

snutij: As far as I know, this is not possible with a submodule command.

What may be possible is to delete all files except fixtures in the [test](https://github.com/Shopify/ruby-lsp/blob/main/bin/test) script. But this is a kind of "workaround" 😄

snutij:
Well, CI is green 🎉 feel free to review it!

lib/ruby_lsp/requests/code_action_resolve.rb
```diff
@@ -51,7 +51,9 @@ def run
         scanner = @document.create_scanner
         start_index = scanner.find_char_position(source_range[:start])
         end_index = scanner.find_char_position(source_range[:end])
-        extracted_source = T.must(@document.source[start_index...end_index])
+        extracted_source = @document.source[start_index...end_index]
+
+        return Error::InvalidTargetRange if extracted_source.nil?
```
vinistock, line 56: Why is this needed? Since `code_action_resolve` requires the user to select something in the editor, I don't think the range would ever be `nil`.

Maybe this needs an improvement in the test instead?

test/expectations/expectations_test_runner.rb
```diff
@@ -91,6 +95,27 @@ def test_#{expectation_suffix}__#{test_name}__does_not_raise
...
+      TEST_YARP_FIXTURES.each do |path|
+        class_eval(<<~RB, __FILE__, __LINE__ + 1)
+          def test_#{expectation_suffix}__#{uniq_name_from_path(path)}__does_not_raise
+            @_path = "#{path}"
+            source = File.read(@_path)
+            run_expectations(source)
+          rescue RubyLsp::Requests::Support::InternalRuboCopError, RubyLsp::Requests::Formatting::Error
+            skip "Fixture requires a fix from Rubocop"
+          rescue NoMethodError => error
+            skip "Fixture requires a fix from SyntaxTree" if error.message.include?("SyntaxTree::MLHSParen")
```
vinistock, line 108: Can you share which scenario caused this one?

Also, we need to raise the error again here since `NoMethodError` could potentially be an issue in the LSP itself.

test/requests/diagnostics_expectations_test.rb
```diff
@@ -15,13 +15,12 @@ def run_expectations(source)
...
+    result = T.let([], T::Array[RubyLsp::Requests::Support::RuboCopDiagnostic])
 
     stdout, _ = capture_io do
-      result = T.cast(
-        RubyLsp::Requests::Diagnostics.new(document).run,
-        T::Array[RubyLsp::Requests::Support::RuboCopDiagnostic],
-      )
+      diagnostics = []
+      diagnostics.push(RubyLsp::Requests::Diagnostics.new(document).run)
+      result = diagnostics.compact.flatten
```
vinistock, line 23: What error is this change fixing? Aren't we able to just do `result = RubyLsp::Requests::Diagnostics.new(document).run`?

test/expectations/expectations_test_runner.rb
snutij: `inlay_hints`, `document_symbol`, `document_highlight`, `document_link`, `code_lens` & `semantic_highlighting` running against: https://github.com/ruby/yarp/blob/main/test/fixtures/methods.txt

Its crash here: 
https://github.com/ruby-syntax-tree/syntax_tree/blob/main/lib/syntax_tree/with_scope.rb#L78-L89 

I can open an issue to report also.

test/requests/diagnostics_expectations_test.rb
snutij: If the document doesn't have valid syntax, the request return early, and this is raising an error below due to `T.must()`
Some fixtures aren't considered as valid on older ruby version, for example: https://github.com/ruby/yarp/blob/main/test/fixtures/whitequark/forwarded_kwrestarg_with_additional_kwarg.txt.

test/expectations/expectations_test_runner.rb
snutij: I tested with your fix and specifying the `main` branch in the `gemfile`, it's working, all tests should be OK after bumping the requirement. 👍  I keep an eye on it, and I'll re-run the CI on this PR after the official release.

bin/test
```diff
@@ -1,4 +1,11 @@
 #!/usr/bin/env bash
+YARP_FIXTURES_DIR=test/fixtures/yarp/test/fixtures
```
kddnewton, line 2: Sorry this is `test/yarp/fixtures` now

snutij: Thank you for raising this change 🙏 I did the change.
For some mysterious reason, two files trigger an error:

```
Assertion failed: (list->size == 0 || newline_offset > list->offsets[list->size - 1]), function yp_newline_list_append, file yp_newline_list.c, line 36.
rake aborted!
SignalException: SIGABRT
```

Files: 
- https://github.com/ruby/yarp/blob/main/test/yarp/fixtures/spanning_heredoc.txt
- https://github.com/ruby/yarp/blob/main/test/yarp/fixtures/newline_terminated.txt

I'll have to dig in this issue, but for now I can't understand why [this line](https://github.com/ruby/yarp/blob/main/src/util/yp_newline_list.c#L36) from YARP is run when we run tests.

If someone has an idea, do not hesitate 👍

snutij: Indeed, it's been corrected. I'm now waiting for a fix in Rubocop before pushing, I've already opened an issue. At this point, I wonder if it wouldn't be better to wait for the YARP migration, to remove the 6 `SyntaxTree` failures at the same time. WDYT?

Anyway, I'm keeping monitoring this PR, it's not abandoned in the wild. 😄

lib/ruby_lsp/requests/support/rubocop_runner.rb
```diff
@@ -20,14 +20,14 @@ class InternalRuboCopError < StandardError
...
+          An internal error occurred: %s.
           Updating to a newer version of RuboCop may solve this.
           For more details, run RuboCop on the command line.
         EOS
 
-        sig { params(rubocop_error: RuboCop::ErrorWithAnalyzedFileLocation).void }
+        sig { params(rubocop_error: StandardError).void }
         def initialize(rubocop_error)
-          message = format(MESSAGE, rubocop_error.cop.name)
+          message = format(MESSAGE, rubocop_error.message)
```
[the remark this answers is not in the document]
snutij: Done! I had to rescue only `StandardError` to avoid the `Lint/ShadowedException` error.

vinistock (approved):
Thank you very much for pushing this over the finish line, I know it took a while.

Amazing contribution 🚀

PR #789, opened 2023-06-27
Create and synchronize custom bundle in the server

vinistock:
### Motivation

Step towards https://github.com/Shopify/vscode-ruby-lsp/issues/421

Currently, we create our custom bundle from the VS Code extension so that the Ruby LSP can use the correct versions for dependencies in the target application. However, doing it in the extension has a few drawbacks:
- It doesn't work for other editors, but this behaviour is desired for anyone using the LSP
- The behaviour is difficult to test from the extension side since it doesn't have direct access to the Ruby environment, it just interacts with Bundler as a CLI

### Implementation

- If the LSP is being executed with `bundle exec`, nothing changes
- If the LSP is executed from a global installation, then `BUNDLE_GEMFILE` is `nil`
    - If that project has no bundle (no `Gemfile.lock`), then we just execute normally
    - Otherwise, we require the custom bundle setup script, which basically does the same as the extension is doing right now (setting up the `.ruby-lsp` folder)

This implementation also includes two bug fixes with the previous approach:
1. To my surprise, `Bundler.locked_gems.dependencies` does not include runtime dependencies declared in the gemspec. This means that we were still creating a `.ruby-lsp` for projects like `ruby-lsp-rails`. I'm not 100% sure if this is intended behaviour or a bug, but I started merging dependencies from both lockfile and gemspec so that we can properly avoid version mismatch errors
2. Our previous solution for handling a custom `BUNDLE_PATH` was to force `bundle install` to use the system path. A few users reported this doesn't work well - especially with docker. What we actually want to do is to turn the relative `BUNDLE_PATH` into an absolute one to avoid creating an incorrect folder under `.ruby-lsp`. For example
    - If `BUNDLE_PATH=vendor/bundle`, then we want to use the top level's `vendor/bundle` folder and not `.ruby-lsp/vendor/bundle`
    - By using the absolute path, we avoid this issue without forcing system paths

### Migration strategy

Nothing should happen until the extension side is updated in https://github.com/Shopify/vscode-ruby-lsp/pull/664.

Projects with the `ruby-lsp` in the bundle will simply continue to work thanks to the check in `exe/ruby-lsp` before re-launching the process.

This change will only be problematic for users on Ruby 2.7 because we will change the extension to run `ruby-lsp` instead of `bundle exec ruby-lsp`. Since they can't upgrade to get the latest executable, it will simply fail.

### Automated Tests

I added tests for this scripts so that we can avoid regressions and more easily maintain this logic.

### Manual tests

This PR on its own (without the extension part), doesn't do much. We will need to publish Ruby LSP v0.7.0 and then test the extension side very thoroughly, then make a preview release so that we can catch bugs ahead of the stable release.

For this PR, the only two scenarios to verify are the `ruby-lsp` included or not in the bundle.
1. Go to a project that has the `ruby-lsp` in the bundle
2. Make sure it boots properly
3. Make sure the extension continues to bundle install normally (until the extension is changed, the new script will actually never be used)
4. Repeat for a project without the `ruby-lsp` in the bundle


andyw8 (commented):
Had a first read through, there's a lot going on but it seems to make sense. I expect you've already verified against a number of repos, if you want to list them then I can pick some others to test with.

vinistock:
> Interesting, I wonder if vscode-ruby-lsp also needs to change, so that it can detect formatters specified in gemspecs:

Yes, it's most likely missing those right now.

>  if you want to list them then I can pick some others to test with.

I added tophat instructions to the PR's description. Sorry, I should've added them from the start!

KaanOzkan (approved):

PR #796, opened 2023-07-04
Indicate that using `bundle exec` is no longer recommended as of v0.7.0

vinistock:
### Motivation

Starting with v0.7.0, we'll ship with the custom bundle functionality from the server and there's no more need to use `bundle exec` or include the gem in the `Gemfile`.

### DIFF
--- a/EDITORS.md
+++ b/EDITORS.md
@@ -11,15 +11,11 @@ new H2 header in this file containing the instructions. -->
 
 ## Emacs Eglot
 
-[Eglot](https://github.com/joaotavora/eglot) runs solargraph server by default. To set it up with ruby-lsp you need to put that in your init file:
+[Eglot](https://github.com/joaotavora/eglot) runs solargraph server by default. To set it up with ruby-lsp you need to
+put that in your init file:
 ```el
 (with-eval-after-load 'eglot
  (add-to-list 'eglot-server-programs '((ruby-mode ruby-ts-mode) "ruby-lsp")))
  ```
-Or if you want to run it with bundler:
-```el
-(with-eval-after-load 'eglot
- (add-to-list 'eglot-server-programs '((ruby-mode ruby-ts-mode) "bundle" "exec" "ruby-lsp")))
-```
- 
+
 When you run `eglot` command it will run `ruby-lsp` process for you.
--- a/README.md
+++ b/README.md
@@ -27,7 +27,11 @@ The gem can be installed by doing
 gem install ruby-lsp
 ```
 
-If you decide to add the gem to the bundle, it is not necessary to require it.
+**NOTE**: starting with v0.7.0, it is no longer recommended to add the `ruby-lsp` to the bundle. The gem will generate a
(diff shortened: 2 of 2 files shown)


README.md
```diff
@@ -27,7 +27,10 @@ The gem can be installed by doing
 gem install ruby-lsp
 ```
 
-If you decide to add the gem to the bundle, it is not necessary to require it.
+**NOTE**: starting with v0.7.0, it is no longer recommended to add the `ruby-lsp` to the bundle. The gem will generate a
+custom bundle in `.ruby-lsp/Gemfile` which is used to the versions of dependencies that should be used (e.g.: RuboCop).
```
st0012, line 31: I don't completely understand what this means. Could you mean this?

```suggestion
custom bundle in `.ruby-lsp/Gemfile` which is used with the versions of dependencies that should be used (e.g.: RuboCop).
```

st0012 (commented):
Should we wait for the release of `v0.7.0` before we merge this?

README.md
andyw8, line 31: Missing a word here I think?
```suggestion
custom bundle in `.ruby-lsp/Gemfile` which is used to specify the versions of dependencies that should be used (e.g.: RuboCop).
```

andyw8 (changes requested):

vinistock:
Yeah, there was a word missing 🤦. Probably a lot easier to understand now, but please let me know if that's not the case.

> Should we wait for the release of v0.7.0 before we merge this?

Given that we mention this applies only to v0.7.0 and up, I think it's okay to merge. Do you prefer to wait?

EDITORS.md
```diff
@@ -11,15 +11,11 @@ new H2 header in this file containing the instructions. -->
 
 ## Emacs Eglot
 
-[Eglot](https://github.com/joaotavora/eglot) runs solargraph server by default. To set it up with ruby-lsp you need to put that in your init file:
+[Eglot](https://github.com/joaotavora/eglot) runs solargraph server by default. To set it up with ruby-lsp you need to
+put that in your init file:
```
bitwise-aiden, line 15:
```suggestion
[Eglot](https://github.com/joaotavora/eglot) runs solargraph server by default. To set it up with ruby-lsp you need to put that in your init file:
```

bitwise-aiden (approved):
Small nit, but lgtm

st0012:
> Given that we mention this applies only to v0.7.0 and up, I think it's okay to merge. Do you prefer to wait?

Would the emacs change work before v0.7.0? If it would, then I'm good with merging it now 👍

vinistock:
That's fair, we don't mention the specific version on the other editors docs. Let's wait a bit until we release v0.7.0.

EDITORS.md
andyw8: (it also makes diffs much easier to read)

vinistock: I think one sentence per line might be a bit too much. As long as it's wrapping in 120 characters and it fits on the screen, it should be fine.

PR #799, opened 2023-07-05
Keep track of nesting when locating nodes

vinistock:
### Motivation

When we integrate the index into the Ruby LSP, we will need to figure out fully qualified name of target constants.

Imagine if someone is trying to go to the definition of Bar in this example:

```ruby
class Foo
  def something
    Bar
  end 
end
```
We first need to look inside the index for `Foo::Bar`. And then if that doesn't exist, we search for `Bar`.

Currently, when we locate the `Bar` node, we don't keep track of the nesting, so we don't know how deep we are in classes and modules. This PR keeps track of it to allow us to use fully qualified names for targets.

### Implementation

Added a new array for nesting and started pushing the names of modules and classes, which we return together with the target and parent information.

### Automated Tests

Added tests.


bitwise-aiden (approved):

st0012 (approved):
👍

PR #801, opened 2023-07-07
Simplify path completion to not use prefix tree

vinistock:
### Motivation

Currently, we build a prefix tree and then search it every time we're serving completion. I've been thinking about whether this is actually what we want to do or not. Creating these trees is super expensive.

Even if we were to maintain them in an index, we'd probably need multiple prefix trees for each thing we want to autocomplete and we'd need to synchronize it - which is also expensive and consumes memory.

I think switching to a simpler approach will prove more beneficial. This is the benchmark of this PR against main.

```
Warming up --------------------------------------
                 new   412.000  i/100ms
Calculating -------------------------------------
                 new      4.144k (± 1.5%) i/s -     21.012k in   5.071885s

Comparison:
                 new:     4143.8 i/s
                 old:        4.7 i/s - 876.83x  slower
```

### Implementation

Instead of building the prefix tree and searching based on the node value, we just search the `$LOAD_PATH` using the `node.value` directly, which only returns results that start with the desired path.


vinistock:
Indeed, when the index is integrated in the Ruby LSP, we won't have to list files anymore. But I'm pretty sure doing a `select` on the indexed files will still be faster than building and then searching the prefix tree. It might not be `876.83x`, but it will still be faster.

Also, when we have the index we'll implement autocomplete for classes, modules and methods. Those will produce even larger prefix trees.

What's the advantage of using the prefix tree? Is there an obvious advantage that's worth paying the performance price?

vinistock:
I'll close this for the time being then. We can do a more in depth exploration once we're ready to implement autocomplete for other things, such as classes and modules.

PR #805, opened 2023-07-10
Add show syntax tree request

vinistock:
### Motivation

Add a show syntax tree request, so that developers can easily inspect the AST of a given file. This is helpful when trying to figure out how Ruby understands code.

### Implementation

The new custom request just returns an object containing the pretty printed AST, so that we can put it in a file and display to the user.

In the future, we can also support
- Showing the AST for a selection
- Using some webview visualization in VS Code to allow navigating the AST alongside the file

### Automated Tests

Added a test for it.

### Manual Tests

2. Open a Ruby file
3. CMD + SHIFT + P: `Ruby LSP: show syntax tree`
4. Verify the syntax tree is displayed properly


lib/ruby_lsp/executor.rb
```diff
@@ -169,6 +169,8 @@ def run(request)
         completion(uri, request.dig(:params, :position))
       when "textDocument/definition"
         definition(uri, request.dig(:params, :position))
+      when "textDocument/showSyntaxTree"
```
andyw8, line 172: Is there any convention to distinguish custom requests from standard requests? I think I saw some LSPs use an underscore prefix.

lib/ruby_lsp/requests/show_syntax_tree.rb
```diff
@@ -0,0 +1,33 @@
...
+    # ![Show syntax tree demo](../../show_syntax_tree.gif)
+    #
+    # Show syntax tree is a custom [LSP
+    # request](https://microsoft.github.io/language-server-protocol/specification#requestMessage) that displays the AST
+    # for the current document in a new tab.
+    #
+    # # Example
+    #
+    # ```ruby
+    # # Executing the Ruby LSP: show syntax tree command will display the AST for the document
```
andyw8, line 15:
```suggestion
    # # Executing the Ruby LSP: Show syntax tree command will display the AST for the document
```

lib/ruby_lsp/executor.rb
Morriar: I like the idea of a prefix to avoid future conflicts 👍

st0012 (approved):

lib/ruby_lsp/executor.rb
vinistock: Added `rubyLsp/` as the prefix for custom requests.

PR #810, opened 2023-07-12
Support selection in show syntax tree

vinistock:
### Motivation

Support handling ranges in show syntax tree. This allows developers to see the AST only for a specific portion of the code.

### Implementation

The only notable part is the decision to artificially remove the initial `program` and `statements` nodes when there's a range. It doesn't make much sense to include them, since the developer is selecting a specific part of the document and those two initial nodes are related to the top level of the file.

### Automated Tests

Added tests.

### Manual Tests

1. Using the branches of both the server and client parts, start the LSP
2. Make a valid selection
3. Execute Show syntax tree
4. Verify the AST only includes what you selected
5. Make an invalid selection
6. Execute Show syntax tree
7. Verify that a message is shown informing of the parse error

### DIFF
--- a/lib/ruby_lsp/executor.rb
+++ b/lib/ruby_lsp/executor.rb
@@ -170,8 +170,13 @@ module RubyLsp
       when "textDocument/definition"
         definition(uri, request.dig(:params, :position))
       when "rubyLsp/textDocument/showSyntaxTree"
-        { ast: Requests::ShowSyntaxTree.new(@store.get(uri)).run }
+        show_syntax_tree(uri, request.dig(:params, :range))
       end
+    end
+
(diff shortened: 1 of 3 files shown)


vinistock:
That's true, the behaviour is not ideal indeed. I don't know of a good way to improve it. If we were to use the document's AST and locate the node using the range, there are some scenarios that wouldn't be super simple. For example

```ruby
# Start selection on the `d` character of `def`
def foo
end

def bar
end
# End selection on the `d` character of `end`
```

In this case, the node that exists at the start position of the range is just the `foo` definition. So we need some algorithm that will find us all nodes in a given range.

Any ideas on how to approach it?

PR #813, opened 2023-07-12
Add extension management

### LINKED ISSUE #590: Inform client of server extensions

vinistock:
### Motivation

Closes #590

It will be good to allow clients to both display which Ruby LSP extensions are being used and also deactivate them if desired. Maybe an extension is failing and we want to disable it temporarily.

This PR adds a custom request that clients can use to discover which extensions have been registered and also starts accepting an initialization option with extension configuration, which clients can use to disable specific extensions.

### Implementation

I split the implementation by commits.
1. Add a custom request that returns the discovered extensions
2. Starts accepting initialization options with extension configuration, so that we can skip activating disabled ones
3. Made an adjustment to our documentation mentioning that extensions should register their functionality all inside of `activate`. We'll need to update `ruby-lsp-rails` to match this, but I couldn't think of another way to de-register a listener that wouldn't involve messing up with the `$LOAD_PATH` and `$LOADED_FEATURES`

I'd appreciate feedback on the approach.

### Automated Tests

Added some tests.

### DIFF
--- a/SERVER_EXTENSIONS.md
+++ b/SERVER_EXTENSIONS.md
@@ -62,9 +62,13 @@ module RubyLsp
     class Extension < ::RubyLsp::Extension
       extend T::Sig
 
-      # Performs any activation that needs to happen once when the language server is booted
+      # Performs any activation that needs to happen once when the language server is booted in addition to registering
+      # listeners and request enhancements
       sig { override.void }
       def activate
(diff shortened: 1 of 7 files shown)


vinistock:
Putting this on pause from now until we have more use cases to learn from.

PR #819, opened 2023-07-17
Make sure to run bundle install even when not altering the custom bundle

vinistock:
### Motivation

There were three issues with the custom bundler setup:
1. When working in the `ruby-lsp` itself, we were comparing a `Pathname` to a `String`, which always returns `false`
2. We were not running bundle install if no updates related to the custom bundle were needed (like creating it or re-copying the `Gemfile.lock`), which means the LSP failed to activate silently
3. We never upgrade the `ruby-lsp` or `debug` automatically if there's a dependency on either of them. For example, when using `ruby-lsp-rails`

### Implementation

Extracted the logic to run bundle install into a lambda, so that we can reuse it in the script. Changed the command to run `bundle update` when the `ruby-lsp` or `debug` are not a part of the `Gemfile`.

Fixed the pathname comparison too.

### Automated Tests

Added a missing scenario and corrected the tests.


lib/ruby_lsp/setup_bundler.rb
```diff
@@ -9,9 +9,41 @@
 # the Ruby LSP without including the gem in their application's Gemfile while at the same time giving us access to the
 # exact locked versions of dependencies.
 
+run_bundle_install = lambda do |dependencies, bundle_gemfile = nil|
+  # If the user has a custom bundle path configured, we need to ensure that we will use the absolute and not relative
+  # version of it when running bundle install. This is necessary to avoid installing the gems under the `.ruby-lsp`
```
andyw8, line 14:
```suggestion
  # version of it when running `bundle install`. This is necessary to avoid installing the gems under the `.ruby-lsp`
```

lib/ruby_lsp/setup_bundler.rb
```diff
@@ -9,9 +9,41 @@
...
+  # folder, which is not the user's intention. For example, if path is configured as `vendor`, we want to install it in
```
andyw8, line 15:
```suggestion
  # folder, which is not the user's intention. For example, if the path is configured as `vendor`, we want to install it in
```

test/setup_bundler_test.rb
```diff
@@ -5,18 +5,22 @@
 
 class SetupBundlerTest < Minitest::Test
   def test_does_nothing_when_running_in_the_ruby_lsp
+    Object.any_instance.expects(:system).with(bundle_install_command)
     run_script
     refute_path_exists(".ruby-lsp")
   end
 
   def test_does_nothing_if_both_ruby_lsp_and_debug_are_in_the_bundle
+    Pathname.any_instance.expects(:basename).returns("fake_project")
```
andyw8, line 14: Is there a risk that globally stubbing `Pathname` here could have unintentional side-effects, e.g. on logging during tests?

test/setup_bundler_test.rb
```diff
@@ -67,12 +85,17 @@ def run_script
     $LOADED_FEATURES.delete(path)
   end
 
-  def bundle_install_command
+  def bundle_install_command(bundle_gemfile = nil, update = true)
```
andyw8, line 88:
```suggestion
  def bundle_install_command(bundle_gemfile = nil, update: true)
```
To make the caller slightly more readable.

andyw8 (approved):

test/setup_bundler_test.rb
vinistock: I don't think so. This test only runs the `setup_bundler.rb` script and the only place we use `basename` is in the specific line we're testing.

Do you see another place where it could accidentally stub the wrong return?

andyw8: I'm thinking that a logging library or something else that's used in tests might rely on `basename` internally.

It's probably fine here, but in large systems I've seen problems caused by `any_instance` so I'm generally cautious of it.

(it's interesting to note that Minitest has [rejected](https://github.com/minitest/minitest/pull/245#issuecomment-148219318) this feature, it's only in Mocha).

vinistock: I tried swapping for `File.basename` but that is used by Minitest itself and stubbing it will break.

I think it should be fine for now. The only way to avoid the stub would be re-writing this to be a class to that it's not invoked in a script style anymore.

PR #822, opened 2023-07-18
Remove old .ruby-lsp folder if `ruby-lsp` and `debug` are in the Gemfile

vinistock:
### Motivation

I'm really sorry about the back and forth and multiple bugfix releases, but I'm trying to make sure we covered every single scenario for removing the custom bundle from the extension using our [beta release](https://github.com/Shopify/vscode-ruby-lsp/releases/tag/v0.4.1).

If users decide to add both `ruby-lsp` and `debug` after the `.ruby-lsp` folder exists, then we need to remove it or else we start the server pointing at the wrong bundle - which might give a bundler error because gems on that bundle are not installed.

Additionally, if either one of the two gems aren't in the top level `Gemfile`, then we can't run `bundle update` immediately after copying the `Gemfile.lock` because Bundler will complain that the gem isn't present in the lock. We can only try to upgrade if the custom `.ruby-lsp/Gemfile.lock` already exists and includes both `ruby-lsp` and `debug`.

### Implementation

Added a fix for both scenarios and bumped the version again.

### Automated Tests

Added a test covering the scenario.

Removed the part of the test related to `update` because it's a bit hard to stub correctly. We want the first invocation of `dependencies` to return empty and the second one to include `ruby-lsp` and `debug`. If you know how to get it right, please let me know.


test/setup_bundler_test.rb
```diff
@@ -6,18 +6,28 @@
...
   def test_does_nothing_if_both_ruby_lsp_and_debug_are_in_the_bundle
-    Object.any_instance.expects(:system).with(bundle_install_command(update: false))
+    Object.any_instance.expects(:system).with(bundle_install_command)
     Bundler::LockfileParser.any_instance.expects(:dependencies).returns({ "ruby-lsp" => true, "debug" => true })
     run_script
     refute_path_exists(".ruby-lsp")
   end
 
+  def test_removes_ruby_lsp_folder_if_both_gems_were_added_to_the_bundle
+    Object.any_instance.expects(:system).with(bundle_install_command)
```
st0012, line 22: Not related to this PR, but I feel we could use `Kernel.system` instead and make these mocks simpler:

```rb
Kernel.expects(:system)
```

st0012 (approved):
Just a small suggestion for future improvement.

test/setup_bundler_test.rb
vinistock: I actually tried that in the beginning, but for some reason it doesn't work. Not sure why.

PR #829, opened 2023-07-20
Refactor listener registration

vinistock:
### Motivation

Allowing extensions to perform listener registration using class methods does not give the LSP enough control to deactivate extensions properly. It also doesn't allow extensions to perform store ahead of time configuration in the `Extension` instance and pass that to their listeners.

This PR proposes a refactor in how listener registration happens. Instead of registering listener classes that we instantiate from the LSP, we invert the responsibility and require the extensions to implement factory methods that will create the listener instances based on parameters. This allows extensions to use information stored in the `Extension` instance in their listeners and also allows us to easily disable extensions, since the factory methods are invoked on every request.

Example
```ruby
class MyExtension < RubyLsp::Extension
  def activate
    @important_info = 123
  end

  # This is invoked on every request. Disabling the extension would then mean just removing
  # it from the list which automatically makes us not invoke these factory methods
  def create_code_lens_listener(uri, emitter, queue)
    MyCodeLensListener.new(@important_info, uri, emitter, queue)
  end
end
```

### Implementation

There are tradeoffs with each approach, so I'd love to hear opinions on this. The idea is that we'd need to provide empty factory methods from the base `Extension` class that are overridden to do the right thing if your extension needs it. I don't love invoking a bunch of empty methods, but I couldn't come up with an alternative.

There's also the question of formatters. Those are still registered at a class level since a factory method doesn't make much sense for them. We just need to know an identifier and an instance of the formatter.

### Automated Tests

Adapted tests.

### DIFF
--- a/lib/ruby_lsp/extension.rb
+++ b/lib/ruby_lsp/extension.rb
@@ -97,8 +97,32 @@ module RubyLsp
     sig { abstract.void }
     def activate; end
 
+    # Each extension should implement `MyExtension#deactivate` and use to perform any clean up, like shutting down a
+    # child process
+    sig { abstract.void }
+    def deactivate; end
+
     # Extensions should override the `name` method to return the extension name
     sig { abstract.returns(String) }
     def name; end
+
+    # Creates a new CodeLens listener. This method is invoked on every CodeLens request
+    sig do
+      overridable.params(
+        uri: String,
+        emitter: EventEmitter,
+        message_queue: Thread::Queue,
+      ).returns(T.nilable(Listener[T::Array[Interface::CodeLens]]))
+    end
+    def create_code_lens_listener(uri, emitter, message_queue); end
+
+    # Creates a new Hover listener. This method is invoked on every Hover request
+    sig do
+      overridable.params(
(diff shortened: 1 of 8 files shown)


Morriar (approved):

lib/ruby_lsp/extension.rb
```diff
@@ -97,8 +97,32 @@ def backtraces
...
+    sig do
+      overridable.params(
+        uri: String,
+        emitter: EventEmitter,
+        message_queue: Thread::Queue,
+      ).returns(T.nilable(Listener[T::Array[Interface::CodeLens]]))
+    end
+    def create_code_lens_listener(uri, emitter, message_queue); end
+
+    # Creates a new Hover listener. This method is invoked on every CodeLens request
```
andyw8, line 119:
```suggestion
    # Creates a new Hover listener. This method is invoked on every Hover request
```

vinistock:
Good point. Changed the documentation to reflect the changes. I'll hold off merging this for a bit to get some bug fixes out before having to bump to `v0.8.0` and ship these breaking changes.

Issue #824, opened 2023-07-19
Unable to run the LSP before doing bundle install


mateusdeap:
### Ruby version

ruby 3.1.2p20 (2022-04-12 revision 4491bb740a) [arm64-darwin21]

### Description

I'm still unsure if this is a bug, but I haven't found anything in the issues relating to this.

It's a very simple thing. I was trying to configure `ruby-lsp` with sublime following the configuration describede in #559. However, sublime was complaining the language server was crashing. When I tried to run the server in the command line, manually, this is what I got:

```
/Users/mateus/.asdf/installs/ruby/3.1.2/lib/ruby/gems/3.1.0/gems/ruby-lsp-0.7.3/lib/ruby_lsp/requests/support/dependency_detector.rb:41:in `direct_dependency?': undefined method `dependencies' for nil:NilClass (NoMethodError)

        Bundler.locked_gems.dependencies.keys.grep(gem_pattern).any?
                           ^^^^^^^^^^^^^
	from /Users/mateus/.asdf/installs/ruby/3.1.2/lib/ruby/gems/3.1.0/gems/sorbet-runtime-0.5.10921/lib/types/private/methods/_methods.rb:277:in `bind_call'
	from /Users/mateus/.asdf/installs/ruby/3.1.2/lib/ruby/gems/3.1.0/gems/sorbet-runtime-0.5.10921/lib/types/private/methods/_methods.rb:277:in `block in _on_method_added'
	from /Users/mateus/.asdf/installs/ruby/3.1.2/lib/ruby/gems/3.1.0/gems/ruby-lsp-0.7.3/lib/ruby_lsp/requests/support/dependency_detector.rb:25:in `detected_test_library'
	from /Users/mateus/.asdf/installs/ruby/3.1.2/lib/ruby/gems/3.1.0/gems/sorbet-runtime-0.5.10921/lib/types/private/methods/_methods.rb:277:in `bind_call'
	from /Users/mateus/.asdf/installs/ruby/3.1.2/lib/ruby/gems/3.1.0/gems/sorbet-runtime-0.5.10921/lib/types/private/methods/_methods.rb:277:in `block in _on_method_added'
	from /Users/mateus/.asdf/installs/ruby/3.1.2/lib/ruby/gems/3.1.0/gems/ruby-lsp-0.7.3/lib/ruby_lsp/executor.rb:16:in `initialize'
[...]

PR #837, opened 2023-07-25
Prevent dependency detector from failing outside of a bundle

vinistock:
### Motivation

The issue #824 was on to something.

We want devs to be able to use the Ruby LSP on scripts and projects that don't have a `Gemfile`, but this code assumes that `locked_gems` will always exist. If there is a `Gemfile`, then I think it's fine to assume that people need to have locked their gems. But we also want to allow people to use the Ruby LSP without a `Gemfile` at all.

### Implementation

Checked if there is a top level `Gemfile` before trying to read locked gems.

### Automated Tests

Added an example. It's a bit of a weak test since it basically mocks the expected response, but it at least serves as documentation.

### DIFF
--- a/lib/ruby_lsp/requests/support/dependency_detector.rb
+++ b/lib/ruby_lsp/requests/support/dependency_detector.rb
@@ -38,7 +38,10 @@ module RubyLsp
 
       sig { params(gem_pattern: Regexp).returns(T::Boolean) }
       def direct_dependency?(gem_pattern)
-        Bundler.locked_gems.dependencies.keys.grep(gem_pattern).any?
+        Bundler.with_original_env { Bundler.default_gemfile } &&
+          Bundler.locked_gems.dependencies.keys.grep(gem_pattern).any?
+      rescue Bundler::GemfileNotFound
+        false
       end
     end
   end
--- a/test/requests/support/dependency_detector_test.rb
+++ b/test/requests/support/dependency_detector_test.rb
@@ -39,5 +39,10 @@ module RubyLsp
 
       assert_equal("rails", DependencyDetector.detected_test_library)
     end
+
+    def test_direct_dependency_returns_false_outside_of_bundle
+      File.expects(:file?).at_least_once.returns(false)
+      refute(DependencyDetector.direct_dependency?(/^ruby-lsp/))
+    end
   end
 end


lib/ruby_lsp/requests/support/dependency_detector.rb
```diff
@@ -38,7 +38,8 @@ def detected_test_library
 
       sig { params(gem_pattern: Regexp).returns(T::Boolean) }
       def direct_dependency?(gem_pattern)
-        Bundler.locked_gems.dependencies.keys.grep(gem_pattern).any?
+        File.exist?("#{Dir.pwd}/Gemfile") &&
+          Bundler.locked_gems.dependencies.keys.grep(gem_pattern).any?
```
paracycle, line 42: This does not consider the `BUNDLE_GEMFILE` env variable nor the fact that the `Gemfile` can be in a parent folder. I think we should rather use the normal `Bundler` methods for this:
```suggestion
        Bundler.default_gemfile &&
          Bundler.locked_gems.dependencies.keys.grep(gem_pattern).any?
      rescue Bundler::GemfileNotFound # Bundler.default_gemfile raises this if no Gemfile found
        nil
```

test/requests/support/dependency_detector_test.rb
```diff
@@ -39,5 +39,10 @@ def test_detects_rails_if_both_rails_and_minitest_are_present
 
       assert_equal("rails", DependencyDetector.detected_test_library)
     end
+
+    def test_direct_dependency_returns_false_outside_of_bundle
+      File.expects(:exist?).with("#{Dir.pwd}/Gemfile").once.returns(false)
```
paracycle, line 44: This should change to `File.file?` since that's what's used inside Bundler: https://github.com/rubygems/bundler/blob/master/lib/bundler/shared_helpers.rb#L239-L243

lib/ruby_lsp/requests/support/dependency_detector.rb
vinistock: This won't work because it always returns `BUNDLE_GEMFILE` if it is set ([here](https://github.com/rubygems/rubygems/blob/e2e2e7cff176ad0d25fa6e9c0d21a500217b47bd/bundler/lib/bundler/shared_helpers.rb#L215-L216)).

It's always going to be set because the LSP will either point to the custom bundle or the main bundle if the gem is in the `Gemfile`, so there's no scenario where this will raise.

paracycle: I am not sure if I understand this. If the Ruby LSP gem is in the `Gemfile`, then we don't want `Bundler.default_gemfile` to fail, since we know that the project uses a `Gemfile`. So it works in that case.

For the case that Ruby LSP is working against the custom bundle, do we not need the custom bundle to be sourcing the original Gemfile as well? In any case, maybe we could be doing all this in a `Bundler.with_original_env` block if we don't want the custom bundle targeting to be a problem.

vinistock: This fix is for when there is no top level `Gemfile`. So, yeah, `Bundler.with_original_env` does the trick in this case since we want to check the `default_gemfile` ignoring the custom bundle.

paracycle (approved):
Thanks!

Issue #838, opened 2023-07-26
OnTypeFormatting for pipe prevents to input closing block arguments


tk0miya:
### Ruby version

3.2.2

### Code snippet

```
10.times do |i|
```

### Description

When I enable the OnTypeFormatting feature, ruby-lsp will append a closing pipe when I open a block argument. But it prevents inputting a closing pipe manually.

On inserting a closing pipe manually after the last block argument, ruby-lsp will append another pipe to the code. So we'll see 3 pipes after the block argument; `|i|||`.

![ruby-lsp](https://github.com/Shopify/ruby-lsp/assets/748828/16bf40c4-79a5-409d-bc5d-91b32b90f9ce)append

### Expected output

This behavior is harmful to me. It would be better to remove needless pipes like #812 does.

PR #839, opened 2023-07-26
Do not append extra pipe if the user adds a closing pipe manually

tk0miya:
### Motivation

If the user adds a closing pipe manually, we currently end up autocompleting in both cases and end up with 3 pipes.

This commits makes it so that if the user inserted a closing pipe to the end of the block argument, we remove the pipe we added automatically and avoid adding yet another one, resulting in two pipes.

refs: #838 

### Implementation

In this scenario, when the user adds a closing pipe manually to the end of the block argument, we end up with two pipes:

1. One in the current position where on type formatting was triggered
2. One in the current position - 1
3. And the opening pipe (and some block arguments)

If we're in this situation, we can remove the pipe we added through autocomplete so that we end up with only one pipe at the end.

### Automated Tests

Added a new example.

### Manual Tests

2. Type `10.times do` in a file
3. Insert a pipe after the space
4. Verify the pipe is autocompleted
5. Insert a block argument `i`
6. Insert another pipe without moving the cursor
7. Verify that one of the extra pipes is removed and you end up with only one closing pipe

ruby-lsp


tk0miya:
I have signed the CLA!

lib/ruby_lsp/requests/on_type_formatting.rb
```diff
@@ -69,10 +69,7 @@ def handle_pipe
 
         # If the current character is a pipe and both previous ones are pipes too, then we autocompleted a pipe and the
         # user inserted a third one. In this case, we need to avoid adding a fourth and remove the previous one
-        if line[@position[:character] - 2] == "|" &&
-            line[@position[:character] - 1] == "|" &&
-            line[@position[:character]] == "|"
-
+        if line[..@position[:character]] =~ /do\s+\|[^|]*?\|\|$/
```
vinistock, line 72: Will this work for one line blocks? E.g.:
```ruby
[].each { || }
```

Also, the comments above this if statement have to be updated given that they no longer reflect the implementation.

test/requests/on_type_formatting_test.rb
```diff
@@ -154,6 +154,39 @@ def test_pipe_is_removed_if_user_adds_manually_after_completion
     assert_equal(expected_edits.to_json, T.must(edits).to_json)
   end
 
+  def test_pipe_is_removed_if_user_adds_manually_after_block_argument
+    document = RubyLsp::Document.new(source: +"", version: 1, uri: "file:///fake.rb")
```
vinistock, line 158: We made a recent change to standardize URI handling, so this needs to become
```suggestion
    document = RubyLsp::Document.new(source: +"", version: 1, uri: URI("file:///fake.rb"))
```

vinistock:
This looks great!

vinistock (approved):

tk0miya:
Thanks!

PR #847, opened 2023-08-01
Use pathnames and bundler methods in custom bundle logic

vinistock:
### Motivation

I noticed an issue with the custom bundle logic and ended up improving 3 separate things:
1. If a project had no bundle (no Gemfile) the logic didn't work at all (original problem)
2. We were concatenating strings to build paths. This is not a good strategy since different operating systems may use different ways of building paths. We should always rely on `Pathname` to do the right thing for us
3. We were again using strings for referencing the main bundle, but we should rely on methods coming from bundler. For example, if a project uses `gems.rb` instead of `Gemfile`, we would fail to set up the custom bundle
4. Bundler actually allows gems to have multiple gemspecs. We were not accounting for that when loading them

### Implementation

1. Removed the special check for the `ruby-lsp` folder. This is actually not necessary, because the Ruby LSP always falls under the scenario of having both `ruby-lsp` and `debug` in the bundle, so we can simplify the logic there
2. Started using pathnames for everything
3. Started using `Bundler.default_gemfile` and `Bundler.default_lockfile` to find the correct bundle files
5. Started returning the correct bundler configuration from `setup!` so that we can use it on the executable and avoid duplicating logic
6. Moved the check for a bundle that hasn't been locked yet inside the setup script

### Automated Tests

Added some more tests.

### DIFF
--- a/lib/ruby_lsp/setup_bundler.rb
+++ b/lib/ruby_lsp/setup_bundler.rb
@@ -4,6 +4,7 @@
 require "sorbet-runtime"
 require "bundler"
 require "fileutils"
+require "pathname"
 
 # This file is a script that will configure a custom bundle for the Ruby LSP. The custom bundle allows developers to use
 # the Ruby LSP without including the gem in their application's Gemfile while at the same time giving us access to the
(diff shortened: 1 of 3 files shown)


lib/ruby_lsp/setup_bundler.rb
```diff
@@ -13,107 +14,133 @@ module RubyLsp
...
-      @dependencies = T.let(load_dependencies, T::Hash[String, T.untyped])
-      @custom_bundle_dependencies = T.let(
-        if File.exist?(".ruby-lsp/Gemfile.lock")
-          Bundler::LockfileParser.new(Bundler.read_file(".ruby-lsp/Gemfile.lock")).dependencies
-        else
-          {}
+
+      # Custom bundle paths
+      @c_dir = T.let(Pathname.new(".ruby-lsp").expand_path(Dir.pwd), Pathname)
+      @c_bundle = T.let(@c_dir + "Gemfile", Pathname)
```
dirceu, line 25: nit: `@bundle` and `@lock` (and their `c_*` versions) would probably be more readable as `@gemfile`, `@lockfile`, and `@custom_*`.

dirceu (approved):

lib/ruby_lsp/setup_bundler.rb
vinistock: Renamed the variables!

Issue #633, opened 2023-04-13
LSP not working correctly for unsaved/temporary files in VSCode


tepsukka:
If I create a new ruby file without saving it in VSCode, the file is not being run through LSP even though it is activated, as you can see in the images below: 

![saved_file](https://user-images.githubusercontent.com/38805997/231679017-afe60a29-469b-4260-96a7-c36683c1b3bc.png)
![unsaved_file](https://user-images.githubusercontent.com/38805997/231679024-91acfbf1-81b2-4377-96b7-94eeed1dbace.png)

You can observe a similar behavior when using VSCode to compare file changes.

Information about my environment:

- Ruby LSP v0.2.4
- ruby 3.1.0p0 (2021-12-25 revision fb4df44d16) [x86_64-linux]
- Rails 7.0.4.2
- Using rbenv as version manager, the project has a bundle.
- Debian GNU/Linux 11

vinistock:
For anyone who ends up looking into this: I would start by verifying if we are receiving `textDocument/didOpen` notifications when these temporary files are created.

It's likely we don't receive it because when you create a temporary file, the language mode is set to plain text. I don't know if we should receive a notification when the language mode is changed to Ruby.

I would check to see if LSPs are supposed to receive some notification when the language mode is changed and how we should handle that.

paracycle:
I looked into this issue and found that it stems from us registering the LSP client specifically for files on disk, as explained here: https://code.visualstudio.com/api/references/document-selector

If we change our registration from `documentSelector: [{ scheme: "file", language: "ruby" }]` to `documentSelector: [{ language: "ruby" }]` or to `documentSelector: [{ scheme: "file", language: "ruby" }, { scheme: "untitled", language: "ruby" }]`, this case immediately starts working. However, the URI of the file is sent as something like `untitled:Untitled-1`, so we need to be careful when parsing URI on the `ruby-lsp` side.

vinistock:
Thinking more about this problem, I think `Document` needs to start storing a `URI` instance rather than a string.

And when we process requests in `Executor`, we should always parse the uri with `URI(uri)` and send the actual object to all requests.

PR #848, opened 2023-08-02
Use URI objects instead of strings

### LINKED ISSUE #709: Error running ruby-lsp in path with spaces

vinistock:
### Motivation

Closes #709
Closes #781

Representing URIs with strings is incorrect. We should always rely on `URI` to get the correct file paths and further we delay this change, the more entangled it might become.

There are already two issues related to this: not handling spaces properly and being unable to support unsaved files (#633).

### Implementation

The most important changes are in `Store` and `Executor`. Everything else is pretty much just adapting the types from `String` to `URI::Generic`.

One thing to note is that for unsaved files the URI is `untitled:Untitled-1`, which has path as `nil` an `opaque` as the filename (`Untitled-1`).

This is a breaking change because extensions will no longer receive a URI string, but a URI object instead.

### Automated Tests

Adapted all tests to use URI objects.

### DIFF
--- a/test/requests/path_completion_test.rb
+++ b/test/requests/path_completion_test.rb
@@ -6,7 +6,7 @@ require "test_helper"
 class PathCompletionTest < Minitest::Test
   def setup
     @message_queue = Thread::Queue.new
-    @uri = "file:///fake.rb"
+    @uri = URI("file:///fake.rb")
     @store = RubyLsp::Store.new
   end
 
@@ -35,7 +35,7 @@ class PathCompletionTest < Minitest::Test
       @store.set(uri: @uri, source: document.source, version: 1)
       RubyLsp::Executor.new(@store, @message_queue).execute({
         method: "textDocument/completion",
-        params: { textDocument: { uri: @uri }, position: end_position },
+        params: { textDocument: { uri: @uri.to_s }, position: end_position },
       }).response
     end
 
@@ -71,7 +71,7 @@ class PathCompletionTest < Minitest::Test
       @store.set(uri: @uri, source: document.source, version: 1)
       RubyLsp::Executor.new(@store, @message_queue).execute({
         method: "textDocument/completion",
-        params: { textDocument: { uri: @uri }, position: end_position },
+        params: { textDocument: { uri: @uri.to_s }, position: end_position },
       }).response
     end
 
@@ -107,7 +107,7 @@ class PathCompletionTest < Minitest::Test
       @store.set(uri: @uri, source: document.source, version: 1)
       RubyLsp::Executor.new(@store, @message_queue).execute({
         method: "textDocument/completion",
-        params: { textDocument: { uri: @uri }, position: end_position },
+        params: { textDocument: { uri: @uri.to_s }, position: end_position },
       }).response
     end
 
@@ -143,7 +143,7 @@ class PathCompletionTest < Minitest::Test
       @store.set(uri: @uri, source: document.source, version: 1)
       RubyLsp::Executor.new(@store, @message_queue).execute({
         method: "textDocument/completion",
-        params: { textDocument: { uri: @uri }, position: end_position },
+        params: { textDocument: { uri: @uri.to_s }, position: end_position },
       }).response
     end
 
@@ -171,7 +171,7 @@ class PathCompletionTest < Minitest::Test
     @store.set(uri: @uri, source: document.source, version: 1)
     RubyLsp::Executor.new(@store, @message_queue).execute({
       method: "textDocument/completion",
-      params: { textDocument: { uri: @uri }, position: end_position },
+      params: { textDocument: { uri: @uri.to_s }, position: end_position },
     }).response
   end
 
@@ -188,7 +188,7 @@ class PathCompletionTest < Minitest::Test
     @store.set(uri: @uri, source: document.source, version: 1)
     response = RubyLsp::Executor.new(@store, @message_queue).execute({
       method: "textDocument/completion",
-      params: { textDocument: { uri: @uri }, position: end_position },
+      params: { textDocument: { uri: @uri.to_s }, position: end_position },
     }).response
     assert_nil(response)
   end
--- a/bin/benchmark
+++ b/bin/benchmark
@@ -62,8 +62,8 @@ def avg_bench(method, params)
   [average, standard_deviation]
 end
 
-FILE_URI = "file://#{File.expand_path(__FILE__)}"
-base_params = { textDocument: { uri: FILE_URI } }
+FILE_URI = URI("file://#{File.expand_path(__FILE__)}")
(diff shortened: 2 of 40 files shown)


test/requests/path_completion_test.rb
```diff
@@ -35,7 +35,7 @@ def test_completion_command
       @store.set(uri: @uri, source: document.source, version: 1)
       RubyLsp::Executor.new(@store, @message_queue).execute({
         method: "textDocument/completion",
-        params: { textDocument: { uri: @uri }, position: end_position },
+        params: { textDocument: { uri: @uri.to_s }, position: end_position },
```
andyw8, line 38: (nit) Could we do the `to_s` conversion in `Executor#run` instead?

andyw8 (approved):

test/requests/path_completion_test.rb
vinistock: We always receive the URIs as strings from the server, so the `Executor` takes them and turns them into `URI` objects.

We end up needing both for tests, but outside of tests we would never receive a uri parameter that isn't a string, so. I don't think it's worth adding logic to handle both cases just.

andyw8: Ah, true, it better reflects the implementation if we stick with the strings.

PR #849, opened 2023-08-02
Test against Ruby head and use 3.2.2

vinistock:
### Motivation

We'll be able to catch more errors/changes if we continuously test against Ruby head than by continuously developing against the same preview version. Also, the preview version is sometimes causing compilation issues for certain gems.


KaanOzkan (approved):
Diff looks good although I'm not sure how beneficial it is because with `continue-on-error` it'll always be marked as 🟢 if I recall correctly. So we have to manually check the output.

PR #850, opened 2023-08-03
Remove Windows test skipping

vinistock:


lib/ruby_lsp/internal.rb
```diff
@@ -10,6 +10,7 @@
 require "cgi"
 
 require "ruby-lsp"
+require "ruby_lsp/requests/support/uri"
```
st0012, line 13: Since it's a monkey patch, I prefer putting it under somewhere like `lib/core_ext/uri.rb` to better express this fact.

lib/ruby_lsp/requests/support/uri.rb
```diff
@@ -0,0 +1,45 @@
...
+        end
+
+        build(scheme: scheme, path: escaped_path)
+      end
+    end
+
+    extend T::Sig
+
+    sig { returns(T.nilable(String)) }
+    def to_path
```
st0012, line 25: I feel this name could be easily confused with the native `path` method and doesn't convey its purpose well. How about `unescaped_path`?

egiurleo (approved):
Nice!

PR #857, opened 2023-08-08
Use hash to compare lockfiles instead of modified timestamps

vinistock:
### Motivation

As pointed out by @paracycle, when switching branches we may be in a situation where the content of the top level lockfile is different, but the modified timestamp is not greater than the custom bundle lockfile. For example, when switching to an old branch where gems are outdated.

### Implementation

A more robust implementation is to compare a digest of the contents of the lockfile, so that if there are any differences, we run the custom bundle logic again.

### Automated Tests

Adapted our existing tests to better cover modifications to the lockfile.

### DIFF
--- a/lib/ruby_lsp/setup_bundler.rb
+++ b/lib/ruby_lsp/setup_bundler.rb
@@ -5,6 +5,7 @@ require "sorbet-runtime"
 require "bundler"
 require "fileutils"
 require "pathname"
+require "digest"
 
 # This file is a script that will configure a custom bundle for the Ruby LSP. The custom bundle allows developers to use
 # the Ruby LSP without including the gem in their application's Gemfile while at the same time giving us access to the
@@ -24,6 +25,7 @@ module RubyLsp
       @custom_dir = T.let(Pathname.new(".ruby-lsp").expand_path(Dir.pwd), Pathname)
       @custom_gemfile = T.let(@custom_dir + "Gemfile", Pathname)
       @custom_lockfile = T.let(@custom_dir + "Gemfile.lock", Pathname)
+      @lockfile_hash_path = T.let(@custom_dir + "main_lockfile_hash", Pathname)
 
       # Regular bundle paths
       @gemfile = T.let(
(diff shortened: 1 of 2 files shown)


lib/ruby_lsp/setup_bundler.rb
```diff
@@ -24,6 +25,7 @@ def initialize(project_path)
       @custom_dir = T.let(Pathname.new(".ruby-lsp").expand_path(Dir.pwd), Pathname)
       @custom_gemfile = T.let(@custom_dir + "Gemfile", Pathname)
       @custom_lockfile = T.let(@custom_dir + "Gemfile.lock", Pathname)
+      @lockfile_hash = T.let(@custom_dir + "lockfile_hash", Pathname)
```
andyw8, line 28:
```suggestion
      @lockfile_hash_path = T.let(@custom_dir + "lockfile_hash", Pathname)
```
(I would think of `lockfile_hash` as the string inside the file).

andyw8: Also, perhaps name it `Gemfile.lock.hash` to indicate its correspondence?

andyw8 (approved):

lib/ruby_lsp/setup_bundler.rb
vinistock: I updated the ivar name and changed the path to `main_lockfile_hash`. I considered `Gemfile.lock.hash`, but my fear was that it would be mistakenly understood as if it was the hash for the custom lockfile, given that they're both in the same folder.

test/setup_bundler_test.rb
```diff
@@ -91,14 +84,38 @@ def test_loading_custom_bundle_dependencies_is_lazy
...
+        capture_subprocess_io do
+          Bundler.with_unbundled_env do
+            # Run bundle install to generate the lockfile
+            system("bundle install")
+
+            # Run the script once to generate a custom bundle
+            run_script
+          end
+        end
+
```
paracycle, line 103: Can we update the `mtime` of the lockfile in between these steps to ensure that it doesn't get copied when all that changes is the last modified time and not the digest? We should be able to do it with `FileUtils.touch("Gemfile.lock", mtime: Time.now + 10 * 60)`

paracycle (approved):
Thanks!

PR #863, opened 2023-08-08
Migrate document to use YARP

andyw8:
Cherry-pick of @vinistock's PR from another branch. I then updated the tests.

`bin/test test/document_test.rb` is passing

### DIFF
--- a/lib/ruby_lsp/document.rb
+++ b/lib/ruby_lsp/document.rb
@@ -102,7 +102,7 @@ module RubyLsp
 
     sig { returns(T::Boolean) }
     def parsed?
-      !@parse_result.value.nil?
+      !!@parse_result.value
     end
 
     sig { returns(Scanner) }
--- a/lib/ruby_lsp/document.rb
+++ b/lib/ruby_lsp/document.rb
@@ -100,11 +100,6 @@ module RubyLsp
       @parse_result.failure?
     end
 
-    sig { returns(T::Boolean) }
-    def parsed?
-      !!@parse_result.value
-    end
-
     sig { returns(Scanner) }
     def create_scanner
       Scanner.new(@source, @encoding)
--- a/lib/ruby_lsp/document.rb
+++ b/lib/ruby_lsp/document.rb
@@ -9,8 +9,11 @@ module RubyLsp
     RangeShape = T.type_alias { { start: PositionShape, end: PositionShape } }
     EditShape = T.type_alias { { range: RangeShape, text: String } }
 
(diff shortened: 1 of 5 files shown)


lib/ruby_lsp/document.rb
```diff
@@ -30,9 +30,12 @@ def initialize(source:, version:, uri:, encoding: Constant::PositionEncodingKind
       @uri = T.let(uri, URI::Generic)
       @unparsed_edits = T.let([], T::Array[EditShape])
       @syntax_error = T.let(false, T::Boolean)
-      @tree = T.let(SyntaxTree.parse(@source), T.nilable(SyntaxTree::Node))
```
andyw8, line 33: Since we are introducing `@parse_result`, I eliminated `@tree`, to reduce the number of state variables needed.

lib/ruby_lsp/document.rb
```diff
@@ -89,20 +92,17 @@ def parse
...
     sig { returns(T::Boolean) }
     def syntax_error?
-      @syntax_error
+      @parse_result.failure?
     end
 
     sig { returns(T::Boolean) }
     def parsed?
-      !@tree.nil?
+      !@parse_result.value.nil?
```
andyw8, line 105: @vinistock you had removed this but I re-added since some other classes still rely on it, but changed the behaviour. We can re-visit later confirm if this is the right approach, I'm just aiming to get the tests green first.

lib/ruby_lsp/document.rb
```diff
@@ -89,20 +92,17 @@ def parse
...
     sig { returns(T::Boolean) }
     def syntax_error?
-      @syntax_error
+      @parse_result.failure?
     end
 
     sig { returns(T::Boolean) }
     def parsed?
-      !@tree.nil?
+      !!@parse_result.value
```
vinistock, line 105: Let's get rid of this method, there's no value in keeping it anymore. We parse the document the moment it's instantiated and because YARP always succeeds in returning an AST, this method will always return `true` no matter what.

test/document_test.rb
```diff
@@ -324,16 +324,14 @@ def test_parsed_returns_true_when_parsed_successfully
       # frozen_string_literal: true
       puts 'hello'
     RUBY
-
-    assert_predicate(document, :parsed?)
   end
 
-  def test_parsed_returns_false_when_parsing_fails
+  def test_parsed_returns_true_when_parsing_fails
```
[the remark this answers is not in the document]
vinistock: We can get rid of this one when removing `parsed?`.

PR #867, opened 2023-08-09
Migrate sorbet support to YARP (Alex's feedback)

andyw8:
@Morriar had left a few suggestions on https://github.com/Shopify/ruby-lsp/pull/862 after it was merged.

### DIFF
--- a/lib/ruby_lsp/requests/support/annotation.rb
+++ b/lib/ruby_lsp/requests/support/annotation.rb
@@ -17,11 +17,30 @@ module RubyLsp
           @receiver = receiver
         end
 
-        sig { returns(T.any(Integer, T::Range[Integer])) }
-        attr_reader :arity
+        sig { params(node: YARP::CallNode).returns(T::Boolean) }
+        def match?(node)
+          receiver_matches?(node) && arity_matches?(node)
+        end
 
-        sig { returns(T::Boolean) }
-        attr_reader :receiver
+        private
+
+        sig { params(node: YARP::CallNode).returns(T::Boolean) }
+        def receiver_matches?(node)
+          node_receiver = node.receiver
+          (node_receiver && @receiver && node_receiver.location.slice == "T") || (!node_receiver && !@receiver)
+        end
+
+        sig { params(node: YARP::CallNode).returns(T::Boolean) }
+        def arity_matches?(node)
+          node_arity = node.arguments&.arguments&.size || 0
+
+          case @arity
+          when Integer
+            node_arity == @arity
+          when Range
+            @arity.cover?(node_arity)
+          end
+        end
       end
     end
   end
--- a/lib/ruby_lsp/requests/support/sorbet.rb
+++ b/lib/ruby_lsp/requests/support/sorbet.rb
@@ -46,33 +46,7 @@ module RubyLsp
             ).returns(T::Boolean)
           end
           def annotation?(node)
-            annotation = ANNOTATIONS[node.name]
-
-            return false unless annotation
-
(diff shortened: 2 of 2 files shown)


lib/ruby_lsp/requests/support/sorbet.rb
```diff
@@ -48,25 +49,35 @@ class << self
...
+          end
+
+          sig { params(arity: Integer, annotation_arity: T.any(Integer, T::Range[Integer])).returns(T::Boolean) }
+          def arity_matches?(arity, annotation_arity)
             case annotation_arity
             when Integer
               arity == annotation_arity
             when Range
               annotation_arity.cover?(arity)
-            else
```
andyw8, line 68: Was unreachable so I removed.

vinistock:
Is this branch properly rebased? There are some conflicts and many commits.

lib/ruby_lsp/requests/support/sorbet.rb
```diff
@@ -10,97 +10,68 @@ class << self
...
-              "requires_ancestor" => Annotation.new(arity: 0),
-              "reveal_type" => Annotation.new(arity: 1, receiver: true),
-              "sealed!" => Annotation.new(arity: 0),
-              "self_type" => Annotation.new(arity: 0, receiver: true),
-              "sig" => Annotation.new(arity: 0),
-              "type_member" => Annotation.new(arity: (0..1)),
-              "type_template" => Annotation.new(arity: 0),
-              "unsafe" => Annotation.new(arity: 1),
-              "untyped" => Annotation.new(arity: 0, receiver: true),
+              "abstract!" => { arity: 0 },
```
paracycle, line 13: I think I would still like to keep these are `Annotation` nodes, but redefine `Annotation` in this file as: `Annotation = Struct.new(:arity, :receiver, keyword_init: true)` (or as a `T::Struct`).

This would make it easier to type the `ANNOTATIONS` has and again make it possible to omit the `receiver` argument in the constructor (since `nil` is falsy). Moreover, it would change all the `[:arity]` accesses (which could return `nil`, btw, if there was a typo), into `.arity` accesses, which would be more type-safe.

lib/ruby_lsp/requests/support/sorbet.rb
```diff
@@ -8,36 +8,41 @@ class Sorbet
         class << self
           extend T::Sig
 
+          class Annotation < T::Struct
+            prop :arity, T.any(Integer, T::Range[Integer])
+            prop :receiver, T.nilable(T::Boolean)
+          end
```
vinistock, line 14: We moved away from using `T::Struct`s for performance reasons in a past PR. Can we make this just a regular Ruby object instead?

Also, just a nitpick, I'd move the class outside of the singleton class block.

andyw8: Updated. Would it be useful to have a cop to warn about `T::Struct`?

lib/ruby_lsp/requests/support/annotation.rb
```diff
@@ -0,0 +1,55 @@
...
+        def initialize(arity:, receiver: false)
+          @arity = arity
+          @receiver = receiver
+        end
+
+        sig { returns(T.any(Integer, T::Range[Integer])) }
+        attr_reader :arity
+
+        sig { returns(T::Boolean) }
+        attr_reader :receiver
```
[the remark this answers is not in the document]
andyw8: Updated to make them private.

vinistock (approved):

Morriar (approved):

Issue #449, opened 2023-01-11
Implement early adoption of YARP


vinistock:
The new Ruby parser https://github.com/Shopify/yarp will provide numerous benefits for the Ruby LSP (most notably error tolerance). We should implement a YARP mode in the Ruby LSP, allowing it to use it as a backend for parsing and to display syntax error diagnostics.

This will provide two main advantages:
1. Assisting YARP development, making it easier to visually inspect the results of parsing files
2. Begin the YARP adoption extremely early, so that as soon as it's production ready, we are all set to make the switch

Note: remember that YARP is still in development, so using it as our parser has to be an opt-in experimental feature and must not replace our defaults for now.

---

We will use the `yarp-clean ` branch for this.

Task breakdown:
- [x] Remove SyntaxTree, use YARP https://github.com/Shopify/ruby-lsp/pull/860
- [x] Migrate Sorbet support https://github.com/Shopify/ruby-lsp/pull/862
- [x] Support - document.rb https://github.com/Shopify/ruby-lsp/pull/863
- [x] Support - highlight_target.rb https://github.com/Shopify/ruby-lsp/pull/866
- [x] Request - code_action_resolve.rb https://github.com/Shopify/ruby-lsp/pull/877
- [x] Request - code_actions.rb - nothing needed
- [x] Request - code_lens.rb - see https://github.com/Shopify/ruby-lsp/compare/main...vs/migrate_to_yarp
- [x] Request - definition.rb https://github.com/Shopify/ruby-lsp/pull/864
- [x] Request - diagnostics.rb https://github.com/Shopify/ruby-lsp/pull/868
- [x] Request - document_highlight.rb - https://github.com/Shopify/ruby-lsp/pull/1005
- [x] Request - document_link.rb - https://github.com/Shopify/ruby-lsp/pull/982
- [x] Request - document_symbol.rb - https://github.com/Shopify/ruby-lsp/pull/894
- [x] Request - folding_ranges.rb - https://github.com/Shopify/ruby-lsp/pull/980
- [x] Request - formatting.rb
- [x] Request - hover.rb - https://github.com/Shopify/ruby-lsp/pull/880
- [x] Request - inlay_hints.rb https://github.com/Shopify/ruby-lsp/pull/871
- [x] Request - on_type_formatting.rb https://github.com/Shopify/ruby-lsp/pull/872
- [x] Request - path_completion.rb https://github.com/Shopify/ruby-lsp/pull/873
[...]

PR #871, opened 2023-08-09
Migrate inlay hints to yarp

andyw8:
Co-authored with @bitwise-aiden 

`bin/test test/requests/inlay_hints_expectations_test.rb` passes


lib/ruby_lsp/requests/inlay_hints.rb
```diff
@@ -36,13 +36,14 @@ def initialize(range, emitter, message_queue)
...
-        emitter.register(self, :on_rescue)
+        emitter.register(self, :on_node)
       end
 
-      sig { params(node: SyntaxTree::Rescue).void }
-      def on_rescue(node)
-        exception = node.exception
-        return unless exception.nil? || exception.exceptions.nil?
+      sig { params(node: T.nilable(YARP::Node)).void }
+      def on_node(node)
```
andyw8, line 43: Since YARP doesn't support `visit_rescue`, I'm not sure if there's a better way to do this?

andyw8: @vinistock reading https://github.com/ruby/yarp/blob/47fe0daddf2a2eab765e6b3ebff744d3fa614786/lib/yarp.rb#L152-L155, it sounds like we would need to implement a number of `visit_` methods, is that the approach we want to take?

andyw8:
Updated to use Vini's version, I had missed checking what was already done.

vinistock (approved):

PR #874, opened 2023-08-09
Add indexer

vinistock:
### Motivation

Add an indexing mechanism based on YARP now that it has been released as a gem, so that we can provide go to definition, hover, workspace symbols, autocomplete, etc.

### Implementation

By commit
- Added YARP as a dependency
- Configured RuboCop and our tests to pick up everything in the indexer folder
- Put the indexer under `lib/ruby_indexer`. Kept the structure as close as possible to a gem so that it can be extracted later


KaanOzkan (approved):

Rakefile
```diff
@@ -8,7 +8,7 @@ require "ruby_lsp/check_docs"
 Rake::TestTask.new(:test) do |t|
   t.libs << "test"
   t.libs << "lib"
-  t.test_files = FileList["test/**/*_test.rb"]
+  t.test_files = FileList["test/**/*_test.rb", "lib/ruby_indexer/test/**/*_test.rb"]
```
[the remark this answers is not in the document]
Morriar: This is how we do with components iirc

lib/ruby_indexer/lib/ruby_indexer/configuration.rb
```diff
@@ -0,0 +1,20 @@
+# typed: strict
+# frozen_string_literal: true
+
+module RubyIndexer
+  class Configuration
+    extend T::Sig
+
+    sig { returns(T::Array[String]) }
+    attr_accessor :files_to_index
```
Morriar, line 9: Is this meant to be changed by the client?

```suggestion
    attr_reader :files_to_index
```

lib/ruby_indexer/lib/ruby_indexer/configuration.rb
```diff
@@ -0,0 +1,20 @@
...
+
+    sig { void }
+    def initialize
+      files_to_index = $LOAD_PATH.flat_map { |p| Dir.glob("#{p}/**/*.rb", base: p) }
```
Morriar, line 13: What about `.ru` and `.rake` files? Should this be configurable? Maybe the constructor could take the list of extensions with a default:

```rb
    def initialize(allowed_extensions: [".rb"])
```

lib/ruby_indexer/lib/ruby_indexer/configuration.rb
```diff
@@ -0,0 +1,20 @@
...
+      files_to_index.concat(Dir.glob("#{Dir.pwd}/**/*.rb"))
+      files_to_index.reject! { |path| path.end_with?("_test.rb") }
```
Morriar, line 15: Same, it seems it should be configurable.

Maybe you can take some inspiration from https://github.com/Shopify/spoom/blob/main/lib/spoom/file_collector.rb.

lib/ruby_indexer/lib/ruby_indexer/index.rb
```diff
@@ -0,0 +1,113 @@
...
+      #  "Foo::Bar" => [#<Entry::Class>],
+      # }
+      @entries = T.let({}, T::Hash[String, T::Array[Entry]])
+
+      # Holds references to where entries where discovered so that we can easily delete them
+      # {
+      #  "/my/project/foo.rb" => ["Foo"],
+      #  "/my/project/bar.rb" => ["Foo::Bar"],
+      # }
+      @files_to_entries = T.let({}, T::Hash[String, T::Array[String]])
```
Morriar, line 25: Should this be pointing to the `entries` rather than namespaces? This would avoid the string comparison to delete each entry?

lib/ruby_indexer/lib/ruby_indexer/index.rb
```diff
@@ -0,0 +1,113 @@
...
+        @entries.delete(fully_qualified_name) if entries.empty?
+      end
+
+      @files_to_entries.delete(path)
+    end
+
+    sig { params(entry: Entry).void }
+    def <<(entry)
+      @entries[entry.name] ||= []
+      T.must(@entries[entry.name]) << entry
```
Morriar, line 46:
```suggestion
      (@entries[entry.name] ||= []) << entry
```

lib/ruby_indexer/lib/ruby_indexer/index.rb
```diff
@@ -0,0 +1,113 @@
...
+
+      @files_to_entries[entry.file_path] ||= []
+      T.must(@files_to_entries[entry.file_path]) << entry.name
```
Morriar, line 49:
```suggestion
      (@files_to_entries[entry.file_path] ||= []) << entry.name
```

lib/ruby_indexer/lib/ruby_indexer/index.rb
```diff
@@ -0,0 +1,113 @@
...
+      end
+
+      class Module < Entry
+        sig { returns(String) }
+        def short_name
+          T.must(@name.split("::").last)
+        end
+      end
+
+      class Class < Module
```
Morriar, line 109: Having `Class` inherit from `Module` may give surprising behavior:

```rb
modules = 0
classes = 0

index.entries.values.flatten do |scope|
  case scope
  when Module
    module += 1
  when Class
    classes += 1
  end
end
```

What is the case scope going to match for classes here?

If you want to merge behavior it may be better to do this:

```rb
class Scope < Entry; end
class Module < Scope; end
class Class < Scope; end
```

lib/ruby_indexer/lib/ruby_indexer/visitor.rb
```diff
@@ -0,0 +1,91 @@
...
+      end
+
+      comments
+    end
+
+    sig { params(name: String).returns(String) }
+    def fully_qualify_name(name)
+      return name if @stack.empty?
+
+      "#{@stack.join("::")}::#{name}"
```
Morriar, line 88: I wonder if names should always/never be fully qualified with a `::` prefix. What are your thoughts about this?

lib/ruby_indexer/ruby_indexer.rb
```diff
@@ -0,0 +1,36 @@
...
+      block.call(configuration)
+    end
+
+    sig { returns(Configuration) }
+    def configuration
+      @configuration ||= T.let(Configuration.new, T.nilable(Configuration))
+    end
+
+    sig { params(paths: T::Array[String]).returns(Index) }
+    def index(paths = configuration.files_to_index)
```
Morriar, line 23: That's a lot of indirection just to pass one value to the indexer.

Why not merging `Configuration` into the `RubyIndexer` and do:

```rb
indexer = RubyIndexer.new(options)
indexer.index(paths)
```

lib/ruby_indexer/lib/ruby_indexer/configuration.rb
vinistock: Yes, it needs to be an accessor so that people can do
```ruby
RubyIndexer.configure |config|
  config.files_to_index = ...
end
```

But notice that this PR is only moving the code we currently have. Improvements are coming in the next follow ups.

lib/ruby_indexer/lib/ruby_indexer/configuration.rb
vinistock: Yes, the follow up PRs will add more configuration options that allow you to add other patterns.

lib/ruby_indexer/lib/ruby_indexer/configuration.rb
vinistock: It is configurable. You can invoke `RubyIndexer.configure` and change it. But again, this PR is only moving the existing code inside, the next ones will tune this.

lib/ruby_indexer/ruby_indexer.rb
vinistock: Yes, this is done in a follow up PR.

lib/ruby_indexer/lib/ruby_indexer/index.rb
vinistock: If we had entries instead of names, how would we find the right entry to remove? We currently, take all entries for a particular class/module
```ruby
@entries[name]
```
And then remove the ones that were pushed by that specific file path.

Maybe I'm not seeing your idea.

Morriar: By comparing instances rather than strings:

```rb
my_entry = Class.new
@entries["Foo"] << my_entry
@entries_by_path["foo.rb"] << my_entry

@entries["Foo"].last.equal?(@entries_by_path["foo.rb"].last) # => true
```

lib/ruby_indexer/lib/ruby_indexer/index.rb
Morriar: Yeah `Namespace` could work. Sorbet uses `ClassOrModule` iirc.

lib/ruby_indexer/lib/ruby_indexer/visitor.rb
Morriar: As a user of the index, I want to find `Foo::Bar`:

```rb
index["Foo::Bar"] => ok
index["::Foo::Bar"] => nil
```

It's kind of surprising.

And it becomes worse if the user start playing with the content of the index:

```rb
index["Foo::Bar"] << SomeClass
# vs. 
index["::Foo::Bar"] << SomeClass
```

So I'm wondering if the index should always standardize the name when passing a namespace to `[]` or `[]=`?

lib/ruby_indexer/lib/ruby_indexer/index.rb
vinistock: Added the `Namespace` concept.

lib/ruby_indexer/lib/ruby_indexer/visitor.rb
vinistock: That's true. I started handling the names prefixed with `::` in the accessor.

As for `<<`, I'll make that private in #875 since we do not want people to manually insert entries.

Morriar (approved):

PR #880, opened 2023-08-10
Migrate hover request to YARP

andyw8:
### Motivation

### Implementation

n/a

### Automated Tests

`bin/test test/requests/hover_expectations_test.rb` is passing

### Manual Tests

n/a

### DIFF
--- a/lib/ruby_lsp/event_emitter.rb
+++ b/lib/ruby_lsp/event_emitter.rb
@@ -40,7 +40,7 @@ module RubyLsp
       when YARP::CallNode
         @listeners[:on_call]&.each { |l| T.unsafe(l).on_call(node) }
       when YARP::ConstantPathNode
-        @listeners[:on_constant_path_node]&.each { |l| T.unsafe(l).on_constant_path_node(node) }
+        @listeners[:on_constant_path]&.each { |l| T.unsafe(l).on_constant_path(node) }
       when YARP::StringNode
         @listeners[:on_string_node]&.each { |l| T.unsafe(l).on_string_node(node) }
       end
--- a/test/requests/hover_expectations_test.rb
+++ b/test/requests/hover_expectations_test.rb
@@ -99,16 +99,16 @@ class HoverExpectationsTest < ExpectationsTestRunner
 
           def initialize(emitter, message_queue)
             super
-            emitter.register(self, :on_command)
+            emitter.register(self, :on_call)
           end
 
-          def on_command(node)
+          def on_call(node)
             T.bind(self, RubyLsp::Listener[T.untyped])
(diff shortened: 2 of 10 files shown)


lib/ruby_lsp/event_emitter.rb
```diff
@@ -40,7 +40,7 @@ def emit_for_target(node)
       when YARP::CallNode
         @listeners[:on_call]&.each { |l| T.unsafe(l).on_call(node) }
       when YARP::ConstantPathNode
-        @listeners[:on_constant_path_node]&.each { |l| T.unsafe(l).on_constant_path_node(node) }
+        @listeners[:on_constant_path]&.each { |l| T.unsafe(l).on_constant_path(node) }
```
vinistock, line 43: I agree with the rename, but this event was already using YARP, which means some other request that was already migrated needs to be updated to match the new event name. It's very likely document symbol, but we need to double-check.

test/requests/hover_expectations_test.rb
```diff
@@ -76,6 +76,7 @@ class Post
       params: { textDocument: { uri: "file:///fake.rb" }, position: { line: 1, character: 2 } },
     }).response
 
+    # debugger
```
vinistock, line 79: ✂️

lib/ruby_lsp/event_emitter.rb
andyw8: Document symbol uses only `:on_constant_path_write` and `on_constant_write`:

https://github.com/Shopify/ruby-lsp/blob/yarp/lib/ruby_lsp/requests/document_symbol.rb

`on_constant_path_node` was introduced in the [Definition Request migration](https://github.com/Shopify/ruby-lsp/pull/864) but it looks like it wasn't actually necessary for that PR.

I verified that `test/requests/definition_expectations_test.rb` and `test/requests/document_symbol_expectations_test.rb` still pass.

vinistock: We definitely need that event in [definition](https://github.com/Shopify/ruby-lsp/blob/82463dc9c7ee9ca7992c2cd6f8fb24fb86d8bcb6/lib/ruby_lsp/requests/definition.rb#L50). It's possible we don't have a test for that scenario.

Don't trust the tests as perfect sources of truth because it's likely we missed other scenarios too. If you identify missing tests, feel free to push those directly to main, so that we avoid any regressions too.

andyw8: Definition was updated in https://github.com/Shopify/ruby-lsp/pull/878 but isn't yet in the YARP branch. I don't want to rebase the `yarp` branch right now, as it will affect the open PRs against that. Once everything is complete we can rebase.

PR #881, opened 2023-08-10
Raise in executor in tests

andyw8:
### Motivation

We currently rescue `StandardError` in `executor`, which covers a wide range of things, including `TypeError`. That can result in test failure messages which are not very informative.

And while working on a YARP migration with @bitwise-aiden we discovered a flawed test which was being masked due to the rescue. 

(I'll fix the linting if we agree to ahead with this).

### Implementation

Be loud if an error occurs while running the tests.

### Automated Tests

Updated

### Manual Tests

n/a

### DIFF
--- a/test/requests/path_completion_test.rb
+++ b/test/requests/path_completion_test.rb
@@ -33,7 +33,6 @@ class PathCompletionTest < Minitest::Test
     result = with_file_structure do
       @store = RubyLsp::Store.new
       @store.set(uri: @uri, source: document.source, version: 1)
-      @executor = RubyLsp::Executor.new(@store, @message_queue)
       run_request(
         method: "textDocument/completion",
         params: { textDocument: { uri: @uri.to_s }, position: end_position },
@@ -70,7 +69,6 @@ class PathCompletionTest < Minitest::Test
 
     result = with_file_structure do
       @store.set(uri: @uri, source: document.source, version: 1)
-      @executor = RubyLsp::Executor.new(@store, @message_queue)
       run_request(
         method: "textDocument/completion",
         params: { textDocument: { uri: @uri.to_s }, position: end_position },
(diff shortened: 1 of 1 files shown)


lib/ruby_lsp/executor.rb
```diff
@@ -25,6 +25,7 @@ def execute(request)
       request_time = Benchmark.realtime do
         response = run(request)
       rescue StandardError, LoadError => e
+        raise e if $RUBY_LSP_TEST
```
andyw8, line 28: Wasn't sure if there was a more 'standard' way to name such a thing.

test/requests/path_completion_test.rb
```diff
@@ -182,7 +182,7 @@ def test_completion_is_not_triggered_if_argument_is_not_a_string
 
     end_position = {
       line: 0,
-      character: document.source.rindex('"'),
+      character: document.source.rindex('o'),
```
andyw8, line 185: Theere is no `"` in the code above so this was returning `nil`.

lib/ruby_lsp/executor.rb
paracycle: I think it is better to make this just:
```suggestion
        raise if $RUBY_LSP_TEST
```
so that the original source of the exception is kept, and it doesn't look like the exception was raised here. Ref: https://www.exceptionalcreatures.com/guides/advanced-rescue-and-raise#reraising-exceptions

andyw8:
@vinistock yeah I like that approached, updated.

test/requests/path_completion_test.rb
```diff
@@ -182,19 +187,28 @@ def test_completion_is_not_triggered_if_argument_is_not_a_string
...
   end
 
   private
 
+  def run_request(method:, params: {})
+    result = @executor.execute({ method: method, params: params })
+    error = result.error
+    raise error if error
+
+    result
```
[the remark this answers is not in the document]
st0012: Yeah I agree with this 👍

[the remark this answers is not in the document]
andyw8: Updated as per suggestion.

andyw8:
Updated with @st0012's suggestion.

PR #885, opened 2023-08-13
Account for lowercase drive letter on Windows

gifvex:
### Motivation

The recent release of https://github.com/Shopify/vscode-ruby-lsp/commit/53fb04ac3c7048ab24dd52b94b4ec6bd50a8b74a made the extension work on Windows. On Windows, VS Code can spawn the terminal used by extensions with a lowecase drive letter, which causes the following error and prevents the extension from starting:
`URI::InvalidComponentError: bad component(expected absolute path component): c:/some/windows/path/to/file.rb`

### Implementation

The drive pattern introduced in #850 was switched to case-insensitive.

### Automated Tests

The tests also introduced in #850 were expanded to cover lowercase drive letters.

### DIFF
--- a/lib/core_ext/uri.rb
+++ b/lib/core_ext/uri.rb
@@ -9,7 +9,7 @@ module URI
       sig { params(path: String, scheme: String).returns(URI::Generic) }
       def from_path(path:, scheme: "file")
         # On Windows, if the path begins with the disk name, we need to add a leading slash to make it a valid URI
-        escaped_path = if /^[A-Z]:/.match?(path)
+        escaped_path = if /^[A-Z]:/i.match?(path)
           DEFAULT_PARSER.escape("/#{path}")
         else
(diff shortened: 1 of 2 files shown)


gifvex:
I have signed the CLA!

gifvex:
Update pushed; moved the CGI.unescape in to_standardized_path to happen before the regex looks for a colon.

https://github.com/Shopify/ruby-lsp/blob/fb1605944f82e3e051f78c0d4729d91b07d83e00/lib/ruby_lsp/requests/diagnostics.rb#L39-L40

The check here to run rubocop failed because the language server protocol can send escaped colons ([example @ learn.microsoft.com](https://learn.microsoft.com/en-us/visualstudio/extensibility/language-server-protocol?view=vs-2022)) and the added `/` wasn't being removed.

PR #891, opened 2023-08-15
Always use complete URI for store key

vinistock:
### Motivation

There are scenarios where we can receive the same file path with different schemes from VS Code. For example, when someone is looking at the diff editor, we receive the same path twice with two different schemes:

```
file:///foo/bar.rb
git:///foo/bar.rb
```

Because we were using the file path as the storage key, these two URIs would conflict and result in inconsistent internal state in the LSP. If you closed the git version, it would delete the file version too and that file would no longer exist (from the LSP's perspective).

### Implementation

The switch to use file paths and untitled names as the storage key was a mistake. We should always use the complete URI so that there are no conflicts and each scheme has a separate version of the document. VS Code sends duplicate requests for each scheme version, so we should respect that and treat them as separate documents.

### Automated tests

Added some examples that catch the issue.

### DIFF
--- a/lib/ruby_lsp/store.rb
+++ b/lib/ruby_lsp/store.rb
@@ -22,25 +22,23 @@ module RubyLsp
 
     sig { params(uri: URI::Generic).returns(Document) }
     def get(uri)
-      path = uri.to_standardized_path
-      return T.must(@state[T.must(uri.opaque)]) unless path
-
-      document = @state[path]
+      document = @state[uri.to_s]
       return document unless document.nil?
 
-      set(uri: uri, source: File.binread(CGI.unescape(path)), version: 0)
-      T.must(@state[path])
+      path = T.must(uri.to_standardized_path)
+      set(uri: uri, source: File.binread(path), version: 0)
+      T.must(@state[uri.to_s])
     end
 
     sig { params(uri: URI::Generic, source: String, version: Integer).void }
     def set(uri:, source:, version:)
       document = Document.new(source: source, version: version, uri: uri, encoding: @encoding)
-      @state[uri.storage_key] = document
+      @state[uri.to_s] = document
     end
 
     sig { params(uri: URI::Generic, edits: T::Array[Document::EditShape], version: Integer).void }
     def push_edits(uri:, edits:, version:)
-      T.must(@state[uri.storage_key]).push_edits(edits, version: version)
+      T.must(@state[uri.to_s]).push_edits(edits, version: version)
     end
 
     sig { void }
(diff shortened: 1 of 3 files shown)


lib/ruby_lsp/store.rb
```diff
@@ -22,25 +22,23 @@ def initialize
 
     sig { params(uri: URI::Generic).returns(Document) }
     def get(uri)
-      path = uri.to_standardized_path
-      return T.must(@state[T.must(uri.opaque)]) unless path
-
-      document = @state[path]
+      document = @state[uri.to_s]
```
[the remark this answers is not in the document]
andyw8: Just to save a little repetition in calling `to_s` each time, not a big deal.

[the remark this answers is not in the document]
vinistock: I tried it, but it ends up not producing the expected results. If a URI is parsed using `URI("path")`, it will produce a `URI::File` object. If we build it using `URI::Generic.build`, it produces a `URI::Generic` object.

In that scenario, even if both URIs have the exact same address and scheme, they will produce different values for `hash` and will not land on the same key. I think we're safer using `to_s`.

Morriar (commented):
You said we're receiving the same path twice but with two different schemes. Are we sure we want to act upon requests on a `git://`? 

Can't we just ignore all requests that are not on a `file://`?

vinistock:
I think it's worth responding to requests on `git` schemes, since it's useful to have consistent highlighting and navigation available when comparing two versions of the same file. If we only support file, then we'd have different features available on each diff side.

lib/ruby_lsp/store.rb
```diff
@@ -22,25 +22,23 @@ def initialize
...
       return document unless document.nil?
 
-      set(uri: uri, source: File.binread(CGI.unescape(path)), version: 0)
-      T.must(@state[path])
+      path = T.must(uri.to_standardized_path)
```
Morriar, line 28: Would it make sense to change this to `to_standardized_uri` and handle URIs everywhere?

Morriar: Depending on the request we use either an URI or a path, this could be confusing.

I was wondering if we shouldn't always use URIs everywhere and provides services to read files etc. from URIs.

vinistock: Indeed, depending on what's happening the LSP needs a file path or a URI. I'm not sure if we could further standardize this, but it would be a significant refactor.

I created #896 to explore that, but I'd like to move forward with the bug fix first since it's breaking fundamental parts of the LSP.

Morriar (approved):
I agree with Andy, the repeated `to_s` is unfortunate.

Related to https://github.com/Shopify/ruby-lsp/issues/896, I wonder if we should create our own abstraction on top of URI that would behaves as we need when performing the hash lookup and that could be used to read/write files from other requests.

PR #903, opened 2023-08-17
Use consistent ordering in expectation files

andyw8:
### Motivation

For consistency and readability, let's ensure the `params` are always listed before the `results`.

I've also included a script to verify this automatically for future, but I'm trying to decide where this should live, as a separate Rake task to run in CI?

### DIFF
--- a/verify_expectations_format.rb
+++ b/verify_expectations_format.rb
@@ -1,0 +1,10 @@
+require 'json'
+
+Dir["test/expectations/**/*.json"].each do |file|
+  json = JSON.parse(File.read(file))
+  params_index = json.keys.index("params")
+  result_index = json.keys.index("result")
+  if params_index && params_index > result_index
+    exit("Params should be before result in #{file}")
+  end
+end
--- a/verify_expectations_format.rb
+++ b/verify_expectations_format.rb
@@ -1,10 +1,14 @@
-require 'json'
+# typed: strict
+# frozen_string_literal: true
+
+require "json"
 
 Dir["test/expectations/**/*.json"].each do |file|
   json = JSON.parse(File.read(file))
   params_index = json.keys.index("params")
   result_index = json.keys.index("result")
   if params_index && params_index > result_index
-    exit("Params should be before result in #{file}")
+    puts "params should be before result in #{file}"
+    exit(1)
   end
 end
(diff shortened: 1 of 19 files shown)


verify_expectations_format.rb
```diff
@@ -0,0 +1,10 @@
+require 'json'
```
andyw8, line 1: I'm trying to decide where this should live, as a Rake task?

verify_expectations_format.rb
```diff
@@ -0,0 +1,14 @@
+# typed: strict
```
andyw8, line 1: (see discussion in PR description)

bitwise-aiden (approved):
Looks good. Would it make sense to have it as a bin stub?

vinistock (commented):
This might be better fixed by a JSON formatter rather than a script. For example, I see that some of the formatting is different for you, since it expands arrays to multiline.

Maybe we can add this our workspace settings (`.vscode/settings.json`) to ensure consistency with everybody working on the Ruby LSP:
```json
{
 "[json]": {
    "editor.defaultFormatter": "vscode.json-language-features"
  },
}
```
And then VS Code has a command to sort JSON keys (which would achieve the desired here, putting params above result). Can you check if there's a way to configure our workspace settings so that it does it automatically?

andyw8:
> And then VS Code has a command to sort JSON keys

TIL! (It's called `JSON: Sort Document`)

> And then VS Code has a command to sort JSON keys (which would achieve the desired here, putting params above result).

The downside of this approach is the `end` of a range will come before `start`.

vinistock:
I don't think working on a perfect solution here is worth the investment. Each request has a different response, so you'd need to define some sort of order schema to know how to order each expectation file.

Similarly to `range`, in code actions it's nicer to read the `title` before `data`. In definition locations, it's nicer to see the `uri` before the `range`. We'd need custom ordering for each part of the requests.

Just accounting for all of these feels like a lot of effort and adding a CI step that will check these and not provide any auto-correct feels like a step backwards in terms of the development experience.

IMO, either we go with something simple and will make things consistent in some aspect and not require a ton of investment or we drop the idea and rely on people ordering the JSON manually.

PR #904, opened 2023-08-17
Simplify constants' `::` prefix handling

st0012:
### Motivation

As discussed in https://github.com/Shopify/ruby-lsp/pull/893#discussion_r1296199461, we could simplify visitor's `::` handling by moving it `fully_qualify_name`.

### Implementation

I decided to handle prefix in `fully_qualify_name` instead of `Index#initialize` because `Index#initialize` wouldn't be able to deal with the name pushed to `IndexVisitor`'s stack.

`Index.<<` is also not ideal as when it's called `entry.name` would already have the prefix.

### DIFF
--- a/lib/ruby_indexer/lib/ruby_indexer/visitor.rb
+++ b/lib/ruby_indexer/lib/ruby_indexer/visitor.rb
@@ -70,10 +70,8 @@ module RubyIndexer
       return unless node.target.parent.nil? || node.target.parent.is_a?(YARP::ConstantReadNode)
 
       name = node.target.location.slice
-      fully_qualified_name = name.start_with?("::") ? name.delete_prefix("::") : fully_qualify_name(name)
-
       comments = collect_comments(node)
-      @index << Index::Entry::Constant.new(fully_qualified_name, @file_path, node.location, comments)
+      @index << Index::Entry::Constant.new(fully_qualify_name(name), @file_path, node.location, comments)
     end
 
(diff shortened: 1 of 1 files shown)


lib/ruby_indexer/lib/ruby_indexer/visitor.rb
```diff
@@ -113,9 +108,11 @@ def collect_comments(node)
...
     sig { params(name: String).returns(String) }
     def fully_qualify_name(name)
-      return name if @stack.empty?
-
-      "#{@stack.join("::")}::#{name}"
+      if @stack.empty? || name.start_with?("::")
+        name
+      else
+        "#{@stack.join("::")}::#{name}"
+      end.delete_prefix("::")
```
vinistock, line 115: We can avoid the `name.start_with?("::")` since we're always deleting the prefix anyway.
```suggestion
      normalized_name = name.delete_prefix("::")
      return normalized_name if @stack.empty?

      "#{@stack.join("::")}::#{normalized_name}"
```

st0012: The current code already always deletes the prefix. `name.start_with?("::")` is to avoid prepending `@stack`, not for the deletion.

vinistock: Sorry, I don't understand. If it always deletes the prefix, when would `name.start_with?("::")` match?

vinistock (approved):

PR #905, opened 2023-08-17
Dup ARGV before the option parser consumes it

vinistock:
### Motivation

I didn't know this, but `OptionParser` actually mutates the `ARGV` array. Because we moved it up in the executable, `ARGV` is always empty when we reach the `bundle exec` part of the script.

This means that we can't use the `--debug` flag to attach the debugger to the server. This PR just dupes `ARGV`, so that we can pass the originals forward.


vinistock:
@andyw8 great catch! Yes it does prevent mutation.

Issue #204, opened 2022-07-15
Add support for workspace symbols to the `ruby-lsp`


vinistock:
If we add the indexer in #199, then it knows which classes, modules, constants and methods are defined in a workspace, which allows us to provide the workspace symbols request.

This request is very similar to document symbols, but instead of being scoped to a single file, it allows fuzzy searching for symbols across the entire project.

Documentation: https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#workspace_symbol

PR #906, opened 2023-08-18
Add Index#fuzzy_search to search for entries based on Jaro-Winkler

vinistock:
### Motivation

First step for #204

If we use a string similarity algorithm to filter and sort results when searching the index, we can show more relevant results on top.

VS Code will actually apply another layer of filtering on top of it, so it might not behave exactly as one would expected (e.g.: it filters out matches with incorrect characters in the middle).

However, I think it may still be beneficial to have this available in the index. It can not only power workspace symbols, but also enable other tooling to search the index with a potentially incorrect query.

**Note**

When we add autocomplete, we will need to explore turning the entries into a Prefix Tree for faster access. However, it would still be nice to offer a slower fuzzy search that doesn't require exact matches. Or maybe there's a way to support that in a Prefix Tree?

### Implementation

The implementation uses the `DidYouMean::JaroWinkler` implementation. Notice that the method is called `distance`, but it's actually computing the similarity. The distance would be `1 - similarity`.

We calculate the similarity for entries based on `query` and their `name`, filter out anything with a similarity lower than 0.7 and sort it based on that similarity score.

### Automated Tests

Added a test showing how to fuzzy search the index.

### DIFF
--- a/lib/ruby_indexer/lib/ruby_indexer/index.rb
+++ b/lib/ruby_indexer/lib/ruby_indexer/index.rb
@@ -5,6 +5,9 @@ module RubyIndexer
   class Index
     extend T::Sig
 
+    # The minimum Jaro-Winkler similarity score for an entry to be considered a match for a given fuzzy search query
+    ENTRY_SIMILARITY_THRESHOLD = 0.7
+
     sig { void }
     def initialize
       # Holds all entries in the index using the following format:
@@ -48,6 +51,21 @@ module RubyIndexer
     sig { params(fully_qualified_name: String).returns(T.nilable(T::Array[Entry])) }
     def [](fully_qualified_name)
       @entries[fully_qualified_name.delete_prefix("::")]
+    end
+
+    # Fuzzy searches index entries based on Jaro-Winkler similarity. If no query is provided, all entries are returned
+    sig { params(query: T.nilable(String)).returns(T::Array[Entry]) }
+    def fuzzy_search(query)
+      return @entries.flat_map { |_name, entries| entries } unless query
+
+      normalized_query = query.gsub("::", "").downcase
+
+      results = @entries.filter_map do |name, entries|
+        similarity = DidYouMean::JaroWinkler.distance(name.gsub("::", "").downcase, normalized_query)
+        [entries, -similarity] if similarity > ENTRY_SIMILARITY_THRESHOLD
+      end
+      results.sort_by!(&:last)
+      results.flat_map(&:first)
     end
 
     # Try to find the entry based on the nesting from the most specific to the least specific. For example, if we have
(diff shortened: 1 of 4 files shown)


andyw8:
RuboCop used to depend on the `jaro_winkler` gem, then switched to use `DidYouMean` which ships with Ruby, could that be an option here? (I haven't looked into what API it offers).

https://github.com/rubocop/rubocop/pull/7931/files

lib/ruby_indexer/lib/ruby_indexer/index.rb
```diff
@@ -50,6 +50,21 @@ def [](fully_qualified_name)
...
+    # Fuzzy searches index entries based on Jaro-Winkler similarity. If no query is provided, all entries are returned
+    sig { params(query: T.nilable(String)).returns(T::Array[Entry]) }
+    def fuzzy_search(query)
+      return @entries.flat_map { |_name, entries| entries } unless query
+
+      normalized_query = query.gsub("::", "").downcase
+
+      results = @entries.filter_map do |name, entries|
+        similarity = DidYouMean::JaroWinkler.distance(name.gsub("::", "").downcase, normalized_query)
+        [entries, -similarity] if similarity > 0.7
```
Morriar, line 62: Should we extract this threshold as a constant and document why we chose this value?

vinistock: Yeah, good point. I added a constant with some comments explaining what it is.

lib/ruby_indexer/lib/ruby_indexer/index.rb
```diff
@@ -5,6 +5,9 @@ module RubyIndexer
   class Index
     extend T::Sig
 
+    # The minimum Jaro-Winkler similarity score for an entry to be considered a match for a given fuzzy search query
+    ENTRY_SIMILARITY_THRESHOLD = 0.7
```
Morriar, line 9: For my own curiosity, is it a standard threshold for other fuzzy matchers? I wonder what the typescript extension uses.

Morriar (approved):

lib/ruby_indexer/lib/ruby_indexer/index.rb
Morriar: For future reference, the [implementation used by VSCode](https://github.com/microsoft/vscode/blob/515c6b204fefa52688c8ff92757ff69d0917a36f/src/vs/base/common/fuzzyScorer.ts) is interesting. It looks similar to Jaro-Winkler with custom heuristic for special chars. May be useful for fuzzing methods with `A#foo` vs. `A::foo`.

vinistock: I spent a long time trying to find it in the repo but clearly your search skills bested mine 😂. Thanks for sharing, this is useful to know.

We ended up deciding to go with the implementation from [DidYouMean](https://github.com/Shopify/ruby-lsp/pull/906#issuecomment-1684104456) to avoid maintaining it, but I do believe that some custom tuning could potentially provide better results.

PR #907, opened 2023-08-18
Add workspace symbol

### LINKED ISSUE #204: Add support for workspace symbols to the `ruby-lsp`

vinistock:
### Motivation

Closes #204 

Add support for the workspace symbol feature. This allows the index to be fuzzy searched by users, allowing them to search for classes, modules or constants defined anywhere including gems.

### Implementation

The implementation basically invokes `Index#fuzzy_search` and then transforms the result into the expected format for the LSP. We use the same trick as we do in definition here to avoid duplicating results for Sorbet users.

### Automated Tests

Since this feature is workspace wide and not related to a single file, I don't think it makes sense to add an expectations style test. I added a regular unit test.

### Manual Tests

1. Enable `rubyLsp.enabledFeatures.workspaceSymbol` in your VS Code settings
2. Launch the LSP on this branch
3. Click CMD + T to start workspace symbol
4. Type some entity name (e.g.: `YARP::CallNode`)
5. Verify you get results from Sorbet and the Ruby LSP (RBI and gem source)


.vscode/settings.json
```diff
@@ -1,13 +1,37 @@
 {
   // Set this value to `verbose` to see the full JSON content of LSP requests and responses
-  "ruby lsp.trace.server": "messages",
+  "ruby lsp.trace.server": "off",
```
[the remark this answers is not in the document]
vinistock: This was a mistake. I changed it only for testing purposes and accidentally committed.

lib/ruby_lsp/requests/workspace_symbol.rb
```diff
@@ -0,0 +1,75 @@
...
+    # ```ruby
+    # # Searching for `Foo` will return this declaration and any others matching Foo in the project
+    # class Foo
+    # end
+    # ```
+    #
+    class WorkspaceSymbol
+      extend T::Sig
+
+      HAS_TYPECHECKER = T.let(DependencyDetector.typechecker?, T::Boolean)
```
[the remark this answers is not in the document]
vinistock: I made it a constant in dependency detector, so that it's not duplicated.

vinistock:
I accidentally force pushed without anything committed and it auto-closed the PR 🤦. Sorry about the confusion, this PR is still valid!

Morriar (approved):

lib/ruby_lsp/requests/support/dependency_detector.rb
```diff
@@ -49,5 +49,7 @@ def direct_dependency?(gem_pattern)
         false
       end
     end
+
+    HAS_TYPECHECKER = T.let(typechecker?, T::Boolean)
```
andyw8, line 53: (nit) I'd usually expect public constants to be defined before methods.

lib/ruby_lsp/requests/workspace_symbol.rb
```diff
@@ -0,0 +1,76 @@
+# typed: strict
+# frozen_string_literal: true
+
+module RubyLsp
+  module Requests
+    # ![Workspace symbol demo](../../workspace_symbol.gif)
+    #
+    # The [workspace symbol](https://microsoft.github.io/language-server-protocol/specification#workspace_symbol)
+    # request allows fuzzy searching declarations in the entire project. On VS Code, use CMD + T to search for symbols.
```
andyw8, line 9: Since `CMD` is Mac-specific, I think it we should either also include the Windows/Linux shortcuts, or not state it at all.

andyw8 (approved):

lib/ruby_lsp/requests/support/dependency_detector.rb
Morriar: You could also replace the constant by memoization in `typechecker?`.

andyw8: Yeah, that would be a nice improvement.

Issue #912, opened 2023-08-21
ruby-lsp cannot be installed on JRuby


okuramasafumi:
### Ruby version

jruby-9.4.3.0

### Description

Since `yarp` `0.6.0` is not compatible with JRuby, `ruby-lsp` now cannot be installed on it.

https://github.com/ruby/yarp/releases/tag/v0.7.0

The new versions of `yarp` will work on JRuby, so just upgrading `yarp` will solve this issue.

### Expected output

Should work on JRuby.

PR #913, opened 2023-08-21
Upgrade to YARP v0.8.0 (indexer changes)

andyw8:
### Motivation

Closes https://github.com/Shopify/ruby-lsp/issues/912

### Implementation

Handle the new `OrWriteNode` in YARP v0.8.0.

### Automated Tests

Updated

### Manual Tests

n/a

### DIFF
--- a/lib/ruby_indexer/test/constant_test.rb
+++ b/lib/ruby_indexer/test/constant_test.rb
@@ -27,8 +27,8 @@ module RubyIndexer
         end
       RUBY
 
-      assert_entry("FOO", Index::Entry::Constant, "/fake/path/foo.rb:0-0:0-8")
-      assert_entry("Bar::FOO", Index::Entry::Constant, "/fake/path/foo.rb:3-2:3-10")
+      assert_entry("FOO", Index::Entry::Constant, "/fake/path/foo.rb:0-0:0-2")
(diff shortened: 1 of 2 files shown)


lib/ruby_indexer/test/constant_test.rb
```diff
@@ -27,8 +27,8 @@ class ::Bar
         end
       RUBY
 
-      assert_entry("FOO", Index::Entry::Constant, "/fake/path/foo.rb:0-0:0-8")
```
andyw8, line 30: Previous we were referencing the whole assignment, but I think it makes more sense to only reference the constant name?

vinistock (approved):
Looks good. Let's just see what's happening with Ruby head

andyw8:
>  I believe the current implementation isn't handling &&= and some other kinds of operator assignments, should we handle that here or separately?

Actually `&&=` is not valid for constants, but I wonder if anything similar is...

vinistock:
I think we can focus this PR on fixing the breaking changes, but we should indeed consider these other assignment operators as other entries for the same constant.

andyw8:
> I think we can focus this PR on fixing the breaking changes, but we should indeed consider these other assignment operators as other entries for the same constant.

https://github.com/Shopify/ruby-lsp/issues/916

Issue #201, opened 2022-07-15
Add hover to the `ruby-lsp`


vinistock:
If https://github.com/Shopify/ruby-lsp/issues/199 works and is successful in providing decent results for finding possible method definitions, we can use the same mechanism to provide hover. This request shows documentation/information about methods and constant when hovering with the cursor.

Note: there may be conflicts between possible methods and constants. It's likely that we'll only be able to provide hover if only a single result is found in the index.

Documentation: https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#textDocument_hover

PR #914, opened 2023-08-21
Ignore magic comments and prejoin comments

vinistock:
### Motivation

In preparation for #201, I identified a couple of things that can be improved in how we index comments.
1. We should skip magic comments like `frozen_string_literal` or `typed` since those aren't actual documentation
2. I can't think of a reason why we'd want the array of comments rather than having the concatenated string already, so we might as well treat them as a full string of documentation

### Implementation

Changed `Entry#comments` to be a string and be collected as such. Also started ignoring the most common magic comments from being indexed.

### Automated Tests

Added a new example and adapted others.

### DIFF
--- a/lib/ruby_indexer/lib/ruby_indexer/visitor.rb
+++ b/lib/ruby_indexer/lib/ruby_indexer/visitor.rb
@@ -105,7 +105,12 @@ module RubyIndexer
         comment = @comments_by_line[line]
         break unless comment
 
-        comments.unshift(comment.location.slice)
+        comment_content = comment.location.slice.chomp
+        next if comment_content.match?(RubyIndexer.configuration.magic_comment_regex)
+
+        comment_content.delete_prefix!("#")
+        comment_content.delete_prefix!(" ")
+        comments.unshift(comment_content)
       end
 
       comments
--- a/lib/ruby_indexer/lib/ruby_indexer/configuration.rb
+++ b/lib/ruby_indexer/lib/ruby_indexer/configuration.rb
@@ -11,6 +11,7 @@ module RubyIndexer
         "included_gems" => Array,
         "excluded_patterns" => Array,
         "included_patterns" => Array,
+        "excluded_magic_comments" => Array,
       }.freeze,
       T::Hash[String, T.untyped],
     )
@@ -25,6 +26,22 @@ module RubyIndexer
       @included_gems = T.let([], T::Array[String])
       @excluded_patterns = T.let(["*_test.rb"], T::Array[String])
       @included_patterns = T.let(["#{Dir.pwd}/**/*.rb"], T::Array[String])
+      @excluded_magic_comments = T.let(
+        [
+          "frozen_string_literal:",
+          "typed:",
+          "compiled:",
+          "encoding:",
+          "shareable_constant_value:",
+          "warn_indent:",
+          "rubocop:",
+          "nodoc:",
+          "doc:",
+          "coding:",
+          "warn_past_scope:",
+        ],
+        T::Array[String],
+      )
     end
 
     sig { void }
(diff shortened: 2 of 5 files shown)


lib/ruby_indexer/lib/ruby_indexer/visitor.rb
```diff
@@ -5,6 +5,9 @@ module RubyIndexer
   class IndexVisitor < YARP::Visitor
     extend T::Sig
 
+    IGNORED_MAGIC_COMMENTS =
+      /^\s*#\s*(frozen_string_literal|typed|compiled|encoding|shareable_constant_value|warn_indent|rubocop|nodoc|doc):/i
```
Morriar, line 9: You're missing `coding` and `warn_past_scope` from [this list](https://github.com/ruby/ruby/blob/v3_0_0/parse.y#L8085-L8094).

A lot of the magic comments defined here are non-standard ones. I wonder if we could just ignore all comments attached to the top of the file and on the form `# <key>:<value>` rather than maintaining a list.

Morriar: > It would also be nice to not show licenses written in comments on top of files. They are usually huge and bury the actual documentation way down.

Don't we already break on empty lines when we collect the comments associated to an entity? If the license is properly separated from the comment of the first entity it should be okay:

```rb
# typed: true
# 
# Some license

# First comment
class Foo; end
```

For the `:nodoc:` etc, could we use the `index.yaml` and default values to make it more configurable?

lib/ruby_indexer/lib/ruby_indexer/index.rb
```diff
@@ -91,10 +91,10 @@ class Entry
       sig { returns(YARP::Location) }
       attr_reader :location
 
-      sig { returns(T::Array[String]) }
+      sig { returns(String) }
```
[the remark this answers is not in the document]
Morriar: We have an example in `rbi` to find `@` annotations: https://github.com/Shopify/rbi/blob/main/lib/rbi/model.rb#L98-L104.

Since we already get lines, it seems less wasteful to just keep them and call `join` when you need the full block instead of joining during the collect then calling `lines` when you need the lines.

lib/ruby_indexer/lib/ruby_indexer/visitor.rb
```diff
@@ -105,7 +122,12 @@ def collect_comments(node)
         comment = @comments_by_line[line]
         break unless comment
 
-        comments.unshift(comment.location.slice)
+        comment_content = comment.location.slice
+        next if comment_content.match?(IGNORED_MAGIC_COMMENTS)
+
+        comment_content.delete_prefix!("#")
+        comment_content.delete_prefix!(" ")
```
andyw8, line 129: nit: the regexp matches on `\s*` after the `#` but we're only removing one space?

Morriar: Removing more space may mess up with Markdown formatting:

```rb
# Here's my method documentation.
# 
# With an example:
# 
#     add(1, 2) # => 3
def add(val1, val2); end
```

lib/ruby_indexer/lib/ruby_indexer/visitor.rb
vinistock: That's a great idea. I made it configurable in the `.index.yml` file.

lib/ruby_indexer/lib/ruby_indexer/index.rb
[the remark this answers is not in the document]
vinistock: That's fair! I reverted back to an array.

Morriar (approved):

PR #915, opened 2023-08-21
Add hover for classes, modules and constants

### LINKED ISSUE #201: Add hover to the `ruby-lsp`

vinistock:
### Motivation

Closes #201

Add hover for classes, modules and constants based on what's indexed. Because Sorbet can fully handle displaying documentation on hover even for gems, I think we can completely skip the functionality if Sorbet is present.

Alternatively, we could instead drop the functionality from Tapioca to collect comments in RBIs and enable this to work like definition. I'd appreciate thoughts on this one.

### Implementation

I highly recommend reviewing by commit:
1. Removed all files related to the Rails documentation feature
2. Remove railties (only related to the Rails documentation feature)
3. Allowed passing fragments to `URI.from_path`, which is necessary for building definition links
4. Added hover based on the index

### What about the Rails documentation feature?

All that functionality will be moved to `ruby-lsp-rails`. I'm marking this PR as a breaking change so that we can easily coordinate moving the feature there, by bumping the `ruby-lsp-rails` requirement only when it's moved.

### Automated Tests

Added an example for each type of indexed symbol.

### DIFF
--- a/bin/benchmark
+++ b/bin/benchmark
@@ -24,9 +24,6 @@ require "benchmark"
 # Run signature blocks ahead of time
 T::Utils.run_all_sig_blocks
 
-# Fetch Rails documents ahead of time
-RubyLsp::Requests::Support::RailsDocumentClient.send(:search_index)
-
 # Build gem file paths ahead of time
 RubyLsp::Requests::DocumentLink.gem_paths
 


lib/ruby_lsp/executor.rb
```diff
@@ -252,16 +252,20 @@ def hover(uri, position)
...
+      target, parent, nesting = document.locate_node(
+        position,
+        node_types: Requests::Hover::ALLOWED_TARGETS,
+      )
 
-      if !Requests::Hover::ALLOWED_TARGETS.include?(target.class) &&
-          Requests::Hover::ALLOWED_TARGETS.include?(parent.class)
+      if (Requests::Hover::ALLOWED_TARGETS.include?(parent.class) &&
+          !Requests::Hover::ALLOWED_TARGETS.include?(target.class)) ||
+          (parent.is_a?(SyntaxTree::ConstPathRef) && target.is_a?(SyntaxTree::Const))
```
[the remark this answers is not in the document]
Morriar: Should we put an issue somewhere to not forget to do this cleaning once everything is migrated?

lib/ruby_lsp/requests/hover.rb
```diff
@@ -61,34 +67,52 @@ def merge_response!(other)
...
+      def find_in_index(name, node)
+        entries = @index.resolve(name, @nesting)
+        return unless entries
+
+        title = +"```ruby\n#{name}\n```"
+        definitions = []
+        content = +""
+        entries.each do |entry|
+          loc = entry.location
+          file_name = Pathname.new(entry.file_path).basename
```
st0012, line 97: I wonder if this should be a method of `Entry` 🤔

test/expectations/hover/documented_module.exp.json
```diff
@@ -0,0 +1,24 @@
...
+    "params": [
+        {
+            "line": 1,
+            "character": 7
+        }
+    ],
+    "result": {
+        "contents": {
+            "kind": "markdown",
+            "value": "```ruby\nFoo\n```\n\n**Definitions**: [fake.rb](file:///fake.rb#L2,1-3,3)\n\n\n\nThis is the documentation for Foo"
```
Morriar, line 11: Can we add a test with multiple entries for the same name?

lib/ruby_lsp/requests/hover.rb
vinistock: Good idea!

lib/ruby_lsp/requests/hover.rb
```diff
@@ -61,34 +67,52 @@ def merge_response!(other)
...
       end
 
       private
 
-      sig { params(name: String, node: SyntaxTree::Node).returns(T.nilable(Interface::Hover)) }
-      def generate_rails_document_link_hover(name, node)
-        urls = Support::RailsDocumentClient.generate_rails_document_urls(name)
-        return if urls.empty?
+      sig { params(name: String, node: SyntaxTree::Node).void }
+      def find_in_index(name, node)
```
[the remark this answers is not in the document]
vinistock: Changed it!

vinistock:
I haven't moved the old hover functionality to `ruby-lsp-rails` yet. Once it's complete, I'll share what it looks like with the two hovers.

lib/ruby_lsp/requests/hover.rb
```diff
@@ -61,34 +67,51 @@ def merge_response!(other)
...
+
+          # We always handle locations as zero based. However, for file links in Markdown we need them to be one based,
+          # which is why instead of the usual subtraction of 1 to line numbers, we are actually adding 1 to columns. The
+          # format for VS Code file URIs is `file:///path/to/file.rb#Lstart_line,start_column-end_line,end_column`
+          uri = URI::Generic.from_path(
+            path: entry.file_path,
+            fragment: "L#{loc.start_line},#{loc.start_column + 1}-#{loc.end_line},#{loc.end_column + 1}",
+          )
+
+          definitions << "[#{entry.file_name}](#{uri})"
```
Morriar, line 106: Are paths too long to be displayed here?

Morriar (approved):

PR #920, opened 2023-08-22
Temporarily disable tests for in-progress YARP migrations

andyw8:
This will allow us to verify that the requests migrated so far remain green, and make it easier to spot regressions when updating the `yarp` branch against the `main` branch.

### DIFF
--- a/test/requests/show_syntax_tree_test.rb
+++ b/test/requests/show_syntax_tree_test.rb
@@ -39,7 +39,7 @@ class ShowSyntaxTreeTest < Minitest::Test
                (6...6)
              ),
              0,
-             \"foo\"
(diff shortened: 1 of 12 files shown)


test/requests/show_syntax_tree_test.rb
```diff
@@ -39,7 +39,7 @@ def test_returns_partial_tree_if_document_has_syntax_error
                (6...6)
              ),
              0,
-             \"foo\"
```
andyw8, line 42: To ensure the linting step passes.

bitwise-aiden (approved):

andyw8:
I see there's a Windows failure, will fix that first.

PR #922, opened 2023-08-22
Make the indexed paths more specific to avoid duplications

vinistock:
### Motivation

Currently, we're indexing the entire `$LOAD_PATH`, but unfortunately that leads to some duplication scenarios.

If we're working on a gem and that gem is also installed in `Bundler.bundle_path`, it will appear twice.

Similarly, if a default gem is added to a project's bundle (e.g.: `gem "irb"`), both the entry under `Bundler.bundle_path` and the entry included in the Ruby installation will be indexed, producing duplicates.

I think we need to be more specific with the paths we index, rather than just doing the entire `$LOAD_PATH`.

### Implementation

Problem 1

Default gems are quite weird. If you inspect your Ruby installation directories, you'll discover that there are the regular gem folders as you'd find for other gems (e.g.: `some-gem-0.1.0`), but those are all empty. The files are actually all moved under `RbConfig::CONFIG["rubylibdir"]`.

In that folder, you'll find both a file for the default require and a folder for the rest of that gem's files. Example:
```
# Default require
bundler.rb
# Other files for the gem
bundler/
```

I think the best path forward is to be explicit about how we index default gems and the gems in the bundle. The code goes through everything under `RbConfig::CONFIG["rubylibdir"]` for indexing default gems and then goes through `Bundler.locked_gems` for the rest. Regular project files are still covered by the default included pattern `Dir.pwd/**/*.rb`.

Problem 2

I also noticed that I made a mistake in #907. I started checking in `Definition` and `WorkspaceSymbol` if an entry was inside the `Bundler.bundle_path` to prevent duplicates with Sorbet. Previously, we were testing it against the `WORKSPACE_URI`, but that's not adequate for when `BUNDLE_PATH` is defined, which could be inside `WORKSPACE_URI` too.

That change fixed the `BUNDLE_PATH` problem, but it then created an issue for default gems, which are always installed under `RbConfig::CONFIG["rubylibdir"]`. I fixed this problem too.

### Automated Tests

Added new tests to ensure that we don't regress.


st0012 (approved):
I made go-to-definition working as expected in IRB. Thanks for the fix 🎉

PR #927, opened 2023-08-23
Expand heredoc test for selection range

andyw8:
### Motivation

Document the current behaviour to assist with the migration to YARP.

### Implementation

In YARP, a heredoc is respresented as a `InterpolatedStringNode`. In the fixture here, `parts` corresponds to the 3 elements of the string.

### Automated Tests

Updated

### Manual Tests

n/a

### DIFF
--- a/test/expectations/selection_ranges/heredoc.exp.json
+++ b/test/expectations/selection_ranges/heredoc.exp.json
@@ -14,7 +14,7 @@
                 },
                 "end": {
                     "line": 1,
-                    "character": 12
+                    "character": 11
                 }
             },
             "parent": {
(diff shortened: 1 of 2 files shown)


vinistock:
The change looks good to me, but I don't understand the intent based on the PR description.

> Document the current behaviour to assist with the migration to YARP.

Where is this documented? And how does changing the fixture assist in the migration?

vinistock (approved):

Morriar:
Would it be useful to add a comment on the test explaining what is happening?

Morriar:
In the fixture file to explain the result we expect at the moment.

andyw8:
Fixtures are often shared by multiple tests so it may get a little noisy if we start adding comments.

One option could be use an `__END__` after the fixture, then allow fuller explanations in text below.

Morriar:
Could we add an extra fixture just to show this behavior so we can safely add a comment into it explaining why this fixture exist, then delete the fixture when the issue is fixed?

PR #930, opened 2023-08-25
Only index gem require paths

vinistock:
### Motivation

Some gems ship their test folders in the release and that can sometimes cause problems. I noticed this because of a gem that tests for bad folder names and actually names a folder as something.rb, which we identify as a file and try to index and then it blows up.

We only care about files that are intended to be required, so we should only index require_paths and not the entire gem structure. This will make indexing faster for gems that ship with their test files and will prevent failures from weird setups.

### Implementation

Instead of indexing full_gem_path/**/.rb, we're now indexing full_gem_path/{lib,app,other_require_paths}/**/*.rb.

### Automated Tests

Added a test to verify we don't try to index any gem's test folders.

### DIFF
--- a/lib/ruby_indexer/lib/ruby_indexer/configuration.rb
+++ b/lib/ruby_indexer/lib/ruby_indexer/configuration.rb
@@ -111,17 +111,17 @@ module RubyIndexer
       end
 
       # Add the locked gems to the list of files to be indexed
-      locked_gems&.each do |spec|
-        next if excluded_gems.include?(spec.name)
+      locked_gems&.each do |lazy_spec|
+        next if excluded_gems.include?(lazy_spec.name)
 
-        full_gem_path = Gem::Specification.find_by_name(spec.name).full_gem_path
+        spec = Gem::Specification.find_by_name(lazy_spec.name)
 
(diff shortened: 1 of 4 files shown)


Morriar (approved):
While I agree with the change, it doesn't seem to handle the problem of parsing directories:

> names a folder as something.rb, which we identify as a file and try to index and then it blows up.

We could find directories named like this under `lib`. Anything we can do to avoid blowing up?

vinistock:
I know someone could name a directory `something.rb`, but I don't think we should account for it for two reasons

- I think the index failing for this scenario is good as a way to not encourage this. I can't see a valid use case for naming directories with a `.rb` extension. Properly handling it legitimizes this usage
- We'll need to pay the price to check if every single path added to the list is a file just to protect ourselves from this bizarre case

I think it's not worth it.

Morriar:
> We'll need to pay the price to check if every single path added to the list is a file just to protect ourselves from this bizarre case

Can't we gracefully catch the exception and continue rather than relying on an arbitrary rule that you can't name a directory `something.rb`?

vinistock:
Yeah, rescuing is lower cost and we can do it. I added a commit to not fail when trying to index a directory.

Morriar (approved):

PR #940, opened 2023-08-28
Fix small issues with end insertion

thomasmarshall:
### Motivation

This PR fixes a couple of small issues with inserting `end` as part of on-type formatting:

- Removes the trailing whitespace after the cursor
- Ensures the cursor is placed correctly when breaking after the keyword
- Ensures the `end` is inserted on a new line rather than replacing one, when breaking after the keyword

<details>
  <summary>Before and after videos</summary>

  https://github.com/Shopify/ruby-lsp/assets/770763/64cd0888-7cba-4ff0-ba40-5f4f536861b7

  https://github.com/Shopify/ruby-lsp/assets/770763/27906b36-a1d7-4d78-b5b1-6de675cd4745

</details>

Happy to split this up if some of these fixes are more contentious than others.

### Implementation

> Removes the trailing white space after the cursor

For some reason, if we try to insert the newline character and the `end` keyword via a single edit without the extra space, [`applyEdit`](https://github.com/Shopify/vscode-ruby-lsp/blob/74e45e754aac281a22d5627c22a1a2a3e73d8b33/src/client.ts#L160) strips the indentation and the cursor ends up at the beginning of the middle line. Inserting the newline via a separate edit doesn't seem to have this problem.

> Ensures the cursor is placed correctly when breaking after the keyword

There was [some discussion](https://github.com/Shopify/ruby-lsp/pull/462#discussion_r1082642170) about whether the cursor should remain where it would normally be if the end insertion didn't happen, but it appears moving it back to where the break happened is the intended behavior. Personally, I would find it more natural to leave the cursor where it is as per the linked comment, but I think the intended behavior makes sense too.

> Ensures the `end` is inserted on a new line rather than replacing one

This just adds a newline character after the `end`.

### Automated Tests

I updated the existing tests for the trailing whitespace issue and added a new test for the cursor placement and newline fixes.

### Manual Tests

> Removes the trailing whitespace after the cursor

Start defining a method and hit return with the cursor immediately after the method name:

The current behavior is that the cursor is in the correct position but there is a trailing whitespace character:

The expected behavior after this change is that the cursor will be at the end of the new line without trailing whitespace:

> Ensures the cursor is placed correctly when breaking after the keyword

Break the line here:

The current behavior is that the cursor incorrectly moves to a position part-way through the second line:

The intended behavior is that the cursor moves back to a position where the user can continue typing the condition:

> Ensures the `end` is inserted on a new line rather than replacing one

Break the line here:

The current behavior is that the `end` replaces the blank line:

I would expect it to have inserted a new line:

### DIFF
--- a/lib/ruby_lsp/requests/on_type_formatting.rb
+++ b/lib/ruby_lsp/requests/on_type_formatting.rb
@@ -117,7 +117,7 @@ module RubyLsp
           move_cursor_to(@position[:line], @indentation + 2)
         elsif next_line.nil? || next_line.strip.empty?
           add_edit_with_text("#{indents}end", { line: @position[:line] + 1, character: @position[:character] })
-          move_cursor_to(@position[:line], @indentation + 3)
+          move_cursor_to(@position[:line] - 1, @indentation + @previous_line.size + 1)
         end
       end
 
--- a/lib/ruby_lsp/requests/on_type_formatting.rb
+++ b/lib/ruby_lsp/requests/on_type_formatting.rb
@@ -116,7 +116,7 @@ module RubyLsp
           add_edit_with_text("#{indents}end")
           move_cursor_to(@position[:line], @indentation + 2)
         elsif next_line.nil? || next_line.strip.empty?
-          add_edit_with_text("#{indents}end", { line: @position[:line] + 1, character: @position[:character] })
+          add_edit_with_text("#{indents}end\n", { line: @position[:line] + 1, character: @position[:character] })
           move_cursor_to(@position[:line] - 1, @indentation + @previous_line.size + 1)
         end
       end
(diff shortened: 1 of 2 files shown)


jacobdaddario:
Oh man, great fix. I've had to disable this because of some of the strange behavior.
> For some reason, if we try to insert the newline character and the end keyword via a single edit without the extra space, [applyEdit](https://github.com/Shopify/vscode-ruby-lsp/blob/74e45e754aac281a22d5627c22a1a2a3e73d8b33/src/client.ts#L160) strips the indentation and the cursor ends up at the beginning of the middle line. Inserting the newline via a separate edit doesn't seem to have this problem.

I looked into this and couldn't figure it out. It's awesome to see it fixed 😄

PR #945, opened 2023-08-29
Do not try to generate test command for temporary files

vinistock:
### Motivation

For unsaved files, the URI has no path and therefore there's no way to run tests defined in them. Currently, we're breaking when generating the test command.

### Implementation

We need to check if there's a path before trying to generate test commands or code lens.

### Automated Tests

Added an example that reproduces the problem.

### DIFF
--- a/lib/ruby_lsp/requests/code_lens.rb
+++ b/lib/ruby_lsp/requests/code_lens.rb
@@ -65,7 +65,7 @@ module RubyLsp
         class_name = node.constant.constant.value
         @class_stack.push(class_name)
 
-        if class_name.end_with?("Test")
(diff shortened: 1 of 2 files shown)


lib/ruby_lsp/requests/code_lens.rb
```diff
@@ -89,7 +89,7 @@ def on_def(node)
         visibility, _ = @visibility_stack.last
         if visibility == "public"
           method_name = node.name.value
-          if method_name.start_with?("test_")
+          if @path && method_name.start_with?("test_")
```
andyw8, line 92: (nit) could this be be part of the `return unless...` guard above?

andyw8 (approved):

PR #947, opened 2023-08-29
Use language_server-protocol gem's SymbolKind constants

st0012:
### Motivation

Instead of maintaining our own list of symbol kinds, we can use the language_server-protocol gem's SymbolKind constants, which will reduce a layer of conversion (symbol to integer) as well.


vinistock (approved):
Nice

PR #949, opened 2023-08-29
Make document symbol extensible

st0012:
### Motivation

By making `DocumentSymbol` request extensible, we can allow extensions to extend it. For example, users would be able to use document symbol feature to list Rails tests defined with `test "name"` too.

### Implementation

I followed the common pattern for extension for this request as well. But for tests I added a `test_extension` helper to encapsulate some common boilerplate and essential cleanups.

### Automated Tests

I added a new test case for the extension.


vinistock (approved):
Nice

Morriar (approved):

PR #951, opened 2023-08-29
Migrate PathCompletion to use the index

### LINKED ISSUE #938: Refactor path completion to use index
Depends on #937

After the index maintains the prefix tree with file paths, we can refactor path completion to not re-build the tree every time but instead use the one from the index.

vinistock:
### Motivation

Closes #938

Start providing require path autocomplete through the index instead of re-building the prefix tree on every request.

### Implementation

Exposed a way to search for require paths from the index and started using it in path completion.

### Automated Tests

Adapted our existing tests.

### Manual Tests

2. Type a require
3. Verify you get correct completion after insert the forward slash


lib/ruby_lsp/requests/path_completion.rb
```diff
@@ -22,33 +22,24 @@ class PathCompletion < Listener
...
-        @tree = T.let(RubyIndexer::PrefixTree.new(collect_load_path_files), RubyIndexer::PrefixTree)
+        @index = index
 
         emitter.register(self, :on_tstring_content)
       end
 
       sig { params(node: SyntaxTree::TStringContent).void }
       def on_tstring_content(node)
-        @tree.search(node.value).sort.each do |path|
+        @index.search_require_paths(node.value).sort!.each do |path|
```
bitwise-aiden, line 36: Are we starting to recommend in place operations to avoid unnecessary array creations?

Also will this be a hot path? Is there a way we can sort only if the contents have changed?

bitwise-aiden: Awesome, thanks for the context

bitwise-aiden (approved):

PR #953, opened 2023-08-30
Enable all Hover tests (YARP migration)

andyw8:
### Motivation

When updating to stay in sync with main, I had temporarily disabled some tests that were tricky to get passing. They are re-enabled in this PR.

Paired on with @vinistock.

### DIFF
--- a/lib/ruby_lsp/requests/hover.rb
+++ b/lib/ruby_lsp/requests/hover.rb
@@ -21,9 +21,10 @@ module RubyLsp
 
       ALLOWED_TARGETS = T.let(
         [
-          YARP::ClassNode,
-          YARP::ModuleNode,
+          YARP::CallNode,
+          YARP::ConstantReadNode,
           YARP::ConstantWriteNode,
+          YARP::ConstantPathNode,
         ],
         T::Array[T.class_of(YARP::Node)],
       )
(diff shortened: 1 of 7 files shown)


test/expectations/definition/class_reference.exp.json
```diff
@@ -8,7 +8,7 @@
                     "character": 2
                 },
                 "end": {
-                    "line": 185,
+                    "line": 187,
```
andyw8, line 11: Because of the two new lines in `lib/ruby_lsp/event_emitter.rb`.

test/expectations/hover/documented_namespaced_class.exp.json
```diff
@@ -17,7 +17,7 @@
             },
             "end": {
                 "line": 1,
-                "character": 14
```
andyw8, line 20: Because of the known off-by-one issue in YARP which will be addressed separately.

lib/ruby_lsp/requests/hover.rb
```diff
@@ -21,9 +21,10 @@ class Hover < Listener
 
       ALLOWED_TARGETS = T.let(
         [
-          YARP::ClassNode,
-          YARP::ModuleNode,
+          YARP::CallNode,
```
andyw8, line 24: I had used `ClassNode` and `ModuleNode` when rebasing against `main`, which was incorrect – it's the _reference_ to the class or module that we want the hover for.

vinistock (approved):

PR #957, opened 2023-08-30
Add autocomplete for classes, modules and constants

### LINKED ISSUE #61: Add completion for classes and modules
Although our LSP won't offer first class completion like Sorbet's LSP (since it would require complete typechecking), we can still offer some simple convenience auto completes.

Some ideas, but not an exhaustive list:
- on typing `do`, offer to complete with two options `|arg| \n end` or just `end`
- on `proc`, `lambda` or `->` offer to complete with `{ |arg| }` or just `{}`
- on `if`, offer `else`, `elsif` and complete the `end`
- offer to add `end` on other keywords, such as `while`, `until`, `for`
- on `require` or `require_relative`, offer a filtered list of possible files

Documentation: https://microsoft.github.io/language-server-protocol/specification#textDocument_completion

In addition, we can investigate how IRB completion works and evaluate if we can somehow use it from inside the Ruby LSP.

vinistock:
### Motivation

Closes #61 

Add completion for classes, modules and constants.

### Implementation

I'll explain the changes per file
1. `Executor`: when a completion request is triggered, we receive the position for the character immediately after the letter the user just typed. So I started checking if the previous character (the one just inserted) is an uppercase character and then we go down the path of completion for classes, modules and constants. Otherwise, we do the previous path completion. Also started passing the `nesting` to the completion request and adapted our capabilities
2. `Completion`: started handling `const` and `const_path_ref` to provide completion for both namespaced and regular constants. These events use the `prefix_search` added in #955 and then build the completions. I also extracted the logic to build markdown content from hover into `Common`, so that we can display the exact same documentation on both hover and completion (the documentation shows up on the side when selecting a completion)

Other than that, I just renamed the request to be more generic, using `Completion` instead of `PathCompletion`.

### Automated Tests

Added tests for both const and const path refs.

### Manual Tests

1. Open a codebase that does not use Sorbet (e.g.: rails/rails)
2. Configure the Ruby LSP to use this branch `"rubyLsp.branch": "vs/add_autocomplete"`
3. Type some uppercase letter somewhere
4. Verify you get completion options
5. Verify you can see the documentation properly when switching options (notice that in VS Code the documentation might be hidden. You need to click the little right arrow on top of a completion item to show docs)


lib/ruby_lsp/executor.rb
```diff
@@ -475,12 +475,18 @@ def completion(uri, position)
...
-      matched, parent = document.locate(
-        T.must(document.tree),
-        char_position,
-        node_types: [SyntaxTree::Command, SyntaxTree::CommandCall, SyntaxTree::CallNode],
-      )
 
+      # When the user types in the first letter of a constant name, we actually receive the position of the next
+      # immediate character. We check to see if the character is uppercase and then remove the offset to try to locate
+      # the node, as it could not be a constant
+      target_node_types = if ("A".."Z").cover?(document.source[char_position - 1])
```
andyw8, line 482: What about constants beginning with letters from other alphabets, e.g. `Á`? 😁

lib/ruby_lsp/requests/support/common.rb
```diff
@@ -74,6 +74,33 @@ def create_code_lens(node, title:, command_name:, arguments:, data:)
             data: data,
           )
         end
+
+        sig { params(title: String, entries: T::Array[RubyIndexer::Index::Entry]).returns(Interface::MarkupContent) }
+        def markdown_from_index_entries(title, entries)
+          markdown_title = +"```ruby\n#{title}\n```"
```
andyw8, line 80: Does this need the `+`? I don't see it being mutated anywhere.

lib/ruby_lsp/requests/support/common.rb
```diff
@@ -74,6 +74,33 @@ def create_code_lens(node, title:, command_name:, arguments:, data:)
...
+              fragment: "L#{loc.start_line},#{loc.start_column + 1}-#{loc.end_line},#{loc.end_column + 1}",
+            )
+
+            definitions << "[#{entry.file_name}](#{uri})"
+            content << "\n\n#{entry.comments.join("\n")}" unless entry.comments.empty?
+          end
+
+          Interface::MarkupContent.new(
+            kind: "markdown",
+            value: "#{markdown_title}\n\n**Definitions**: #{definitions.join(" | ")}\n\n#{content}",
```
andyw8, line 101: (nit) Perhaps we can write this a heredoc to be a little more readable?

test/requests/completion_test.rb
```diff
@@ -193,6 +193,65 @@ def test_completion_is_not_triggered_if_argument_is_not_a_string
...
+    RubyLsp::DependencyDetector.const_set(:HAS_TYPECHECKER, false)
+    document = RubyLsp::Document.new(source: +<<~RUBY, version: 1, uri: @uri)
+      class Foo::Bar
+      end
+
+      module Foo
+        B
+      end
+
+      Foo::B
```
Morriar, line 230: I think we should also add a `::Bar` constant to show the current behavior

lib/ruby_lsp/requests/completion.rb
```diff
@@ -0,0 +1,121 @@
...
+          Constant::CompletionItemKind::CLASS
+        when RubyIndexer::Index::Entry::Module
+          Constant::CompletionItemKind::MODULE
+        when RubyIndexer::Index::Entry::Constant
+          Constant::CompletionItemKind::CONSTANT
+        else
+          Constant::CompletionItemKind::REFERENCE
+        end
+
+        insertion_text = first_entry.name.dup
```
Morriar, line 103: Is this dup necessary? Is `name` mutable?

lib/ruby_lsp/executor.rb
vinistock: How do we get all upper case characters including the ones with accents in Ruby? I never seen this done before. And we need to return the list of trigger characters to the editor.

andyw8: With a little help from ChatGPT 😁 

```
# Define the range of Unicode characters
unicode_range = 0..0x10FFFF

# Initialize an array to store the uppercase characters
uppercase_chars = []

# Iterate over the unicode range
unicode_range.each do |i|
  char = [i].pack('U*')
  if char =~ /[[:upper:]]/
    uppercase_chars << char
  end
rescue # to avoid 'invalid byte sequence in UTF-8' for some values
  nil 
end.compact

# Output the uppercase characters
puts uppercase_chars
```

(1951 results)

vinistock: I created #990 to explore this in more detail. For now, let's ship without unicode support.

andyw8 (approved):

Morriar (approved):

lib/ruby_lsp/executor.rb
paracycle: It is usually best to ask Ruby to compile a snippet with the given input and see if Ruby treats it as a constant:
```ruby
pry> RubyVM::InstructionSequence.compile("Ã = 42", nil, nil, 0, false).to_a.dig(-1, -2, 0) == :setconstant
=> true
pry> RubyVM::InstructionSequence.compile("_FOO = 42", nil, nil, 0, false).to_a.dig(-1, -2, 0) == :setconstant
=> false
pry> RubyVM::InstructionSequence.compile("FOO = 42", nil, nil, 0, false).to_a.dig(-1, -2, 0) == :setconstant
=> true
pry> RubyVM::InstructionSequence.compile("Ҷ = 42", nil, nil, 0, false).to_a.dig(-1, -2, 0) == :setconstant
=> true
```

PR #958, opened 2023-08-31
Use parse! for OptionParser

vinistock:
### Motivation

In #905, we ended up deciding to go with `parse` instead of `parse!` because it didn't mutate `ARGV`. However, it also doesn't invoke the `on` blocks and so `ruby-lsp --version` or `ruby-lsp --branch` don't work.

### Implementation

Changed the implementation to dup the `ARGV` first and then just used `parse!`.

### Automated Tests

I think we need to extract a `CLI` class to encapsulate all of these and make it easier to test. However, I don't think it's worth the investment immediately, let's just fix this for now and think about the `CLI` later.

### DIFF
--- a/exe/ruby-lsp
+++ b/exe/ruby-lsp
@@ -3,6 +3,7 @@
 
 require "optparse"
 
+original_args = ARGV.dup
 options = {}
 parser = OptionParser.new do |opts|
   opts.banner = "Usage: ruby-lsp [options]"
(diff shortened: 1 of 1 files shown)


andyw8 (approved):
Should we report this is a bug (or documention issue) for `OptionParser`? It seems very non-obvious.

st0012 (approved):

PR #960, opened 2023-08-31
Sync `yarp` branch with `main`

andyw8:
This brings in the latest changes from main. There were few conflicts in DocumentSymbol due to #947 and #949, so I thought it best to have a review of how that was resolved.


test/requests/code_lens_expectations_test.rb
```diff
@@ -85,6 +85,26 @@ def test_bar; end
     assert_empty(response)
   end
 
+  def test_no_code_lens_for_unsaved_files
+    skip
```
andyw8, line 89: Note: Marked as `skip`

PR #961, opened 2023-08-31
Return `VOID` by default for unhandled requests

### LINKED ISSUE #882: INVALID_SERVER_MESSAGE
### Ruby version

3.2.2

### Code snippet

The exact code listed here: https://github.com/Shopify/ruby-lsp/issues/188

### Description

I see someone else in https://github.com/Shopify/ruby-lsp/issues/188 had the same issue, I also have it.

I'm using Neovide/Neovim

```
      ruby_ls = {
        enableExperimentalFeatures = true,
        formatter = 'auto',
        rubyVersionManager = 'rbenv',
      },
```
and the on_attach method I pass the code in the docs:

```
_timers = {}
local function ruby_setup_diagnostics(client, buffer)
  if require('vim.lsp.diagnostic')._enable then
    return
  end

  local diagnostic_handler = function()
    local params = vim.lsp.util.make_text_document_params(buffer)
    client.request('textDocument/diagnostic', { textDocument = params }, function(err, result)
      if err then
        local err_msg = string.format('diagnostics error - %s', vim.inspect(err))
        vim.lsp.log.error(err_msg)
      end
      if not result then
        return
      end
      vim.lsp.diagnostic.on_publish_diagnostics(
        nil,
        vim.tbl_extend('keep', params, { diagnostics = result.items }),
        { client_id = client.id }
      )
    end)
  end

  diagnostic_handler() -- to request diagnostics on buffer when first attaching

  vim.api.nvim_buf_attach(buffer, false, {
    on_lines = function()
      if _timers[buffer] then
        vim.fn.timer_stop(_timers[buffer])
      end
      _timers[buffer] = vim.fn.timer_start(200, diagnostic_handler)
    end,
    on_detach = function()
      if _timers[buffer] then
        vim.fn.timer_stop(_timers[buffer])
      end
    end,
  })
end
```

### Expected output

I would presume not an error

vinistock:
### Motivation

Closes #882

Currently, if we receive a request we don't handle, we are returning `nil` back to the editor. Based on the issue reported, not every editor handles this and some completely stop the execution of the LSP.

This is probably happening when the server makes a [request to the client to initiate progress](https://github.com/Shopify/ruby-lsp/blob/d3610077a097adaaf04d75d92e0f857ac8151876/lib/ruby_lsp/executor.rb#L514). Because this is not a notification, but a request to the client, it actually sends back a response. When we receive that response back, we should not return `nil` back to the editor, since it does not expect anything back.

Generally, I don't think the LSP should ever return anything back to the editor for requests we don't handle explicitly.

### Implementation

Added an `else` branch to our case statement to always return `VOID` for unhandled requests.

### Automated Tests

Added a test verifying the new behaviour. Also moved the `private` since there were some helper methods defined outside of that section.

### DIFF
--- a/lib/ruby_lsp/executor.rb
+++ b/lib/ruby_lsp/executor.rb
@@ -178,6 +178,8 @@ module RubyLsp
         workspace_symbol(request.dig(:params, :query))
       when "rubyLsp/textDocument/showSyntaxTree"
         show_syntax_tree(uri, request.dig(:params, :range))
+      else
+        VOID
       end
     end
 
--- a/test/executor_test.rb
+++ b/test/executor_test.rb
@@ -217,6 +217,15 @@ class ExecutorTest < Minitest::Test
     end
   end
 
+  def test_returns_void_for_unhandled_request
+    executor = RubyLsp::Executor.new(@store, @message_queue)
+
+    result = executor.execute(method: "anything/not/existing", params: {})
+    assert_same(RubyLsp::VOID, result.response)
+  end
+
+  private
+
   def with_uninstalled_rubocop(&block)
     rubocop_paths = $LOAD_PATH.select { |path| path.include?("gems/rubocop") }
     rubocop_paths.each { |path| $LOAD_PATH.delete(path) }
@@ -236,8 +245,6 @@ class ExecutorTest < Minitest::Test
   rescue NameError
     # Depending on which tests have run prior to this one, `RuboCopRunner` may or may not be defined
   end
-
-  private
 
   def stub_dependencies(rubocop:, syntax_tree:)
     dependencies = {}


Morriar (approved):
Any way we could have leveraged static typing to avoid this?

I see the return being typed as `T.untyped`.

Morriar:
> I'm not sure how we'd go about typing in this case. The method returns a different type depending on what request we received, which is why it's currently untyped.

A first option would be to use `T.any(...)` this would work better with responses such as `T::Array[Interface::WorkspaceSymbol]`:

```rb
POSSIBLE_RESPONSE_TYPES = T.type_alias { T.any(all the types we handle right now) }
```

A more involved solution would be to create type wrappers with a common ancestor:

```rb
Do the returned types have a common ancestor we could use?

If not, we could add one ourselves:

```rb
module ResponseType
  interface!
end

class ::Interface::SemanticTokens
  include ResponseType
end
```

I wonder if we could handle the arrays the same way:

```rb
class WorkspaceSymbols < Array
  include ResponseType

  Elem = type_member{{fixed: Interface::WorkspaceSymbol}}
end
```

And interesting case would be to avoid implicit `nil` returns to avoid mistake, maybe with something like this:

```rb
class NilReponse < NilClass
  include ResponseType
end
```

So it would force us to use `return NilResponse.new` rather than just `return` and catch branches where we just forgot to return something.

vinistock:
There's no common ancestor unfortunately. We could try using the `T.any`, but I'm not sure it would make much of a difference in terms of catching errors.

The untyped response is only used when writing it to STDOUT to send it back to the editor [here](https://github.com/Shopify/ruby-lsp/blob/main/lib/ruby_lsp/server.rb#L145-L168).

PR #965, opened 2023-09-01
Use Spoom backtrace filter

andyw8:
### Motivation

Use the new Sorbet backtrace filter added in https://github.com/Shopify/spoom/pull/457

### Implementation

`bundle update spoom` then update `test_helper.rb` and `require.rb`

### Automated Tests

n/a

### Manual Tests

n/a


Morriar:
One question though, currently Spoom as a runtime dependency on `sorbet-static`, are we sure we want to import this transitive dependency to the Ruby LSP?

We _could_ make Spoom not depend on `sorbet-static` in its Gemfile and just assume it would be here, but that causes problems regarding versions requirements that we would need to address first.

vinistock (approved):

andyw8:
I don't understand why it's failing on Window...

```
D:/a/ruby-lsp/ruby-lsp/test/test_helper.rb:15:in `require': cannot load such file -- spoom/backtrace_filter/minitest (LoadError)
````

**Comment by vinistock** [comment]
Because [Spoom is not available on Windows](https://github.com/Shopify/ruby-lsp/blob/59b19990639e69d92669134f60486b4d5de3f3b1/Gemfile#L23). That will need to be a conditional require.

File: test/test_helper.rb [hunk 1]
```diff
@@ -29,15 +35,6 @@ class Test
...
-  extend T::Sig
-
-  sig { override.params(bt: T.nilable(T::Array[String])).returns(T::Array[String]) }
-  def filter(bt)
-    super.select do |line|
-      SORBET_PATHS.none? { |path| line.include?(path) }
-    end
-  end
+if defined? Spoom::BacktraceFilter::Minitest
+  Minitest.backtrace_filter = Spoom::BacktraceFilter::Minitest.new
```
vinistock, line 39: How about avoiding two checks and just doing everything in the same block?

```ruby
begin
  require "spoom/backtrace_filter/minitest"
  Minitest.backtrace_filter = Spoom::BacktraceFilter::Minitest.new
rescue LoadError
  # Tapioca (and thus Spoom) is not available on Windows
end
```

PR #971, opened 2023-09-05
Run Tapoica gem verify prior to release

andyw8:
### Motivation

As discussed with @Morriar.

Outdated RBIs could mean that typechecking failures are missed. But updating the RBIs every time a gem is updated is tedious. As a middle-ground, let's verify just prior to release.

### Implementation

Add a step to the release script. 

### Automated Tests

n/a

### Manual Tests

This is a bit awkward to verify without actually making a release - I suggest that we observe it next time we do a release, since it is low risk.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -13,6 +13,15 @@ jobs:
       - uses: actions/checkout@v3
         name: Checkout
 
+      - name: Set up Ruby
+        uses: ruby/setup-ruby@v1
+        with:
+          ruby-version: 3.2
+          bundler-cache: true
+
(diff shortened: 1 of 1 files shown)


andyw8:
(if this is agreed to, I'll add a step to the release instructions for the `tapioca gem` update).

andyw8:
If https://github.com/Shopify/ruby-lsp/pull/985 is merged then I think this will still useful as a safety check before release (it shouldn't ever fail to verify).

vinistock:
Given that we have the automatic RBI updates and this should never fail, what's the benefit of having the step?

Is it worth the risk of potentially delaying a release for any unexpected issues?

Issue #808, opened 2023-07-10
Allow more requests to be extended


vinistock:
In #782, we encapsulated the logic to execute multiple listeners in the base implementations. Let's generalize this even more to be implemented in the parent `Listener` class and let's expose a configuration to allow requests to be extended.

Something like this
```ruby
class Hover < Listener
  extensible!
end
```

Then let's allow the following requests to be enhanced by extensions:
- semantic highlighting
- document symbols - https://github.com/Shopify/ruby-lsp/pull/949
- definition - https://github.com/Shopify/ruby-lsp/pull/989

st0012:
Today at the DX stream meeting we talked about that on some requests, we may need to expose more info or even make certain data mutable to our extensions. I think semantic highlight may be one of them.

Consider this:

```rb
class Foo < ApplicationRecord
  before_save :bar
end
```

Currently `before_save` is marked as a `method` with no modifiers by `ruby-lsp`. If `ruby-lsp-rails` wants to enhance it by assigning modifiers (e.g. declaration), the current way of merging response won't be able to achieve it.

@vinistock can you check if my understanding on this case is correct. And if it is, should we still make it extensible before we discuss further about the data flow between ruby-lsp/extensions?

vinistock:
Yes, your understanding is correct. We'll indeed need to think of a better strategy if two extensions want to modify the same semantic token. Let's put this on pause while we plan of a better abstraction for merging responses from extensions.

st0012:
Waiting for https://github.com/Shopify/ruby-dev-exp-issues/issues/1050

PR #972, opened 2023-09-05
Make requests easier to extend

st0012:
### Motivation

This is part of https://github.com/Shopify/ruby-lsp/issues/808

Currently, to make a listener-based request extensible, we need to:

- Update its constructor to initialise an array of external listeners
- Implement a few interfaces like `merge_response!` and `merge_external_listeners_responses!`
- Remember to call `merge_external_listeners_responses!` in the `executor.rb` before accessing the response

It's not easy to remember all of them at the same time. And when it's not implemented correctly, we'd rely on runtime information to figure out what's missing.

So with this change, I want to utilise Sorbet as much as possible, while reducing the APIs that need to be implemented.

### Implementation

- When extending a listener class, we change its parent class from `Listener` to the new `ExtensibleListener` class. Sorbet would then require a few more interfaces on its child class than `Listener`'s.
- To make `ExtensibleListener#response` merge external responses automatically, I changed `Listener`'s response interface to `_response`. This will prevent individual listener classes from overriding the `response` method when declaring its `ResponseType`.
- I also changed a few utility scripts (e.g. doc check) as `ExtensibleListener` is `Listener`'s child class but is abstract.

One shortfall of this approach is that if `initialize_external_listener` uses ivars to initialise external listeners, `super` needs to be called **after** the ivars are initialised. And this could not be type-checked. (Ideally we should prepend a module for such order-dependant requirement, but it's not supported by Sorbet)

### Automated Tests

All existing tests should pass

### DIFF
--- a/lib/ruby_lsp/listener.rb
+++ b/lib/ruby_lsp/listener.rb
@@ -20,10 +20,15 @@ module RubyLsp
       @message_queue = message_queue
     end
 
+    sig { returns(ResponseType) }
+    def response
+      _response
+    end
+
     # Override this method with an attr_reader that returns the response of your listener. The listener should
     # accumulate results in a @response variable and then provide the reader so that it is accessible
     sig { abstract.returns(ResponseType) }
-    def response; end
+    def _response; end
   end
 
   # ExtensibleListener is an abstract class to be used by requests that accept extensions.
@@ -58,9 +63,9 @@ module RubyLsp
     end
 
     sig { returns(ResponseType) }
-    def merged_response
+    def response
       merge_external_listeners_responses! unless @response_merged
-      response
+      super
     end
 
     sig do
--- a/lib/ruby_lsp/listener.rb
+++ b/lib/ruby_lsp/listener.rb
@@ -18,7 +18,6 @@ module RubyLsp
     def initialize(emitter, message_queue)
       @emitter = emitter
       @message_queue = message_queue
-      @external_listeners = T.let([], T::Array[RubyLsp::Listener[ResponseType]])
     end
 
     # Override this method with an attr_reader that returns the response of your listener. The listener should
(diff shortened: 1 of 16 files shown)


lib/ruby_lsp/listener.rb
```diff
@@ -18,13 +18,39 @@ class Listener
     def initialize(emitter, message_queue)
       @emitter = emitter
       @message_queue = message_queue
-      @external_listeners = T.let([], T::Array[RubyLsp::Listener[ResponseType]])
+    end
+
+    sig { returns(ResponseType) }
+    def response
+      _response
```
vinistock, line 25: Can we use `alias` or `alias_method` instead?

lib/ruby_lsp/listener.rb
```diff
@@ -18,13 +18,39 @@ class Listener
...
+  class ExtensibleListener < Listener
+    extend T::Sig
+    extend T::Generic
+
+    ResponseType = type_member
+
+    abstract!
+
+    sig { params(emitter: EventEmitter, message_queue: Thread::Queue).void }
+    def initialize(emitter, message_queue)
```
vinistock, line 44: Let's a comment here calling out the fact that `super` needs to be invoked last because it depends on instance variables being initialized. The new Sorbet version is already on main, so hovering over `super` will show docs.

test/requests/code_lens_expectations_test.rb
```diff
@@ -127,8 +127,10 @@ def name
       end
 
       def create_code_lens_listener(uri, emitter, message_queue)
+        raise "uri can't be nil" unless uri
```
vinistock, line 130: Let's leverage Sorbet for this instead. We can just add a signature and we shouldn't need this check.

vinistock (commented):
Nice. This will really help ensuring consistency in extensible requests.

test/requests/code_lens_expectations_test.rb
st0012: I've tried it but maybe I was doing something wrong because it's like wrestling with Sorbet with little benefit due to the dynamic class definition:

1. We need to add `extend T::Sig` to `CodeLensExpectationsTest` instead of the extension test class.
2. In this case, we need to add `override` to the signature of `create_code_lens_listener`. But Sorbet would complain that there's nothing to override about, which I think is also due to dynamic class definition?
    Screenshot 2023-09-07 at 12 20 23
3. If there's any runtime type checking error, it will be swallowed in `RubyLsp::Result` instead of being raised

    Screenshot 2023-09-07 at 12 21 31

So eventually I decided that it's not worth the effort here.

PR #974, opened 2023-09-05
Correct create_document_symbol_listener's signature

st0012:
The listener should be expected to return an array of `DocumentSymbol`, instead of a nilable one.

I tried to capture by making tests `typed: true`, but currently we can only create the test extension class with `Class.new(RubyLsp::Extension)`, which seems to Sorbet's ability to type check the method in tests.

### DIFF
--- a/lib/ruby_lsp/extension.rb
+++ b/lib/ruby_lsp/extension.rb
@@ -130,7 +130,7 @@ module RubyLsp
       overridable.params(
         emitter: EventEmitter,
         message_queue: Thread::Queue,
-      ).returns(T.nilable(Listener[T.nilable(Interface::DocumentSymbol)]))
+      ).returns(T.nilable(Listener[T::Array[Interface::DocumentSymbol]]))
     end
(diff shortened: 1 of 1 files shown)


lib/ruby_lsp/extension.rb
```diff
@@ -130,7 +130,7 @@ def create_hover_listener(emitter, message_queue); end
       overridable.params(
         emitter: EventEmitter,
         message_queue: Thread::Queue,
-      ).returns(T.nilable(Listener[T.nilable(Interface::DocumentSymbol)]))
+      ).returns(T.nilable(Listener[T::Array[Interface::DocumentSymbol]]))
```
vinistock, line 133: I wonder if we should use `Requests::DocumentSymbol::ResponseType` instead to prevent any inconsistencies again.

vinistock: Oh, okay. Yeah, that makes sense, you shouldn't have access to type variables outside of their scope.

vinistock (approved):

PR #976, opened 2023-09-06
Add fixture for class reference test

andyw8:
### Motivation

Currently we test against Ruby LSP's own code, which means a change to the implementation which alters the length of the class will break the test.

### Implementation

Introduce a new fixture.

### Automated Tests

Updated

### Manual Tests

n/a


vinistock (approved):
Nice. Thanks for adding those! 🚀

PR #977, opened 2023-09-06
Increase telemetry time allowance in CI

andyw8:
### Motivation

Reduce intermittent CI failures such as [this](https://github.com/Shopify/ruby-lsp/actions/runs/6099767290/job/16552664051).

### Implementation

Increase the timeout in CI.

### Automated Tests

n/a

### Manual Tests

n/a


vinistock (approved):

test/integration_test.rb
```diff
@@ -387,7 +387,7 @@ def assert_telemetry(request, expected_uri = @uri.path.sub(Dir.home, "~"))
     assert_equal(expected_uri, telemetry_response.dig(:params, :uri))
     assert_equal(RubyLsp::VERSION, telemetry_response.dig(:params, :lspVersion))
     assert_equal(request, telemetry_response.dig(:params, :request))
-    assert_in_delta(0.5, telemetry_response.dig(:params, :requestTime), 2)
+    assert_in_delta(0.5, telemetry_response.dig(:params, :requestTime), ENV["CI"] ? 5 : 2)
```
Morriar, line 390: This is a strange usage of delta? We expect the query to be processed in 0.5s +/- 5s?

vinistock: That's a good point. The idea was just to check that this was present and that the value made sense, but with all of the CI variation this assertion is making basically no sense.

What do people think about changing this to be
```ruby
assert_instance_of(Float, telemetry_response.dig(:params, :requestTime))
```
Just to check that there's some float value in there.

Morriar: Yeah, I think the `assert_instance_of` will show the intent better 👍

vinistock (approved):

PR #978, opened 2023-09-06
Add YARP RBI annotations

vinistock:
### Motivation

I created a scrappy YARP template to generate RBI annotations for the entire tree. It's not perfect and it required some manual adjustments, but this includes types for most of the tree and will help a lot in the migration.


vinistock:
This is the [branch](https://github.com/ruby/yarp/compare/main...vinistock:yarp:vs/add_rbi_template) where I added the template.

After we're done with migrating to YARP, we can take a look at polishing it and making it a part of YARP itself.

PR #980, opened 2023-09-06
Migrate folding range to YARP

vinistock:
### Motivation

Migrate folding range to YARP.

### Implementation

Folding range was one of the first requests we implemented and it was quite messy. I took the opportunity to greatly simplify how it works, completely removing the concept of `PartialRange` and simplifying how we handle call nodes. I'll add comments on the code to make it clearer.

### Automated Tests

Had to adjust some tests.


lib/ruby_lsp/requests/folding_ranges.rb
```diff
@@ -18,276 +18,154 @@ module Requests
...
-            location = node.location
-            add_lines_range(location.start_line, location.end_line - 1)
-          end
-        when SyntaxTree::DefNode
-          add_def_range(node)
-        when SyntaxTree::StringConcat
-          add_string_concat(node)
-          return
+      sig { void }
+      def push_comment_ranges
```
vinistock, line 40: Comments aren't nodes in YARP, so we need to go through them separately.

lib/ruby_lsp/requests/folding_ranges.rb
```diff
@@ -18,276 +18,154 @@ module Requests
...
-        )
-      end
-
       sig { params(document: Document).void }
       def initialize(document)
         super
 
         @ranges = T.let([], T::Array[Interface::FoldingRange])
-        @partial_range = T.let(nil, T.nilable(PartialRange))
+        @requires = T.let([], T::Array[YARP::CallNode])
```
vinistock, line 26: Previously, we needed partial ranges for comments and requires. Now that comments are no longer nodes, we only need it for requires.

We basically accumulate require call nodes into this array until we find anything that isn't a require. Then we push a range for all of those require nodes.

lib/ruby_lsp/requests/folding_ranges.rb
```diff
@@ -18,276 +18,154 @@ module Requests
...
+      sig { override.params(node: T.nilable(YARP::Node)).void }
+      def visit(node)
+        emit_requires_range unless node.is_a?(YARP::CallNode)
 
-        if kind.nil?
-          emit_partial_range
-          return true
-        end
+        case node
+        when YARP::ArrayNode, YARP::BlockNode, YARP::CaseNode, YARP::ClassNode, YARP::ForNode, YARP::HashNode,
```
vinistock, line 73: I got rid of the constants listing all nodes because that makes it impossible for Sorbet to understand and help with typing due to the splat operator.

Putting them directly here means we get correct type and need fewer annotations.

test/expectations/folding_ranges/call_chained.exp.json
```diff
@@ -1,5 +1,25 @@
 {
   "result": [
+    {
```
vinistock, line 3: We had complicated logic to prevent all of these nodes from `0` to something. However, VS Code just ignores all of them and considers only the widest range to apply.

It's definitely not worth keeping all of the complex logic we had since it's already taken care by the client.

test/expectations/folding_ranges/def_multiline_params.exp.json
```diff
@@ -1,5 +1,10 @@
 {
   "result": [
+    {
```
vinistock, line 3: We weren't emitting a range for multiline parameters, which is wrong.

test/expectations/folding_ranges/defs_multiline_params.exp.json
```diff
@@ -1,5 +1,10 @@
 {
   "result": [
+    {
+      "startLine": 0,
```
vinistock, line 4: Same here, we were missing the multiline parameters range.

lib/ruby_lsp/requests/folding_ranges.rb
bitwise-aiden: Just to confirm, if you had some requires separated by a comment, that would be two require ranges?

vinistock: Yes, that's right. If we find any nodes that aren't a require, we consider that to be the end of the require range and start a new one if necessary.

lib/ruby_lsp/requests/folding_ranges.rb
vinistock: The new node in YARP for heredocs is the `InterpolatedStringNode`. It needs to be handled in its own branch because in YARP the location of the heredoc is not based on the opening and closing delimiters.

lib/ruby_lsp/requests/folding_ranges.rb
```diff
@@ -18,276 +18,154 @@ module Requests
...
-
-        sig { returns(Interface::FoldingRange) }
-        def to_range
-          Interface::FoldingRange.new(
-            start_line: @start_line,
-            end_line: @end_line,
-            kind: @kind,
+      sig { void }
+      def emit_requires_range
+        if @requires.length > 1
```
andyw8, line 57: Since we have `next if chunk.length == 1` above, is this check necessary?

lib/ruby_lsp/requests/folding_ranges.rb
```diff
@@ -18,276 +18,154 @@ module Requests
...
-        params = node.params
-        return unless params
+      sig { params(node: YARP::CallNode).returns(T::Boolean) }
+      def require?(node)
+        message = node.message
+        return false unless message == "require" || message == "require_relative"
 
-        params_location = params.location
+        receiver = node.receiver
+        return false unless receiver.nil? || receiver.slice == "Kernel"
```
andyw8, line 138: Why was the `Kernel` check not needed previously?

test/expectations/folding_ranges/call_chained.exp.json
andyw8: I wonder if other editors behave the same...

lib/ruby_lsp/requests/folding_ranges.rb
```diff
@@ -18,276 +18,154 @@ module Requests
...
-        return if statements.empty?
+      sig { params(node: YARP::Node, other: YARP::Node).returns(T::Boolean) }
+      def same_lines?(node, other)
+        loc = node.location
+        other_loc = other.location
+
+        loc.start_line == other_loc.start_line && loc.end_line == other_loc.end_line
+      end
+
+      sig { params(node: T.any(YARP::IfNode, YARP::InNode, YARP::RescueNode, YARP::WhenNode)).void }
```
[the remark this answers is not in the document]
vinistock: Do you mean in the `when` statement? You can't use a type alias in a case statement, so that wouldn't work.

test/expectations/folding_ranges/call_chained.exp.json
vinistock: I'm not sure, but also can't think of an alternative behaviour for editors. What would they do in the case of conflicting ranges?

lib/ruby_lsp/requests/folding_ranges.rb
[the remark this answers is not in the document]
andyw8: Ah, TIL.

andyw8 (approved):

PR #982, opened 2023-09-07
Migrate DocumentLink request to YARP

andyw8:
### Motivation

### Implementation

The previous approach was based on parsing the document for comments, but with YARP we don't have a Comment node, so instead we visit the critical nodes and then check for attached comments.

### Automated Tests

I added extra examples to the existing fixtures.

### DIFF
--- a/lib/ruby_lsp/document.rb
+++ b/lib/ruby_lsp/document.rb
@@ -37,7 +37,7 @@ module RubyLsp
       @parse_result.value
     end
 
-    # sig { returns(YARP::ProgramNode) }
+    sig { returns(T::Array[YARP::Comment]) }
     def comments
       @parse_result.comments
     end
--- a/lib/ruby_lsp/document.rb
+++ b/lib/ruby_lsp/document.rb
@@ -30,6 +30,7 @@ module RubyLsp
       @uri = T.let(uri, URI::Generic)
       @unparsed_edits = T.let([], T::Array[EditShape])
       @parse_result = T.let(YARP.parse(@source), YARP::ParseResult)
+      @parse_result.attach_comments!
     end
 
     sig { returns(YARP::ProgramNode) }
--- a/lib/ruby_lsp/document.rb
+++ b/lib/ruby_lsp/document.rb
@@ -30,12 +30,16 @@ module RubyLsp
       @uri = T.let(uri, URI::Generic)
       @unparsed_edits = T.let([], T::Array[EditShape])
       @parse_result = T.let(YARP.parse(@source), YARP::ParseResult)
-      @parse_result.attach_comments!
     end
 
     sig { returns(YARP::ProgramNode) }
     def tree
       @parse_result.value
+    end
+
+    # sig { returns(YARP::ProgramNode) }
+    def comments
+      @parse_result.comments
     end
 
     sig { params(other: Document).returns(T::Boolean) }
--- a/lib/ruby_lsp/executor.rb
+++ b/lib/ruby_lsp/executor.rb
@@ -97,7 +97,7 @@ module RubyLsp
         # Run listeners for the document
         emitter = EventEmitter.new
         document_symbol = Requests::DocumentSymbol.new(emitter, @message_queue)
-        document_link = Requests::DocumentLink.new(uri, emitter, document.comments, @message_queue)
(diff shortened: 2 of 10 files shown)


lib/ruby_lsp/document.rb
```diff
@@ -30,6 +30,7 @@ def initialize(source:, version:, uri:, encoding: Constant::PositionEncodingKind
       @uri = T.let(uri, URI::Generic)
       @unparsed_edits = T.let([], T::Array[EditShape])
       @parse_result = T.let(YARP.parse(@source), YARP::ParseResult)
+      @parse_result.attach_comments!
```
andyw8, line 33: @vinistock do you think we should always attach the comments, or make it conditional for requests that need it?

vinistock: It could indeed have a performance cost because it basically analyzes the entire AST to attach the comments.

Given that the only request that needs comments is document link, which we plan on removing in the future anyway, I'm not sure we want to do it.

Tapioca always adds the magic source comment immediately above the class/module/constant/method, so we always know where to find it without paying this cost.

What do you think about avoiding it for now to get the most performance we can out of YARP?

In that case, we can pass the comments array to document link's initializer, turn the array into a hash of `line_number => comment` and then just access it directly when finding nodes.

lib/ruby_lsp/requests/document_link.rb
```diff
@@ -86,12 +86,43 @@ def initialize(uri, emitter, message_queue)
         @gem_version = T.let(version_match && version_match[0], T.nilable(String))
         @_response = T.let([], T::Array[Interface::DocumentLink])
 
-        emitter.register(self, :on_comment)
+        emitter.register(self, :on_def, :on_class, :on_module, :on_constant_write, :on_constant_path_write)
```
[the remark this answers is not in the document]
vinistock: Sorbet desugars attributes into simple method definitions and so does Tapioca. We don't need to worry about attributes.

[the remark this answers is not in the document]
andyw8: Technically we do, but it looks like `attr_` calls are transpiled to `def`s in the RBIs.

lib/ruby_lsp/document.rb
andyw8: I'll push a commit to show how that would look, then we can decide.

andyw8: Pushed.

lib/ruby_lsp/requests/document_link.rb
```diff
@@ -85,6 +92,12 @@ def initialize(uri, emitter, message_queue)
         version_match = path ? /(?<=%40)[\d.]+(?=\.rbi$)/.match(path) : nil
         @gem_version = T.let(version_match && version_match[0], T.nilable(String))
         @_response = T.let([], T::Array[Interface::DocumentLink])
+        @comments = T.let(
```
andyw8, line 95: Not sure what to name the instance variable, `@lines_to_comments`?

lib/ruby_lsp/executor.rb
```diff
@@ -97,7 +97,7 @@ def run(request)
         # Run listeners for the document
         emitter = EventEmitter.new
         document_symbol = Requests::DocumentSymbol.new(emitter, @message_queue)
-        document_link = Requests::DocumentLink.new(uri, emitter, @message_queue)
+        document_link = Requests::DocumentLink.new(uri, emitter, @message_queue, document.comments)
```
vinistock, line 100: We've been putting extra arguments other than `emitter` and `message_queue` first. Can we put this after `uri` please?

vinistock (approved):
Looks good! Just some minor adjustments

lib/ruby_lsp/executor.rb
```diff
@@ -97,7 +97,7 @@ def run(request)
         # Run listeners for the document
         emitter = EventEmitter.new
         document_symbol = Requests::DocumentSymbol.new(emitter, @message_queue)
-        document_link = Requests::DocumentLink.new(uri, emitter, @message_queue)
+        document_link = Requests::DocumentLink.new(uri, emitter, document.comments, @message_queue)
```
vinistock, line 100: The order doesn't look right here.

PR #985, opened 2023-09-07
Automate Dependabot RBI updates

andyw8:
Based on the approach being used for non-public projects. It can simplified since we are only targeting gems, not apps.

Verified in my fork: https://github.com/andyw8/ruby-lsp/pull/1

### DIFF
--- a/.github/workflows/dependabot_update_rbis.yml
+++ b/.github/workflows/dependabot_update_rbis.yml
@@ -14,7 +14,7 @@ jobs:
   update-gem-rbis:
     runs-on: ubuntu-latest
     timeout-minutes: 15
-    if: ${{ github.actor == '"dependabot[bot]"' }}
+    if: ${{ github.actor == 'dependabot[bot]' }}
     steps:
       - name: Checkout PR branch
         uses: actions/checkout@v4
@@ -36,7 +36,7 @@ jobs:
         id: setup
         with:
           bundler-cache: true
-          ruby-version: ${{ steps.ruby.outputs.version }}
+          ruby-version: ${{ steps.ruby.outputs.ruby_version }}
       - name: Update gem RBIs
         run: |
           bin/tapioca gem
--- a/.github/workflows/dependabot_update_rbis.yml
+++ b/.github/workflows/dependabot_update_rbis.yml
@@ -1,0 +1,57 @@
+name: Update gem RBIs on Dependabot PRs
+
+on: workflow_call
+
+permissions:
+  contents: write
+  pull-requests: write
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.head_ref }}
+  cancel-in-progress: true
+
+jobs:
+  update-gem-rbis:
+    runs-on: ubuntu-latest
+    timeout-minutes: 15
+    if: ${{ github.actor == '"dependabot[bot]"' }}
+    steps:
+      - name: Checkout PR branch
+        uses: actions/checkout@v4
+        with:
+          ref: ${{ github.head_ref }}
+      - name: Get Ruby version from dev.yml
+        id: ruby
+        shell: ruby -ryaml {0}
+        run: |
+          exit 1 unless File.exist?("dev.yml")
+          yaml = YAML.respond_to?(:unsafe_load_file) ? YAML.unsafe_load_file("dev.yml") : YAML.load_file("dev.yml")
+          ruby_version = yaml["up"]
+            &.find {|h| Hash === h && h.key?("ruby") }
+            &.dig("ruby")
+            .then {|c| Hash === c ? c["version"] : c }
+          exit 1 unless ruby_version
+          File.write(ENV.fetch("GITHUB_OUTPUT"), "ruby_version=#{ruby_version}\n", mode: "a")
+      - uses: ruby/setup-ruby@v1
+        id: setup
+        with:
+          bundler-cache: true
+          ruby-version: ${{ steps.ruby.outputs.version }}
+      - name: Update gem RBIs
+        run: |
+          bin/tapioca gem
+      - name: Check for modified files
+        id: git_modified_files_check
+        # If there are changes in the sorbet directory, `echo -n` will return true (zero exit code)
+        run: |
+          status=$(git status sorbet --porcelain=v1)
+          echo "changes=$([ -n "$status" ]; echo $?)" >> $GITHUB_OUTPUT
+      - name: Push changes
+        if: ${{ steps.git_modified_files_check.outputs.changes == 0 }}
+        run: |
+          git config user.name github-actions
+          git config user.email github-actions@github.com
+          git add sorbet/*
+          date="$(date -u +"%Y-%m-%d")"
+          git commit -m "Update RBI files for gems ${date} [dependabot skip]"
+          git push
--- a/.github/workflows/dependabot_update_rbis_run.yml
+++ b/.github/workflows/dependabot_update_rbis_run.yml
@@ -1,0 +1,6 @@
+name: Update gem RBIs on Dependabot PRs
+on: [pull_request]
+
+jobs:
+  dependabot_update_rbis:
+    uses: Shopify/ruby-lsp/.github/workflows/dependabot_update_rbis.yml@main
--- a/.github/workflows/dependabot_update_rbis_run.yml
+++ b/.github/workflows/dependabot_update_rbis_run.yml
@@ -1,5 +1,14 @@
 name: Update gem RBIs on Dependabot PRs
-on: [pull_request]
+
+on:
+  pull_request:
+    types: [opened, synchronize]
+    paths:
+      - 'Gemfile.lock'
+
+permissions:
+  contents: write
+  pull-requests: write
 
 jobs:
   dependabot_update_rbis:
--- a/sorbet/tapioca/config.yml
+++ b/sorbet/tapioca/config.yml
@@ -3,8 +3,9 @@ gem:
     "psych": "false"
   # Add your `gem` command parameters here:
   #
-  # exclude:
-  # - gem_name
+  # rdoc causes a failure 'webrick is not found.' when running `tapioca gem`.
+  exclude:
+  - rdoc
   # doc: true
   # workers: 5
 dsl:


.github/workflows/dependabot_update_rbis.yml
```diff
@@ -0,0 +1,71 @@
+name: Update gem RBIs on Dependabot PRs
```
andyw8, line 1: Once we are happy with this, it can be moved to another repo for sharing.

.github/workflows/dependabot_update_rbis.yml
```diff
@@ -0,0 +1,71 @@
...
+          ruby-version: 3.2
+      - name: Ruby setup error details
+        if: ${{ failure() && steps.setup.conclusion == 'failure' }}
+        run: |
+          echo "A failure during the Ruby setup step could indicate that the permissions" \
+            " for this repo are not configured correctly." \
+          exit 1
+      - name: Update gem RBIs
+        run: |
+          ${{ inputs.tapioca_env_vars }} bin/tapioca gem --workers=1
```
andyw8, line 56: I was getting a failure when running with multiple workers.

.github/workflows/dependabot_update_rbis.yml
```diff
@@ -0,0 +1,71 @@
...
+          bundler-cache: true
+          ruby-version: 3.2
+      - name: Ruby setup error details
+        if: ${{ failure() && steps.setup.conclusion == 'failure' }}
+        run: |
+          echo "A failure during the Ruby setup step could indicate that the permissions" \
+            " for this repo are not configured correctly." \
+          exit 1
+      - name: Update gem RBIs
+        run: |
```
andyw8, line 55: Should we also run `tapoica dsl` here?

.github/workflows/dependabot_update_rbis.yml
```diff
@@ -0,0 +1,71 @@
...
+        description: A space-delimited list of environment variables and their values, to be passed to the `tapioca gem` command
+        type: string
+        required: false
+      valid_actors:
+        description: A stringified list of GitHub actors that can trigger this job
+        type: string
+        default: '["dependabot[bot]"]'
+        required: false
+
+permissions:
```
andyw8, line 20: TODO: check these are appropriate.

.github/workflows/dependabot_update_rbis.yml
```diff
@@ -0,0 +1,71 @@
...
+    steps:
+      - name: Checkout PR branch
+        uses: actions/checkout@v4
+        with:
+          ref: ${{ github.head_ref }}
+      - name: Set up any manually specified dependencies
+        if: ${{ inputs.dependency_setup_command != '' }}
+        run: ${{ inputs.dependency_setup_command }}
+        env:
+          DEBIAN_FRONTEND: noninteractive
```
vinistock, line 42: Is this necessary?

.github/workflows/dependabot_update_rbis.yml
vinistock: I don't think it's necessary. We haven't been using DSL RBIs in the Ruby LSP.

Gemfile
```diff
@@ -23,4 +23,5 @@ group :development do
   gem "tapioca", "~> 0.11", require: false, platforms: NON_WINDOWS_PLATFORMS
   gem "rdoc", require: false
   gem "psych", "~> 5.1", require: false
+  gem "webrick"
```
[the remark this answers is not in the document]
vinistock: Hopefully there is cause we don't want to add a webserver as a dependency 😂

.github/workflows/dependabot_update_rbis.yml
andyw8: I was thinking more for if we generalized this to be a shared action.

.github/workflows/dependabot_update_rbis.yml
andyw8: As we don't have extra dependencies to install, we can remove this step.

.github/workflows/dependabot_update_rbis.yml
Morriar: Wouldn't that fail for Ruby-LSP?

irrc when the project is not a Rails app `tapioca dsl --verify` will return the exit code 1 with this error:

```
No classes/modules can be matched for RBI generation.
Please check that the requested classes/modules include processable DSL methods.
```

andyw8: Ah yes it fails, I thought it would just be a no-op if there weren't any.

Isn't is possible to have DSL compilers in non-Rails projects?

sorbet/tapioca/config.yml
```diff
@@ -3,8 +3,9 @@ gem:
     "psych": "false"
   # Add your `gem` command parameters here:
   #
-  # exclude:
-  # - gem_name
+  # rdoc causes a failure 'webrick is not found.' when running `tapioca gem`.
+  exclude:
+  - rdoc
```
andyw8, line 8: https://github.com/Shopify/tapioca/issues/1647

.github/workflows/dependabot_update_rbis.yml
Morriar: Not at the moment but we're working on it.

If you want to make this sharable, you _could_ look for the `config/application.rb` file to exist before running `tapicoa dsl`.

.github/workflows/dependabot_update_rbis.yml
```diff
@@ -0,0 +1,53 @@
...
+      - name: Checkout PR branch
+        uses: actions/checkout@v4
+        with:
+          ref: ${{ github.head_ref }}
+      - uses: ruby/setup-ruby@v1
+        id: setup
+        with:
+          bundler-cache: true
+          ruby-version: 3.2
+      - name: Ruby setup error details
```
KaanOzkan, line 28: We should get the error message in the step prior this can be removed.

.github/workflows/dependabot_update_rbis.yml
KaanOzkan: Ruby-lsp is already public. Why not have this in tapioca from the start?

.github/workflows/dependabot_update_rbis.yml
```diff
@@ -0,0 +1,53 @@
...
+    steps:
+      - name: Checkout PR branch
+        uses: actions/checkout@v4
+        with:
+          ref: ${{ github.head_ref }}
+      - uses: ruby/setup-ruby@v1
+        id: setup
+        with:
+          bundler-cache: true
+          ruby-version: 3.2
```
KaanOzkan, line 27: This should be customizable

.github/workflows/dependabot_update_rbis.yml
```diff
@@ -0,0 +1,53 @@
...
+      - name: Ruby setup error details
+        if: ${{ failure() && steps.setup.conclusion == 'failure' }}
+        run: |
+          echo "A failure during the Ruby setup step"
+          exit 1
+      - name: Update gem RBIs
+        run: |
+          bundle exec tapioca gem
```
KaanOzkan, line 35:
```suggestion
          bin/tapioca gem
```
For consistency

.github/workflows/dependabot_update_rbis.yml
KaanOzkan: I really don't think running `tapioca dsl` is necessary. And for non simple apps this will fail because we can't boot it. IMO if someone wants to update DSLs they can copy the workflow and modify it. We want a workflow that's easy to maintain for us and just works.

.github/workflows/dependabot_update_rbis.yml
KaanOzkan: If the goal of this workflow to be picked up by others they could need this step.

.github/workflows/dependabot_update_rbis_run.yml
```diff
@@ -0,0 +1,9 @@
+name: Update RBIs
+on:
+  pull_request:
+    branches:
+      - main
```
KaanOzkan, line 5:
```suggestion
    types: [opened, synchronize]
    paths:
      - 'Gemfile.lock'
```

I don't think we need to filter for a specific branch.

.github/workflows/dependabot_update_rbis_run.yml
```diff
@@ -0,0 +1,9 @@
...
+
+jobs:
+  verify:
```
KaanOzkan, line 8: Why the name `verify`?

.github/workflows/dependabot_update_rbis_run.yml
```diff
@@ -0,0 +1,9 @@
+name: Update RBIs
```
KaanOzkan, line 1:
```suggestion
name: Update gem RBIs on Dependabot PRs
```
I'd like to have the reusable workflow in tapioca and name this more specifically.

.github/workflows/dependabot_update_rbis.yml
andyw8: Just so that if we need to tweak anything, then it's all in the one place, rather than having to make changes across two repos.

.github/workflows/dependabot_update_rbis.yml
andyw8: What do we think about committing a `.ruby-version` for public repos? Then we can remove this and settig and it will pick that up by default.

KaanOzkan: Yes we can use that to customize it. It'd have to be highlighted in the docs since it's a mandatory onboarding step.

I don't think removing the setting is easiest. We can keep the setting so that we have a logic for looking at `.ruby-version` and then falling back to reading `dev.yml` if it doesn't exist. When a new shopify repo onboards it means one less change to remember.

.github/workflows/dependabot_update_rbis.yml
andyw8: Fair enough, removed to keep things simple for now.

.github/workflows/dependabot_update_rbis_run.yml
andyw8: Fixed

.github/workflows/dependabot_update_rbis.yml
vinistock: This comment was resolved, but the code wasn't updated.

.github/workflows/dependabot_update_rbis.yml
vinistock: In my opinion, the Ruby version should be configurable. If users of the workflow want to read `.ruby-version` or any other configuration file, then they should be able to do that from their usage of the workflow.

.github/workflows/dependabot_update_rbis.yml
andyw8: Fixed (I committed on GitHub.com but force pushed before pulling).

.github/workflows/dependabot_update_rbis.yml
KaanOzkan: Some gems that are not released as binary but compiled require certain packages to be present. We won't run into this with our simple repos but it would be needed depending on the gem. It could be a future improvement too.

.github/workflows/dependabot_update_rbis.yml
andyw8: For now I've updated it to read from `dev.yml`, similar to how the internal action works. We can later consider options for supporting other configurations. It would nice to have it consistent with the [options available through setup-ruby.](https://github.com/ruby/setup-ruby#supported-version-syntax)

.github/workflows/dependabot_update_rbis.yml
```diff
@@ -0,0 +1,57 @@
...
+  pull-requests: write
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.head_ref }}
+  cancel-in-progress: true
+
+jobs:
+  update-gem-rbis:
+    runs-on: ubuntu-latest
+    timeout-minutes: 30
```
KaanOzkan, line 16: Nit: Timeout can be significantly reduced given the average `bundle install` time on ruby-lsp. 10-15 mins would be enough buffer.

KaanOzkan (approved):

.github/workflows/dependabot_update_rbis.yml
andyw8: 👍 Other projects may need be a bit longer than Ruby LSP, but 15m seems plenty, updated.

vinistock (approved):
I think it's fine to give it a test here, but once we conclude it's working, we need to move it to Tapioca before people start using this action

andyw8:
I pushed a commit with a few small updates, after testing in my fork, and comparing agains the internal version.

Example of the RBI update successfully added: https://github.com/andyw8/ruby-lsp/pull/9/commits

andyw8:
Draft PR for moving this to Tapioca once we're happy with it: https://github.com/Shopify/tapioca/pull/1657

PR #986, opened 2023-09-07
Update text references from Syntax Tree to YARP

andyw8:

### DIFF
--- a/lib/ruby_lsp/requests/base_request.rb
+++ b/lib/ruby_lsp/requests/base_request.rb
@@ -20,7 +20,7 @@ module RubyLsp
       sig { abstract.returns(Object) }
       def run; end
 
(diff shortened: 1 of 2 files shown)


lib/ruby_lsp/requests/base_request.rb
```diff
@@ -20,7 +20,7 @@ def initialize(document)
       sig { abstract.returns(Object) }
       def run; end
 
-      # Syntax Tree implements `visit_all` using `map` instead of `each` for users who want to use the pattern
+      # YARP implements `visit_all` using `map` instead of `each` for users who want to use the pattern
```
andyw8, line 23: Still true:
https://github.com/ruby/yarp/blob/718e4bd1e5f5ba9f5dce90766536dffdeefe3f2d/lib/yarp.rb#L195

PR #989, opened 2023-09-08
Make Definition extensible

st0012:
### Motivation

This is part of https://github.com/Shopify/ruby-lsp/issues/808

### Implementation

1. Make `Definition` inherit `ExtensibleListener`
2. Implement all the required interfaces + move `super` call of the constructor after other ivars are initialised 
3. Add `Extension#create_definition_listener`

### Automated Tests

I added 1 test for the extension. But given `Definition#merge_response!` needs to handle a few more cases for combination of response types (`nil | Location | Array of Location`), we **could** use more tests cover them.

However, all the cases should already be checked by Sorbet and it'd be very lengthy to make tests populate different definition results **while** calling extensions. So I'm not sure if it'd be worth the effort.


vinistock (approved):
🚀

PR #995, opened 2023-09-11
Automatically bump version after release

vinistock:
### Motivation

In order to compare the performance of our latest release vs main, we need main to be eagerly updated with the next version number after a release happens. This PR adds automation to bump the patch version automatically after a release.

**NOTE**: it always bumps the patch version just so that we can have a different version show up in the metrics. It does not mean we can't manually change the version in case necessary.

### Implementation

The action
1. Bumps the version using awk
2. Runs bundle, creates a branch and commits both `VERSION` and `Gemfile.lock`
3. Creates a PR using that new branch
4. Approves and turns on auto-merge for the PR

### DIFF
--- a/.github/workflows/version_bump.yml
+++ b/.github/workflows/version_bump.yml
@@ -1,0 +1,86 @@
+name: Release
+
+on:
+  release:
+    types: [published]
+
+jobs:
+  publish:
+    runs-on: ubuntu-latest
+
+    steps:
+      - uses: actions/checkout@v3
+        name: Checkout
+
+      - name: Set up Ruby
+        uses: ruby/setup-ruby@v1
+        with:
+          ruby-version: 3.2.2
+          bundler-cache: true
+          cache-version: 1
+
+      - name: Bump patch version
+        id: version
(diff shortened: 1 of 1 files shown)


.github/workflows/version_bump.yml
```diff
@@ -0,0 +1,83 @@
...
+            const response = await github.rest.pulls.create({
+              owner: context.repo.owner,
+              repo: context.repo.repo,
+              head: automated-bump-version-${{ steps.version.outputs.VERSION }},
+              base: main,
+              title: "Bump version to v${{ steps.version.outputs.VERSION }}"
+            });
+
+            const pullRequestId = response.data.id;
+            const pullRequestNumber = response.data.number;
```
Morriar, line 53: Should we at least display the PR number for debugging?

Morriar (approved):

.github/workflows/version_bump.yml
vinistock: I put in some console logs. Is that what you meant?

PR #997, opened 2023-09-12
Upgrade YARP dependency to v0.11.0

vinistock:
### Motivation

There's one breaking change that impacts us in v0.11.0: variables and constant names where changed to symbols instead of strings.

Additionally, the off-by-one error in end columns has been fixed.

### Implementation

1. Upgrade the YARP requirement
2. Upgrade the YARP RBI and the annotations generated by our scrappy template
3. Fix the error in our indexer by turning constant names into strings
4. Fix all off-by-one error locations that we introduced in tests


andyw8 (approved):

Morriar:
Should we migrate to 0.11 while at it?

Issue #464, opened 2023-01-20
Fix behaviour of DocumentLink for when a gem has been removed


andyw8:
_Originally posted by @vinistock in https://github.com/Shopify/ruby-lsp/pull/463#discussion_r1083056874_
            

 > I think the mistake was actually the other way around. If the source comment is for a gem that is no longer a part of the bundle, then `GEM_TO_VERSION_MAP[uri.gem_name]` will return `nil`.

> Instead of doing the `T.must`, we need to return early in that case.

andyw8:
@haroon26 done!

vinistock:
@haroon26 are you looking into this? Let us know if you need any help. Otherwise, we can remove the assignment to let somebody else pick it up.

haroon26:
Hello @vinistock, I got busy with my work so completely forgot about it. I will resolve it on priority this week. Thank you.

PR #998, opened 2023-09-13
Fix DocumentLink behaviour for removed gem

andyw8:
Closes https://github.com/Shopify/ruby-lsp/issues/464

Co-authored by @haroon26.

### DIFF
--- a/test/fixtures/source_comment.rb
+++ b/test/fixtures/source_comment.rb
@@ -9,3 +9,7 @@ end
 # source://syntax_tree//lib/syntax_tree.rb#39
 def baz
 end
+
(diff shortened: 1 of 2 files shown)


test/fixtures/source_comment.rb
```diff
@@ -9,3 +9,7 @@ def bar
 # source://syntax_tree//lib/syntax_tree.rb#39
 def baz
 end
+
+# source://deleted//lib/foo.rb.rb#1
```
andyw8, line 13: Since `test/expectations/document_link/source_comment.exp.json` remains unchanged, it verifies the new behaviour is correct.

PR #884, opened 2023-08-11
Migrate semantic highlighting to YARP

andyw8:
### Motivation

Partially addresses https://github.com/Shopify/ruby-lsp/issues/449

PR #1000, opened 2023-09-13
Migrate semantic highlighting to YARP

vinistock:
### Motivation

Closes #884

Migrate semantic highlighting to YARP.

### Implementation

This PR adds a parameter scope implementation. This is a much simpler implementation than the scopes we contributed to SyntaxTree. It only tracks the parameters for the current scope and nothing else - which we use to highlight parameters consistently.

Other than that, the rest is just adding the necessary visits and changing node types.

### Automated Tests

Most tests didn't need any changes, but some new tokens are now being pushed in a few cases. For example, `<<` used to not be considered a `CallNode`, but it now is and so we get a token for that.


KaanOzkan (approved):

lib/ruby_lsp/parameter_scope.rb
```diff
@@ -0,0 +1,33 @@
...
+
+    sig { params(name: T.any(Symbol, String)).returns(Symbol) }
+    def type_for(name)
+      parameter?(name) ? :parameter : :variable
+    end
+
+    sig { params(name: T.any(Symbol, String)).returns(T::Boolean) }
+    def parameter?(name)
+      sym = name.to_sym
+      @parameters.include?(sym) || (!@parent.nil? && @parent.parameter?(sym))
```
andyw8, line 30:
```suggestion
      @parameters.include?(sym) || @parent&.parameter?(sym)
```

lib/ruby_lsp/requests/semantic_highlighting.rb
```diff
@@ -339,38 +391,47 @@ def add_token(location, type, modifiers = [])
...
+        # The regexp needs to be the receiver of =~ for local variable capture
+        return unless receiver.is_a?(YARP::RegularExpressionNode)
+
+        content = receiver.content
+        loc = receiver.content_loc
+
+        # For each capture name we find in the regexp, look for a local in the current_scope
+        Regexp.new(content, Regexp::FIXEDENCODING).names.each do |name|
+          # The +3 is to compensate for the "(?<" part of the capture name
+          capture_name_offset = T.must(content.index("(?<#{name}>")) + 3
```
andyw8, line 414: Not a blockers, but it seems there are some other ways of representing named captures:

https://rubyapi.org/3.2/o/regexp#method-i-named_captures:~:text=Named%20groups%20can%20be%20backreferenced%20with%20%5Ck%3Cname%3E%2C%20where%20name%20is%20the%20group%20name.

andyw8 (approved):

lib/ruby_lsp/parameter_scope.rb
```diff
@@ -0,0 +1,33 @@
...
+
+module RubyLsp
+  class ParameterScope
+    extend T::Sig
+
+    sig { returns(T.nilable(ParameterScope)) }
+    attr_reader :parent
+
+    sig { params(parent: T.nilable(ParameterScope)).void }
+    def initialize(parent = nil)
```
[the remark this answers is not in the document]
vinistock: We don't pass an argument when we instantiate it for the first time in `SemanticHighlighting#initialize`. It would also not be correct to always require a parent given that the global scope doesn't have one.

lib/ruby_lsp/parameter_scope.rb
vinistock: That fails typechecking because instead of returning `T::Boolean` it returns `T.nilable(T::Boolean)`.

test/requests/support/semantic_token_encoder_test.rb
```diff
@@ -96,15 +96,12 @@ def test_encoded_modifiers_with_some_modifiers
   private
 
   def stub_token(start_line, start_column, length, type, modifier)
+    location = YARP::Location.new("", 123, 123)
+    location.expects(:start_line).returns(start_line).at_least_once
```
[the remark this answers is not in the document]
vinistock: It gets triggered more than once, so this is necessary. By default, Mocha fails if an expectation gets hit twice.

lib/ruby_lsp/requests/semantic_highlighting.rb
vinistock: I think the `\k<>` is a backreference to the named capture and not another way to create named captures. The idea is for you to do this

```ruby
/(?<group>(something|other))(\k<group>)/
#  ^ this captures group       ^ this references the captured group
```

Issue #966, opened 2023-09-01
Handle private constants in the index


vinistock:
Currently, we're not tracking whether a constant is marked as private or not. It's important to keep this information in the index, since you should only be able to see hover, go to definition or receive autocomplete suggestions for public constants or private constants while being inside the namespace that owns them.

Add a `visibility` field to our `Entry` class and start tracking invocations to `private_constant`.

PR #1004, opened 2023-09-14
Account for private constants in features

vinistock:
### Motivation

Closes https://github.com/Shopify/ruby-lsp/issues/966

We should only provide definition, completion and hover for private constants if the reference happens inside the exact same namespace as they are defined. Otherwise, it's a reference to a constant that doesn't actually exist.

Similarly, private constants should never be shown in workspace symbol.

### Implementation

I split the PR by commits to make it easier to review.

On each request, started skipping adding results if we find a non-private reference to a private constant.

### Automated Tests

Added tests for all features.


KaanOzkan (approved):
Nice comments

test/requests/completion_test.rb
```diff
@@ -294,6 +294,58 @@ module Foo
...
+      method: "textDocument/completion",
+      params: { textDocument: { uri: @uri.to_s }, position: end_position },
+    )
+    assert_equal(["CONST"], result.map { |completion| completion.text_edit.new_text })
+  ensure
+    RubyLsp::DependencyDetector.const_set(:HAS_TYPECHECKER, true)
+  end
+
+  def test_completion_private_constants_from_different_namespace
+    RubyLsp::DependencyDetector.const_set(:HAS_TYPECHECKER, false)
```
andyw8, line 324: Not in this PR, but since we use this a lot I wonder we should extract it out, e.g.:

```ruby
def with_typechecker_disabled(&block)
  RubyLsp::DependencyDetector.const_set(:HAS_TYPECHECKER, false)
  yield
ensure
  RubyLsp::DependencyDetector.const_set(:HAS_TYPECHECKER, true)
end
```

test/requests/workspace_symbol_test.rb
```diff
@@ -95,4 +95,17 @@ def test_finds_default_gem_symbols
...
+  def test_does_not_include_private_constants
+    @index.index_single(RubyIndexer::IndexablePath.new(nil, "/fake.rb"), <<~RUBY)
+      class Foo
+        CONSTANT = 1
+        private_constant(:CONSTANT)
+      end
+    RUBY
+
+    result = RubyLsp::Requests::WorkspaceSymbol.new("Foo::CONSTANT", @index).run
+    assert_equal(1, result.length)
```
andyw8, line 108: I may be misunderstanding the test, but if we're looking up `Foo::CONSTANT` then why is `Foo` in the results?

andyw8: Ah, I see. That may be worth a code comment.

andyw8 (approved):

PR #1005, opened 2023-09-14
Migrate document highlight to YARP

vinistock:
### Motivation

Migrate document highlight to use YARP. I based this branch on top of the semantic highlighting one because it allows us to use the LSP and so it's significantly easier to verify the changes.

### Implementation

This one was a bit painful because our previous implementation was basically based on `SyntaxTree::Ident`, which is a concept that doesn't exist in YARP.

I had to list every possible node for local vars, instance vars, class vars, constants, constant paths, etc.

To maintain the precision of the highlights, I also change the match to be hold a location rather than a node. This is necessary because YARP doesn't produce nodes for certain things we want to highlight, like method declaration names. We wouldn't want to highlight the entire method node, only the `name_loc`.

### Automated Tests

Had to adapt a few. See the comments.


test/expectations/document_highlight/class_and_reference.exp.json
```diff
@@ -19,6 +19,19 @@
             },
             "kind": 3
         },
+        {
```
vinistock, line 22: Because of the structure of the tree, we get a duplicate highlight for classes. One occurs because of the `ClassNode` and another because of the `ConstantReadNode` inside of it.

However, the editor dedups this and only shows the write highlight, which results in exactly the same experience for developers.

test/expectations/document_highlight/module_and_reference.exp.json
```diff
@@ -19,6 +19,19 @@
             },
             "kind": 3
         },
+        {
```
vinistock, line 22: Same thing as `class_and_reference`.

lib/ruby_lsp/requests/document_highlight.rb
```diff
@@ -47,19 +47,29 @@ def initialize(target, parent, emitter, message_queue)
 
         highlight_target =
           case target
-          when *DIRECT_HIGHLIGHTS
+          when YARP::GlobalVariableReadNode, YARP::GlobalVariableAndWriteNode, YARP::GlobalVariableOperatorWriteNode,
```
andyw8, line 50: If wonder if there's some way to structure this list would be less sensitive to changes in YARP (e.g. renamed, added, or deleted node types). Maybe @kddnewton has thoughts?

kddnewton: I see what you mean, but hopefully the AST should be pretty stable going forward. Most of the churn has been while we've been getting it up and running.

vinistock:
Yeah, document highlight needs a revisit to improve the implementation. For the YARP migration, let's move forward with the request as is to minimize the changes.

After the migration is shipped, we can revisit some requests that can be improved - especially considering YARP's tree structure.

PR #1010, opened 2023-09-14
Exclude transitive development dependencies

vinistock:
### Motivation

Previously, we were only automatically ignoring direct development dependencies. That's not ideal because transitive development dependencies are generally not wanted in indexing. This PR excludes transitive dependencies from the default indexing, to produce more accurate results and faster indexing times.

Devs can still manually include any development dependency through configuration if desired.

### Implementation

The logic is basically this:
1. Split all dependencies into development only and others
2. Create a list of gemspec dependencies for the current gem if you're working on one
3. Go through each transitive development dependency and check
    - Is this transitive dependency included in `others`? That means the transitive dependency is included directly as part of another group, so we can't exclude
    - Is this transitive dependency included as a transitive dependency of another gem that's not in the development group? In that case, the transitive gem is included outside of development and we can't exclude
    - Is the transitive dependency included as a gemspec dependency for the gem currently being worked on? If true, then we can't exclude
    - All other transitive dependencies are excluded

### Automated Tests

Enhanced our test to show that we're no longer including a transitive dependency, but still including gemspec dependencies.

### DIFF
--- a/lib/ruby_indexer/lib/ruby_indexer/configuration.rb
+++ b/lib/ruby_indexer/lib/ruby_indexer/configuration.rb
@@ -18,11 +18,7 @@ module RubyIndexer
 
     sig { void }
     def initialize
-      excluded_gem_names = Bundler.definition.dependencies.filter_map do |dependency|
-        dependency.name if dependency.groups == [:development]
-      end
-
-      @excluded_gems = T.let(excluded_gem_names, T::Array[String])
+      @excluded_gems = T.let(initial_excluded_gems, T::Array[String])
       @included_gems = T.let([], T::Array[String])
       @excluded_patterns = T.let([File.join("**", "*_test.rb")], T::Array[String])
       path = Bundler.settings["path"]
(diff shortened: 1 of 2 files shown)


lib/ruby_indexer/lib/ruby_indexer/configuration.rb
```diff
@@ -174,5 +170,39 @@ def apply_config(config)
...
+    sig { returns(T::Array[String]) }
+    def initial_excluded_gems
+      excluded, others = Bundler.definition.dependencies.partition do |dependency|
+        dependency.groups == [:development]
+      end
+
+      # When working on a gem, we need to make sure that its gemspec dependencies can't be excluded. This is necessary
+      # because Bundler doesn't assign groups to gemspec dependencies
+      this_gem = Bundler.definition.dependencies.find { |d| d.to_spec.full_gem_path == Dir.pwd }
+      others.concat(this_gem.to_spec.dependencies) if this_gem
```
KaanOzkan, line 183: I'm confused by the `this_gem` logic, could you elaborate a bit. I feel like `Dir.pwd` will always return `ruby-lsp` so we will add its dependencies to `others`?

vinistock: Yeah, I couldn't think of a better name for it.

Basically, if you're working on a gem, we need to prevent the gem's gemspec dependencies from being excluded accidentally.

For example, in the Ruby LSP we have `sorbet-static-and-runtime` in the Gemfile, but `sorbet-runtime` as a gemspec dependency. Because `sorbet-runtime` is a transitive dependency of `sorbet-static-and-runtime` and that gem is in the development group, we would accidentally exclude a real dependency.

KaanOzkan: Ah I see. Thanks, I was mistaken `Dir.pwd` does return directory of the main application.

KaanOzkan (approved):

st0012 (approved):

PR #1012, opened 2023-09-15
Invert response merging responsibility

vinistock:
### Motivation

This PR proposes a way to invert the responsibility for response merging, so that extensions have more control over how they merge themselves into the final result.

### Implementation

The idea is that merging the response is no longer implemented in our base listeners. They're instead implemented in extension listeners.

To make sure that extensions are doing the right thing, we created a separate class `ExtensionListener`, which should be the only one used by extensions. We also marked our other listener classes as private constants, so that only the Ruby LSP can use them.

Let me know what you folks think of the design. This gives more power to extensions, but of course they may accidentally mutate the response in a weird way.

As a follow up, we could write response wrappers for the requests we allow extending, where we hide the original response to minimize potential impact.

### Automated Tests

Because `Listener` is now a private constant, I suggest we namespace all of our tests with `RubyLsp`. It makes it easier to refer to constants inside the namespace anyyway.

### DIFF
--- a/lib/ruby_lsp/listener.rb
+++ b/lib/ruby_lsp/listener.rb
@@ -30,6 +30,7 @@ module RubyLsp
     sig { abstract.returns(ResponseType) }
     def _response; end
   end
+  private_constant(:Listener)
 
   # ExtensibleListener is an abstract class to be used by requests that accept extensions.
   class ExtensibleListener < Listener
@@ -51,7 +52,7 @@ module RubyLsp
         Extension.extensions.filter_map do |ext|
           initialize_external_listener(ext)
         end,
-        T::Array[RubyLsp::Listener[ResponseType]],
+        T::Array[RubyLsp::ExtensionListener[ResponseType]],
       )
     end
 
@@ -59,7 +60,7 @@ module RubyLsp
     # response to the editor including the results of all extensions
     sig { void }
     def merge_external_listeners_responses!
-      @external_listeners.each { |l| merge_response!(l) }
+      @external_listeners.each { |l| l.merge_response(_response) }
     end
 
     sig { returns(ResponseType) }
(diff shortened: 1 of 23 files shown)


lib/ruby_lsp/listener.rb
```diff
@@ -69,14 +70,22 @@ def response
...
+    extend T::Sig
+    extend T::Helpers
+    extend T::Generic
+
+    ResponseType = type_member
+
+    abstract!
+
+    sig { abstract.params(current_response: ResponseType).void }
+    def merge_response(current_response); end
```
st0012, line 89: To features like `SemanticHighlighting` or `DocumentSymbol`, it will be more like updating the request with additional information instead of merging a separate response. Perhaps we should use a different verb here?

Morriar: Good point.

* `handle_response`
* `process_response`
* `decorate_response`
* `extend_response`

?

vinistock: I find merge to be intuitive, but I don't have strong opinions. Which one do people find the best?

test/requests/code_action_resolve_expectations_test.rb
```diff
@@ -4,69 +4,71 @@
 require "test_helper"
 require "expectations/expectations_test_runner"
 
-class CodeActionResolveExpectationsTest < ExpectationsTestRunner
-  expectations_tests RubyLsp::Requests::CodeActionResolve, "code_action_resolve"
+module RubyLsp
```
andyw8, line 7: 👍 I always prefer namespacing tests like this. I think it would have been preferable as as separate PR, but no need to change now.

lib/ruby_lsp/listener.rb
andyw8: I like `decorate_response` and `extend_response` as they both indicate that we are augmenting the original response.

andyw8 (approved):

vinistock:
Closing this for now. We'll revisit addon response merging in more detail and make sure the interface sets up authors for success.

PR #750, opened 2023-06-07
WIP: Refactor folding ranges to use emitter pattern (alternative approach)

andyw8:

PR #1013, opened 2023-09-15
Make folding range a listener

### LINKED ISSUE #610: Migrate requests to new emitter listener strategy
After #607 goes in, we can start refactoring other requests to use the listener pattern. This will allow us to better understand what is missing for extensions to do their work, but it'll also allow us to execute all nonpositional requests in one round of visits, which should be great for performance.

Note: certain requests are simply not a good fit for the listener pattern (e.g.: formatting, on type formatting, code actions, code action resolve). We should keep those as they are for now.

```[tasklist]
### Tasks
- [ ] https://github.com/Shopify/ruby-lsp/issues/655
- [ ] https://github.com/Shopify/ruby-lsp/issues/677
- [ ] https://github.com/Shopify/ruby-lsp/issues/646
- [ ] https://github.com/Shopify/ruby-lsp/issues/678
- [ ] https://github.com/Shopify/ruby-lsp/issues/659
- [ ] https://github.com/Shopify/ruby-lsp/issues/806
- [ ] https://github.com/Shopify/ruby-lsp/issues/661
- [ ] https://github.com/Shopify/ruby-lsp/issues/664
- [ ] https://github.com/Shopify/ruby-lsp/issues/702
```

### LINKED ISSUE #677: Migrate folding range

vinistock:
### Motivation

Closes #677
Closes #610
Closes #750

After migrating to YARP, turning folding range into a listener became a little easier.

### Implementation

Since we discovered that the clients dedup many responses, I think we should avoid extra complexity in the server just to avoid duplicate items. Some tests now include a few duplicates, but this allows us to get a clean listener implementation for folding range.

In terms of the migration, the only odd thing for this one is that we need to emit partial ranges and collect comments after receiving all events. I did this in the `_response` implementation, let me know if you have concerns.

### Automated Tests

Updated the tests where duplication happens.

### DIFF
--- a/lib/ruby_lsp/event_emitter.rb
+++ b/lib/ruby_lsp/event_emitter.rb
@@ -251,5 +251,101 @@ module RubyLsp
       @listeners[:on_block_local_variable]&.each { |l| T.unsafe(l).on_block_local_variable(node) }
       super
     end
+
+    sig { override.params(node: YARP::IfNode).void }
+    def visit_if_node(node)
+      @listeners[:on_if]&.each { |l| T.unsafe(l).on_if(node) }
+      super
+    end
+
+    sig { override.params(node: YARP::InNode).void }
+    def visit_in_node(node)
+      @listeners[:on_in]&.each { |l| T.unsafe(l).on_in(node) }
+      super
+    end
+
+    sig { override.params(node: YARP::WhenNode).void }
+    def visit_when_node(node)
+      @listeners[:on_when]&.each { |l| T.unsafe(l).on_when(node) }
+      super
+    end
+
+    sig { override.params(node: YARP::InterpolatedStringNode).void }
+    def visit_interpolated_string_node(node)
+      @listeners[:on_interpolated_string]&.each { |l| T.unsafe(l).on_interpolated_string(node) }
+      super
+    end
+
+    sig { override.params(node: YARP::ArrayNode).void }
+    def visit_array_node(node)
+      @listeners[:on_array]&.each { |l| T.unsafe(l).on_array(node) }
+      super
+    end
+
+    sig { override.params(node: YARP::CaseNode).void }
+    def visit_case_node(node)
+      @listeners[:on_case]&.each { |l| T.unsafe(l).on_case(node) }
+      super
+    end
+
+    sig { override.params(node: YARP::ForNode).void }
+    def visit_for_node(node)
+      @listeners[:on_for]&.each { |l| T.unsafe(l).on_for(node) }
+      super
+    end
+
+    sig { override.params(node: YARP::HashNode).void }
+    def visit_hash_node(node)
+      @listeners[:on_hash]&.each { |l| T.unsafe(l).on_hash(node) }
+      super
+    end
+
+    sig { override.params(node: YARP::SingletonClassNode).void }
+    def visit_singleton_class_node(node)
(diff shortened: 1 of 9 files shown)


lib/ruby_lsp/requests/folding_ranges.rb
```diff
@@ -15,38 +15,206 @@ module Requests
...
 
-      sig { override.returns(T.all(T::Array[Interface::FoldingRange], Object)) }
-      def run
-        visit(@document.tree)
+      sig { override.returns(ResponseType) }
+      def _response
         push_comment_ranges
         emit_requires_range
-        @ranges
+        @_response
```
st0012, line 63: I remember we wanted `Listener#response` to be idempotent. If that's still the case, we should probably add a flag to guard the above 2 method calls?

vinistock: Good point! I added a guard to make it idempotent.

lib/ruby_lsp/event_emitter.rb
```diff
@@ -251,5 +251,101 @@ def visit_block_local_variable_node(node)
       @listeners[:on_block_local_variable]&.each { |l| T.unsafe(l).on_block_local_variable(node) }
       super
     end
+
```
andyw8, line 254: Not for this PR, but as followup maybe we can order these alphabetically?

lib/ruby_lsp/executor.rb
```diff
@@ -105,6 +104,7 @@ def run(request)
 
         # Store all responses retrieve in this round of visits in the cache and then return the response for the request
         # we actually received
+        document.cache_set("textDocument/foldingRange", folding_range.response)
```
andyw8, line 107:
```suggestion
        document.cache_set("textDocument/foldingRanges", folding_range.response)
```

lib/ruby_lsp/executor.rb
```diff
@@ -96,6 +94,7 @@ def run(request)
 
         # Run listeners for the document
         emitter = EventEmitter.new
+        folding_range = Requests::FoldingRanges.new(document.parse_result.comments, emitter, @message_queue)
```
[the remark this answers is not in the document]
andyw8: Actually, to match the LSP spec, maybe we keep the name as `folding_range`, but rename the class?

andyw8 (approved):

lib/ruby_lsp/event_emitter.rb
vinistock: We're doing work to ship the event emitter in YARP, so maybe let's hold off since we may be able to get rid of this entire file.

lib/ruby_lsp/executor.rb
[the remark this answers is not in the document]
vinistock: Yeah, in the beginning of the project we named a few requests wrong (using plural where it should've been singular).

Let's handle that in a separate PR though.

st0012, line 97: Not directly related to this PR, but now with 5 requests depending on this code path, I feel we'll need more sophisticated error handling so failures to propagate too much, especially if we decided to invert the response updating logic to extensions.

st0012 (approved):

lib/ruby_lsp/executor.rb
vinistock: Yeah, agreed. We need to think about how we could collect errors on a per listener basis.

PR #1014, opened 2023-09-17
Auto indent after end keyword

### LINKED ISSUE #272: Auto format indentation on type
When wrapping some code in a block, like adding a conditional around it, we can use `onTypeFormatting` to automatically fix the indentation of the body.

For example,
```ruby
# Original

puts "a"

# Wrapped
if true
puts "a"
end # <<- the moment the user finishes typing this end, we can determine the if statement's body
    # and push everything two characters forward
```

Same thing for any other code that can have a block inside, like `do`, `class`, `module`, loops, etc.

giovannism20:
### Motivation

Closes #272 

### Implementation

I wrote something based on actual code and not too complex. Basically, I just verified the exact position and back to the place where the indentation was wrong, adjusted it, and then back to the final position.

### Automated Tests

Added test case for formatting indentation.


lib/ruby_lsp/requests/on_type_formatting.rb
```diff
@@ -171,6 +175,25 @@ def find_indentation(line)
...
+
+        return unless start_line_index
+
+        start_line_indentation = find_indentation(@lines[start_line_index])
+
+        (start_line_index + 1...@position[:line] - 1).each do |i|
+          add_edit_with_text("  ", { line: i, character: 0 })
+        end
+
+        move_cursor_to(@position[:line] - 1, 3)
```
vinistock, line 193: Should we move back to where the `end` token finished? Where is this intended to land?

```suggestion
        move_cursor_to(@position[:line], @position[:character])
```

lib/ruby_lsp/requests/on_type_formatting.rb
```diff
@@ -171,6 +175,25 @@ def find_indentation(line)
...
+
+        add_edit_with_text("\n" + (" " * (start_line_indentation + 2)), @position)
```
vinistock, line 195: If I understand correctly, the code above that does `(start_line_index + 1...@position[:line] - 1).each` is the part pushing the indentation.

What's this line pushing forward? The next line after the `end`?

vinistock (commented):
This is looking great!

lib/ruby_lsp/requests/on_type_formatting.rb
giovannism20: Not only the next line after the end but all lines between "end" and current position

vinistock:
We need to add `d` as a completion character in the on type formatter provider configuration [here](https://github.com/Shopify/ruby-lsp/blob/18f8610d02297dd545bcb67d870696a754bc4f88/lib/ruby_lsp/executor.rb#L622).

Also, can we change the implementation to use node locating? I think it's going to be more robust. Something like this

```ruby
# ...
        when "d"
          auto_indent_after_end_keyword
# ...

      def auto_indent_after_end_keyword
        # We need to look for the end in the current line, not the previous one
        current_line = @lines[@position[:line]]
        return unless current_line&.strip == "end"

        # Instead of relying on regexes, we can try to find a Statements node on top
        # of the triggered position
        target, _parent, _nesting = T.cast(
          @document.locate_node(@position, node_types: [YARP::StatementsNode]),
          [T.nilable(YARP::StatementsNode), T.nilable(YARP::Node), T::Array[String]],
        )
        return unless target

        # If there is a statements node, then we want to push every node inside body
        # two characters forward
        target.body.each do |node|
          add_edit_with_text(...)
        end

        move_cursor_to(@position[:line], @position[:character])
      end
```

giovannism20:
Ok, I'll change it!

giovannism20:
Should I close this pull request ?
@vinistock

lib/ruby_lsp/requests/on_type_formatting.rb
```diff
@@ -171,6 +173,27 @@ def find_indentation(line)
 
         count
       end
+
+      sig { void }
+      def auto_indent_after_end_keyword
```
vinistock, line 178:
```ruby
      def auto_indent_after_end_keyword
        current_line = @lines[@position[:line]]
        return unless current_line&.strip == "end"

        target, _parent, _nesting = @document.locate_node({
          line: @position[:line],
          character: @position[:character] - 1,
        })

        statements = case target
        when Prism::IfNode, Prism::UnlessNode, Prism::ForNode, Prism::WhileNode, Prism::UntilNode
          target.statements
        end
        return unless statements

        statements.body.each do |node|
          loc = node.location
          next unless loc.start_column == @indentation

          add_edit_with_text("  ", { line: loc.start_line - 1, character: 0 })
        end

        move_cursor_to(@position[:line], @position[:character])
      end
```

test/requests/on_type_formatting_test.rb
```diff
@@ -363,4 +363,32 @@ def test_breaking_line_immediately_after_keyword
 
     assert_equal(expected_edits.to_json, T.must(edits).to_json)
   end
+
+  def test_auto_indent_after_end_keyword
```
vinistock, line 367:
```ruby
 def test_auto_indent_after_end_keyword
    document = RubyLsp::RubyDocument.new(source: +"if foo\nbar\nend", version: 1, uri: URI("file:///fake.rb"))
    edits = RubyLsp::Requests::OnTypeFormatting.new(document, { line: 2, character: 2 }, "d").run

    expected_edits = [
      {
        range: { start: { line: 1, character: 0 }, end: { line: 1, character: 0 } },
        newText: "  ",
      },
      {
        range: { start: { line: 2, character: 2 }, end: { line: 2, character: 2 } },
        newText: "$0",
      },
    ]

    assert_equal(expected_edits.to_json, T.must(edits).to_json)
  end
```

lib/ruby_lsp/requests/on_type_formatting.rb
```diff
@@ -185,6 +187,31 @@ def find_indentation(line)
 
         count
       end
+
+      def auto_indent_after_end_keyword
```
vinistock, line 191: You still need the Sorbet signature though. Typechecking won't pass without. It should be `sig { void }`.

vinistock:
Linting failed. You can fix that by running `bundle exec rubocop -a`.

vinistock (approved):
Thank you very much for the contribution!

Issue #908, opened 2023-08-18
Handle constant aliases in the indexer


vinistock:
Currently, the indexer does not properly take aliases into account. I think the best way of storing this is to create an `Alias` entry type that basically just proxies to the real entry. That way we know it's an alias, but also know what it resolves to.

One concern is how to properly handle if we discover the alias before we discover the aliased constant. For example

```ruby
# file 1. Gets indexed first
Alias = SomeClass

# file 2. Gets indexed after
class SomeClass
end
```

For this case, we will not know which constant is proxied by the alias when we discover it. We may need to make the `Alias` entry just keep the name for which it aliases to and then resolve it once someone tries to access it, when the index is fully built.

vinistock:
Yes, you should, but that's not enough to fully handle aliases.

For example, in the Ruby LSP's own code, we alias `RubyLsp::Interface` to `LanguageServer::Protocol::Interface`. This means that when we do this

```ruby
module RubyLsp
  Interface::Hover
end
```
We need to be able to resolve `Interface::Hover` to the correct class, which is `LanguageServer::Protocol::Interface::Hover`. For simpler cases, where you're just referencing the alias constant directly, there's no need to know what it proxies to. But these more complex cases are currently unhandled.

Morriar:
Ah, makes sense.

Note that constant definitions can be conditional, for example when depending on architecture or or dependencies:

```rb
if rand(0..1) # could be for a specific arch or if a dependency is loaded
   X = String
else
   X = Integer
end
```

vinistock:
Yeah, I'm not sure how well we can handle condition declarations like those. Without a proper typechecker, we'll probably not be able to understand cases that aren't trivial. We might be better off ignoring those if they are inside conditionals.

PR #1023, opened 2023-09-18
Handle constant aliases and ability to follow them

### LINKED ISSUE #916: Indexer: Handle other operator writers, e.g. `&&=`
_Originally posted by @vinistock in https://github.com/Shopify/ruby-lsp/issues/913#issuecomment-1686925762_
 
> I think we can focus this PR on fixing the breaking changes, but we should indeed consider these other assignment operators as other entries for the same constant.

vinistock:
### Motivation

Closes #916

First step for #908 

We need to handle constant aliases properly, so that we can provide go to definition, completion and hover for them.

I'd appreciate feedback on the proposed API here.

### Implementation

The idea of the implementation is the following:

- When we find a constant alias, there's a chance we haven't indexed the target constant yet, so we can't eagerly try to resolve it
- Instead, we place `UnresolvedAlias` entries in the index, which remember the target name and the nesting where the alias was found
- Then, I'm proposing the addition of `follow_alias`, which does two things:
    - Turns `UnresolvedAlias` entries into a concrete `Alias`. This is done by taking the target and nesting and then resolving to what target the alias refers to, since at this point we finished indexing
    - Then we loop through all of the concrete aliases to figure out which constant the alias ultimately points to (since Ruby allows you to have an alias that points to another alias)

### Automated Tests

Added tests that show the behaviour of the proposed API.

### DIFF
--- a/lib/ruby_indexer/lib/ruby_indexer/index.rb
+++ b/lib/ruby_indexer/lib/ruby_indexer/index.rb
@@ -5,6 +5,8 @@ module RubyIndexer
   class Index
     extend T::Sig
 
+    class UnresolvableAliasError < StandardError; end
+
     # The minimum Jaro-Winkler similarity score for an entry to be considered a match for a given fuzzy search query
     ENTRY_SIMILARITY_THRESHOLD = 0.7
 
@@ -125,13 +127,23 @@ module RubyIndexer
     # 3. Baz
     sig { params(name: String, nesting: T::Array[String]).returns(T.nilable(T::Array[Entry])) }
     def resolve(name, nesting)
-      (nesting.length + 1).downto(0).each do |i|
-        prefix = T.must(nesting[0...i]).join("::")
-        full_name = prefix.empty? ? name : "#{prefix}::#{name}"
-        entries = @entries[full_name]
-        return entries if entries
+      nesting.length.downto(0).each do |i|
+        namespace = T.must(nesting[0...i]).join("::")
+        full_name = namespace.empty? ? name : "#{namespace}::#{name}"
+
+        # If we find an entry with `full_name` directly, then we can already return it, even if it contains aliases -
+        # because the user might be trying to jump to the alias definition.
+        #
+        # However, if we don't find it, then we need to search for possible aliases in the namespace. For example, in
+        # the LSP itself we alias `RubyLsp::Interface` to `LanguageServer::Protocol::Interface`, which means doing
+        # `RubyLsp::Interface::Location` is allowed. For these cases, we need some way to realize that the
+        # `RubyLsp::Interface` part is an alias, that has to be resolved
+        entries = @entries[full_name] || @entries[follow_aliased_namespace(full_name)]
+        return entries.map { |e| e.is_a?(Entry::UnresolvedAlias) ? resolve_alias(e) : e } if entries
       end
 
+      nil
+    rescue UnresolvableAliasError
       nil
     end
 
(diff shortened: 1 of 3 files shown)


lib/ruby_indexer/lib/ruby_indexer/visitor.rb
```diff
@@ -78,28 +78,33 @@ def handle_private_constant(node)
...
-      @index << Index::Entry::Constant.new(fully_qualify_name(node.name.to_s), @file_path, node.location, comments)
-    end
+      value = node.value
+      name = fully_qualify_name(node.name.to_s)
 
-    sig do
-      params(
-        node: T.any(YARP::ConstantPathWriteNode, YARP::ConstantPathOrWriteNode),
-      ).void
+      @index << if value.is_a?(YARP::ConstantReadNode) || value.is_a?(YARP::ConstantPathNode)
```
Morriar, line 87: What about `FOO = BAR = 42` should this be considered as an alias?

lib/ruby_indexer/lib/ruby_indexer/index.rb
```diff
@@ -199,6 +251,47 @@ class Class < Namespace
...
+          params(
+            target: String,
+            nesting: T::Array[String],
+            name: String,
+            file_path: String,
+            location: YARP::Location,
+            comments: T::Array[String],
+          ).void
+        end
+        def initialize(target, nesting, name, file_path, location, comments) # rubocop:disable Metrics/ParameterLists
```
Morriar, line 274: It's a lot of parameters, I wonder if we should use keywords instead? It may require some refactoring.

lib/ruby_indexer/lib/ruby_indexer/index.rb
```diff
@@ -199,6 +251,47 @@ class Class < Namespace
...
+        end
+      end
+
+      class Alias < Entry
+        extend T::Sig
+
+        sig { returns(String) }
+        attr_reader :target
+
+        sig { params(target: String, unresolved_alias: UnresolvedAlias).void }
```
Morriar, line 288: Since the alias is "resolved", why not having it pointing to the constant directly? It would make following the aliases easier.

lib/ruby_indexer/lib/ruby_indexer/index.rb
```diff
@@ -152,6 +185,25 @@ def index_single(indexable_path, source = nil)
       # If `path` is a directory, just ignore it and continue indexing
     end
 
+    private
+
+    # Resolves an alias updating the index with a concrete alias entry that points to the fully qualified target
+    sig { params(alias_entries: T::Array[Entry::UnresolvedAlias]).returns(T.nilable(T::Array[Entry])) }
```
Morriar, line 191: Why do we pass an array rather than a single one?

lib/ruby_indexer/lib/ruby_indexer/index.rb
vinistock: Let's do the refactor in a different PR, since it will require quite a few changes.

lib/ruby_indexer/lib/ruby_indexer/index.rb
vinistock: Because we may have multiple entries for the same name. Having the resolved fully qualified name allows us to fetch those entries that are stored in the index.

lib/ruby_indexer/lib/ruby_indexer/index.rb
```diff
@@ -152,6 +152,28 @@ def index_single(indexable_path, source = nil)
       # If `path` is a directory, just ignore it and continue indexing
     end
 
+    private
+
+    # Attempts to resolve an UnresolvedAlias into a resolved Alias. If the unresolved alias is pointing to a constant
+    # that doens't exist, then we return the same UnresolvedAlias
```
andyw8, line 158:
```suggestion
    # that doesn't exist, then we return the same UnresolvedAlias
```

lib/ruby_indexer/lib/ruby_indexer/index.rb
```diff
@@ -152,6 +152,28 @@ def index_single(indexable_path, source = nil)
...
+    sig { params(entry: Entry::UnresolvedAlias).returns(T.any(Entry::Alias, Entry::UnresolvedAlias)) }
+    def resolve_alias(entry)
+      target = resolve(entry.target, entry.nesting)
+      return entry unless target
+
+      target_name = T.must(target.first).name
+      resolved_alias = Entry::Alias.new(target_name, entry)
+
+      # Replace the UnresolvedAlias by a resolved one so that we don't have to do this again later
+      original_entries = T.must(@entries[entry.name])
```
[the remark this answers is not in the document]
vinistock: I'm really not a fan of that style, to be honest.

lib/ruby_indexer/lib/ruby_indexer/visitor.rb
```diff
@@ -36,9 +36,16 @@ def visit(node)
...
+        name = fully_qualify_name(node.name.to_s)
+        add_constant(node, name)
+      when YARP::ConstantPathWriteNode, YARP::ConstantPathOrWriteNode, YARP::ConstantPathOperatorWriteNode,
+        YARP::ConstantPathAndWriteNode
+
+        # ignore variable constants like `var::FOO` or `self.class::FOO`
+        return unless node.target.parent.nil? || node.target.parent.is_a?(YARP::ConstantReadNode)
+
+        name = fully_qualify_name(node.target.location.slice)
+        add_constant(node, name)
```
Morriar, line 48: Are we missing the handling of constant target nodes? `A, B = 42` and `FOO::A, FOO::B = 42`?

Morriar (approved):

PR #1032, opened 2023-09-21
Use a needs_parsing boolean instead of unparsed edits array

vinistock:
### Motivation

I was looking at this code and there's really no reason why we need the `unparsed_edits` array. We only use it to signal that there has been a new edit to a document and that it needs to be parsed, which can be achieve by just using a boolean.

### Implementation

Switched to use a boolean to indicate whether we need to parse the document or not.

### Automated Tests

Our existing test was weak, so I improved the verification to make sure we aren't parsing if no edits were made.

### DIFF
--- a/lib/ruby_lsp/document.rb
+++ b/lib/ruby_lsp/document.rb
@@ -28,7 +28,7 @@ module RubyLsp
       @source = T.let(source, String)
       @version = T.let(version, Integer)
       @uri = T.let(uri, URI::Generic)
-      @unparsed_edits = T.let([], T::Array[EditShape])
+      @needs_parsing = T.let(false, T::Boolean)
       @parse_result = T.let(YARP.parse(@source), YARP::ParseResult)
     end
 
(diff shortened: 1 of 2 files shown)


lib/ruby_lsp/document.rb
```diff
@@ -87,15 +87,15 @@ def push_edits(edits, version:)
...
       @cache.clear
     end
 
     sig { void }
     def parse
-      return if @unparsed_edits.empty?
+      return unless @needs_parsing
 
-      @unparsed_edits.clear
+      @needs_parsing = false
```
andyw8, line 98: Instead of the instance variable, would it be feasible to set `@parse_result` to nil in `push_edits` and use that to determine whether we need to parse again?

vinistock: That would be risky since the LSP is multi-threaded. If we switch to a different thread that is executing another request and requires the parse result, then it would break.

PR #1034, opened 2023-09-22
Index constant targets

vinistock:
### Motivation

Address https://github.com/Shopify/ruby-lsp/pull/1023#discussion_r1334414933

Handle constant target node indexing. E.g.:

```ruby
A, B = [1, 2]
```

### Implementation

Because the assignment may be pointing to another constant (creating an alias), we need to handle the `MultiWriteNode` to also check the possible values.

I went with only trying to figure out the right value assigned to each constant on array literals since other nodes are considerably more difficult.

### Automated Tests

Added a test that displays the behaviour.

### DIFF
--- a/lib/ruby_indexer/lib/ruby_indexer/visitor.rb
+++ b/lib/ruby_indexer/lib/ruby_indexer/visitor.rb
@@ -30,6 +30,26 @@ module RubyIndexer
       add_index_entry(node, Index::Entry::Module)
     end
 
+    sig { override.params(node: YARP::MultiWriteNode).void }
+    def visit_multi_write_node(node)
+      value = node.value
+      values = value.is_a?(YARP::ArrayNode) && value.opening_loc ? value.elements : []
+
+      node.targets.each_with_index do |target, i|
+        current_value = values[i]
+        # The moment we find a splat on the right hand side of the assignment, we can no longer figure out which value
+        # gets assigned to what
+        values.clear if current_value.is_a?(YARP::SplatNode)
+
+        case target
+        when YARP::ConstantTargetNode
+          add_constant(target, fully_qualify_name(target.name.to_s), current_value)
+        when YARP::ConstantPathTargetNode
+          add_constant(target, fully_qualify_name(target.slice), current_value)
+        end
+      end
+    end
+
     sig { override.params(node: YARP::ConstantPathWriteNode).void }
     def visit_constant_path_write_node(node)
       # ignore variable constants like `var::FOO` or `self.class::FOO`
@@ -138,12 +158,15 @@ module RubyIndexer
           YARP::ConstantPathOrWriteNode,
           YARP::ConstantPathOperatorWriteNode,
           YARP::ConstantPathAndWriteNode,
+          YARP::ConstantTargetNode,
+          YARP::ConstantPathTargetNode,
         ),
         name: String,
+        value: T.nilable(YARP::Node),
       ).void
     end
-    def add_constant(node, name)
-      value = node.value
+    def add_constant(node, name, value = nil)
+      value = node.value unless node.is_a?(YARP::ConstantTargetNode) || node.is_a?(YARP::ConstantPathTargetNode)
       comments = collect_comments(node)
 
       @index << case value
--- a/lib/ruby_indexer/test/constant_test.rb
+++ b/lib/ruby_indexer/test/constant_test.rb
@@ -324,5 +324,68 @@ module RubyIndexer
       assert_entry("F::G", Index::Entry::Constant, "/fake/path/foo.rb:4-0:4-10")
       assert_entry("H::I", Index::Entry::Constant, "/fake/path/foo.rb:5-0:5-9")
     end
+
+    def test_indexing_constant_targets
+      index(<<~RUBY)
+        module A
+          B, C = [1, Y]
+          D::E, F::G = [Z, 4]
+          H, I::J = [5, B]
+          K, L = C
+        end
+
+        module Real
+          Z = 1
+          Y = 2
+        end
+      RUBY
+
+      assert_entry("A::B", Index::Entry::Constant, "/fake/path/foo.rb:1-2:1-3")
+      assert_entry("A::C", Index::Entry::UnresolvedAlias, "/fake/path/foo.rb:1-5:1-6")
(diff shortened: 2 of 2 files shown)


lib/ruby_indexer/lib/ruby_indexer/visitor.rb
```diff
@@ -46,6 +46,18 @@ def visit(node)
 
         name = fully_qualify_name(node.target.location.slice)
         add_constant(node, name)
+      when YARP::MultiWriteNode
```
Morriar, line 49: I guess we need this gymnastic of accessing the node targets manually rather than visiting down the tree just to access the value?

I wonder if our simplified visitor could descend into two ConstantWriteNode to make it easier to handle for free?

lib/ruby_indexer/test/constant_test.rb
```diff
@@ -306,5 +306,24 @@ module A
       constant = @index["A::M"].first
       assert_instance_of(Index::Entry::Constant, constant)
     end
+
+    def test_indexing_constant_targets
+      index(<<~RUBY)
+        module A
+          B, C = [1, Y]
+          D::E, F::G = [Z, 4]
```
Morriar, line 314: Let's mix a constant and a constant path in the same multi node:

```suggestion
          D::E, F::G = [Z, 4]
          H, I::J = B, C
```

lib/ruby_indexer/lib/ruby_indexer/visitor.rb
```diff
@@ -46,6 +46,18 @@ def visit(node)
...
+        value = node.value
+        values = value.is_a?(YARP::ArrayNode) ? value.elements : Array.new(node.targets.length, value)
```
Morriar, line 51: I think you can write this and not have an ArrayNode: `A, B = C`

vinistock: Yes. Added a test for this as well. In that case, `A` is an `UnresolvedAlias` and `B` is just a declared constant for which the value is `nil`.

lib/ruby_indexer/lib/ruby_indexer/visitor.rb
vinistock: Yes, it's because we need the value. That would simplify the code a bit indeed, but we'd need to be careful about missing values.

E.g.:
```ruby
# In this case, the first ConstantWrite visit would have value C and the second nil
A, B = C
```

lib/ruby_indexer/lib/ruby_indexer/visitor.rb
```diff
@@ -46,6 +46,18 @@ def visit(node)
 
         name = fully_qualify_name(node.target.location.slice)
         add_constant(node, name)
+      when YARP::MultiWriteNode
+        value = node.value
+        values = value.is_a?(YARP::ArrayNode) ? value.elements : [value]
+
+        node.targets.each_with_index do |target, i|
```
kddnewton, line 53: This is unfortunately only going to work in a subset of cases, and going to be wrong in many. Consider these:

```ruby
Foo, *, Bar = baz
Foo, = bar
(Foo, Bar) = baz

Foo, Bar = *baz, qux
Foo, Bar = [baz, *qux]
```

All of these are wrong at the moment, and these are just off the top of my head. I'm sure there are more edge cases here to consider.

If you want to handle the subset of cases that it looks like you're trying to, you will need to validate that:

* the RHS is an array literal with no splats
* the LHS has no parentheses or splats
* the number of targets and values lines up 1:1

kddnewton (changes requested):

lib/ruby_indexer/lib/ruby_indexer/visitor.rb
vinistock: We have both of those cases in our tests. Unless I'm missing something, they should be working as we expect.

The first one is not a constant target, so all we do is check if the right hand side is a constant literal. The second case, `Foo` will be an alias to `Baz` and `Bar` will be `nil`, which works in the current setup.

kddnewton: No I mean in this case `Foo = 1` and `Bar = 2`. `Foo` is not equal to `Baz` in this case.

vinistock: Oh, sorry. I had our test style in mind when reading the example and thought each line was a different case. I now see what you meant to say.

Indeed, there's no way to handle proper destructuring like that unless it's an array literal. I changed the code so that in your example, `Baz`, `Foo` and `Bar` and all considered constants - since the right hand side is not an array literal.

kddnewton (changes requested):

lib/ruby_indexer/lib/ruby_indexer/visitor.rb
vinistock: I think the implementation is conservative enough. I think it's significantly more likely that we'd miss constants rather than adding confusing results. We only consider the right hand side if it's an array literal with no splats - and we only do so to check if there's a constant alias inside.

For all other cases, we ignore the right hand side and just remember where the constant was declared, which allows us to provide definition, hover, completion, etc.

I would rather start like this and deal with corner cases we encounter, rather than miss the constant declarations.

lib/ruby_indexer/lib/ruby_indexer/visitor.rb
```diff
@@ -46,6 +46,23 @@ def visit(node)
...
+          current_value = values[i]
+          # The moment we find a splat on the right hand side of the assignment, we can no longer figure out which value
+          # gets assigned to what
+          values.clear if current_value.is_a?(YARP::SplatNode)
+
+          case target
+          when YARP::ConstantTargetNode
+            add_constant(target, fully_qualify_name(target.name.to_s), current_value)
+          when YARP::ConstantPathTargetNode
+            add_constant(target, fully_qualify_name(target.slice), current_value)
```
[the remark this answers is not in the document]
vinistock: https://github.com/ruby/prism/pull/1619

lib/ruby_indexer/lib/ruby_indexer/visitor.rb
kddnewton: Currently you're not considering the right hand side only if it's an array literal. Consider:

```ruby
foo = bar, baz
```

that gives you an array node on the RHS. If you want to only consider array literals you need to check if `opening_loc` is set on it.

PR #1039, opened 2023-09-25
Remove dependency on syntax_tree

vinistock:
### Motivation

I noticed that we forgot to remove our dependency on Syntax Tree during the YARP migration. Now that we use YARP for the foundational part of the functionality, we should treat Syntax Tree like any other formatter.

If the gem is available in the bundle, we use register it as a possible formatter.

### Implementation

Used the same technique we use for RuboCop. We try to require the gem. If it's there, it becomes an available formatter. Otherwise, we just ignore it.

I removed the Syntax Tree dependency and moved it to the Gemfile, since we need it for testing purposes.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -23,4 +23,6 @@ group :development do
   gem "tapioca", "~> 0.11", require: false, platforms: NON_WINDOWS_PLATFORMS
   gem "rdoc", require: false
   gem "psych", "~> 5.1", require: false
+
+  gem "syntax_tree", ">= 6.1.1", "< 7"
 end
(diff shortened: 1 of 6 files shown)


Morriar (commented):
Should we also remove https://github.com/Shopify/ruby-lsp/blob/main/sorbet/rbi/gems/syntax_tree%406.1.1.rbi?

Morriar (approved):
> That's not yet possible because Spoom depends on Syntax Tree, so the RBI would come back.

Oh good one 👍

PR #1041, opened 2023-09-27
Remove overriding `visit` and `on_node`

kddnewton:
### Motivation

In the next version of YARP, I'm splitting up `Visitor` versus `Compiler`. The former is a visitor that defines each of the visit methods as:

```ruby
def visit_child_nodes(node)
  node.compact_child_nodes.each { |node| node.accept(self) }
end
```

the latter is a visitor that defines each of the visit methods as:

```ruby
def visit_child_nodes(node)
  node.compact_child_nodes.map { |node| node.accept(self) }
end
```

notably neither of these methods call through the `visit` method. This is because it's actually quite a bit faster to skip that step, since that's going to just immediately call `accept` anyway.

Previously, there was an implicit assumption within `ruby-lsp` that you could override `#visit` and it would pass through that method every time a node was going to be visited. Because that implicit assumption will no longer be true, this will break and be hard to fix. I started out by trying to show how to work around this, and ended up just implementing it.

The other piece of this is the `on_node` event emitter event. I don't want to support that event in YARP because it slows down all of the dispatchers by attempting to fire events on every node. Basically this ended up splitting up a lot of overall visits into specialized visits.

### Implementation

`lib/ruby_indexer/lib/ruby_indexer/visitor.rb` - instead of overriding the main `visit` method, we override the specific visit methods. This one was relatively straight-forward, splitting up the `case` statement into each specialized visit method.
`lib/ruby_lsp/requests/document_highlight.rb` - instead of using the `on_node` method, specialize the request based on the node that is currently being highlighted. We do this by creating a different kind of listener for each different kind of highlight. Each of them highlights based on the value of the wrapped node. This also makes it much more explicit which nodes are being highlighted per target. This also fixes some subtle bugs, like constant paths not being highlighted because of whitespace (`Foo::Bar` wouldn't match `Foo :: Bar`, but now it does) or method names being highlighted on constants (`def Foo()` would match `Foo`).
`lib/ruby_lsp/requests/folding_ranges.rb` - removed `on_node` which wasn't being fired.
`lib/ruby_lsp/requests/selection_ranges.rb` - instead of overriding the main `visit` method, use a queue to inline the previous recursive function.

### Automated Tests

Since this is a refactor, I assume the test coverage is sufficient to cover this. I hope that's true?

### Manual Tests

To manually test, it will be jump-to-definition, document highlight, and selection ranges.


lib/ruby_indexer/lib/ruby_indexer/visitor.rb
```diff
@@ -60,6 +79,26 @@ def visit_all(nodes)
...
+    sig do
+      params(node: T.any(
+        YARP::ConstantPathWriteNode,
+        YARP::ConstantPathAndWriteNode,
+        YARP::ConstantPathOrWriteNode,
+        YARP::ConstantPathOperatorWriteNode,
+      )).void
+    end
+    def add_constant_path_write_node(node)
+      return unless node.target.parent.nil? || node.target.parent.is_a?(YARP::ConstantReadNode)
```
vinistock, line 96: Let's maintain the comment.
```suggestion
      # ignore variable constants like `var::FOO` or `self.class::FOO`
      return unless node.target.parent.nil? || node.target.parent.is_a?(YARP::ConstantReadNode)
```

lib/ruby_lsp/requests/folding_ranges.rb
```diff
@@ -166,11 +166,6 @@ def on_begin(node)
         add_simple_range(node)
       end
 
-      sig { params(node: YARP::Node).void }
-      def on_node(node)
```
vinistock, line 170: In order to group requires for a folding range, we need to accumulate call nodes that match `require` or `require_relative`. We keep expanding the range until we find any other type of node, which is on this removed `on_node` method. Then we emit the range.

We might be missing a test for this. Is it working for this case?

```ruby
require "foo"
require "bar"

class Baz
  def something; end
end

require "qux"
require "zod"
```
In that example, you should be have 3 folding ranges:
1. The two requires for foo and bar
2. The class
3. The two other requires for qux and zod

lib/ruby_lsp/requests/document_highlight.rb
```diff
@@ -3,86 +3,739 @@
...
+        sig do
+          type_parameters(:ResponseType)
+            .params(
+              target: T.nilable(YARP::Node),
+              parent: T.nilable(YARP::Node),
+              emitter: EventEmitter,
+              message_queue: Thread::Queue,
+            ).returns(Listener[T::Array[Interface::DocumentHighlight]])
+        end
+        def for(target, parent, emitter, message_queue)
```
vinistock, line 708: We've been keeping all request implementation consistent, with

```ruby
Request.new(params).run
```

So, we'd need to keep the consistency here too.

lib/ruby_lsp/requests/document_highlight.rb
```diff
@@ -3,86 +3,739 @@
...
+          case target
```
vinistock, line 709: I'm a bit unsure about the document highlight refactor. Our current structure really doesn't account for having multiple listeners for the same request, as you probably notice when repeating the same documentation multiple times. And we still end up with a large case/when to select the right listener.

I think we can probably achieve a good refactor while keeping a single listener if we just focus on getting rid of `on_node`. I think it'd be fine to have the listener be instantiated with the target and then handle all of the different events where we look for highlights.

PR #1042, opened 2023-09-27
Refactor dependency detector

andyw8:
### Motivation

The current DependencyDetector is somewhat awkward to work with, and because it cannot be instantiated, we are repeated several checks unnecessarily.

### Implementation

Convert into a class. Move its use up to the `executor` layer.

### Automated Tests

Updated

### Manual Tests

Verified locally.

### DIFF
--- a/lib/ruby_lsp/requests/completion.rb
+++ b/lib/ruby_lsp/requests/completion.rb
@@ -30,15 +30,17 @@ module RubyLsp
         params(
           index: RubyIndexer::Index,
           nesting: T::Array[String],
+          type_checker: T::Boolean,
           emitter: EventEmitter,
           message_queue: Thread::Queue,
         ).void
       end
-      def initialize(index, nesting, emitter, message_queue)
+      def initialize(index, nesting, type_checker, emitter, message_queue)
         super(emitter, message_queue)
         @_response = T.let([], ResponseType)
         @index = index
         @nesting = nesting
+        @type_checker = type_checker
 
         emitter.register(self, :on_string, :on_constant_path, :on_constant_read)
       end
@@ -53,7 +55,7 @@ module RubyLsp
       # Handle completion on regular constant references (e.g. `Bar`)
       sig { params(node: YARP::ConstantReadNode).void }
       def on_constant_read(node)
-        return if DependencyDetector::HAS_TYPECHECKER
+        return if @type_checker
 
         name = node.slice
         candidates = @index.prefix_search(name, @nesting)
@@ -65,7 +67,7 @@ module RubyLsp
       # Handle completion on namespaced constant references (e.g. `Foo::Bar`)
       sig { params(node: YARP::ConstantPathNode).void }
       def on_constant_path(node)
-        return if DependencyDetector::HAS_TYPECHECKER
+        return if @type_checker
 
         name = node.slice
 
--- a/lib/ruby_lsp/requests/completion.rb
+++ b/lib/ruby_lsp/requests/completion.rb
@@ -30,17 +30,16 @@ module RubyLsp
         params(
           index: RubyIndexer::Index,
           nesting: T::Array[String],
-          typechecker: T::Boolean,
           emitter: EventEmitter,
           message_queue: Thread::Queue,
         ).void
       end
-      def initialize(index, nesting, typechecker, emitter, message_queue)
+      def initialize(index, nesting, emitter, message_queue)
         super(emitter, message_queue)
         @_response = T.let([], ResponseType)
         @index = index
         @nesting = nesting
-        @typechecker = typechecker
+        @typechecker = T.let(DependencyDetector.instance.typechecker?, T::Boolean)
 
         emitter.register(self, :on_string, :on_constant_path, :on_constant_read)
       end
--- a/lib/ruby_lsp/requests/completion.rb
+++ b/lib/ruby_lsp/requests/completion.rb
@@ -53,7 +53,7 @@ module RubyLsp
       # Handle completion on regular constant references (e.g. `Bar`)
       sig { params(node: YARP::ConstantReadNode).void }
       def on_constant_read(node)
-        return if DependencyDetector.instance.typechecker?
+        return if DependencyDetector.instance.typechecker
 
         name = node.slice
         candidates = @index.prefix_search(name, @nesting)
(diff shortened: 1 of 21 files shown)


lib/ruby_lsp/requests/hover.rb
```diff
@@ -71,21 +71,21 @@ def merge_response!(other)
 
       sig { params(node: YARP::ConstantReadNode).void }
       def on_constant_read(node)
-        return if DependencyDetector::HAS_TYPECHECKER
+        return if DependencyDetector.new.typechecker?
```
andyw8, line 74: Instead of this,  we can pass in `typechecker?` as a boolean from `Executor`.

lib/ruby_lsp/requests/definition.rb
```diff
@@ -26,25 +26,29 @@ class Definition < ExtensibleListener
       sig { override.returns(ResponseType) }
       attr_reader :_response
 
+      # rubocop:disable Metrics/ParameterLists
```
andyw8, line 29: After this, we are thinking of introduce some kind of context object to group common parameters together, and reduce the number of things we need to pass around.

test/test_helper.rb
```diff
@@ -25,7 +25,14 @@
 
 module Minitest
   class Test
+    extend T::Sig
+
     Minitest::Test.make_my_diffs_pretty!
+
+    sig { void }
+    def stub_typechecking
```
andyw8, line 33: (not sure if this name is clear enough)

lib/ruby_lsp/requests/support/dependency_detector.rb
```diff
@@ -2,54 +2,64 @@
...
-        else
-          "none"
-        end
-      end
+  class DependencyDetector
+    extend T::Sig
+
+    sig { void }
+    def initialize
+      @dependency_keys = T.let(nil, T.nilable(T::Array[String]))
```
vinistock, line 10: I think we can move a lot of things into this initializer and eagerly discover everything we need, so that we don't have to constantly grep on an array.

For example

```ruby
keys = dependency_keys

@detected_formatter = ... # check which formatter exists and set it
@has_typechecker = ...
@test_framework = ...
```

I also think this class might be safe enough to turn into a singleton. That way, we don't have to worry about passing it as arguments. E.g.:

```ruby
class DependencyDetector
  include Singleton
  # ...
end

# In requests
DependencyDetector.instance.has_typechecker?
```

vinistock (commented):
This is looking good. One thing about naming conventions: I think both `type checker` and `typechecker` are correct, but we usually prefer `typechecker`.

This PR is mixing both a little bit. Can we standardize on `typechecker`?

lib/ruby_lsp/requests/completion.rb
```diff
@@ -39,6 +39,7 @@ def initialize(index, nesting, emitter, message_queue)
         @_response = T.let([], ResponseType)
         @index = index
         @nesting = nesting
+        @typechecker = T.let(DependencyDetector.instance.typechecker?, T::Boolean)
```
andyw8, line 42: I'm wondering whether it's worth storing this in an instance var, or just calling the singleton each time.

sorbet/rbi/shims/singleton.rbi
```diff
@@ -0,0 +1,5 @@
+# typed: true
+
+module Singleton
```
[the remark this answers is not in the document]
andyw8: (just noticed we all use this in `formatting_test_runner` but mark it unsafe)

test/test_helper.rb
```diff
@@ -25,7 +25,19 @@
...
+      RubyLsp::DependencyDetector.instance.stubs(:typechecker?).returns(false)
+    end
+
+    sig { void }
+    def teardown
+      Singleton.__init__(RubyLsp::DependencyDetector)
+    end
```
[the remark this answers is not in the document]
andyw8: > Are there too many tests that depend on the state of the singleton being reset? Can we reset the singleton on each test that requires it instead?

We could do that, but I think has a similar problem that if we introduce a new test, then we need to remember to add that teardown. And since it's order dependant, CI may still pass.

Another approach would be to switch the tests to use `ActiveSupport::TestCase`. That provides a `teardown {}` method that takes a block, which automatically calls `super`. I know it's another dependency to add, but it would only be required for tests.

lib/ruby_lsp/requests/support/dependency_detector.rb
```diff
@@ -1,55 +1,80 @@
...
+
+    sig { returns(String) }
+    attr_reader :detected_formatter
+
+    sig { returns(String) }
+    attr_reader :detected_test_library
+
+    sig { void }
+    def initialize
+      @dependency_keys = T.let(nil, T.nilable(T::Array[String]))
```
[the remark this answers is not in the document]
andyw8: It's used for the memoization. We _could_ restructure the class to eliminate it, but think it's more readable this way.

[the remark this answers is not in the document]
vinistock: You shouldn't need it here in the initialize if you do this in the memoized method.

```ruby
sig { returns(T::Array[String]) }
def something
  @something ||= T.let(whatever, T.nilable(T::Array[String]))
end
```

sorbet/rbi/shims/singleton.rbi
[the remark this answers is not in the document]
vinistock: Yeah, we can push an RBI change to Sorbet.

test/requests/completion_test.rb
```diff
@@ -198,7 +199,7 @@ def test_completion_is_not_triggered_if_argument_is_not_a_string
   end
 
   def test_completion_for_constants
-    RubyLsp::DependencyDetector.const_set(:HAS_TYPECHECKER, false)
+    stub_no_typechecker
```
vinistock, line 202: Can we do this in the `setup` method and avoid the duplicates?

test/requests/workspace_symbol_test.rb
```diff
@@ -5,14 +5,10 @@
 
 class WorkspaceSymbolTest < Minitest::Test
   def setup
-    RubyLsp::DependencyDetector.const_set(:HAS_TYPECHECKER, false)
+    RubyLsp::DependencyDetector.instance.stubs(typechecker: false)
```
vinistock, line 8: Does invoking the new `stub_typechecker` method work here?

test/test_helper.rb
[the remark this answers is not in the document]
vinistock: IMO, let's not add this teardown. The test for the dependency detector itself can do this in its own teardown.

For all other tests that depend on a specific value being returned by something, we should use stubbing.

st0012 (approved):

PR #1043, opened 2023-09-27
Add support for indexing methods with no parameters

andyw8:

### DIFF
--- a/lib/ruby_indexer/lib/ruby_indexer/index.rb
+++ b/lib/ruby_indexer/lib/ruby_indexer/index.rb
@@ -274,6 +274,52 @@ module RubyIndexer
       class Constant < Entry
       end
 
+      class Method < Entry
+        extend T::Sig
+
+        sig { returns(T::Array[Symbol]) }
+        attr_reader :parameters
+
+        sig do
+          params(
+            name: String,
+            file_path: String,
+            location: YARP::Location,
+            comments: T::Array[String],
+            parameters_node: YARP::ParametersNode,
+          ).void
+        end
+        def initialize(name, file_path, location, comments, parameters_node)
+          super(name, file_path, location, comments)
+          @parameters = T.let(list_params(parameters_node), T::Array[Symbol])
+          @arity = T.let(arity(parameters_node), T::Range[Integer])
+        end
+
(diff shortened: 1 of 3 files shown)


andyw8:
@vinistock I pushed an update to cover some more agument types. The typecheck is failing, it looks like the RBI needs corrected in a few places.

lib/ruby_indexer/lib/ruby_indexer/index.rb
```diff
@@ -280,6 +280,73 @@ class Class < Namespace
...
+        sig { params(name: Symbol).void }
+        def initialize(name:)
+          @name = name
+        end
+      end
+
+      class RequiredParameter < Parameter
+      end
+
+      class Method < Entry
```
Morriar, line 301: Should we add a flag for instance vs. singleton method?

andyw8: We've updated to use separate classes to distinguish each.

lib/ruby_indexer/test/method_test.rb
```diff
@@ -0,0 +1,73 @@
...
+        end
+      RUBY
+
+      assert_entry("bar", Index::Entry::InstanceMethod, "/fake/path/foo.rb:1-2:2-5")
+    end
+
+    def test_singleton_method_using_self_receiver
+      index(<<~RUBY)
+        class Foo
+          def self.bar
```
andyw8, line 22: We're not yet covering using `class << self`, that will be in a follow-up.

lib/ruby_indexer/lib/ruby_indexer/index.rb
```diff
@@ -301,6 +301,81 @@ class Class < Namespace
...
+        sig do
+          params(
+            name: String,
+            file_path: String,
+            location: Prism::Location,
+            comments: T::Array[String],
+            parameters_node: T.nilable(Prism::ParametersNode),
+          ).void
+        end
+        def initialize(name, file_path, location, comments, parameters_node)
```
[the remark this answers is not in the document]
andyw8: Yes, that will be part of the follow-up work.

PR #1045, opened 2023-09-28
Add support for finding class, module and constant references

### LINKED ISSUE #202: Add find references to the `ruby-lsp`
We can use a similar mechanism of #199 and find possible references if we store those in the index. These will not always be correct as type checking is needed to provide accurate results.

Documentation: https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#textDocument_references

vinistock:
### Motivation

Closes #202 

At least from a classes, modules and constants perspective. We can create a new issue for handling methods.

This PR uses the work in #1044 to add support for the references request for classes, modules and constants.

### Implementation

The majority of it is pretty standard for adding new requests. In terms of the implementation, the LSP receives an option inside the `context` parameter about whether to include the declaration as a reference or not.

The implementation:
1. Resolves the constant under the cursor using the index
2. Adds all declaration locations as references if that option is `true`
3. Asks the indexer to find other references to the fully qualified name

### Automated Tests

Added tests.


lib/ruby_indexer/lib/ruby_indexer/references_visitor.rb
```diff
@@ -0,0 +1,63 @@
...
+    end
+
+    sig { override.params(node: YARP::ConstantPathNode).void }
+    def visit_constant_path_node(node)
+      name = node.slice
+      check_reference_for(name, node)
+    end
+
+    sig { override.params(node: YARP::ClassNode).void }
+    def visit_class_node(node)
```
Morriar, line 34: Aren't we going to miss the references in the superclass declaration?

vinistock:
I'm going to put this on hold for the moment. Finding references like this takes around 40 seconds, which also slows down Sorbet's response because VS Code keeps waiting for all LSPs to return before showing results.

I did extensive profiling of our indexing mechanism and the biggest gains we can have come from making Prism instantiate `Prism::Location` and `Prism::Source` lazily (we spend more than 40% of our time creating instances and on GC).

When we can make those instantiations lazily, indexing will be significantly sped up. Then we can make a better decision and maybe even collect references in the first indexing pass.

Issue #1059, opened 2023-10-04
Indexer Gemspec Awareness Fails on Gem::MissingSpecError


jacobdaddario:
### Ruby version

3.2.2

### Code snippet

The issue can be reproduced by running the Ruby LSP on a Rails application that does not download `tzinfo-data` as part of its bundle. This occurs due to the platform check on the dependency in the default Rails gemfile.

### Description

When starting up the Ruby LSP for a project that raises a `Gem::MissingSpecError` on the following line, indexing fails and the indexing request hangs indefinitely.
https://github.com/Shopify/ruby-lsp/blob/1922f0c26ad4d8c3dfd7a3c8ca830a940ab7df95/lib/ruby_indexer/lib/ruby_indexer/configuration.rb#L182

I've confirmed this is the case in the following fashion.
1. Opened ruby-lsp on my local machine and modified the downloaded gem to log progress through the code.
2. Identified that the linked line of code is the last thing to be executed before indexer execution halts.
3. Confirmed that `tzinfo-data` is raising an error when running the line of code in a Rails console.
4. Confirmed that commenting out the dependency allows the indexer to work as normal.
5. Confirmed that rescuing `Gem::MissingSpecError` **with `tzinfo-data` included in the bundle** also works.

### Expected output

The Indexer should not fail during initial indexing. Potential error cases should be handled gracefully.

PR #1060, opened 2023-10-04
Fix Indexing Gemspec Awareness Bug

jacobdaddario:
### Motivation

Solves #1059

Fixes an issue in which the Indexer would hang on startup of the Ruby LSP.

### Implementation

Rescued an error that occurred [here](https://github.com/Shopify/ruby-lsp/blob/1922f0c26ad4d8c3dfd7a3c8ca830a940ab7df95/lib/ruby_indexer/lib/ruby_indexer/configuration.rb#L182), such that it no longer halts execution of the initial indexing request.

### Automated Tests

I was looking for some tests around gemspec awareness in the `configuration_spec` file, but I couldn't find anything. I'd like to add a test if there's an appropriate place to do so.

### Manual Tests

1. Boot up a rails app that does not download `tzinfo-data` as a dependency, despite being included in the bundle
2. Confirm the indexing request works correctly and completes with features that depend on the index working correctly.


vinistock:
Thank you for the fix!

Issue #1057, opened 2023-10-02
Use Rubocop severity for non-correctable issues


nhumble-sa:
### I have checked that this feature is not already implemented

- [X] This feature does not exist

### Use case

I have customized the severity of several Rubocop rules to lower their severity to `INFORMATION`, but they show up as warnings when run with ruby-lsp.

### Description

When running diagnostics, non-correctable Rubocop offenses have their severity explicitly set to `WARNING` instead of taking the configured severity from Rubocop.

### Implementation

Simply do no override the Rubocop severity. The code already gets the severity for auto-correctable offenses. The place where this is set is here:
https://github.com/Shopify/ruby-lsp/blob/db3e7129e199ca419f288b44351abb7f45a5edc3/lib/ruby_lsp/requests/support/rubocop_diagnostic.rb#L54

The other branch of that conditional just uses the severity from Rubocop:
https://github.com/Shopify/ruby-lsp/blob/db3e7129e199ca419f288b44351abb7f45a5edc3/lib/ruby_lsp/requests/support/rubocop_diagnostic.rb#L51

andyw8:
Completed in https://github.com/Shopify/ruby-lsp/pull/1064

PR #1064, opened 2023-10-05
Use RuboCop's severity for all issues.

nhumble-sa:
### Motivation

Non auto-correctable RuboCop offenses were all flagged as WARNINGs rather than using the configured RuboCop severity [see issue](https://github.com/Shopify/ruby-lsp/issues/1057)

### Implementation

I pulled the logic for the severity outside the auto-correctable condition check.

### Automated Tests

I modified an existing test that relied on the old behavior to expect the new behavior.

### Manual Tests

I added an entry to a Gemfile for a ruby project I have that used the `path:` to link to my local instance of the gem. I was able to validate the new behavior in the context of VS Code.


nhumble-sa:
I have signed the CLA!

vinistock (approved):

Issue #1072, opened 2023-10-10
Fix breaking changes related to the next Prism upgrade


vinistock:
The latest version of Prism drops the ability to override `visit`. Let's address those breaking changes so that we don't fall behind in upgrades.

This PR https://github.com/Shopify/ruby-lsp/pull/1041 has much of the work that is needed, but also includes other refactors that aren't strictly related to the breaking change. Let's focus on the breaking changes only so that we don't get blocked.

PR #1079, opened 2023-10-10
Remove `on_node` from folding range

vinistock:
### Motivation

Part of #1072

Remove `on_node` usage from folding range.

### Implementation

This actually caught a bug we had too. We defined the `on_node` method, but didn't actually register it on the event emitter.

The unfortunate part of the implementation is that we have to repeat `emit_requires_range` on every single event that isn't a require invocation.

### Automated Tests

Enhanced our existing test.

### DIFF
--- a/lib/ruby_lsp/requests/folding_ranges.rb
+++ b/lib/ruby_lsp/requests/folding_ranges.rb
@@ -98,81 +98,92 @@ module RubyLsp
 
       sig { params(node: YARP::ArrayNode).void }
       def on_array(node)
+        emit_requires_range
         add_simple_range(node)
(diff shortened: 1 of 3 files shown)


lib/ruby_lsp/requests/folding_ranges.rb
```diff
@@ -203,6 +215,7 @@ def on_call(node)
           return
         end
 
+        emit_requires_range
```
st0012, line 218: Based on the old implementation: `emit_requires_range unless node.is_a?(YARP::CallNode)`, I thought call node should be skipped?

PR #1082, opened 2023-10-11
Emit requires range when trying to add any other range

vinistock:
### Motivation

Applying Ufuk's suggestion from here https://github.com/Shopify/ruby-lsp/pull/1079#discussion_r1355085173.

I agree, the code is simpler this way.


Morriar (approved):

paracycle (commented):
Thanks!

PR #1085, opened 2023-10-12
Remove node override selection ranges

vinistock:
### Motivation

Part of #1072 

Since we're already going to have to change the implementation on selection range, I think it's worth matching what the [specification](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#textDocument_selectionRange) and make selection ranges match the AST exactly.

This simplifies the request and better conforms to the spec.

### Implementation

Since we can't override `visit` anymore, it's not possible to implement this request with a visitor unfortunately. We use a queueing mechanism instead where we keep track of the current node and its parent and push ranges for all of the nodes

### Automated Tests

Fixed the tests to match the new behaviour.

### DIFF
--- a/lib/ruby_lsp/requests/selection_ranges.rb
+++ b/lib/ruby_lsp/requests/selection_ranges.rb
@@ -20,123 +20,35 @@ module RubyLsp
     #   puts "Hello, world!" # --> Cursor is on this line
     # end
     # ```
-    class SelectionRanges < BaseRequest
+    class SelectionRanges
       extend T::Sig
-
-      NODES_THAT_CAN_BE_PARENTS = T.let(
-        [
-          YARP::ArgumentsNode,
-          YARP::ArrayNode,
-          YARP::AssocNode,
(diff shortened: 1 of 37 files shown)


lib/ruby_lsp/requests/selection_ranges.rb
```diff
@@ -20,123 +20,35 @@ module Requests
...
-        else
-          create_selection_range(node.location, @stack.last)
+        # [node, parent]
+        queue = [[@document.tree, nil]]
+
+        until queue.empty?
+          node, parent = queue.shift
+          next unless node
+
+          range = RubyLsp::Requests::Support::SelectionRange.new(
```
st0012, line 43: nit: is the `RubyLsp::Requests` namespace needed?

st0012 (approved):

st0012:
BTW, since this conforms the spec better, maybe it categorises as a fix?

PR #1087, opened 2023-10-12
Use Ruby to bump version in workflow

vinistock:
### Motivation

Our workflow to bump the version automatically failed again https://github.com/Shopify/ruby-lsp/actions/runs/6486332799/job/17614311626.

I give up on using AWK, let's just use Ruby to bump the version. Also, put a proper name in the workflow.

### DIFF
--- a/.github/workflows/version_bump.yml
+++ b/.github/workflows/version_bump.yml
@@ -1,4 +1,4 @@
-name: Release
+name: Version bump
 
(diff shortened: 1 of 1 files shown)


st0012 (approved):
Should we use this to bump the version instead of manually changing it locally and push? Asking because it's different from what the release doc says.

PR #1090, opened 2023-10-12
Migrate to Prism

st0012:
### Motivation

Since `YARP` has been renamed to `Prism`, we should also migrate our dependency to the new gem.

Closes https://github.com/Shopify/ruby-dev-exp-issues/issues/1062

### Implementation

- Install the new gem and generate new rbi files
- Update a few tests that's affected by other new `Prism` commits (e.g. inspection change)
- Update the git submodule for fixtures

Note: Because `rbi` still uses `yarp` as dependency, we still have it in `Gemfile.lock` as well as its rbi files.

### Automated Tests

All existing tests should pass


test/expectations/expectations_test_runner.rb
```diff
@@ -93,7 +93,7 @@ def test_#{expectation_suffix}__#{test_name}__does_not_raise
         end
       end
 
-      Dir.glob(TEST_YARP_FIXTURES).each do |path|
+      Dir.glob(TEST_Prism_FIXTURES).each do |path|
```
vinistock, line 96:
```suggestion
      Dir.glob(TEST_PRISM_FIXTURES).each do |path|
```

vinistock (approved):
Awesome. Thank you for tackling this!

Issue #662, opened 2023-04-27
Support specifying a custom bundler group to install from Gemfile


TSMMark:
### I have checked that this feature is not already implemented

- [X] This feature does not exist

### Use case

I have a lot of gems in my gemfile. Only like 5% of them are actually needed to ruby Ruby LSP in VS Code. E.g. rubocop, rubocop-minitest, rubocop-performance, rubocop-graphql

### Description

I have the necessary gems in a bundler group

```
group :test, :vscode do
```

I'd like to be able to tell Ruby-LSP to only install the vscode group. 

```
Bundler.require(:vscode)
```

### Implementation

When ruby-lsp installs gems in `.ruby-lsp/gemfile`, it could install only the configured group, e.g.:

```
BUNDLE_WITH="vscode" bundle install
```

TSMMark:
Would maintainers accept a PR for this?

TSMMark:
Nice! A custom activation to specify `BUNDLE_WITH` would be perfect actually. Looking forward to the merge and release of https://github.com/Shopify/vscode-ruby-lsp/pull/583

Feel free to tag me on it if you want some testers eventually

vinistock:
Hello!

I don't think the Ruby LSP should be concerned with Gemfile groups in any way. If a gem can't be installed in all platforms, then I believe the solution is to scope to the platforms where it can be installed.

For example, Sorbet doesn't support Windows and so we [scope it to all other platforms](https://github.com/Shopify/ruby-lsp/blob/3e6b20fcf21cbb7254758b1d9b68ec76ff53c0f7/Gemfile#L22). In your case, you'd need to scope it to platforms that do not include Darwin ARM.

I also don't understand how you don't have the same issue when running `bundle install` in your terminal - since that's all the LSP is doing. Are you running bundle install with some configuration when you do it manually?

jfahrenkrug:
Hello and thank you for your helpful reply.

I'll answer your second question first: I ran `bundle config set --local without 'production'` before running `bundle install`. It works fine. But I don't think ruby-lsp takes that config setting into account. That would actually be nice if it did.

Thank you for your suggestion about the `platform` setting in the Gemfile. I'll have to try that. There's no obvious "non-mac-os" platform that would also include Ubuntu, so I'll have to do some experimenting. If you happen to know what I'd have to specify to get "every Unix platform except for macOS (or macOS ARM)".... :) Thanks!

vinistock:
[...]

PR #1093, opened 2023-10-13
Respect local bundler configuration

vinistock:
### Motivation

Closes https://github.com/Shopify/ruby-lsp/issues/662, Closes https://github.com/Shopify/vscode-ruby-lsp/issues/597

For local Bundler configuration, it uses the relative path considering `BUNDLE_GEMFILE`. For example, if the root of the project is `rails`, then it'll be in `rails/.bundle`.

Because we set `BUNDLE_GEMFILE` to the custom bundle, then it'll use the relative path based on that, resulting in `rails/.ruby-lsp/.bundle`, which is incorrect.

### Implementation

When we are setting up the custom bundle, we need to check if the current project has a `.bundle` folder. In that case, we can set `BUNDLE_APP_CONFIG` to the absolute path to that folder to avoid having bundler incorrectly set it inside the `.ruby-lsp` folder.

### Automated Tests

Added a test demonstrating the behavior.

### DIFF
--- a/lib/ruby_lsp/setup_bundler.rb
+++ b/lib/ruby_lsp/setup_bundler.rb
@@ -46,9 +46,9 @@ module RubyLsp
       @dependencies = T.let(load_dependencies, T::Hash[String, T.untyped])
     end
 
-    # Setups up the custom bundle and returns the `BUNDLE_GEMFILE` and `BUNDLE_PATH` that should be used for running the
-    # server
-    sig { returns([String, T.nilable(String)]) }
+    # Sets up the custom bundle and returns the `BUNDLE_GEMFILE`, `BUNDLE_PATH` and `BUNDLE_APP_CONFIG` that should be
+    # used for running the server
+    sig { returns([String, T.nilable(String), T.nilable(String)]) }
     def setup!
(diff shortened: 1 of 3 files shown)


lib/ruby_lsp/setup_bundler.rb
```diff
@@ -46,9 +46,9 @@ def initialize(project_path, branch: nil)
       @dependencies = T.let(load_dependencies, T::Hash[String, T.untyped])
     end
 
-    # Setups up the custom bundle and returns the `BUNDLE_GEMFILE` and `BUNDLE_PATH` that should be used for running the
-    # server
-    sig { returns([String, T.nilable(String)]) }
+    # Setups up the custom bundle and returns the `BUNDLE_GEMFILE`, `BUNDLE_PATH` and `BUNDLE_APP_CONFIG` that should be
```
andyw8, line 49:
```suggestion
    # Sets up the custom bundle and returns the `BUNDLE_GEMFILE`, `BUNDLE_PATH` and `BUNDLE_APP_CONFIG` that should be
```

andyw8 (approved):

PR #1096, opened 2023-10-13
Indexing performance improvements

vinistock:
### Motivation

After profiling our indexing code a bit, I noticed we were spending an abnormal amount of time in `is_a?`. Unfortunately, those were coming from the `sorbet-runtime`. In particular, the `T.let` and signatures for `PrefixTree::Node` had significant impact on the indexing time.

The reason is because we need to instantiate an enormous amount of nodes for the prefix tree, since each character in the name becomes a new node. That means that initializing a new `Node` is critical for the performance of the index.

### Implementation

- Used signature without runtime in two critical methods. I'm not sure why we had checks happening despite setting the check level to `never` to be honest
- Switched a few method calls to instance variables to avoid a few dispatches
- Removed the `T.let` inside the `Node` initializer

This change shoves 10+ seconds from indexing time in Core.

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -33,6 +33,7 @@ Sorbet/TrueSigil:
   Include:
     - "test/**/*.rb"
     - "lib/ruby_indexer/test/**/*.rb"
+    - "lib/ruby_indexer/lib/ruby_indexer/prefix_tree.rb"
   Exclude:
     - "**/*.rake"
     - "lib/**/*.rb"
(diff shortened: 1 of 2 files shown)


Morriar (commented):
What do you think about turning of `sorbet-runtime` in "production" mode like we do in Tapioca: https://github.com/Shopify/tapioca/blob/main/exe/tapioca instead?

vinistock:
We already do it. Copied directly from the exact source you linked https://github.com/Shopify/ruby-lsp/blob/da6725db348569c4300fe625a240ea5660b6ec83/exe/ruby-lsp#L66.

I don't know for sure why I kept seeing so many `is_a?` in the profile even after signatures unwrapped.

Morriar:
It's possible the `is_a?` comes from the `T.let` rather than the sigs in which case we could keep them but remove the `T.let`?

vinistock:
That's fair. I brought back the regular sigs, let's give that a try and profile again.

PR #1106, opened 2023-10-16
match rubocop diagnostic severities

maxnotarangelo:
### Motivation

Rubocop has an `info` severity level (which doesn't fail a lint check, see https://github.com/rubocop/rubocop/pull/9391), which I think should be given a light treatment in a code editor. As an example, I have the `LineLength` cop set to this severity level. Here's the existing behavior:

Screenshot 2023-10-16 at 11 53 07 AM

And here's the new behavior:

Screenshot 2023-10-16 at 12 09 14 PM

### Implementation

I matched the rubocop diagnostic severities (see here: https://github.com/rubocop/rubocop/blob/affbe3b3a66ce15534021548e282270c4a688962/lib/rubocop/lsp/severity.rb#L8-L15).

### Automated Tests

The existing code was untested.

### Manual Tests

I tested this in my own project with the appropriate rubocop setting.

### DIFF
--- a/lib/ruby_lsp/requests/support/rubocop_diagnostic.rb
+++ b/lib/ruby_lsp/requests/support/rubocop_diagnostic.rb
@@ -9,9 +9,9 @@ module RubyLsp
 
         RUBOCOP_TO_LSP_SEVERITY = T.let(
           {
-            convention: Constant::DiagnosticSeverity::INFORMATION,
-            info: Constant::DiagnosticSeverity::INFORMATION,
+            info: Constant::DiagnosticSeverity::HINT,
             refactor: Constant::DiagnosticSeverity::INFORMATION,
+            convention: Constant::DiagnosticSeverity::INFORMATION,
             warning: Constant::DiagnosticSeverity::WARNING,
             error: Constant::DiagnosticSeverity::ERROR,
             fatal: Constant::DiagnosticSeverity::ERROR,


maxnotarangelo:
I have signed the CLA!

maxnotarangelo:
In the Ruby LSP repo it doesn't look like there's a line length rule, but if you add the following to `.rubocop.yml`
```yml
Layout/LineLength:
  Max: 80
  Severity: info
```

it will show the errors but still exit with 0:

```sh
$ bundle exec rubocop
...
104 files inspected, 1003 offenses detected
$ echo $?
0
```

maxnotarangelo:
> If you manually configure the severity to be `info`, yes. But by default, line length is not `info` is it? Our RuboCop configuration inherits from the Shopify style-guide. We don't change the severity there as far as I know, but RuboCop still exits with a non-zero status.

No, it's not, so in this repo if a line is too long you'll get a non-zero status, and you'll get a blue underline on lines that are too long in your editor.

> Do you have line length manually configured to `info`?

Yes, in my project we do.

> Is there an example of a cop that has severity `info` by default?

No, none of Rubocop's rules have it as the default. I think it's specifically for this scenario, where you want to report on something (e.g. in the editor) but not enforce it.

> The key thing here is: we would rather err on the side of showing more errors on the editor than accidentally let developers push code that will fail on CI.
> 
> So we need to be sure that cops with severity `info` really won't fail CI and which cops have this severity by default - since depending on which ones are using it, we may want to change our style-guide to a different severity.

Right, and if this PR gets merged, we'll still show the same blue/yellow/red underlines on any code that's going to fail CI. This PR just makes it so that if someone changes the severity of a rubocop rule to `info` (i.e. they _don't_ want violations of the rule to fail CI) the editor will do something more subtle than it does with the `INFORMATION` severity level.

And I think more subtlety is important, because the blue underlines are pretty annoying if you're not planning to fix them. The sidebar also gets pretty hard to read, like below, where rules with an `info` level severity according to rubocop are making it pretty hard to see which lines of code actually changed (which are the slightly-less-vibrant blue color on the left):

Screenshot 2023-10-17 at 5 12 18 PM

PR #1107, opened 2023-10-16
Reduce `Completion#top_level?` complexity

### LINKED ISSUE #1062: High CPU usage
### Ruby version

3.0.6

### Code snippet

_No response_

### Description

High CPU usage.

I noticed this issue today when quickly correcting an error reported by YARP, typically using copilot's autocomplete. Ruby starts consuming a substantial amount of CPU for an extended period, ultimately causing the Ruby LSP to crash completely. Here's a video demonstrating this behavior:

https://github.com/Shopify/ruby-lsp/assets/42195321/aa3892fc-8fe2-4001-9b1c-500cc9b06eaa

### Expected output

Should have detected the correction change quickly.

vinistock:
### Motivation

Closes #1062 (again 😅)

We were spending an abnormal amount of time trying to figure out if we needed to dereference a constant because of a conflict.

For example

```ruby
class Bar; end

module Foo
  class Bar; end

  # When completion is triggered here for `Bar`, we need to use `::Bar` or else it conflicts
  # with `Foo::Bar`
  B
end

# When completion is triggered here, then we can use `Bar`
B
```

### Implementation

Instead of checking all possible entries for conflicts, we can just check if there are any conflicts when combining the suggestion name with the nesting.

I also removed an incorrect `+1` on the prefix search implementation. There's no need for that `+1`.

### Automated Tests

Added more tests.

### DIFF
--- a/lib/ruby_indexer/lib/ruby_indexer/index.rb
+++ b/lib/ruby_indexer/lib/ruby_indexer/index.rb
@@ -95,7 +95,7 @@ module RubyIndexer
     # ```
     sig { params(query: String, nesting: T::Array[String]).returns(T::Array[T::Array[Entry]]) }
     def prefix_search(query, nesting)
-      results = (nesting.length + 1).downto(0).flat_map do |i|
+      results = nesting.length.downto(0).flat_map do |i|
         prefix = T.must(nesting[0...i]).join("::")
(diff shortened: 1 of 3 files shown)


andyw8 (approved):
Should we verify on core?

vinistock:
I did. It reduces some cases where it may take minutes to auto-complete to seconds.

PR #1110, opened 2023-10-16
Add Folding Ranges for Lambda Literals

tomascco:
### Motivation

Lambda literals like the one below don't show folding ranges.

```ruby
lambda = ->(item) do
  puts item
end
```

### Implementation

This PR hooks on `Prism::LambdaNode` enter event and add a simple handle,  like the one for blocks.

### Automated Tests

Yes, I wrote expectations.  

### Manual Tests

1. Open the editor with this build
2. Ensure that the server is active
3.  Write the following code:

```ruby
lambda = ->(item) do
  puts item
end
```
4. And ensure that the folding region appears.

### DIFF
--- a/lib/ruby_lsp/requests/folding_ranges.rb
+++ b/lib/ruby_lsp/requests/folding_ranges.rb
@@ -54,6 +54,7 @@ module RubyLsp
           :on_string_concat_node_enter,
           :on_def_node_enter,
           :on_call_node_enter,
+          :on_lambda_node_enter,
         )
       end
 
@@ -200,6 +201,11 @@ module RubyLsp
 
         location = node.location
         add_lines_range(location.start_line, location.end_line - 1)
+      end
+
+      sig { params(node: Prism::LambdaNode).void }
+      def on_lambda_node_enter(node)
+        add_simple_range(node)
       end
 
       private
--- a/test/expectations/folding_ranges/lambda_literal_braces.exp.json
+++ b/test/expectations/folding_ranges/lambda_literal_braces.exp.json
@@ -1,0 +1,9 @@
+{
+  "result": [
+    {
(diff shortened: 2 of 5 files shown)


lib/ruby_lsp/requests/folding_ranges.rb
```diff
@@ -202,6 +203,11 @@ def on_call_node_enter(node)
         add_lines_range(location.start_line, location.end_line - 1)
       end
 
+      sig { params(node: Prism::LambdaNode).void }
+      def on_lambda_node_enter(node)
+        add_simple_range(node)
```
tomascco, line 208: Is this sufficient? Do we need to also handle multiline parameters here?

paracycle: It depends on the behaviour you want for how something like:
```ruby
foo = -> (
  bar,
  baz
) do
  # something
end
```
to fold.

Should it be:
```ruby
foo = -> (
  ...
end
```
or
```ruby
foo = -> (
  bar,
  baz
) do
  ...
end
```
If you want the latter behaviour, you need to do handling similar to the one `on_def_node_enter` does, and add a test case to ensure that behaviour.

tomascco: For now I'll stick with this basic behavior and if anyone asks for the folding of multiline arguments in lambdas I'll be happy to implement it.

I've also corrected the rubocop offense. Thank you for the reply.

vinistock: Yeah, for multiline parameters, we'll need something similar to what we have implemented in `on_def`.

But we can definitely move forward with this and address the parameters in the next PR.

PR #1111, opened 2023-10-17
add documentation link to rubocop diagnostics

dleavitt:
### Motivation

RuboCop has nice documentation for all its built-in rules. The doc url for each of these rules is programmatically available, and LSP diagnostics have a special slot designed exactly for this.*

Screen Shot 2023-10-17 at 00 46 08

*Not a lawyer, but I believe this is sufficient to mirandize the user against any subsequent autocorrections RuboCop may undertake.

### Implementation

Ideally RuboCop would attach this info to the offense, but for now I'm using the global cop registry to look up the documentation url at the time the diagnostic is generated. 

I don't _think_ this causes RuboCop to read YAML files from disk, change global state, or print things to stdout, but I have been burned in the past.

### Automated Tests

Yup.

### Manual Tests

Rule w/ doc link:

1. Write some code that's in violation of a built-in RuboCop rule.
2. Mouse over that code.
3. See this, hopefully:

Screen Shot 2023-10-17 at 00 46 08

Rule w/o doc link:

In this repo

1. Add a Rake rule, e.g. `Rake/Desc: { Enabled: true }`
2. Open `Rakefile`
3. Add a task without a `desc`
4. Mouse over the task name
5. See this:

Screen Shot 2023-10-17 at 00 45 56

### DIFF
--- a/lib/ruby_lsp/requests/support/rubocop_diagnostic.rb
+++ b/lib/ruby_lsp/requests/support/rubocop_diagnostic.rb
@@ -52,10 +52,17 @@ module RubyLsp
 
           message += "\n\nThis offense is not auto-correctable.\n" unless @offense.correctable?
 
+          cop = RuboCop::Cop::Registry.global.find_by_cop_name(@offense.cop_name)
+          code = if cop&.documentation_url
+            { value: @offense.cop_name, target: cop.documentation_url }
+          else
+            @offense.cop_name
+          end
+
           Interface::Diagnostic.new(
             message: message,
             source: "RuboCop",
-            code: @offense.cop_name,
+            code: code,
             severity: severity,
             range: Interface::Range.new(
               start: Interface::Position.new(
--- a/test/expectations/diagnostics/def_bad_formatting.exp.json
+++ b/test/expectations/diagnostics/def_bad_formatting.exp.json
@@ -61,7 +61,10 @@
         }
       },
       "severity": 3,
-      "code": "Style/FrozenStringLiteralComment",
+      "code": {
+        "value": "Style/FrozenStringLiteralComment",
+        "target": "https://docs.rubocop.org/rubocop/cops_style.html#stylefrozenstringliteralcomment"
+      },
       "source": "RuboCop",
       "message": "Style/FrozenStringLiteralComment: Missing frozen string literal comment.",
       "data": {
@@ -110,7 +113,10 @@
         }
       },
       "severity": 3,
-      "code": "Layout/EmptyLinesAroundMethodBody",
+      "code": {
+        "value": "Layout/EmptyLinesAroundMethodBody",
+        "target": "https://docs.rubocop.org/rubocop/cops_layout.html#layoutemptylinesaroundmethodbody"
+      },
       "source": "RuboCop",
       "message": "Layout/EmptyLinesAroundMethodBody: Extra empty line detected at method body beginning.",
       "data": {
@@ -159,7 +165,10 @@
         }
       },
       "severity": 3,
-      "code": "Layout/EmptyLines",
+      "code": {
+        "value": "Layout/EmptyLines",
+        "target": "https://docs.rubocop.org/rubocop/cops_layout.html#layoutemptylines"
+      },
       "source": "RuboCop",
       "message": "Layout/EmptyLines: Extra blank line detected.",
       "data": {
@@ -208,7 +217,10 @@
         }
       },
       "severity": 3,
-      "code": "Layout/IndentationWidth",
+      "code": {
+        "value": "Layout/IndentationWidth",
+        "target": "https://docs.rubocop.org/rubocop/cops_layout.html#layoutindentationwidth"
+      },
       "source": "RuboCop",
       "message": "Layout/IndentationWidth: Use 2 (not 0) spaces for indentation.",
       "data": {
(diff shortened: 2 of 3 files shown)


vinistock:
According to the [specification](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#diagnostic), `code` can only be a string or an integer - not an object. Did you mean to use `codeDescription`?

> *Not a lawyer, but I believe this is sufficient to mirandize the user against any subsequent autocorrections RuboCop may undertake.

I don't understand. Can you elaborate what you mean by this? How is this related to the documentation links?

I like this addition. My only concern is: how fast is it to run `RuboCop::Cop::Registry.global.find_by_cop_name(@offense.cop_name)`? Is it a hash lookup or is it searching in an array? RuboCop diagnostics are already painfully slow - and run every time a text edit is made - so we really can't take a performance hit here.

dleavitt:
> According to the [specification](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#diagnostic), code can only be a string or an integer - not an object. Did you mean to use codeDescription?

You're absolutely right, I was looking at the vscode api, not the lsp protocol.

https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/vscode/index.d.ts#L6753-L6768

I'll switch to the LSP version.

> I like this addition. My only concern is: how fast is it to run...

I'll double check. 

>> *Not a lawyer, but I believe this is sufficient to mirandize the user against any subsequent autocorrections RuboCop may undertake.

> I don't understand. Can you elaborate what you mean by this? How is this related to the documentation links?

A RuboCop joke that didn't land. Disregard.

dleavitt:
Okay, I've updated the diagnostic to use the `codeDescription` field rather than the vscode-specific `code` object.

> My only concern is: how fast is it to run...

Looks like it's just a hash lookup: https://github.com/rubocop/rubocop/blob/master/lib/rubocop/cop/registry.rb#L254

It appears that all the config is already loaded into memory at the point the diagnostic is generated, so overhead should just be the hash lookup and a few method calls. e.g. `@enrollment_queue` here is already empty, etc:
https://github.com/rubocop/rubocop/blob/master/lib/rubocop/cop/registry.rb#L283

dleavitt:
I have signed the CLA!

vinistock:
> Looks like it's just a hash lookup: https://github.com/rubocop/rubocop/blob/master/lib/rubocop/cop/registry.rb#L254

Not quite just the hash lookup. It invokes `to_h` and that invokes [other things](https://github.com/rubocop/rubocop/blob/6713311554d74f55fed0df20ab2bf2298d672006/lib/rubocop/cop/registry.rb#L174).

Considering that will be executed for every single violation in diagnostics, let's create our own in memory map of `cop name -> doc url`, so that we're 100% sure is just a hash access with no other dispatches. Something like this

```ruby
class RuboCopDiagnostic
  # Let's do the work ahead of time to get a hash that is: { "cop" => "doc_url" }
  # and then use that directly
  COP_TO_DOC_URL = Registry.global.to_h
end
```

dleavitt:
Will do! THough FYI I think we're good unless the implementation of `to_h` changes - `@enrollment_queue` always appears to be empty at this stage, which is the only thing that would trigger more work.

dleavitt:
Cached cops in a constant as requested.

lib/ruby_lsp/requests/support/rubocop_diagnostic.rb
```diff
@@ -19,6 +19,12 @@ class RuboCopDiagnostic
           T::Hash[Symbol, Integer],
         )
 
+        # Cache cops to attach URLs to diagnostics. Only built-in cops for now.
+        COP_TO_DOC_URL = T.let(
+          RuboCop::Cop::Registry.global.to_h,
```
dleavitt, line 24: note that invoking to the RuboCop global registry at the top-level is sensitive to load order.

vinistock: I think it should be fine. We require `rubocop` before this class is loaded.

lib/ruby_lsp/requests/support/rubocop_diagnostic.rb
```diff
@@ -19,6 +19,12 @@ class RuboCopDiagnostic
...
+          T::Hash[String, [T.class_of(RuboCop::Cop::Base)]],
```
vinistock, line 25: Are the values in the hash always an array with a single element? Or is it just a regular array? If it's the latter, then the type should be this. Otherwise, just ignore.

```suggestion
          T::Hash[String, T::Array[T.class_of(RuboCop::Cop::Base)]],
```

vinistock (approved):
Very nice. Thank you for the contribution!

lib/ruby_lsp/requests/support/rubocop_diagnostic.rb
dleavitt: Yep, it's also best if this file isn't required until after RuboCop has been _run_ for the first time (which is the case right now). This is because the first run is when non-built-in cops get added to the global registry.

E.g. in the ruby lsp codebase `RuboCop::Cop::Registry.global.length` is 544 before this line and 635 after it: 

https://github.com/Shopify/ruby-lsp/blob/5b732ebe1d6c4dbaf40a4852946fd1de89830590/lib/ruby_lsp/requests/support/rubocop_runner.rb#L90

Doc links for non-built-in cops don't even work right now, so it's sort of moot, but seems worth being aware of!

andyw8 (approved):

vinistock (approved):

PR #1112, opened 2023-10-17
Extract Entry from Indexer

andyw8:
### Motivation

As we add more indexing behaviour, it's becomes more awkward to have everything in a single file. Let's start to split this up.

### Implementation

Extract `Entry` into its own file.

### Automated Tests

Updated to reflect changed namespace.

### Manual Tests

n/a


st0012 (approved):

lib/ruby_indexer/lib/ruby_indexer/index/entry.rb
```diff
@@ -0,0 +1,207 @@
+# typed: strict
+# frozen_string_literal: true
+
+module RubyIndexer
+  class Index
```
vinistock, line 5: How about we get rid of the `Index` namespace while we're at it? We don't get much out of it.

lib/ruby_indexer/lib/ruby_indexer/index/entry.rb
vinistock, line 1: I think the extra `index` folder in the middle might be a bit much, given that there will always only be a single file inside.

lib/ruby_indexer/lib/ruby_indexer/index/entry.rb
andyw8: 👍 updated

PR #1113, opened 2023-10-17
Index singleton methods with `class << self`

andyw8:
### Motivation

Currently we only support indexing singleton methods defined as `self.foo`.

### Implementation

We need to keep track of whether we are within a singleton class node, so I added a new instance variable to track that.

### Automated Tests

Added

### Manual Tests

- [x] run against `Shopify/shopify` to ensure no crashes.

### DIFF
--- a/lib/ruby_indexer/lib/ruby_indexer/visitor.rb
+++ b/lib/ruby_indexer/lib/ruby_indexer/visitor.rb
@@ -132,16 +132,19 @@ module RubyIndexer
     def visit_def_node(node)
       method_name = node.name.to_s
       comments = collect_comments(node)
-      case node.receiver
+      entry_class = case node.receiver
       when nil
-        @index << if @singleton
-          Entry::SingletonMethod.new(method_name, @file_path, node.location, comments, node.parameters)
+        if @singleton
+          Entry::SingletonMethod
         else
-          Entry::InstanceMethod.new(method_name, @file_path, node.location, comments, node.parameters)
+          Entry::InstanceMethod
         end
       when Prism::SelfNode
-        @index << Entry::SingletonMethod.new(method_name, @file_path, node.location, comments, node.parameters)
+        Entry::SingletonMethod
+      else
+        return
       end
+      @index << entry_class.new(method_name, @file_path, node.location, comments, node.parameters)
     end
 
     private
--- a/lib/ruby_indexer/lib/ruby_indexer/visitor.rb
+++ b/lib/ruby_indexer/lib/ruby_indexer/visitor.rb
@@ -10,7 +10,7 @@ module RubyIndexer
       @index = index
       @file_path = file_path
       @stack = T.let([], T::Array[String])
-      @singleton_class = T.let(false, T::Boolean)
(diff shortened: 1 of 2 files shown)


lib/ruby_indexer/lib/ruby_indexer/visitor.rb
```diff
@@ -124,12 +132,19 @@ def visit_call_node(node)
...
-        @index << Entry::InstanceMethod.new(method_name, @file_path, node.location, comments, node.parameters)
+        if @singleton_class # i.e. `class << self`
+          Entry::SingletonMethod
+        else
+          Entry::InstanceMethod
+        end
       when Prism::SelfNode
-        @index << Entry::SingletonMethod.new(method_name, @file_path, node.location, comments, node.parameters)
+        Entry::SingletonMethod
+      else
```
andyw8, line 144: We don't yet handle things like `def Foo.bar`. The next step will be record the namespace along with the definition.

lib/ruby_indexer/lib/ruby_indexer/visitor.rb
```diff
@@ -25,6 +26,13 @@ def visit_class_node(node)
       add_class_entry(node)
     end
 
+    sig { override.params(node: Prism::SingletonClassNode).void }
+    def visit_singleton_class_node(node)
+      @singleton_class = true
```
paracycle, line 31: What about usages of:
```ruby
class << Foo
  def bar
  end
end
```
syntax? That opens up the singleton class of `Foo` and the scope inside defines stuff on there.

So, I think this flag shouldn't a boolean, but store the name of the singleton class that's opened (which should be accessible via `node.expression`) or `nil`.

andyw8: 👍 Updated, although for now I'm storing the node rather than the name of the class, in case we need anything from that. Once we get further along in https://github.com/Shopify/ruby-lsp/pull/1121 I'll have better understanding of if that's necessary.

vinistock:
That's a good point. We can probably just store `Method` entries and differentiate based on the owner.

lib/ruby_indexer/lib/ruby_indexer/visitor.rb
```diff
@@ -124,12 +132,19 @@ def visit_call_node(node)
     def visit_def_node(node)
       method_name = node.name.to_s
       comments = collect_comments(node)
-      case node.receiver
+      entry_class = case node.receiver
       when nil
-        @index << Entry::InstanceMethod.new(method_name, @file_path, node.location, comments, node.parameters)
+        if @singleton_class_node # i.e. `class << self`
+          Entry::SingletonMethod
```
paracycle, line 138: This is not necessarily true, and depends on the `class << xxx` call:
```ruby
obj = Object.new
obj.foo # NoMethodErro
class << obj
  def foo = 42
end
obj.foo # => 42
```

In this case, the method is added to the singleton class of **an instance**, so it ends up being an instance method, not a singleton class method.

andyw8:
Closing in favour of https://github.com/Shopify/ruby-lsp/pull/1250 (WIP) since `main` has changed a lot.

PR #1114, opened 2023-10-17
RuboCop Code Action "Disable for this line"

### LINKED ISSUE #1099: Rubocop Diagnostics Quick Fix Code Action "Disable for this line"

### Use case

It's useful to be able to disable project lint rules on a case-by-case basis - sometimes a rule is worth having but does not apply everywhere. Every linter I've used, including Rubocop, supports inline annotations to disable a rule.

ruby-lsp provides a code action to autocorrect violations but no code actions to disable the rule for a line or file. Doing so by hand involves copying the rule id from within the tooltip, remembering the annotation syntax, etc. Not the end of the world, but a bit of a flow-breaker.

### Description

Add two code actions, one that disables a rubocop rule for a line and the other for an entire file. 

Like this:
Screen Shot 2023-10-15 at 15 06 28

Other linters I've used with vscode support this: eslint, stylelint, and solargraph [via an extension](https://marketplace.visualstudio.com/items?itemName=aki77.rubocop-quick-fix).

### Implementation

There's some prior art here, which is in typescript but seems almost line-for-line what would be needed:
https://github.com/aki77/vscode-rubocop-quick-fix/blob/main/src/QuickFixProvider.ts#L21-L63

Presumably something similar would be added here (or next to it):
https://github.com/Shopify/ruby-lsp/blob/87be915e0ec159b87c7f0ca5dcc582826aad13cd/lib/ruby_lsp/requests/support/rubocop_diagnostic.rb#L29-L46

dleavitt:
### Motivation

Adds a code action that disables the RuboCop rule for a line:

https://github.com/Shopify/ruby-lsp/assets/71686/72b2f0f0-4aac-49d1-a78e-509e740625b7

fixes #1099.

### Automated Tests

See notes inline.

### Manual Tests

1. Check out this branch in vscode
2. Restart the LSP server
3. Edit `.rubocop.yml` to stop excluding `"test/expectations/**/*"`
4. Edit `.vscode/settings.json` and disable `editor.formatOnSave` for `[ruby]
5. Open the `test/expectations/code_actions_formatting` directory.
6. Open files in there and run the "Disable X" code actions on the offenses.


lib/ruby_lsp/requests/code_actions.rb
```diff
@@ -38,14 +38,19 @@ def initialize(document, range, context)
...
-          next if code_action.nil?
+        code_actions = diagnostics.flat_map do |diagnostic|
+          diagnostic_code_actions = diagnostic.dig(:data, :code_actions) || []
+          diagnostic_code_actions.filter_map do |code_action|
+            # We want to return only code actions that are within range or that do not have any edits, such as refactor
+            # code actions
+            range = code_action.dig(:edit, :documentChanges, 0, :edits, 0, :range)
+            unless cover?(range)
+              # Seems to work?
+              code_action[:title] += " RANGE?"
```
dleavitt, line 49: Stop suppressing code actions based on request range (but annotate them so you can see that they would have been suppressed.)

So far I haven't found any that don't work, so the range check may be unnecessary or too aggressive. May become more clear when testing for the branch is cleaned up.

lib/ruby_lsp/requests/code_actions.rb
```diff
@@ -38,14 +38,19 @@ def initialize(document, range, context)
...
+            end
 
-          # We want to return only code actions that are within range or that do not have any edits, such as refactor
-          # code actions
-          range = code_action.dig(:edit, :documentChanges, 0, :edits, 0, :range)
-          code_action if diagnostic.dig(:data, :correctable) && cover?(range)
```
dleavitt, line 48: we filter out the autocorrect actions upstream of this, since the rubocop comment actions are applicable even if the autocorrect isn't.

lib/ruby_lsp/requests/support/rubocop_diagnostic.rb
```diff
@@ -69,13 +61,33 @@ def to_lsp_diagnostic
             ),
             data: {
               correctable: @offense.correctable?,
-              code_action: to_lsp_code_action,
+              code_actions: to_lsp_code_actions,
```
dleavitt, line 64: changed hash key here - hopefully that's not part of any external api?

lib/ruby_lsp/requests/support/rubocop_diagnostic.rb
```diff
@@ -88,6 +100,79 @@ def offense_replacements
...
+            ),
+          )
+        end
+
+        sig { returns(T::Array[Interface::TextEdit]) }
+        def line_disable_comment
+          eol = Interface::Position.new(
+            line: @offense.line - 1,
+            # TODO: what's the real way to do this?
+            character: @offense.source_line.bytes.length,
```
dleavitt, line 128: Finds the end of the line, but probably not the best way to do it.

`Offense#source_line` appears to be `@private`. Not sure if `bytes.length` is correct (but seems to work.)

Also doesn't handle multiline quotes, etc.

lib/ruby_lsp/requests/support/rubocop_diagnostic.rb
```diff
@@ -88,6 +100,79 @@ def offense_replacements
...
+          )
+          new_text = if @offense.source_line.include?(" # rubocop:disable ")
+            ",#{@offense.cop_name}"
```
dleavitt, line 131: for when there's already a rubocop comment on the line.

lib/ruby_lsp/requests/support/rubocop_diagnostic.rb
```diff
@@ -88,6 +100,79 @@ def offense_replacements
...
+
+          edit = Interface::TextEdit.new(
+            range: Interface::Range.new(start: eol, end: eol),
+            new_text: new_text,
+          )
+          [edit]
+        end
+
+        sig { returns(NilClass) }
+        def docTestEsque_DELETE_ME # rubocop:disable Naming/MethodName
```
dleavitt, line 144: sandbox for rubocop violations

lib/ruby_lsp/requests/support/rubocop_diagnostic.rb
```diff
@@ -102,6 +114,136 @@ def offense_replacements
...
+                  edits: file_disable_comments,
+                ),
+              ],
+            ),
+          )
+        end
+
+        sig { returns(T::Array[Interface::TextEdit]) }
+        def file_disable_comments
+          block_disable_comments(0, @document.source.lines.length - 1)
```
dleavitt, line 181: maybe this should look for magic comments and attempt to put the rubocop comment below?

lib/ruby_lsp/requests/code_actions.rb
```diff
@@ -55,14 +49,6 @@ def run
 
       private
 
-      sig { params(range: T.nilable(Document::RangeShape)).returns(T::Boolean) }
-      def cover?(range)
```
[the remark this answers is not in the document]
vinistock: I think this is fine 👍

lib/ruby_lsp/requests/support/rubocop_diagnostic.rb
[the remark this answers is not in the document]
vinistock: Yeah, can we start just with the implementation for disabling a cop for the current line?

The file disable is getting a bit complicated. We shouldn't need to pass `@document` to a diagnostic, so I think we may need a few refactors before we can push that one forward.

test/requests/code_actions_formatting_test.rb
```diff
@@ -0,0 +1,173 @@
+# typed: true
+# frozen_string_literal: true
+
+require "test_helper"
+require "expectations/expectations_test_runner"
+
+# Tests RuboCop disable directives - before/after on whole file
+class CodeActionsFormattingTest < Minitest::Test
+  EXPECTATIONS_DIR = File.expand_path(File.join(ExpectationsTestRunner::TEST_EXP_DIR, "code_actions_formatting"))
```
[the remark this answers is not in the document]
vinistock: I think the idea is good, but we should try to keep a similar set up as the rest of the expectation runner tests. Let's put fixtures inside `test/fixtures` and only expectations in `test/expectations/code_actions_formatting`.

We can override the way we verify expectations to match your idea here. There are a few examples in other test files.

test/requests/code_actions_formatting_test.rb
```diff
@@ -0,0 +1,173 @@
...
+
+  sig do
+    params(
+      diagnostic_code: String,
+      code_action_title: String,
+      source: String,
+      expected: String,
+    ).returns(T.untyped)
+  end
+  def assert_corrects_to_expected(diagnostic_code, code_action_title, source, expected)
```
vinistock, line 127: I think this would be simpler if we just applied all available code actions for a fixture and then verified that the `document.source` matches the expectation file.

We lose a bit of granularity in the testing, but I think that's fine.

vinistock (commented):
This is really good. If we move the disable for line to a different PR, we can probably move forward with the line disable already with a few adjustments to the tests

dleavitt:
> If we move the disable for line to a different PR

Will do.

dleavitt:
Thanks for the review, got most of your feedback in and I think I'm ready to turn this one over to you guys. I'll open a separate PR for the "disable for entire file" action.

lib/ruby_lsp/requests/support/rubocop_diagnostic.rb
```diff
@@ -102,6 +114,136 @@ def offense_replacements
...
+              range: Interface::Range.new(start: reenable_pos, end: reenable_pos),
+              new_text: "\n# rubocop:enable #{@offense.cop_name}",
+            )
+          end
+
+          [disable_comment, reenable_comment]
+        end
+
+        sig { params(line: String).returns(Integer) }
+        def length_of_line(line)
```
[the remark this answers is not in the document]
vinistock: Indeed, we'll need to refactor this later. Some sort of globally available context that will know the encoding, so that everywhere in the LSP we can always ask what's the correct length for the current encoding.

For now, let's leave it was is.

lib/ruby_lsp/requests/support/rubocop_diagnostic.rb
```diff
@@ -19,30 +19,21 @@ class RuboCopDiagnostic
           T::Hash[Symbol, Integer],
         )
 
-        sig { params(offense: RuboCop::Cop::Offense, uri: URI::Generic).void }
-        def initialize(offense, uri)
+        sig { params(document: Document, offense: RuboCop::Cop::Offense, uri: URI::Generic).void }
+        def initialize(document, offense, uri)
+          @document = document
```
vinistock, line 24: Can we receive just the encoding here for now? We're going to need to do some refactors on this, but I'd prefer if the diagnostics didn't have a reference to the document.

lib/ruby_lsp/requests/support/rubocop_diagnostic.rb
```diff
@@ -102,6 +113,64 @@ def offense_replacements
...
+            " # rubocop:disable #{@offense.cop_name}"
+          end
+
+          eol = Interface::Position.new(
+            line: @offense.line - 1,
+            character: length_of_line(@offense.source_line),
+          )
+
+          # TODO: probably better to use a block comment for multi-line
+          # offenses. Harder though; need to do indentation and such.
```
vinistock, line 150: Is there a scenario where we are unable to disable a cop for a multiline offense? If not, then I think it's fine to support only trailing comments.

test/requests/code_actions_formatting_test.rb
```diff
@@ -0,0 +1,106 @@
...
+    )
+
+    diagnostics = RubyLsp::Requests::Diagnostics.new(document).run
+    diagnostic = T.must(T.must(diagnostics).find { |d| d.code == diagnostic_code })
+    range = diagnostic.range.to_hash.transform_values(&:to_hash)
+    result = RubyLsp::Requests::CodeActions.new(document, range, {
+      diagnostics: [JSON.parse(T.must(diagnostic).to_json, symbolize_names: true)],
+    }).run
+
+    # deal with weird typing in CodeActions#run lying about its type
```
vinistock, line 75: Can you explain what you mean with this? What's the type here and what error does it cause?

test/requests/code_actions_formatting_test.rb
```diff
@@ -0,0 +1,106 @@
...
+    untyped_result = T.let(result, T.untyped)
+    selected_action = untyped_result.find do |ca|
+      code_action = T.let(ca, T.untyped)
+      code_action.respond_to?(:[]) && code_action[:title] == code_action_title
```
vinistock, line 79: Why do we need to check if it responds to `[]` here?

vinistock (commented):
This is looking good. Just a few questions, but we should be able to merge soon

dleavitt:
Pushed a few small changes and clarifications.

vinistock (approved):
Thank you so much for the contribution and the detailed PR!

The team had a discussion and we agreed that we can refactor the dependency on `document` later, possibly using the code action resolve request to handle the quickfixes that disable cops.

The failures on Ruby head are related to Prim's breaking changes that are not yet released. No worries about those.

Issue #1115, opened 2023-10-18
Request textDocument/codeAction failed.


vitaliiorlov:
### Ruby version

3.2.2

### Description

When I updated the rubocop gem from 1.56.4 to the 1.57.1 version, the ruby-lsp stopped working.
If I downgrade the rubocop gem to the 1.56.4 version, the ruby-lsp starts working again after VScode relaunching.

attaching VScode output:
[ruby-lsp-report.txt](https://github.com/Shopify/ruby-lsp/files/12998106/ruby-lsp-report.txt)

### Expected output

ruby-lsp works with the latest version of rubocop gem.

vitaliiorlov:
Update: now it stopped working with rubocop 1.56.4 version too.

vitaliiorlov:
@andyw8 no, it doesn't.
![image](https://github.com/Shopify/ruby-lsp/assets/1788994/1be27cfd-8f19-4c0c-a591-97fdaf97cbda)

andyw8:
Ok, could you please try save [this](https://gist.github.com/andyw8/630bb27450592f9b05ad591cfbf54d03) file into your repo, then run:

`bundle exec ruby index_check.rb`

It will print out the path last file indexed before the crash. It's likely from a gem.

andyw8:
Sorry, `.ruby-lsp` (hyphen).

andyw8:
Sorry, I was checking against `main` but that isn't yet released, please try this instead:

```
require "ruby-lsp"
require "ruby_lsp/internal"

module RubyIndexer
  index = Index.new

  RubyIndexer.configuration.indexables.each do |indexable|
    puts "indexing: #{indexable.full_path}"
    content = File.read(indexable.full_path)
    visitor = IndexVisitor.new(index, YARP.parse(content), indexable.full_path)
    visitor.run
  end
end
```

st0012:
PR for the above: https://github.com/getsentry/sentry-ruby/pull/2146

vitaliiorlov:
Cool!
Thanks, @andyw8 and @st0012 !

Issue #1063, opened 2023-10-04
Segfault on startup with 0.11.1


DeMarko:
### Operating System

Mac OS 13.6 (22G120)

### Ruby version

3.2.2

### Project has a bundle

- [X] Has bundle

### Ruby version manager being used

rbenv

### Description

This started yesterday for me after restarting my VSCode. I noticed that on start I would see two notices of the Ruby LSP failing when I started up my editor. 
![Screenshot 2023-10-04 at 5 21 04 PM](https://github.com/Shopify/vscode-ruby-lsp/assets/173373/a63cadab-dfd9-4545-8554-40a98d5ae9f8)

<details><summary>Details</summary>

Ruby LSP> Skipping custom bundle setup since /Users/danneljurado/dev/kickstarter/.ruby-lsp/Gemfile.lock already exists and is up to date
Ruby LSP> Running bundle install for the custom bundle. This may take a while...
Resolving dependencies...
The Gemfile's dependencies are satisfied
Starting Ruby LSP...
/Users/danneljurado/.rbenv/versions/3.2.2/lib/ruby/gems/3.2.0/gems/ruby-lsp-0.11.1/lib/ruby_indexer/lib/ruby_indexer/index.rb:164: [BUG] Segmentation fault at 0x000000011dddd000
ruby 3.2.2 (2023-03-30 revision e51014f9c0) +YJIT [x86_64-darwin22]
<output omitted>
[IMPORTANT]
Don't forget to include the Crash Report log file under
DiagnosticReports directory in bug reports.

[Error - 5:04:37 PM] Client Ruby LSP: connection to server is erroring. Shutting down server.
[Error - 5:04:37 PM] Sending document notification textDocument/didOpen failed
Error: write EPIPE
	at WriteWrap.onWriteComplete [as oncomplete] (node:internal/stream_base_commons:94:16)
[Error - 5:04:37 PM] Request textDocument/diagnostic failed.
  Message: Cannot call write after a stream was destroyed
  Code: -32099 
[Error - 5:04:37 PM] Document pull failed for text document file:///Users/danneljurado/dev/kickstarter/app/mailers/user_mailer/presenters.rb
  Message: Cannot call write after a stream was destroyed
  Code: -32099 
[Error - 5:04:37 PM] Connection to server got closed. Server will not be restarted.
[Error - 5:04:37 PM] Request textDocument/documentSymbol failed.
  Message: Cannot call write after a stream was destroyed
  Code: -32099 
[...]

PR #1117, opened 2023-10-18
Add `ruby-lsp-doctor` tool for troubleshooting indexing problems

andyw8:
### Motivation

We have a script for troubleshooting issues related to indexing, but it's sensitive to the version of Ruby LSP is use, so https://github.com/Shopify/ruby-lsp/issues/1115 had a lot of back-and-forward. To make things easier in future, let's ship a command people can easily run.

### Implementation

Adapted from https://github.com/Shopify/ruby-lsp/issues/1063#issuecomment-1749018562

### Automated Tests

I added a step to CI.

### Manual Tests

Run `./exe/ruby-lsp-index-info`.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -42,3 +42,6 @@ jobs:
 
       - name: Run tests
         run: bundle exec rake
+
+      - name: Run index troubleshooting tool
+        run: ./exe/ruby-lsp-doctor
--- a/ruby-lsp.gemspec
+++ b/ruby-lsp.gemspec
@@ -14,7 +14,7 @@ Gem::Specification.new do |s|
 
   s.files = Dir.glob("lib/**/*.rb") + ["README.md", "VERSION", "LICENSE.txt"]
   s.bindir = "exe"
-  s.executables = ["ruby-lsp", "ruby-lsp-check"]
+  s.executables = ["ruby-lsp", "ruby-lsp-check", "ruby-lsp-index-info"]
   s.require_paths = ["lib"]
 
   s.add_dependency("language_server-protocol", "~> 3.17.0")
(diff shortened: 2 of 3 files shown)


ruby-lsp.gemspec
```diff
@@ -14,7 +14,7 @@ Gem::Specification.new do |s|
 
   s.files = Dir.glob("lib/**/*.rb") + ["README.md", "VERSION", "LICENSE.txt"]
   s.bindir = "exe"
-  s.executables = ["ruby-lsp", "ruby-lsp-check"]
+  s.executables = ["ruby-lsp", "ruby-lsp-check", "ruby-lsp-index-info"]
```
andyw8, line 17: What might be a better name for this?

vinistock: How about `ruby-lsp-debug`?

andyw8: How about `ruby-lsp-doctor` (like `brew doctor`) to avoid any confusion with the `debug` gem?

andyw8: My thought was that it might become a more general purpose tool for troubleshooting setups in future.

vinistock: Yeah, I agree with that. We may want to expand that for other functionality and renaming the executable later may prove painful.

st0012: I understand the argument but now it doesn't start the LSP server. If I saw it and assumed it's like `ruby-lsp` but with debugging output, I'd be very confused to see it just exits after indexing.

.github/workflows/ci.yml
```diff
@@ -42,3 +42,7 @@ jobs:
 
       - name: Run tests
         run: bundle exec rake
+
+      - name: Run index troubleshooting tool
+        if: matrix.os != 'windows-latest'
```
andyw8, line 47: I could not figure how why, but on Windows this fails with `bundler: not executable`.

vinistock (approved):

PR #1125, opened 2023-10-19
Declare all default gem dependencies in the gemspec

vinistock:
### Motivation

Closes https://github.com/Shopify/vscode-ruby-lsp/issues/848

@paracycle pointed out that the rubygems issue is likely because we depend on many default gems, but don't have them declared as runtime dependencies in the gemspec.

I tested this and it does make the issue go away.

### Implementation

1. Declared all dependencies in the gemspec
2. Added two missing requires


vinistock:
As discussed in https://github.com/Shopify/vscode-ruby-lsp/issues/848, this is not really the right solution.

PR #1126, opened 2023-10-20
Make the experimental flag install beta versions of the server

vinistock:
### Motivation

Instead of having the `branch` option to run the LSP from a branch, we can make the `experimental` flag install beta versions of the gem. This will help us get early feedback on non-stable features.

It's also easier to analyze in our metrics since they will have distinct version names.

### Implementation

Swapped the `branch` option for `experimental`. When the `experimental` flag is `true`, we run `bundle update ruby-lsp --pre`, which makes bundler accept prereleases.

### Automated Tests

Adapted tests.

### DIFF
--- a/test/setup_bundler_test.rb
+++ b/test/setup_bundler_test.rb
@@ -270,6 +270,38 @@ class SetupBundlerTest < Minitest::Test
     end
   end
 
+  def test_install_prerelease_versions_if_experimental_is_true
+    Dir.mktmpdir do |dir|
+      Dir.chdir(dir) do
+        File.write(File.join(dir, "Gemfile"), <<~GEMFILE)
+          source "https://rubygems.org"
+          gem "rdoc"
+        GEMFILE
+
+        capture_subprocess_io do
+          Bundler.with_unbundled_env do
+            # Run bundle install to generate the lockfile
+            system("bundle install")
+
+            # Run the script once to generate a custom bundle
+            run_script
+          end
+        end
+
+        capture_subprocess_io do
+          Object.any_instance.expects(:system).with(
+            bundle_env(".ruby-lsp/Gemfile"),
+            "((bundle check && bundle update ruby-lsp debug --pre) || bundle install) 1>&2",
+          )
+
+          Bundler.with_unbundled_env do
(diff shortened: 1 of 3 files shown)


andyw8:
Are we potentially losing a useful development feature of being able to point to specific branches for testing?

vinistock:
That's a fair point. Should we keep both? @st0012 what do you think?

vinistock:
Fair, I agree. I changed the implementation to only add the `experimental` flag, without removing `branch`.

test/setup_bundler_test.rb
```diff
@@ -270,6 +270,38 @@ def test_creates_custom_bundle_with_specified_branch
...
+
+            # Run the script once to generate a custom bundle
+            run_script
+          end
+        end
+
+        capture_subprocess_io do
+          Object.any_instance.expects(:system).with(
+            bundle_env(".ruby-lsp/Gemfile"),
+            "((bundle check && bundle update ruby-lsp debug --pre) || bundle install) 1>&2",
```
andyw8, line 294: Is there a risk of unintentionally using a pre-release version of `debug`?

andyw8 (approved):

test/setup_bundler_test.rb
vinistock: If there's a prerelease version of debug, it will definitely be installed. There's no way to prevent this though - unless we never upgraded debug.

vinistock: Yes, but that'll run `bundle update` twice, which is slow.

andyw8: Yeah, but if it only impacts the experimental flag then it's not so critical.

I assume `bundle update debug --pre ruby-lsp` isn't possible, i.e. `--pre` applies to everything?

Not a blocker anyway, feel free to ship.

vinistock: I think `--pre` applies to everything. To make it run two bundle updates only for the experimental flag will involve complicating the code a little more, since now we're just appending `--pre` at the end.

Let's ship like this for now and revisit in the future if it becomes an issue.

PR #1147, opened 2023-10-26
added ruby-lsp version to start up message

### LINKED ISSUE #1146: Show gem version in startup message
If we add the version to the startup message, e.g. `Starting Ruby LSP... (v1.2.3)`, then when someone shares the logs we won't need to separately ask about the version in use.

haiderashfaq:
closes #1146 

Added `VERSION` to start up messages

### DIFF
--- a/lib/ruby_lsp/server.rb
+++ b/lib/ruby_lsp/server.rb
@@ -58,7 +58,7 @@ module RubyLsp
 
     sig { void }
     def start
-      warn("Starting Ruby LSP...(v#{VERSION})")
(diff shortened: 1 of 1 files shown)


lib/ruby_lsp/server.rb
```diff
@@ -58,7 +58,7 @@ def initialize
 
     sig { void }
     def start
-      warn("Starting Ruby LSP...")
+      warn("Starting Ruby LSP...(v#{VERSION})")
```
vinistock, line 61:
```suggestion
      warn("Starting Ruby LSP v#{VERSION}...")
```

PR #1148, opened 2023-10-27
Provide constant type differentiation in semantic highlighting request

### LINKED ISSUE #1128: Semantic highlighting marking a class usage as a constant
### Ruby version

3.2.2

### Code snippet

This may not be the most lightweight way to reproduce it, but it's visible.
<details><summary>Dockerfile</summary>

```Dockerfile
# Use Alpine as the base image
FROM alpine:latest

# Set the working directory
WORKDIR /root

# Install necessary packages
RUN apk add --no-cache \
  git \
  curl \
  build-base \
  zlib-dev \
  openssl-dev \
  readline-dev \
  linux-headers \
  libffi-dev \
  cmake \
  automake \
  autoconf \
  libtool \
  pkgconf \
  gettext-tiny-dev \
  ruby \
  ruby-dev \
  bash \
  yaml-dev

# Clone and build Neovim from source
RUN git clone https://github.com/neovim/neovim.git && \
  cd neovim && \
  make CMAKE_BUILD_TYPE=Release && \
  make install

# Clone the LazyVim starter repository
RUN git clone https://github.com/LazyVim/starter ~/.config/nvim

# Copy the content of custom_plugins.lua from an external file
COPY custom_plugins.lua /root/.config/nvim/lua/plugins/custom_plugins.lua

# Create example.rb inline
RUN echo -e "class Foo\n  def self.bar\n    Foo.new # `Foo` should be the same color in both lines\n  end\nend" > ~/example.rb

# Create Gemfile inline
RUN echo "source 'https://rubygems.org'" > ~/Gemfile

# Install bundler and gems specified in the Gemfile
RUN gem install bundler && \
  bundle install

# Start Neovim headlessly and install plugins
RUN nvim -es -u ~/.config/nvim/init.lua -i NONE -c "Lazy" -c "qa"

# Start a shell when the Docker container is run
CMD ["sh"]
```

</details>

```lua
-- store this in the same folder of the Dockerfile as `custom_plugins.lua`
return {
    {
        "williamboman/mason.nvim",
        opts = function(_, opts)
            vim.list_extend(opts.ensure_installed, {
                "ruby-lsp",
            })
        end,
    },
    {
        "neovim/nvim-lspconfig",
        opts = {
            servers = {
                ruby_ls = {},
            },
        },
    },
}
```

Run:
```bash
docker build -t example .
docker run -ti example
# inside
nvim example.rb
```

Wait a bit and you'll see some notifications like `installing ruby_ls`, `ruby_ls installed`, and some seconds later the server will be started automatically. At that point you'll see the color of the second `Foo` change.

### Description

I wanted to replace solargraph with `ruby_ls` with neovim (and LazyVim). When switching I noticed that the highlighting was different. I've noticed that this is most probably because `ruby_ls` semantic highlighting.

I'm using https://github.com/maxmx03/solarized.nvim with `neo` theme, although I can reproduce it with other themes.
After using `:Inspect`, i've noticed the following

Screenshot 2023-10-20 at 20 30 14

`Foo` here is highlighted as a a namespace / constant, where on declaration it's as a class:

Screenshot 2023-10-20 at 20 36 11

### Expected output

I would expect that in both cases `Foo` is the same color (orange, as a `Type`)

mateusdeap:
### Motivation

Fixes #1128 (Potentially)

This PR aims to modify the `SemanticHighlighting` class so that it can properly determine what type of constant is being referenced in a given piece of code. The details and discussion can be found in the issue, but we want syntax highlighting to know that `Foo` in `Foo.new` below is referencing the class `Foo` and isn't just some other constant `Foo`, for example:

```ruby
class Foo
  THIS_IS_A_CONSTANT = 1

  def self.create_instance
    Foo.new
  end
end
```

### Implementation

Following @vinistock's recommendations, I made sure an instance of `RubyIndexer::Index` is given to the request and created a variable to keep track of nesting. Additionally subscribed to events needed to keep track of this nesting.

At the time of submission, the PR is incomplete, however. I wasn't able to figure out how to properly set up the index in the tests, so the lines where I call `index_single` are definitely something that needs fixing.

Finally, one thing discussed in the issue is that, even if this is implemented correctly, it might be too resource intensive, so if this gets finished, benchmarks will be required

### Automated Tests

Haven't added any new tests so far but only tried to modify the current test to reflect the needed changes.

### Manual Tests

The issue was opened referencing neovim, so I'm not sure if this affects VS Code. However the request doesn't have this specific capability so I think it's surely possible to replicate this by opening the example file provided in the issue.

### DIFF
--- a/lib/ruby_lsp/server.rb
+++ b/lib/ruby_lsp/server.rb
@@ -58,7 +58,7 @@ module RubyLsp
 
     sig { void }
     def start
-      warn("Starting Ruby LSP...")
+      warn("Starting Ruby LSP v#{VERSION}...")
 
       # Requests that have to be executed sequentially or in the main process are implemented here. All other requests
       # fall under the else branch which just pushes requests to the queue


vinistock (commented):
Keeping track of the nesting is looking good. Now, when we find a constant reference, we need to use the constant name, the nesting and the index to determine what the constant is declared as.

```ruby
def on_constant_read_node_enter(node)
  entries = @index.resolve(node.name, @nesting)
  # if entries is nil, we didn't find the declaration for it
  # it might be defined with meta-programming or using C code
  unless entries
    # Push a constant semantic token
  end

  # Otherwise, we can check the type of entry to determine the
  # type of constant
  first_entry = T.must(entries.first)

  case first_entry
  when RubyIndexer::Entry::Class
    # push a class.declaration token
  when RubyIndexer::Entry::Module
    # push a namespace.declaration token
  when RubyIndexer::Entry::Constant
    # push a constant.declaration token
  end
end
```

PR #1153, opened 2023-10-30
Improve document symbols of singleton class operators and methods

vinistock:
### Motivation

It was brought up that we weren't properly handling singleton class operator nodes in document symbol and indeed we were missing a few things. This PR addresses the singleton node aspects we were missing.

### Implementation

This PR
- Starts handling the `class << something` nodes and pushing symbols for it
- Starts prefixing methods defined inside these blocks with `(expression.)`, where the expression is the right hand side of `class <<`
- Changes the symbol kind of singleton methods from `method` to `function`, which better represents them and allows us to differentiate the two

### Automated Tests

Added a new fixture and fixed an existing one.

### DIFF
--- a/lib/ruby_lsp/requests/document_symbol.rb
+++ b/lib/ruby_lsp/requests/document_symbol.rb
@@ -73,6 +73,8 @@ module RubyLsp
           :on_module_node_leave,
           :on_instance_variable_write_node_enter,
           :on_class_variable_write_node_enter,
+          :on_singleton_class_node_enter,
+          :on_singleton_class_node_leave,
         )
       end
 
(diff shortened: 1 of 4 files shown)


lib/ruby_lsp/requests/document_symbol.rb
```diff
@@ -163,10 +182,15 @@ def on_module_node_enter(node)
...
         receiver = node.receiver
+        previous_symbol = @stack.last
 
         if receiver.is_a?(Prism::SelfNode)
           name = "self.#{node.name}"
-          kind = Constant::SymbolKind::METHOD
+          kind = Constant::SymbolKind::FUNCTION
+        elsif previous_symbol.is_a?(Interface::DocumentSymbol) && previous_symbol.name.start_with?("<<")
+          _operator, expression = previous_symbol.name.split(" ")
+          name = "(#{expression}.)#{node.name}"
```
paracycle, line 192: I am not sure adding the `(expression.)` prefix makes this more readable. Since these methods will be listed under the `<< expression` symbol anyway, I think it is fine to just list them with `node.name` only.

RubyMine seems to do it like that too, btw.

vinistock: @nirvdrum since you originally reported this, do you have any preferences / opinions? I'm fine either way.

rafaelfranca: In the document symbol I'd just have `node.name` but I do see value in inline hits given some of those `class << self` can span multiple lines and you can't actually see anymore a method is inside a `class << self`. I do understand it is confusing though since `class << self; def self.foo; end` has a different result to `class << self; def foo; end`. Maybe there is other kind of hint that would make easy to know exactly that method is.

paracycle (approved):

PR #1161, opened 2023-11-01
Lint RBI shims in CI

andyw8:
### Motivation

Ensure RBIs are linted in CI.

### Implementation

* Add a CI step
* Fix existing offenses

### Automated Tests

n/a

### Manual Tests

n/a


sorbet/rbi/shims/prism.rbi
```diff
@@ -89,11 +89,6 @@ module Prism
     def end_column; end
   end
 
-  class Comment
```
andyw8, line 92: Duplicate method.

.github/workflows/ci.yml
```diff
@@ -40,6 +40,10 @@ jobs:
       - name: Lint Ruby files
         run: bundle exec rubocop
 
+      - name: Lint RBI shims
+        if: matrix.os != 'windows-latest'
```
andyw8, line 44: I'm unsure why this fails on Windows:
https://github.com/Shopify/ruby-lsp/actions/runs/6723189944/job/18272722122

andyw8: > We don't install Tapioca, Sorbet, Spoom, etc on Windows because Sorbet is no compatible. That's likely why.

But how would that result a `Layout/EndOfLine` failure?

> I think this should be a separate job from CI. There's no reason why we would check shims for every OS and every Ruby version.

Yeah, I think we can avoid of lot of those runs. I've updated `ci.yml` to split the job into two (it might be better to have this in a separate PR though?).

.github/workflows/ci.yml
```diff
@@ -3,6 +3,34 @@ name: CI
...
+      - uses: actions/checkout@v3
+        with:
+          submodules: "recursive"
+
+      - name: Set up Ruby
+        uses: ruby/setup-ruby@v1
+        with:
+          ruby-version: "3.2"
+          bundler-cache: true
+          cache-version: 1
```
andyw8, line 18: (I'll remove this)

vinistock (approved):

.github/workflows/ci.yml
vinistock: Maybe we lint the shims and that fails because on Windows the end of line is expected to be `\r\n` rather then `\n`. Not sure.

PR #1171, opened 2023-11-07
Use a queue instead of visitor for indexing

### LINKED ISSUE #1164: Indexing fails due to stack level too deep error
### Operating System

macOS Monterey

### Ruby version

3.1.1p18

### Project has a bundle

- [X] Has bundle

rbenv

### Description

Indexing fails at 18% on intitialization of ruby-lsp.  It reaches a StackLevelTooDeep error while indexing the Prism gem.  

```
#<Thread:0x0000000108307a38 /Users/user/.rbenv/versions/3.1.1/lib/ruby/gems/3.1.0/gems/ruby-lsp-0.12.2/lib/ruby_lsp/executor.rb:213 run> terminated with exception (report_on_exception is true):
/Users/user/.rbenv/versions/3.1.1/lib/ruby/gems/3.1.0/gems/prism-0.15.1/lib/prism/visitor.rb:23:in `each': stack level too deep (SystemStackError)
	from /Users/user/.rbenv/versions/3.1.1/lib/ruby/gems/3.1.0/gems/prism-0.15.1/lib/prism/visitor.rb:23:in `visit_child_nodes'
	from /Users/user/.rbenv/versions/3.1.1/lib/ruby/gems/3.1.0/gems/prism-0.15.1/lib/prism/node.rb:14924:in `accept'
	from /Users/user/.rbenv/versions/3.1.1/lib/ruby/gems/3.1.0/gems/prism-0.15.1/lib/prism/visitor.rb:23:in `block in visit_child_nodes'
	from /Users/user/.rbenv/versions/3.1.1/lib/ruby/gems/3.1.0/gems/prism-0.15.1/lib/prism/visitor.rb:23:in `each'
	from /Users/user/.rbenv/versions/3.1.1/lib/ruby/gems/3.1.0/gems/prism-0.15.1/lib/prism/visitor.rb:23:in `visit_child_nodes'
	from /Users/user/.rbenv/versions/3.1.1/lib/ruby/gems/3.1.0/gems/prism-0.15.1/lib/prism/node.rb:14924:in `accept'
	from /Users/user/.rbenv/versions/3.1.1/lib/ruby/gems/3.1.0/gems/prism-0.15.1/lib/prism/visitor.rb:23:in `block in visit_child_nodes'
	from /Users/user/.rbenv/versions/3.1.1/lib/ruby/gems/3.1.0/gems/prism-0.15.1/lib/prism/visitor.rb:23:in `each'
	 ... 528 levels...
	from /Users/user/.rbenv/versions/3.1.1/lib/ruby/gems/3.1.0/gems/ruby-lsp-0.12.2/lib/ruby_indexer/lib/ruby_indexer/index.rb:169:in `index_all'
	from /Users/user/.rbenv/versions/3.1.1/lib/ruby/gems/3.1.0/gems/sorbet-runtime-0.5.11114/lib/types/private/methods/_methods.rb:277:in `bind_call'
	from /Users/user/.rbenv/versions/3.1.1/lib/ruby/gems/3.1.0/gems/sorbet-runtime-0.5.11114/lib/types/private/methods/_methods.rb:277:in `block in _on_method_added'
	from /Users/user/.rbenv/versions/3.1.1/lib/ruby/gems/3.1.0/gems/ruby-lsp-0.12.2/lib/ruby_lsp/executor.rb:215:in `block in perform_initial_indexing'
```

vinistock:
### Motivation

Closes #1164

This is an alternative to #1170.

Although using visitors allows us to neatly organize our code, for performance critical operations like indexing we pay a significant price due to the recursive visits. In the associated issue, we do so much recursion we actually raise a stack error.

To prevent these issues in the future, I believe it would be worth using a queue mechanism only for the indexer (LSP requests will remain as visitors).

This not only fixes the issue of hitting too many levels and raising stack errors, but it also speeds up indexing significantly (25% on a few benchmarks I ran).

### Feedback

Do people agree with switching from visitor to queue for the indexer? Any major concerns?

### Implementation

Turned our index visitor into a collector using a queue. The code style is, of course, not as neatly organized - but it's also not terrible.

### Automated Tests

Added a test that currently raises the stack error on main.

### DIFF
--- a/lib/ruby_indexer/lib/ruby_indexer/collector.rb
+++ b/lib/ruby_indexer/lib/ruby_indexer/collector.rb
@@ -2,8 +2,10 @@
 # frozen_string_literal: true
 
 module RubyIndexer
-  class IndexVisitor < Prism::Visitor
+  class Collector
     extend T::Sig
+
+    LEAVE_EVENT = T.let(Object.new.freeze, Object)
 
     sig { params(index: Index, parse_result: Prism::ParseResult, file_path: String).void }
     def initialize(index, parse_result, file_path)
@@ -16,22 +18,62 @@ module RubyIndexer
         end,
         T::Hash[Integer, Prism::Comment],
       )
+      @queue = T.let([], T::Array[Object])
 
       super()
     end
 
-    sig { override.params(node: Prism::ClassNode).void }
-    def visit_class_node(node)
-      add_class_entry(node)
+    sig { params(node: Prism::Node).void }
+    def collect(node)
+      @queue = [node]
+
+      until @queue.empty?
+        node_or_event = @queue.shift
+
+        case node_or_event
+        when Prism::ProgramNode
+          @queue << node_or_event.statements
+        when Prism::StatementsNode
+          T.unsafe(@queue).prepend(*node_or_event.body)
+        when Prism::ClassNode
+          add_class_entry(node_or_event)
+        when Prism::ModuleNode
+          add_module_entry(node_or_event)
+        when Prism::MultiWriteNode
+          handle_multi_write_node(node_or_event)
+        when Prism::ConstantPathWriteNode
+          handle_constant_path_write_node(node_or_event)
+        when Prism::ConstantPathOrWriteNode
+          handle_constant_path_or_write_node(node_or_event)
+        when Prism::ConstantPathOperatorWriteNode
+          handle_constant_path_operator_write_node(node_or_event)
+        when Prism::ConstantPathAndWriteNode
+          handle_constant_path_and_write_node(node_or_event)
+        when Prism::ConstantWriteNode
+          handle_constant_write_node(node_or_event)
+        when Prism::ConstantOrWriteNode
+          name = fully_qualify_name(node_or_event.name.to_s)
+          add_constant(node_or_event, name)
+        when Prism::ConstantAndWriteNode
+          name = fully_qualify_name(node_or_event.name.to_s)
+          add_constant(node_or_event, name)
+        when Prism::ConstantOperatorWriteNode
+          name = fully_qualify_name(node_or_event.name.to_s)
+          add_constant(node_or_event, name)
+        when Prism::CallNode
+          handle_call_node(node_or_event)
+        when Prism::DefNode
+          handle_def_node(node_or_event)
+        when LEAVE_EVENT
+          @stack.pop
+        end
+      end
     end
 
-    sig { override.params(node: Prism::ModuleNode).void }
-    def visit_module_node(node)
-      add_module_entry(node)
-    end
+    private
 
-    sig { override.params(node: Prism::MultiWriteNode).void }
-    def visit_multi_write_node(node)
+    sig { params(node: Prism::MultiWriteNode).void }
+    def handle_multi_write_node(node)
       value = node.value
       values = value.is_a?(Prism::ArrayNode) && value.opening_loc ? value.elements : []
 
@@ -50,8 +92,8 @@ module RubyIndexer
       end
     end
 
-    sig { override.params(node: Prism::ConstantPathWriteNode).void }
-    def visit_constant_path_write_node(node)
+    sig { params(node: Prism::ConstantPathWriteNode).void }
+    def handle_constant_path_write_node(node)
       # ignore variable constants like `var::FOO` or `self.class::FOO`
       target = node.target
       return unless target.parent.nil? || target.parent.is_a?(Prism::ConstantReadNode)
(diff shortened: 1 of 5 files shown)


lib/ruby_indexer/lib/ruby_indexer/collector.rb
```diff
@@ -204,20 +226,26 @@ def add_constant(node, name, value = nil)
     sig { params(node: Prism::ModuleNode).void }
     def add_module_entry(node)
       name = node.constant_path.location.slice
-      return visit_child_nodes(node) unless /^[A-Z:]/.match?(name)
+      unless /^[A-Z:]/.match?(name)
```
paracycle, line 229: Not related to this PR, but is this trying to check if something is a constant? If so, this will be missing with cases where Unicode characters are being used in constant names.

For example:
```ruby
module Çava; end
defined? Çava # => "constant"
/^[A-Z:]/.match?(Çava.name) # => false
```

lib/ruby_indexer/lib/ruby_indexer/collector.rb
```diff
@@ -204,20 +226,26 @@ def add_constant(node, name, value = nil)
...
-      @stack.pop
+      @queue.unshift(node.body, LEAVE_EVENT)
     end
 
     sig { params(node: Prism::ClassNode).void }
     def add_class_entry(node)
       name = node.constant_path.location.slice
-      return visit_child_nodes(node) unless /^[A-Z:]/.match?(name)
+
+      unless /^[A-Z:]/.match?(name)
```
paracycle, line 245: Same comment as above

lib/ruby_indexer/lib/ruby_indexer/collector.rb
```diff
@@ -227,10 +255,15 @@ def add_class_entry(node)
...
+        fully_qualify_name(name),
+        @file_path,
+        node.location,
+        comments,
+        parent_class,
+      )
       @stack << name
-      visit(node.body)
-      @stack.pop
+      @queue.unshift(node.body, LEAVE_EVENT)
```
paracycle, line 266: Instead of creating a general purpose event name, I wonder if we should be queueing `Proc` instances, so that the collector can handle `Proc`s and call them.

That way, you will be able to queue any custom behaviour that you want without having to add extra event names.

This line would become:
```suggestion
      @queue.unshift(node.body, -> { @stack.pop })
```

paracycle (approved):
Looks good to me, I have a few comments

lib/ruby_indexer/lib/ruby_indexer/collector.rb
vinistock: If you're okay with it, I would defer that decision until we have a use case for it. Currently, there's no other action we want to take other than popping the stack and I'm not sure about the performance impact of the procs.

lib/ruby_indexer/lib/ruby_indexer/collector.rb
```diff
@@ -16,22 +18,62 @@ def initialize(index, parse_result, file_path)
...
+      @queue = [node]
+
+      until @queue.empty?
+        node_or_event = @queue.shift
+
+        case node_or_event
+        when Prism::ProgramNode
+          @queue << node_or_event.statements
+        when Prism::StatementsNode
+          T.unsafe(@queue).unshift(*node_or_event.body)
```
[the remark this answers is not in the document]
paracycle: Oh right, the name `unshift` always confuses me. Can we start using `prepend` instead, since it is an alias to `unshift`.

st0012:
Will the performance difference still be this big when we start indexing methods? If the difference would shrink to, like 10%, then that's something I'm willing to pay in order to have 1 single pattern in the codebase.

vinistock:
> Will the performance difference still be this big when we start indexing methods? If the difference would shrink to, like 10%, then that's something I'm willing to pay in order to have 1 single pattern in the codebase.

We already index methods. We're just missing a few argument handling scenarios + accounting for singleton methods, but we're already indexing all methods.

Also, I don't see why adding methods would reduce the benefit. If anything, it would increase the benefit, because handling methods would mean we would be doing more recursion.

lib/ruby_indexer/lib/ruby_indexer/collector.rb
vinistock: I put a comment in our issue with those findings - I didn't know we wouldn't need to register all of the unicode characters, that's great.

Yeah, we were using the regex to discard usages of `module foo::Bar`, but we can indeed do something better with the node types.

st0012:
> We already index methods. We're just missing a few argument handling scenarios + accounting for singleton methods, but we're already indexing all methods.

Sorry I wasn't aware of it. Just checked the code again and see it now.

> If anything, it would increase the benefit

👍

st0012 (approved):

Issue #1007, opened 2023-09-14
Support completing heredoc delimiter in on type formatting


vinistock:
It'd be nice to get the closing delimiter of a heredoc automatically inserted - similar to what we do for `end` tokens.

For example
```ruby
<<~RUBY # the moment the dev breaks the line we insert the closing RUBY delimiter
```

PR #1181, opened 2023-11-13
feat: Add support for autocompleting HEREDOC delimiter

DRBragg:
### Motivation

We'd like to add support for autocompleting HEREDOC delimiters as requested in #1007 

### Implementation

We added to the `\n` branch of the case statement within the `run` method in the `OnTypeFormatting` class to compare our HEREDOC regex to the previous line. We handled that conditional with a new method to insert the delimiter. 

### Manual Tests

- Open a new Ruby file
- Type `str = <<~STR`
- Hit enter/return
- Observe that `STR` is added to the file.


vinistock (approved):
Awesome! Thank you for the contribution ❤️

Issue #1055, opened 2023-10-02
Add support for providing features on ERB files


vinistock:
I think there's a path to add support for ERB without a ton of effort. We can try to require `erb` from the main bundle and provide features only if it is a dependency.

Basically, we can use the ERB gem to parse the `.erb` file and extract the Ruby code. Then we parse the Ruby code with Prism and provide features for it.

```ruby
require "erb"
erb = "<% x = 32 %><p><%= x %></p>"
ruby_code = ERB.new(erb).src # => "#coding:UTF-8\n_erbout = +'';  x = 32 ; _erbout.<< \"<p>\".freeze; _erbout.<<(( x ).to_s); _erbout.<< \"</p>\".freeze; _erbout"

Prism.parse(ruby_code) # => AST
```

In terms of the codebase, I think the best way forward would be turning creating an abstract parent class `Document` and turning our existing document class into `RubyDocument`. Then we can create an `ErbDocument` class to handle the specific details of ERB.

### Notes

Diagnostics, formatting and on type formatting need to happen differently on ERB files. For example, if some is using `[].each do %>` and breaks the line, we need to insert `<% end %>` and not just `end`. Similarly, we can see if erb lint is available to provide diagnostics and potentially format the code too.

It might be worth separating these requests into `RubyFormatting` and `ErbFormatting`.

vinistock:
Attempt at erasing the HTML

```ruby
# typed: strict
# frozen_string_literal: true

require "prism"
require "strscan"

original = File.read("index.html.erb")
scanner = StringScanner.new(original)
output = +""

until scanner.eos?
  non_ruby_code = scanner.scan_until(/<%(-|=)?/)
  break unless non_ruby_code

  output << non_ruby_code.gsub(/[^\n]/, " ")

  ruby_code = scanner.scan_until(/(-)?%>/)
  break unless ruby_code

  output << ruby_code[...-2]
  output << "  "
end

puts output

p Prism.parse(output)
```

vinistock:
We can use VS Code's embedded language features to forward HTML operations to the built-in language server and get features: https://code.visualstudio.com/api/language-extensions/embedded-languages.

vinistock:
Initial support for ERB has been shipped in #2235, so I'm closing this issue in favour of more specific ones if we need them.

PR #1184, opened 2023-11-13
Add ERB Support

marcoroth:
Address #1055


natematykiewicz:
Hey @marcoroth! What's the status of this one? I'm really wishing that Ruby LSP supported [ERB Lint](https://github.com/Shopify/erb-lint).

marcoroth:
We are currently missing the right tools to fully provide really useful support for ERB. This PR only adds super basic support.

I guess the question is if we want to finish up and merge the most minimalistic ERB support (which is implemented here) or if we should wait until we can build the needed tools.

I do have some ideas for the tooling and needed infrastructure, but it needs more coordination and work, which is out of scope for this PR.

Issue #1104, opened 2023-10-16
Some highlights coloured incorrectly (shorthand hash syntax)


wJoenn:
### Ruby version

3.3.2

### Code snippet

```ruby
def some_method
  "do something"
end

some_hash = { some_method:, some_key: "dark blue" }

SOME_CONSTANT = some_hash
```

### Description

A constant is highlighted in green (like a class)
And when using a method as an implicit value inside an hash the key is highlighted in yellow (like a method)

![image](https://github.com/Shopify/ruby-lsp/assets/75388869/bcea3e97-9463-436e-a983-26d109dc7684)

### Expected output

I think the constant should be blue like variables and the key should be dark blue like a symbol

vinistock:
Thank you for the report.

Note that the Ruby LSP does not control which colours are used to highlight the code. The Ruby LSP only provides semantic tokens and it's the theme's responsibility to decide how to highlight them.

The Ruby LSP semantic token information differentiates between variables and constants. If you want them to be the same colour, that would require adjusting your theme.

Concerning the short hand hash syntax, Ruby itself considers the key to be a method call, which is why it gets highlighted as one. Example:

```shell
pp Prism.parse("{ foo: }").value


@ ProgramNode (location: (1,0)-(1,8))
├── locals: []
└── statements:
    @ StatementsNode (location: (1,0)-(1,8))
    └── body: (length: 1)
        └── @ HashNode (location: (1,0)-(1,8))
            ├── opening_loc: (1,0)-(1,1) = "{"
            ├── elements: (length: 1)
            │   └── @ AssocNode (location: (1,2)-(1,6))
            │       ├── key:
            │       │   @ SymbolNode (location: (1,2)-(1,6))
            │       │   ├── opening_loc: ∅
            │       │   ├── value_loc: (1,2)-(1,5) = "foo"
            │       │   ├── closing_loc: (1,5)-(1,6) = ":"
            │       │   └── unescaped: "foo"
            │       ├── value:
            │       │   @ ImplicitNode (location: (1,2)-(1,6))
            │       │   └── value:
            │       │       @ CallNode (location: (1,2)-(1,6))
            │       │       ├── receiver: ∅
            │       │       ├── call_operator_loc: ∅
            │       │       ├── message_loc: (1,2)-(1,5) = "foo"
            │       │       ├── opening_loc: ∅
            │       │       ├── arguments: ∅
            │       │       ├── closing_loc: ∅
            │       │       ├── block: ∅
[...]

Issue #834, opened 2023-07-24
Allow Code Lens features to be configured at a more granular level


andyw8:
https://ruby-dx.slack.com/archives/C0515KVRNE4/p1690113893796309

e.g. allow the Gemfile Code Lens to be disabled independently from the test running ones.

We would still want to have a single [Code Lens request](https://github.com/Shopify/ruby-lsp/blob/main/lib/ruby_lsp/requests/code_lens.rb), but adjusted to accept a configuration option.

A corresponding change will be needed for [vscode-ruby-lsp](https://github.com/Shopify/vscode-ruby-lsp).

We may also need to to consider how to 'migrate' users to the new configuration.

simonc:
Hi there! 🙂

I would have been really interested in this feature. Being able to disable the Gemfile Code Lens would be 👌 

Cheers ❤️

vinistock:
@snutij sorry about the delay. To support this we will need to start receiving inlay hint configurations from the initialize options, similar to what we do for [formatter](https://github.com/Shopify/ruby-lsp/blob/9a2afd2e91e3d5ee6d9ec21e1fc5f63e30d553e6/lib/ruby_lsp/executor.rb#L597).

We'd need to store the configuration for inlay hints in the `Store` class and then pass it to inlay hint [here](https://github.com/Shopify/ruby-lsp/blob/9a2afd2e91e3d5ee6d9ec21e1fc5f63e30d553e6/lib/ruby_lsp/executor.rb#L396).

I think we can define the options structure to be like this

```ruby
{
  inlayHint: {
    gemLinks: false,
    implicitRescue: true,
    implicitHashValue: false,
  }
}
```

I'd start with only the rescue hint enabled by default since it seems this feature is quite controversial. After the server supports the new config, then we need to change the extension to surface this in VS Code.

We will need a new [setting](https://github.com/Shopify/vscode-ruby-lsp/blob/64a004cab6683ddb9bb7deda0bafb54ae454309c/package.json#L100) using an object type, where each property is the name of the hint with a boolean value (to match the server). Then we need to read the configured hints and send them to the server as part of the [initialization options](https://github.com/Shopify/vscode-ruby-lsp/blob/64a004cab6683ddb9bb7deda0bafb54ae454309c/src/client.ts#L115).

PR #1186, opened 2023-11-14
Add inlay hint for omitted hash values

RyanBrushett:
### Motivation

[In this discussion](https://github.com/Shopify/ruby-lsp/issues/1104#issuecomment-1766733528), @andyw8 mentions how we could add an inlay hint for an omitted hash value to be a bit clearer about what's going on.

### Implementation

- Followed how the inlay hint for implied `rescue StandardError` works
- I added an example but it's kind of hard to write out how it works... let me know if I could make it better
- I think there's only three Node types that `node.value` could be when the `node` is an `ImplicitNode` but let me know if I'm missing anything

### Automated Tests

I added some expectation tests!

### Manual Tests

You could test this by opening the `ruby-lsp` project in VS code, restart the extension so it loads the development version, and then write some code with implied hash values. This should do it:

```ruby
some_variable = "hello"
{some_variable:} # hint will be added here
```

### Also

❤️

### DIFF
--- a/lib/ruby_lsp/requests/inlay_hints.rb
+++ b/lib/ruby_lsp/requests/inlay_hints.rb
@@ -18,6 +18,16 @@ module RubyLsp
     #   puts "handle some rescue"
     # end
     # ```
+    #
+    # # Example
+    #
+    # ```ruby
+    # var = "foo"
+    # {
+    #   var: var, # Label "var" goes here in cases where the value is omitted
+    #   a: "hello",
+    # }
+    # `
     class InlayHints < Listener
       extend T::Sig
       extend T::Generic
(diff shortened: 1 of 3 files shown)


Morriar (approved):
Thanks!

davidalejandroaguilar:
Thanks for this PR! However, how can we disable this? The reason I omit the value is because I don't want it to be shown 😃 

(I'm guessing this PR is the one that added this):

image

image

vinistock:
There's currently no way to configure inlay hints with that level of granularity. We still need to accept a more complex configuration for it.

vinistock:
It's already being worked on as part of https://github.com/Shopify/ruby-lsp/issues/834, which is related.

Issue #899, opened 2023-08-17
Properly handle methods for definition, completion and hover


vinistock:
Depends on #898 and #1333

Use the index to provide go to definition for methods.

Might depend on exploring using a control flow graph (CFG) to implement reaching definitions. While we do not provide typechecking, using reaching definitions might allow us to be more precise.

For example

```ruby
a = "something"
# We know that `a` is a string because it's a literal, so we know 100% sure this has to be
# String#upcase. If we have an implementation of reaching definitions, we can make this work
a.upcase
```

Checklist

- [x] Attributes. The LSP protocol has special types for them, so we should create a separate index entry
- [ ] Taking `new` invocations to `initialize`. For completion, it's actually the opposite, we need to use the signature for `initialize`, but complete to `new`
- [x] Method aliases through `alias_method` and `alias`

Kyrluckechuck:
Just bumping this since it's not stale and I don't want it marked as closed without being formally acknowledged! 😅

andyw8:
It's not usable yet, but a first step towards this was merged today: https://github.com/Shopify/ruby-lsp/pull/1043

Drowze:
Method completion and jump-to-method-definition is the only thing currently holding me to other alternatives. I am anxiously looking towards this feature to jump on ruby-lsp boat! 🛥️ 🛥️ 

Thank you very much for your hard work on this project!

vinistock:
@rosslyn-beckynorville I encourage you (and others!) to give it a try right now. Here's a little progress report so that people can understand the current state.

Note that all features are supported both for project declarations and gems, so you can use these features while navigating inside a gem's source code as well.

In addition to that, note that meta-programming constructs that cannot be analyzed statically are impossible to support for any of the features, so those are excluded. For example

```ruby
some_var_that_depends_on_user_input.const_get(:Whatever).send(other_var, ...)
```

**Go to definition support**

1. Constants: fully supported
2. Instance variables: fully supported
3. Class instance variables: almost fully supported, we need to merge https://github.com/Shopify/ruby-lsp/pull/2214 to properly analyze singleton inheritance
4. Methods: methods are the most difficult to support because they require much more complex type flow analysis than other Ruby constructs. Let's break down how it currently works:
    - Methods invoked directly on `self`: fully supported (also excludes meta-programming, like invoking a method with `instance_exec` thus mutating the meaning of `self`)
[...]

PR #1195, opened 2023-11-20
Support methods invoked on self

vinistock:
### Motivation

Big step towards #899 

I got frustrated that we keep getting feedback about missing method support and powered through some basic implementations.

This PR adds support for

- Workspace symbols (for all methods)
- Definition, completion and hover for methods invoked directly on `self` (since the receiver is known)

We are not taking inheritance into account yet, this is just a first step. The same ideas from this PR can be applied to singleton methods once #1113 is merged.

Then we'll just be missing inheritance and methods with unknown receivers.

### Implementation

I recommend reviewing per commit. For the most part, all implementations just use `resolve_method` or check method owners when necessary.

The only exception is workspace symbol. We were just not supporting the method kind, which completely implements the functionality.

### Automated Tests

Added tests for all of these.

### DIFF
--- a/lib/ruby_lsp/requests/completion.rb
+++ b/lib/ruby_lsp/requests/completion.rb
@@ -127,7 +127,7 @@ module RubyLsp
       sig { params(node: Prism::CallNode).void }
       def on_call_node_enter(node)
         return if DependencyDetector.instance.typechecker
-        return if node.receiver
+        return unless self_receiver?(node)
 
         name = node.message
         return unless name
--- a/lib/ruby_lsp/requests/completion.rb
+++ b/lib/ruby_lsp/requests/completion.rb
@@ -6,9 +6,14 @@ module RubyLsp
     # ![Completion demo](../../completion.gif)
     #
     # The [completion](https://microsoft.github.io/language-server-protocol/specification#textDocument_completion)
-    # suggests possible completions according to what the developer is typing. Currently, completion is support for
-    # - require paths
-    # - classes, modules and constant names
+    # suggests possible completions according to what the developer is typing.
+    #
+    # Currently supported targets:
+    # - Classes
+    # - Modules
+    # - Constants
+    # - Require paths
+    # - Methods invoked on self only
     #
     # # Example
     #
@@ -45,6 +50,7 @@ module RubyLsp
           :on_string_node_enter,
           :on_constant_path_node_enter,
           :on_constant_read_node_enter,
+          :on_call_node_enter,
         )
       end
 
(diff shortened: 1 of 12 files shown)


lib/ruby_lsp/requests/hover.rb
```diff
@@ -95,6 +96,25 @@ def on_constant_path_node_enter(node)
         generate_hover(node.slice, node.location)
       end
 
+      sig { params(node: Prism::CallNode).void }
+      def on_call_node_enter(node)
+        return if DependencyDetector.instance.typechecker
+        return if node.receiver
```
Morriar, line 102: I think we can also consider calls on explicit `self` when the receiver is a `SelfNode`.

lib/ruby_lsp/requests/support/common.rb
```diff
@@ -82,7 +85,7 @@ def markdown_from_index_entries(title, entries)
           markdown_title = "```ruby\n#{title}\n```"
           definitions = []
           content = +""
-          entries.each do |entry|
+          Array(entries).each do |entry|
```
Morriar, line 88: Interesting. This wasn't caught by Sorbet before?

test/requests/hover_expectations_test.rb
```diff
@@ -61,6 +61,38 @@ class A
...
+    store = RubyLsp::Store.new
+
+    uri = URI("file:///fake.rb")
+    source = <<~RUBY
+      class A
+        # Hello from `foo`
+        def foo; end
+
+        def bar
+          foo
```
Morriar, line 75: Let's add `self.foo` as well.

lib/ruby_lsp/requests/hover.rb
```diff
@@ -95,6 +96,25 @@ def on_constant_path_node_enter(node)
...
+
+        message = node.message
+        return unless message
+
+        target_method = @index.resolve_method(message, @nesting.join("::"))
```
Morriar, line 107: I wonder if we can also try to resolve `new` to `initialize` here?

test/requests/workspace_symbol_test.rb
```diff
@@ -105,4 +105,16 @@ class Foo
     assert_equal(1, result.length)
     assert_equal("Foo", T.must(result.first).name)
   end
+
+  def test_returns_method_symbols
+    @index.index_single(RubyIndexer::IndexablePath.new(nil, "/fake.rb"), <<~RUBY)
+      class Foo
+        def bar; end
```
Morriar, line 112: Let's also add an example for `def initialize`?

lib/ruby_lsp/requests/definition.rb
```diff
@@ -75,8 +80,52 @@ def merge_response!(other)
...
+      sig { params(node: Prism::ConstantReadNode).void }
+      def on_constant_read_node_enter(node)
+        find_in_index(node.slice)
+      end
+
+      private
+
+      sig { params(node: Prism::CallNode).void }
+      def handle_method_definition(node)
+        return if node.receiver
```
Morriar, line 105: Same remark for explicit `self`.

lib/ruby_lsp/requests/definition.rb
```diff
@@ -75,8 +80,52 @@ def merge_response!(other)
...
+
+        message = node.message
+        return unless message
+
+        target_method = @index.resolve_method(message, @nesting.join("::"))
```
Morriar, line 110: Same question about `new <> initialize`.

lib/ruby_lsp/requests/hover.rb
vinistock: Good point. Added that.

test/requests/hover_expectations_test.rb
vinistock: Let's please add singleton methods after #1113. There are a few refactors we need to do for that PR that will influence the implementation (e.g.: getting rid of the `SingletonMethod` class and just storing the right owner).

lib/ruby_lsp/requests/hover.rb
vinistock: Let's also wait for the singleton class support since it requires handling the receiver in the same way. I added an item to our checklist in #899 so that we do not forget.

test/requests/workspace_symbol_test.rb
vinistock: Done!

lib/ruby_lsp/requests/definition.rb
vinistock: I added on the issue so that we don't forget.

test/requests/hover_expectations_test.rb
vinistock: Oh, sorry, I got confused. Added the test cases for explicit `self.foo`.

lib/ruby_lsp/requests/completion.rb
```diff
@@ -118,8 +124,48 @@ def on_constant_path_node_enter(node)
         end
       end
 
+      sig { params(node: Prism::CallNode).void }
+      def on_call_node_enter(node)
+        return if DependencyDetector.instance.typechecker
+        return if node.receiver
```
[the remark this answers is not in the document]
vinistock: Ah, good catch. Moved it into `Common` and started using it for completion.

Morriar (approved):

PR #1201, opened 2023-11-22
Avoid Bundler platform issues by moving sorbet out of development group

andyw8:
### Motivation

Due to recent changes in Bundler, it is causing platforms to be removed:

https://github.com/dependabot/dependabot-core/issues/8444

(can replicate locally with `bundle update --bundler`).

### Implementation

Avoid having `sorbet` in a group in the `Gemfile`. Since this a gem, the `gemspec` is sufficient to declare development vs runtime dependencies.

I initially tried complete removing the `development` group but it's causing some failures in `lib/ruby_indexer/test/configuration_test.rb`, so this is a less disruptive approach for now.

### Automated Tests

n/a

### Manual Tests

n/a

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -19,10 +19,11 @@ group :development do
   gem "rubocop-minitest", "~> 0.33.0", require: false
   gem "rubocop-rake", "~> 0.6.0", require: false
   gem "rubocop-sorbet", "~> 0.7", require: false
-  gem "sorbet-static-and-runtime", platforms: NON_WINDOWS_PLATFORMS
   gem "tapioca", "~> 0.11", require: false, platforms: NON_WINDOWS_PLATFORMS
   gem "rdoc", require: false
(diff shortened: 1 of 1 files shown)


lib/ruby_indexer/test/configuration_test.rb
```diff
@@ -14,8 +14,8 @@ def test_load_configuration_executes_configure_block
       indexables = @config.indexables
 
       assert(indexables.none? { |indexable| indexable.full_path.include?("test/fixtures") })
-      assert(indexables.none? { |indexable| indexable.full_path.include?("minitest-reporters") })
-      assert(indexables.none? { |indexable| indexable.full_path.include?("ansi") })
+      assert(indexables.any? { |indexable| indexable.full_path.include?("minitest-reporters") })
```
andyw8, line 17: tbc

PR #1202, opened 2023-11-22
Store owner for singleton methods

andyw8:
### Motivation

Step towards https://github.com/Shopify/ruby-lsp/issues/899

Co-authored with @vinistock 

### Implementation

Track singleton methods using an owner field on `Method` instances.

Note: We can't use the name `singleton_class` as it's already defined by Ruby, hence the ugly `singleton_klass`.

### Automated Tests

Updaetd

### Manual Tests

n/a

### DIFF
--- a/lib/ruby_indexer/lib/ruby_indexer/entry.rb
+++ b/lib/ruby_indexer/lib/ruby_indexer/entry.rb
@@ -87,22 +87,6 @@ module RubyIndexer
       end
     end
 
-    # class SingletonClass
-    #   extend T::Sig
-
-    #   sig { returns(Namespace) }
-    #   attr_reader :attached_class
-
-    #   sig { returns(String) }
-    #   attr_reader :name
-
-    #   sig { params(name: String, attached_class: Namespace).void }
-    #   def initialize(name, attached_class)
-    #     @name = T.let("Class:#{name}", String)
-    #     @attached_class = attached_class
-    #   end
-    # end
-
     class Constant < Entry
     end
 


lib/ruby_indexer/lib/ruby_indexer/entry.rb
```diff
@@ -40,6 +40,15 @@ class Namespace < Entry
 
       abstract!
 
+      sig { returns(SingletonClass) }
+      attr_reader :singleton_klass
+
+      sig { params(name: String, file_path: String, location: Prism::Location, comments: T::Array[String]).void }
+      def initialize(name, file_path, location, comments)
+        super
+        @singleton_klass = T.let(SingletonClass.new(name, self), SingletonClass)
```
Morriar, line 49:
```suggestion
        @singleton_class = T.let(SingletonClass.new(name, self), SingletonClass)
```

As the other one is `attached_class`.

lib/ruby_indexer/lib/ruby_indexer/entry.rb
```diff
@@ -69,6 +78,23 @@ class Class < Namespace
       def initialize(name, file_path, location, comments, parent_class)
         super(name, file_path, location, comments)
         @parent_class = T.let(parent_class, T.nilable(String))
+        @singleton_klass = T.let(SingletonClass.new(name, self), SingletonClass)
```
Morriar, line 81: Isn't it already done in the `Namespace` parent?

lib/ruby_indexer/lib/ruby_indexer/entry.rb
```diff
@@ -69,6 +78,23 @@ class Class < Namespace
...
+      end
+    end
+
+    class SingletonClass
```
Morriar, line 85: I'm not sure about this abstraction.

In Ruby a singleton class is a class like a real instance of `Class`:

```rb
Foo.class # => Class
Foo.class.ancestors # => [Class, Module, Object, PP::ObjectMixin, Kernel, BasicObject]

Foo.class.class # => Class
Foo.class.class.ancestors # => [Class, Module, Object, PP::ObjectMixin, Kernel, BasicObject]
```

Having `SingletonClass` as a standalone will make it harder to resolve the methods properly when we call things coming from `Class` for example `Foo.name`.

I wonder if `SingletonClass` could simply inherit from `Class` or if we actually need a `SingletonClass` at all? From what I see we only have this abstraction so we store `attached_class` but it doesn't seem to be used anywhere at the moment? Or could we make `Class` old a nilable `@attached_class` and implement `singleton?` depending on if the attribute holds a value or not?

lib/ruby_indexer/lib/ruby_indexer/entry.rb
vinistock: We can't call this `@singleton_class` or else its accessor would be called `singleton_class` and that conflicts with the Ruby internal method. It also throws Sorbet off.

lib/ruby_indexer/lib/ruby_indexer/entry.rb
andyw8: 👍 Now changed to only have `Class`.

lib/ruby_indexer/lib/ruby_indexer/entry.rb
```diff
@@ -40,6 +40,21 @@ class Namespace < Entry
...
 
+      sig { returns(Entry::Class) }
+      attr_reader :singleton_klass
+
+      sig { params(name: String, file_path: String, location: Prism::Location, comments: T::Array[String]).void }
+      def initialize(name, file_path, location, comments)
+        super
+
+        singleton = if name.start_with?("Class:")
+          T.cast(self, Entry::Class)
```
Morriar, line 51: Why do we need this recursive link?

lib/ruby_indexer/lib/ruby_indexer/entry.rb
andyw8: Yup, fixed.

lib/ruby_indexer/lib/ruby_indexer/entry.rb
vinistock: Because the singleton class of a singleton class is itself. And if we didn't have the check we would have an infinite loop.

Morriar: Yeah I think creating the singleton class only when we need it makes more sense

PR #1203, opened 2023-11-22
Disregard quotes in string node location for require path completion

vinistock:
### Motivation

Closes https://github.com/Shopify/vscode-ruby-lsp/issues/900

Finally managed to figure out what was wrong with our require path completion. Prism includes a string's quotes as part of their location, but we only want to replace the contents of the string.

Because of this off by one discrepancy, the range we were returning to the editor didn't fully match the replacement and VS Code refused to apply the completion.

### Implementation

We need to add one to the start character and remove one from the end character to account for the quotes.

### Automated Tests

Fixed the tests, which I mistakenly changed during the Prism migration.

### DIFF
--- a/lib/ruby_lsp/requests/completion.rb
+++ b/lib/ruby_lsp/requests/completion.rb
@@ -168,10 +168,16 @@ module RubyLsp
 
       sig { params(label: String, node: Prism::StringNode).returns(Interface::CompletionItem) }
       def build_completion(label, node)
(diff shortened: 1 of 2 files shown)


st0012 (approved):

vinistock:
@paracycle correctly pointed out we can use the `content_loc`, which then means our completion works properly even for less conventional require syntaxes like

```ruby
require %q{ruby_lsp/addon}
```

andyw8 (approved):

PR #1207, opened 2023-11-22
Temporarily disable CI for Windows with `x64-mingw-ucrt`

andyw8:
### Motivation

Recent changes in Bundler are causing the `x64-mingw-ucrt` platform to be removed, either when running running in Dependabot, or running `bundle update --bundler` locally. This breaks CI for Windows on Ruby 3.1 and newer.

Every Dependabot PR currently fails because of this, which is quite disruptive.

(I initially thought this was a Dependabot issue, so had opened https://github.com/dependabot/dependabot-core/issues/8444).

I've tried various things to solve it, but wasn't able to yet yet. Since this a recent Bundler change, which will likely impact others too, I'm inclined to hold off doing more in case the behaviour gets reverted or altered. We can revisit this in a week or two.

### Implementation

This PR disables CI for Windows where `x64-mingw-ucrt` is in use, and updates the `Gemfile.lock` as generated by `bundle update --bundler`.

Although disabling Windows does introduce some small risk, very little of Ruby LSP's behaviour is platform-dependent.

### Automated Tests

n/a

### Manual Tests

n/a

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -108,7 +108,6 @@ GEM
 
 PLATFORMS
   arm64-darwin
-  x64-mingw-ucrt
   x64-mingw32
   x86_64-darwin
   x86_64-linux
@@ -133,4 +132,4 @@ DEPENDENCIES
   tapioca (~> 0.11)
 
 BUNDLED WITH
-   2.4.10
+   2.4.22


vinistock (changes requested):
This is basically removing Windows entirely from CI. We can't do that or we risk getting into a replace where we push more Windows incompatible changes like we did in the past.

We should open an issue in rubygems asking if this is the new intended behaviour or a bug and ask for guidance on how to get CI passing.

andyw8:
As per Slack discussion, https://github.com/rubygems/rubygems/pull/7172 may solve this but until then we'll manually run and commit `bundle lock --add-platform x64-mingw-ucrt` for Dependabot PRs.

PR #1208, opened 2023-11-22
Support attributes

vinistock:
### Motivation

Step towards #899 

Add support for indexing attributes and using those in features. Definition and hover work without modification, there are only minor modifications in completion and workspace symbol.

### Implementation

- Started indexing attribute calls
- Started using the property type for attributes in workspace symbol
- Allowed completion to return attributes

### Automated Tests

Added tests.

### DIFF
--- a/lib/ruby_lsp/requests/completion.rb
+++ b/lib/ruby_lsp/requests/completion.rb
@@ -137,15 +137,11 @@ module RubyLsp
 
         receiver = T.must(receiver_entries.first)
 
-        candidates = T.cast(
-          @index.prefix_search(name),
-          T::Array[T::Array[RubyIndexer::Entry::Member]],
-        )
-        candidates.each do |entries|
-          entry = entries.find { |e| e.owner&.name == receiver.name }
+        @index.prefix_search(name).each do |entries|
+          entry = entries.find { |e| e.is_a?(RubyIndexer::Entry::Member) && e.owner&.name == receiver.name }
           next unless entry
 
-          @_response << build_method_completion(entry, node)
+          @_response << build_method_completion(T.cast(entry, RubyIndexer::Entry::Member), node)
         end
       end
 


lib/ruby_indexer/lib/ruby_indexer/entry.rb
```diff
@@ -93,6 +93,33 @@ def initialize(name:)
...
     end
 
+    class Attribute < Entry
+      extend T::Sig
+
+      sig { returns(T.nilable(Entry::Namespace)) }
+      attr_reader :owner
+
+      sig { returns(T::Array[Parameter]) }
+      attr_reader :parameters
```
Morriar, line 103: Do we really need to store this? I believe it's just used for completion right? Why not move this to a method and call it when needed?

lib/ruby_indexer/lib/ruby_indexer/entry.rb
```diff
@@ -93,6 +93,33 @@ def initialize(name:)
     class RequiredParameter < Parameter
     end
 
+    class Attribute < Entry
```
Morriar, line 96: I'm not sure about the name. It seems we're more interested in the accessor rather than the real attribute that is behind. Do we plan to store in the index the location of an attribute initialization like `@foo = "foo"`? In which case we would need to rename this class anyway.

We kinda see this naming problem when you set the symbol kind: `PROPERTY`, I would expect the property to be the accessors and the real attribute to be a `FIELD`.

What do you think of renaming it as `Accessor` or maybe `AttributeAccessor`?

lib/ruby_indexer/lib/ruby_indexer/collector.rb
```diff
@@ -141,8 +141,18 @@ def handle_constant_write_node(node)
 
     sig { params(node: Prism::CallNode).void }
     def handle_call_node(node)
-      message = node.message
-      handle_private_constant(node) if message == "private_constant"
+      message = node.name
```
Morriar, line 144: We may want to check that the receiver is `null` or explicit `self` because Ruby is fun:

```rb
class Foo; end

class Bar
   Foo.attr_reader :foo
end

Bar.foo # => undefined method `foo' for Bar:Class (NoMethodError)
```

lib/ruby_indexer/lib/ruby_indexer/entry.rb
vinistock: Yeah, good point. No need to waste memory on every attribute.

lib/ruby_indexer/lib/ruby_indexer/entry.rb
vinistock: That's fair. We will eventually want to support instance variables. I renamed to `Accessor`.

lib/ruby_lsp/requests/completion.rb
```diff
@@ -137,7 +137,10 @@ def on_call_node_enter(node)
 
         receiver = T.must(receiver_entries.first)
 
-        candidates = T.cast(@index.prefix_search(name), T::Array[T::Array[RubyIndexer::Entry::Method]])
+        candidates = T.cast(
+          @index.prefix_search(name),
+          T::Array[T.any(T::Array[RubyIndexer::Entry::Method], T::Array[RubyIndexer::Entry::Accessor])],
```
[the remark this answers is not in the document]
vinistock: I like that! Added a `Member` abstract class that represents both accessors and methods. Also simplified that type, which no longer requires a union.

PR #1223, opened 2023-11-28
Move ownership of message_queue from listeners to addon

### LINKED ISSUE #1205: Pass notification queue to addons upon activation
For addons that perform slow tasks, such as typechecking, we will want them to push diagnostic results asynchronously to prevent blocking the LSP from answering other requests.

We should pass our message queue on activation, so that the addon can keep a reference to it.

vinistock:
### Motivation

Closes #1205

We initially put the ownership of the message queue on listeners. That wasn't the best design for a few reasons:

1. It forces every listener to receive the message queue even if it won't use it
2. Addons don't have a non-hacky way of keeping a reference to the message queue if they need to send async responses back to the editor (which would be necessary for a typechecker or any slow computing of diagnostics)

If we invert the ownership, then the addon can decide when to use the message queue. It still allows addons to use is in listeners if desired, but gives enough flexibility for returning async responses.

We realized the need for these changes when hacking on a Steep addon.

### Implementation

The first commit has the actual change and the second one just fixes all declarations and call sites.

The idea is that we send the message queue during addon activation instead of passing it to all listeners. Then the addon can store the queue in an instance variable and reuse it.

### Automated Tests

Adjusted tests.

### DIFF
--- a/ADDONS.md
+++ b/ADDONS.md
@@ -63,8 +63,8 @@ module RubyLsp
       extend T::Sig
 
       # Performs any activation that needs to happen once when the language server is booted
-      sig { override.void }
-      def activate
+      sig { override.params(message_queue: Thread::Queue).void }
+      def activate(message_queue)
       end
 
       # Performs any cleanup when shutting down the server, like terminating a subprocess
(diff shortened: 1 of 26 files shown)


st0012:
> It still allows addons to use is in listeners if desired, but gives enough flexibility for returning async responses.

Maybe let's not expose message queue at all until it's actually needed? I don't see any addons using it now. 

I understand that it's going to be needed at some point. But before that happens, we may go through other design changes on APIs and exposing it now could make such changes harder without obvious benefits. WDYT?

st0012 (approved):

vinistock:
Chatted with Stan through Slack about how adding this will allow us to better experiment with typechecker addons, such as a Steep addon.

PR #1238, opened 2023-12-05
Allow hover addons to specify documentation groups

vinistock:
### Motivation

One of the limitations we discussed about addons is that they can't decide how their content is inserted into the hover response.

Instead of making an ordering system, which is probably more complex and harder to control, I suggest that we use content categories so that addons can define how their response is merged.

I'd appreciate feedback on the approach.

### Implementation

The idea is that hover addons now just return a `HoverResponse`, which is just the markdown string with a category associated to it.

We then use the category to merge the contents into a final hover response and return that to the editor. I believe this is a decent way of providing flexibility while at the same time maintaining consistency regardless of which addons are available.

### Automated Tests

Enhanced our current test.

### DIFF
--- a/lib/ruby_lsp/utils.rb
+++ b/lib/ruby_lsp/utils.rb
@@ -92,4 +92,36 @@ module RubyLsp
       @configuration[:enableAll] || @configuration[feature]
     end
   end
+
+  class HoverResponse
+    extend T::Sig
+
+    CATEGORIES = T.let([:link, :documentation, :signature], T::Array[Symbol])
+
+    sig { returns(String) }
+    attr_reader :markdown_content
+
+    sig { params(category: Symbol, markdown_content: String).void }
+    def initialize(category, markdown_content)
+      raise ArgumentError, "Invalid category: #{category}" unless CATEGORIES.include?(category)
+
+      @category = category
+      @markdown_content = markdown_content
+    end
+
+    sig { returns(T::Boolean) }
+    def link?
+      @category == :link
+    end
+
+    sig { returns(T::Boolean) }
+    def documentation?
+      @category == :documentation
+    end
+
+    sig { returns(T::Boolean) }
+    def signature?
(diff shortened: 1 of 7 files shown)


lib/ruby_lsp/utils.rb
```diff
@@ -92,4 +92,36 @@ def enabled?(feature)
...
+    extend T::Sig
+
+    CATEGORIES = T.let([:link, :documentation, :signature], T::Array[Symbol])
+
+    sig { returns(String) }
+    attr_reader :markdown_content
+
+    sig { params(category: Symbol, markdown_content: String).void }
+    def initialize(category, markdown_content)
+      raise ArgumentError, "Invalid category: #{category}" unless CATEGORIES.include?(category)
```
andyw8, line 106: Would it be less fragile if we allow any category name, but treat it as lowest precedence if not recognized? Then an addon can introduce a new category without breaking older versions of the extension.

vinistock: I feel like we'd end up in a similar situation though. If two addons want to create their own categories, how do we determine the order?

Also, if an addon wants to put something at the bottom part, they can always just use `documentation`. I feel like if they wanted a new category it's because they would like to be above `documentation`, but then how do we know where exactly?

vinistock: I would vote for keeping it so that we have a deterministic order for the hover content. If we come to the conclusion that we missed too many categories, we can revert this decision.

andyw8 (approved):

vinistock:
I'll hold off merging this for a bit. @st0012 had a great idea on how to better separate concerns and refactor our listeners/requests concepts.

If the idea goes smoothly and doesn't uncover any blockers, then I'll adapt this PR to follow the same structure.

vinistock:
This will be implemented in a different PR.

PR #1241, opened 2023-12-06
Add initial support for signature help

### LINKED ISSUE #200: Add signature help to the `ruby-lsp`
If #199 works and is successful in providing decent results for finding possible method definitions, we can use the same mechanism to provide signature help. This request shows the method signature for the invocation at the given cursor position, displaying arguments.

Note: there may be conflicts between possible methods. It's likely that we'll only be able to provide signature help if only a single result is found in the index.

Documentation: https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#textDocument_signatureHelp

vinistock:
### Motivation

Closes #200

We're currently inserting a method's parameters as part of completion, but that's not how the specification prescribes handling parameters.

Completion should only insert the method name and then signature help shows the parameter information while the user types.

This PR removes the parameters from completion and adds initial support for signature help.

### Implementation

The idea is that we discover which method the user is invoking and then display the information.

The main source of difficulty implementing this for Ruby are keyword and optional parameters, because it means that we can't trust the order of arguments to define the `active_parameter`.

For example

```ruby
def foo(a, b = 123, c:, d:)
end

foo(a, c:,) # => current active parameter is `d` because we skipped `b`
foo(a, d:) # => current parameter is `c` because keyword parameters don't require an order
```

Currently, we are simply not handling this complexity so that we can start simple. We'll need to design some mechanism to understand what the current active parameter is.

### Automated Tests

Added tests - including the cases where we are not handling parameters correctly yet.


andyw8:
Note: You need to add `"signatureHelp": true` to the enabledFeatures list in `settings.json` to enable.

vinistock:
I hadn't caught those! I'll see how we can improve these cases.

> Typing the arguments quickly can result in the highlighted argument being wrong.

This one may not be easy to fix. It might just be the delay from the LSP processing multiple requests for each keypress. Does it just lag?

lib/ruby_lsp/requests/completion.rb
```diff
@@ -125,7 +125,7 @@ def on_constant_path_node_enter(node)
 
       sig { params(node: Prism::CallNode).void }
       def on_call_node_enter(node)
-        return if DependencyDetector.instance.typechecker
+        # return if DependencyDetector.instance.typechecker
```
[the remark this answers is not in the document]
vinistock: Yes, oops.

lib/ruby_indexer/lib/ruby_indexer/entry.rb
```diff
@@ -84,6 +84,8 @@ class Parameter
       sig { returns(Symbol) }
       attr_reader :name
 
+      alias_method :decorated_name, :name
```
Morriar, line 87: We should add a comment explaining the difference with `name` and when one should be used or the other.

lib/ruby_lsp/executor.rb
```diff
@@ -692,6 +727,12 @@ def initialize_request(options)
         )
       end
 
+      signature_help_provider = if enabled_features["signatureHelp"]
+        Interface::SignatureHelpOptions.new(
+          trigger_characters: ["(", " ", ","],
```
Morriar, line 732: Should we also trigger on `:` like if I'm typing `bar(a:` wouldn't the help disappear?

lib/ruby_indexer/lib/ruby_indexer/entry.rb
vinistock: Added comments to both.

lib/ruby_lsp/executor.rb
vinistock: It's already triggered automatically. It seems editors consider `:` as part of identifiers.

lib/ruby_indexer/lib/ruby_indexer/entry.rb
Morriar: Should the keyword parameter `foo` be decorate as `foo:`?

lib/ruby_lsp/executor.rb
Morriar: Should we add a comment about this for future us?

Morriar (approved):

lib/ruby_indexer/lib/ruby_indexer/entry.rb
vinistock: Ah, yeah. Good catch.

PR #1242, opened 2023-12-10
Handle `#to_spec #=> nil` in `RubyIndexer::Configuration`

coffeejunk:
### Motivation

`ruby-lsp` would crash on boot if a prerelease gem is in the `Gemfile`.

Screenshot 2023-12-10 at 4 34 39 PM

### Implementation

Add `&` to handle `nil` values.

### Automated Tests

-

### Manual Tests

I modified the installed gem and `ruby-lsp` boots without issue.

### DIFF
--- a/lib/ruby_indexer/lib/ruby_indexer/configuration.rb
+++ b/lib/ruby_indexer/lib/ruby_indexer/configuration.rb
@@ -193,7 +193,7 @@ module RubyIndexer
       # When working on a gem, we need to make sure that its gemspec dependencies can't be excluded. This is necessary
       # because Bundler doesn't assign groups to gemspec dependencies
       this_gem = Bundler.definition.dependencies.find do |d|
-        d.to_spec.full_gem_path == Dir.pwd
+        d.to_spec&.full_gem_path == Dir.pwd
       rescue Gem::MissingSpecError
         false
       end


coffeejunk:
I have signed the CLA!

andyw8:
🤔 I think we'd still want to index gems that are prereleases.

Which particular gem was causing the issue?

vinistock:
@coffeejunk what version of bundler is your project locked to? I'm surprised that it doesn't return a gem path for prerelease entries. Have you tried upgrading Bundler?

I agree with Andy, we should try to discover if there's a way of ensuring those are indexed too.

If you'd like a temporary workaround, you can exclude that gem from indexing by creating a `.index.yml` file in the project root with

```ruby
excluded_gems:
  - "select2-rails"
```

coffeejunk:
Hm, yeah that makes sense.

```
 cat Gemfile.lock | tail -2
BUNDLED WITH
   2.2.22
```

and 

```
```

andyw8:
That's quite old (July 2021), can you try update it? (`bundle update --bundler`)

andyw8 (approved):
On further checking, this relates to an [known existing problem](https://github.com/Shopify/ruby-lsp/pull/1159).

Let's proceed with this an investigate separately in https://github.com/Shopify/ruby-lsp/issues/1246.

PR #1247, opened 2023-12-12
Remove request logic from CodeLens listener

st0012:


lib/ruby_lsp/requests/request.rb
```diff
@@ -0,0 +1,36 @@
...
+  module Requests
+    # :nodoc:
+    class Request
+      extend T::Sig
+      extend T::Generic
+
+      abstract!
+
+      sig { abstract.returns(T.anything) }
+      def response; end
```
vinistock, line 14: What do you think about maintaining the requests' main method as `run`? That will reduce the amount of changes, but I also think it's a better name then `response` in this case.

lib/ruby_lsp/requests/request.rb
```diff
@@ -0,0 +1,36 @@
...
+
+      ResponseType = type_member
+
+      abstract!
+
+      sig { params(listeners: T::Array[Listener[ResponseType]]).void }
+      def initialize(listeners)
+        super()
+        @listeners = listeners
+      end
```
vinistock, line 30: Instead of initializing a request with a list of listeners, could we instantiate the listeners inside the request itself? Is there any value I'm missing with making the listeners an initialization parameter?

This is what I'm thinking:

```ruby
module Requests
  class CodeLens < Request
    def initialize(...)
      @listeners = instantiate_all_listeners
    end

    def run
      dispatcher = Prism::Dispatcher.new
      # ...
      response
    end
  end
end
```

In this case, we wouldn't even need the `ListenerBasedRequest` class.

lib/ruby_lsp/requests/request.rb
st0012: I extracted the standardisation changes to #1281. But I prefer `response` over `run`.

Morriar (commented):
Once all requests are proper descendants of `Request` you should be able to enforce overrides on `provider`:

```rb
class Request
  extend T::Sig
  extend T::Helpers

  abstract!

  class << self
    extend T::Sig

    sig { abstract.returns(T.attached_class) }
    def provider; end
  end
end

class RequestA < Request
  class << self
    sig { override.returns(T.attached_class) }
    def provider
      new
    end
  end
end

class RequestB < Request # Error: Missing definition for abstract method `Request.provider`
end
```

[[sorbet.run]](https://sorbet.run/#%23%20typed%3A%20true%0A%0Aclass%20Request%0A%20%20extend%20T%3A%3ASig%0A%20%20extend%20T%3A%3AHelpers%0A%0A%20%20abstract!%0A%0A%20%20sig%20%7B%20abstract.returns%28T.anything%29%20%7D%0A%20%20def%20response%3B%20end%0A%0A%20%20class%20%3C%3C%20self%0A%20%20%20%20extend%20T%3A%3ASig%0A%0A%20%20%20%20sig%20%7B%20abstract.returns%28T.attached_class%29%20%7D%0A%20%20%20%20def%20provider%3B%20end%0A%20%20end%0Aend%0A%0Aclass%20RequestA%20%3C%20Request%0A%20%20sig%20%7B%20override.returns%28String%29%20%7D%0A%20%20def%20response%0A%20%20%20%20%22foo%22%0A%20%20end%0A%0A%20%20class%20%3C%3C%20self%0A%20%20%20%20sig%20%7B%20override.returns%28T.attached_class%29%20%7D%0A%20%20%20%20def%20provider%0A%20%20%20%20%20%20new%0A%20%20%20%20end%0A%20%20end%0Aend%0A%0Aclass%20RequestB%20%3C%20Request%0A%20%20sig%20%7B%20override.returns%28Integer%29%20%7D%0A%20%20def%20response%0A%20%20%20%2042%0A%20%20end%0Aend%0A%0Ap1%20%3D%20RequestA.provider%0AT.reveal_type%28p1%29%0A%0Ap2%20%3D%20RequestB.provider%0AT.reveal_type%28p2%29)

st0012:
@Morriar But provider is not to return the request itself. It's for options like `Interface::CodeLensOptions.new(resolve_provider: false)` or just a hash. Also, not every request needs it. For example, `definition` and `selection_range` don't need to define it.

lib/ruby_lsp/listeners/code_lens.rb
```diff
@@ -0,0 +1,241 @@
...
+      BASE_COMMAND = T.let(
+        begin
+          Bundler.with_original_env { Bundler.default_lockfile }
+          "bundle exec ruby"
+        rescue Bundler::GemfileNotFound
+          "ruby"
+        end + " -Itest ",
+        String,
+      )
+      GEMFILE_NAME = T.let(
```
vinistock, line 22: This was moved into utils, so we can get rid of this definition.

lib/ruby_lsp/requests/code_lens.rb
```diff
@@ -68,209 +47,25 @@ def provider
...
-          command += " --testcase " + "/#{Shellwords.escape(class_name)}/"
-
-          if method_name
-            command += " --name " + Shellwords.escape(method_name)
-          end
+      sig { override.returns(ResponseType) }
+      def response
+        result = []
+        @listeners.each do |listener|
+          result.concat(listener.response) if listener.response
```
vinistock, line 66: Do we need the nilable check here? We default to returning empty arrays for code lens listeners.

lib/ruby_lsp/requests/hover.rb
```diff
@@ -28,119 +30,61 @@ def provider
...
-      def generate_hover(name, location)
-        entries = @index.resolve(name, @nesting)
-        return unless entries
+        return unless first_response
 
-        # We should only show hover for private constants if the constant is defined in the same namespace as the
-        # reference
-        first_entry = T.must(entries.first)
-        return if first_entry.visibility == :private && first_entry.name != "#{@nesting.join("::")}::#{name}"
+        # TODO: other_responses should never be nil. Check Sorbet
```
vinistock, line 82: Should this comment be removed? I think the `T.must` below addresses it, right?

test/requests/document_symbol_expectations_test.rb
```diff
@@ -24,6 +24,28 @@ def test_document_symbol_addons
     end
   end
 
+  def run_expectations(source)
```
vinistock, line 27: It looks like the document symbol request is still instantiated the same way it was before. Why is overriding `run_expectations` necessary?

lib/ruby_lsp/requests/request.rb
vinistock: I still find it odd to name the request's main method as `response`. It hints that requests do not control their own flow of execution, like listeners.

I really think `run` or `perform` would be a better fit. @andyw8 do you have thoughts on the naming?

vinistock: Yeah, yeah. Let's do it in the next PR, no need to do it in this one.

My opinion is that having `response` for listeners and `run/perform` for requests makes the naming a bit clearer. It also avoids confusion since many listeners and requests are named the same.

Please, do move forward with this PR and we can change the name of the request methods in the next.

lib/ruby_lsp/requests/hover.rb
st0012: I think the comment means that we shouldn't need to wrap `T.must` around `other_responses` as it's a splat assignment:

```rb
        first_response, *other_responses = responses
```

It should always be an array regardless of `responses`'s value:

```
ruby-lsp(main):008> x, *y = a; y
=> []
ruby-lsp(main):009> a = nil
=> nil
ruby-lsp(main):010> x, *y = a; y
=> []
ruby-lsp(main):011> 
```

lib/ruby_lsp/requests/completion.rb
```diff
@@ -43,265 +45,65 @@ def provider
...
-          full_name = prefix.empty? ? entry_name : "#{prefix}::#{entry_name}"
-          next if full_name == entry_name
-
-          return true if @index[full_name]
-        end
-
-        false
+      sig { override.returns(ResponseType) }
+      def response
+        @dispatcher.dispatch_once(@target)
```
vinistock, line 105: This line is failing because `@dispatcher` is `nil` on certain conditions due to the early returns in `initialize`.

I think we should move most of the code in the `initialize` method to `response` and ensure that the key instance variables are always set.

vinistock (commented):
Caught one bug while tophatting

lib/ruby_lsp/requests/hover.rb
st0012: Yes there's an 4-year old issue for it: https://github.com/sorbet/sorbet/issues/2020
To be fair, it was fixed but then the fix was reverted.

lib/ruby_lsp/requests/completion.rb
st0012: I prefer not moving most of them into `response` as it means we'd need to initialise listeners there too, which doesn't align with other requests.
I've rearranged the ordering in `initialize` a bit instead so all the ivars used in `response` should always be defined.

vinistock (approved):

lib/ruby_lsp/requests/request.rb
st0012: I've opened https://github.com/Shopify/ruby-lsp/pull/1297 for renaming.

PR #1249, opened 2023-12-13
Enable Minitest/AssertPredicate cop

andyw8:
See https://github.com/Shopify/ruby-lsp/pull/1245#discussion_r1424702695

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -18,6 +18,9 @@ AllCops:
     - "test/fixtures/**/*"
     - "test/expectations/**/*"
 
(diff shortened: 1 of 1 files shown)


andyw8:
It's from `rubocop-minitest` rather than `rubocop-shopify`, so we don't have a standard config for that.

andyw8:
It's from `rubocop-minitest` rather than `rubocop`, so we don't have a standard config for that.

andyw8:
But there is some exploration for other plugin configs: https://github.com/Shopify/ruby-style-guide/pull/498

PR #1254, opened 2023-12-15
Remove aliased shapes

vinistock:
### Motivation

Removed our type aliases with shapes as they present considerable performance overhead.

### DIFF
--- a/lib/ruby_lsp/document.rb
+++ b/lib/ruby_lsp/document.rb
@@ -8,10 +8,6 @@ module RubyLsp
 
     abstract!
 
-    PositionShape = T.type_alias { { line: Integer, character: Integer } }
-    RangeShape = T.type_alias { { start: PositionShape, end: PositionShape } }
-    EditShape = T.type_alias { { range: RangeShape, text: String } }
-
     sig { returns(Prism::ParseResult) }
     attr_reader :parse_result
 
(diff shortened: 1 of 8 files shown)


lib/ruby_lsp/document.rb
```diff
@@ -80,7 +76,7 @@ def cache_get(request_name)
       @cache[request_name]
     end
 
-    sig { params(edits: T::Array[EditShape], version: Integer).void }
+    sig { params(edits: T::Array[T::Hash[Symbol, T.untyped]], version: Integer).void }
```
Morriar, line 79: I wonder if we could create a real `Edit` class instead?

vinistock: We receive these parameters from the `language-server_protocol` gem as hashes, so that would be a much more involved refactor. We'd need to start instantiating specialized objects for each type of request.

To be honest, I think it's a worth while contribution for the gem since passing hashes around is not very type friendly. But it'll take considerably more effort than this

andyw8 (approved):

lib/ruby_lsp/document.rb
paracycle: Yes, I've been meaning to do this refactor for the gem, actually. Maybe I can get this done this week.

Issue #687, opened 2023-05-03
Add CI against JRuby head


vinistock:
We talked to the JRuby folks and they fixed the mismatch in their Ripper implementation when parsing optional arguments. We should introduce `jruby-head` to our CI to make sure everything else is passing.

Notes:
- We don't need to run benchmarks on JRuby as well, only CRuby
- We need to avoid requiring `debug` on JRuby since it is CRuby specific

PR #1257, opened 2023-12-17
[Draft] Add jruby to CI

tachyons:
Add JRuby and JRuby CI to expand the platform coverage.

### Manual Tests

Fixes https://github.com/Shopify/ruby-lsp/issues/687

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -35,7 +35,7 @@ jobs:
       fail-fast: false
       matrix:
         os: [ubuntu-latest, macos-latest, windows-latest]
-        ruby: ["3.0", "3.1", "3.2", "head"]
+        ruby: ["3.0", "3.1", "3.2", "head", "jruby-head"]
         include:
           - ruby: "head"
             experimental: true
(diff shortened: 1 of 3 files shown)


.github/workflows/ci.yml
```diff
@@ -35,7 +35,7 @@ jobs:
       fail-fast: false
       matrix:
         os: [ubuntu-latest, macos-latest, windows-latest]
-        ruby: ["3.0", "3.1", "3.2", "head"]
+        ruby: ["3.0", "3.1", "3.2", "head", "jruby", "jruby-head"]
```
vinistock, line 38: I think just `jruby-head` is probably fine

vinistock (commented):
Looks like CI is having platform issues. Not sure how you solve them exactly, is there a generic java platform that's not tied to a specific version?

.github/workflows/ci.yml
tachyons: @vinistock I tried that too, But it was still failing with cryptic sorbet error. I haven't worked with sorbet before to debug that.

tachyons:
Screenshot 2023-12-19 at 7 40 08 AM
This is the error I am getting, Same issue is reported by another user [here](https://github.com/Shopify/vscode-ruby-lsp/issues/920)

andyw8:
Thank you, it seems this related to the [patch](https://github.com/Shopify/ruby-lsp/blob/main/lib/core_ext/uri.rb) we have for `URI::Generic.from_path`.

As Sorbet [does not officially support](https://github.com/sorbet/sorbet/blob/master/docs/JRuby.md) JRuby, this may be a difficult one to solve...

PR #1261, opened 2023-12-18
Move provider definition to individual requests

st0012:
### Motivation

While working on #1247, I found that we try to access some request's constants directly from the executor for building the providers. I think this is a leaky abstraction and requests should simply have a `provider` method to return the associated options or provider.

### Implementation

1. Define a `Request.provider` method on request classes that need a provider
2. Simply call the method from the executor


andyw8 (approved):
I didn't 🎩 but I like the improvement.

vinistock (approved):
Nice

PR #1264, opened 2023-12-20
Start keeping a stack of namespaces instead of namespace names

paracycle:
### Motivation

The current method collection was assigning the wrong owner for methods defined after a namespace was entered and then exited, since the `@current_owner` was not being "popped" at all.

For example:
```ruby
module Foo
  module Bar
  end

  def foo
  end
end
```
would end up assigning `Foo::Bar` as the owner of the method `foo`, when it should actually be `Foo`.

### Implementation

This allows us to use the same stack for both the constant and method collection. For constants, keeping the current namespace on the stack allows us to refer to the current namespace name instantly, without having to build it from the stack entries. For methods, we need to keep the current namespace on the stack so that we can properly find the owner, i.e. the namespace that the method is defined in.

### Automated Tests

Added a few automated tests to test the initial symptom, plus a few edge cases.

### DIFF
--- a/lib/ruby_indexer/lib/ruby_indexer/index.rb
+++ b/lib/ruby_indexer/lib/ruby_indexer/index.rb
@@ -255,7 +255,7 @@ module RubyIndexer
     # that doesn't exist, then we return the same UnresolvedAlias
     sig { params(entry: Entry::UnresolvedAlias).returns(T.any(Entry::Alias, Entry::UnresolvedAlias)) }
     def resolve_alias(entry)
-      target = resolve(entry.target, entry.nesting)
+      target = resolve(entry.target, Array(entry.namespace&.name))
       return entry unless target
 
       target_name = T.must(target.first).name
--- a/lib/ruby_indexer/test/constant_test.rb
+++ b/lib/ruby_indexer/test/constant_test.rb
@@ -198,7 +198,7 @@ module RubyIndexer
 
       unresolve_entry = @index["A::FIRST"].first
       assert_instance_of(Entry::UnresolvedAlias, unresolve_entry)
-      assert_equal(["A"], unresolve_entry.nesting)
+      assert_equal("A", unresolve_entry.namespace.name)
       assert_equal("B::C", unresolve_entry.target)
 
       resolved_entry = @index.resolve("A::FIRST", []).first
@@ -224,7 +224,7 @@ module RubyIndexer
 
       unresolve_entry = @index["A::ALIAS"].first
       assert_instance_of(Entry::UnresolvedAlias, unresolve_entry)
-      assert_equal(["A"], unresolve_entry.nesting)
+      assert_equal("A", unresolve_entry.namespace.name)
       assert_equal("B", unresolve_entry.target)
 
       resolved_entry = @index.resolve("ALIAS", ["A"]).first
@@ -237,7 +237,7 @@ module RubyIndexer
 
       unresolve_entry = @index["Other::ONE_MORE"].first
       assert_instance_of(Entry::UnresolvedAlias, unresolve_entry)
-      assert_equal(["Other"], unresolve_entry.nesting)
+      assert_equal("Other", unresolve_entry.namespace.name)
       assert_equal("A::ALIAS", unresolve_entry.target)
 
       resolved_entry = @index.resolve("Other::ONE_MORE::C", []).first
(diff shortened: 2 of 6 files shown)


lib/ruby_indexer/test/constant_test.rb
```diff
@@ -305,6 +307,25 @@ module A
...
+      assert_instance_of(Entry::UnresolvedAlias, unresolve_entry)
+      assert_nil(unresolve_entry.namespace)
+      assert_equal("O::P", unresolve_entry.target)
+
+      resolved_entry = @index.resolve("N", []).first
+      assert_instance_of(Entry::Alias, resolved_entry)
+      assert_equal("O::P", resolved_entry.target)
+
+      # When we are resolving A::I::J, we invoke `resolve("K::L", ["A"])`, which recursively resolves A::K::L too.
+      # Therefore, both A::I::J and A::K::L point to A::M by the end of the previous resolve invocation
```
st0012, line 322: I think these comments are mistakenly copied here?

lib/ruby_indexer/test/constant_test.rb
```diff
@@ -305,6 +307,25 @@ module A
...
+      resolved_entry = @index["O::P"].first
+      assert_instance_of(Entry::Alias, resolved_entry)
+      assert_equal("Q::R", resolved_entry.target)
+
+      constant = @index["Q::R"].first
+      assert_instance_of(Entry::Constant, constant)
```
st0012, line 328: Can we have a new test case for these new assertions? It took me a while to realise they actually don't rely on other parts of the fixture.

st0012 (commented):
Left a few comments on tests. Otherwise the fix looks good 👍

lib/ruby_indexer/lib/ruby_indexer/index.rb
```diff
@@ -255,7 +255,7 @@ def resolve_method(method_name, receiver_name)
     # that doesn't exist, then we return the same UnresolvedAlias
     sig { params(entry: Entry::UnresolvedAlias).returns(T.any(Entry::Alias, Entry::UnresolvedAlias)) }
     def resolve_alias(entry)
-      target = resolve(entry.target, entry.nesting)
+      target = resolve(entry.target, Array(entry.namespace&.name))
```
vinistock, line 258: While this might work, the idea behind the second argument is for it to be an array containing each part of the name.

I.e.:
```ruby
"Foo::Bar::Baz"
# Should be
["Foo", "Bar", "Baz"]
```

What do you think about adding a method named `nesting` in `UnresolvedAlias` that just splits the name?

```ruby
class UnresolvedAlias
  def nesting
    @namespace.name.split("::")
  end
end
```

vinistock (commented):
This is great ❤️

lib/ruby_indexer/lib/ruby_indexer/index.rb
paracycle: Sure, I can make that change but I think this part of the indexer is fundamentally broken, and needs a refactor anyway. The problem is that Ruby nesting is not as simple as looking at each part of the namespace, since implicit namespaces work differently:
```ruby
module Foo
  A = 1
end

module Foo::Bar
  puts A
  # => uninitialized constant Foo::Bar::A (NameError)
end
```

For more reference see: https://www.writesoftwarewell.com/nested-modules-in-ruby/

Thus, what we really need to do is to store the "nesting" as Ruby would have and resolve against that nesting stack.

lib/ruby_indexer/test/constant_test.rb
paracycle: I don't understand how these are not related. This test case is testing all kinds of ways constant aliases could be defined, but it was never checking for top level constant alias definitions, which were broken.

My changes just add top-level constant alias definition examples to this method as well.

lib/ruby_indexer/lib/ruby_indexer/index.rb
vinistock: I think that's already what happens in main with the naming stack. Unless I'm missing something, I don't think it's broken.

We don't split namespaces that are defined as `module Foo::Bar`, so I believe the resolution is correct. Although we should add a test for that specific example.

The changes in this PR might require some extra refactors to make it fully work. For example, if we found an alias like this

```ruby
module First
  class Foo::Bar
    ALIAS_CONST = SOME_OTHER_THING
  end
end
```

We correctly store the nesting as `["First", "Foo::Bar"]`. When resolving `SOME_OTHER_THING`, we'd then search it in `First::Foo::Bar`, then `First` and finally on the top level.

And the features that use `resolve` do send the right nesting. For example, definition would invoke it like `@index.resolve("SOME_OTHER_THING", ["First", "Foo::Bar"])` - since we also don't split those names when locating target nodes for features.

However, with these changes we no longer have a reference to the nesting where things are found for unresolved symbols. We only have the entries with their fully qualified names in the stack, which as you pointed out is not enough to reconstruct the nesting where they were originally found.

To make the change in this PR and not break unresolved aliases, I think we need to change the entries to store their nesting instead of the fully qualified name, so that we can work back the nesting for aliases. Either that or maintain the original naming stack, but that seems more wasteful.

paracycle: Right, that makes sense. That also means that we are missing a lot of test cases that should have broken because of my changes in this PR, exercising specifically those kinds of nesting cases.

PR #1267, opened 2023-12-22
Choose other test framework when ruby-lsp-rails not present

tristil:
### Motivation

ruby-lsp chooses 'rails' framework when rails is present in the gem dependencies list, so that ruby-lsp-rails can add its own testing commands. But this can be confusing if somebody installs ruby-lsp without also installing ruby-lsp-rails on a rails project, because the test commands will not show up at all. Ideally a dialogue in the relevant extension would prompt the user to install ruby-lsp-rails, but ruby-lsp should select the available testing framework.

### Implementation

If ruby-lsp-rails is not present look for minitest or test-unit instead.

### Automated Tests

I updated the tests to handle the new case. There is one failing test that was preexisting.

### Manual Tests

I tested this inside Vscode

### DIFF
--- a/lib/ruby_lsp/requests/support/dependency_detector.rb
+++ b/lib/ruby_lsp/requests/support/dependency_detector.rb
@@ -40,7 +40,7 @@ module RubyLsp
     def detect_test_library
       # A Rails app may have a dependency on minitest, but we would instead want to use the Rails test runner provided
       # by ruby-lsp-rails.
-      if direct_dependency?(/^rails$/)
+      if direct_dependency?(/^rails$/) && direct_dependency?(/^ruby-lsp-rails$/)
         "rails"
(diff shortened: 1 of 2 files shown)


tristil:
I have signed the CLA!

andyw8:
Hi @tristil,

Am I correct in thinking that your Rails app uses "minitest style" tests?, e.g.:

```ruby
def test_foo_is_true
  # ...
end
```

rather than ActiveSupport declarative tests, e.g.:

```ruby
test "foo is true" do
 # ...
end
```

tristil:
@andyw8 For the most part, yes

andyw8:
This should shortly be made unnecessary by https://github.com/Shopify/ruby-lsp/pull/1381.

PR #1274, opened 2023-12-30
Support running tests in Rails engines 

tristil:
### Motivation

In Rails projects with engines, you need to cd into the engine in order to run a test, or else requires, etc. will not work.

### Implementation

This change detects when a file has `engines` in its path and has a Rails dependency and alters the command to cd in (and out, afterwards) of the engine. This allows running a test and sending the test to terminal. The special characters involved break debug, which is fixed in the second commit. The second commit escapes the command when passing to `rdbg`.

### Automated Tests

Tests added

### Manual Tests

I am able to run tests in an engine using VS Code

### DIFF
--- a/lib/ruby_lsp/listeners/code_lens.rb
+++ b/lib/ruby_lsp/listeners/code_lens.rb
@@ -198,7 +198,14 @@ module RubyLsp
 
       sig { params(class_name: String, method_name: T.nilable(String)).returns(String) }
       def generate_test_command(class_name:, method_name: nil)
-        command = BASE_COMMAND + T.must(@path)
+        cd_command =
+          if DependencyDetector.instance.direct_dependency?(/^rails$/) && T.must(@path).include?("/engines/")
+            engine_name = T.must(T.must(@path).split("/engines/").last).split("/").first
+            "cd engines/#{engine_name} && "
+          else
+            ""
+          end
+        command = cd_command + BASE_COMMAND + T.must(@path)
 
         case DependencyDetector.instance.detected_test_library
         when "minitest"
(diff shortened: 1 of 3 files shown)


st0012:
@tristil have you tried installing the [`ruby-lsp-rails`](https://github.com/Shopify/ruby-lsp-rails) addon? With it, the code lens feature would use `bin/rails test` to execute the tests instead of `ruby -Itest`, which may take the engine setup into consideration.

tristil:
@st0012 I tried it out and it didn't work. When running engine tests manually we always cd into the engine directory.

st0012:
Can you explain how it didn't work? Was the `bin/rails test` command didn't work, or the path was not correct, or was it something else?

tristil:
@st0012 My colleague suggested that instead of looking for `engines` directory we should look for presence of `.gemspec` IOW, instead of looking at this as an engine-specific fix it is really a way to cd into a subproject like inline gems.

tristil:
Some context is that we are a big company with a Rails monolith that is divided into 40+ engines. So although it's true that we could open vscode, etc. in each engine it would be a lot of context switching especially since there are interactions between engines and the top level app.

tristil:
@vinistock Sorry, I should have left a comment about why I was closing. Basically I think this is too specific to my organization's needs, plus we are moving away from engines to Packwerk packs. But: the underlying problem is that `ruby-lsp` wants there to be a Gemfile at the root of a project, even if there is also a `.gemspec` file, and in general it doesn't understand how to load an engine in isolation (or at least I couldn't figure it out). If it did know how to do these things then we could have set up our app as a multi root project as recommended and we wouldn't need special handling. IOW if it were possible for ruby-lsp to work with a single engine (or a single inline gem without an explicit Gemfile) then multi root project would have served our needs.

PR #1276, opened 2024-01-02
Allow bypassing typechecker in dependency detector

vinistock:
### Motivation

It's becoming difficult to dogfood our own feature implementations since we delegate a lot to Sorbet. Let's allow the server to pretend there's no typechecker, so that we can enable that for development and use all of our own features.

### Implementation

Just added an environment variable.


lib/ruby_lsp/requests/support/dependency_detector.rb
```diff
@@ -61,7 +61,8 @@ def direct_dependency?(gem_pattern)
 
     sig { returns(T::Boolean) }
     def detect_typechecker
-      direct_dependency?(/^sorbet/) || direct_dependency?(/^sorbet-static-and-runtime/)
+      !ENV["RUBY_LSP_BYPASS_TYPECHECKER"] &&
+        (direct_dependency?(/^sorbet/) || direct_dependency?(/^sorbet-static-and-runtime/))
```
andyw8, line 65:
```suggestion
      return false if ENV["RUBY_LSP_BYPASS_TYPECHECKER"]
      
      direct_dependency?(/^sorbet/) || direct_dependency?(/^sorbet-static-and-runtime/)
```

andyw8 (approved):

PR #1280, opened 2024-01-05
Add support for hover and go to definition when typechecker enabled but typed: false

jscharf:
### Motivation

This is a follow up to https://github.com/Shopify/ruby-lsp/pull/1245 by @SeanKG and I. 

We did another pair programming session today to continue that work to add hover and go to definition support for when the type checker is enabled but the file is `typed: false`. 

A lot of our files are currently `typed: false` but we still wanted to get the editor benefits of the Ruby LSP despite that.

### Implementation

The implementation was very similar to https://github.com/Shopify/ruby-lsp/pull/1245. We added an extra parameter to the `initialize` method in both `hover.rb` and `definition.rb` to store whether the type checker is enabled or not, and then we replaced existing instances of `DependencyDetector.instance.typechecker` with that newly provided instance variable `@typechecker_enabled`.

There is also a small refactor to `common.rb` included in these changes so that we could modify the code we needed to modify. We pulled out the call to `DependencyDetector.instance.typechecker` because it didn't relate to whether the caller was defined in a gem or not. So now the `defined_in_gem?` method looks like this
```ruby
def defined_in_gem?(file_path)
  BUNDLE_PATH &&
    !file_path.start_with?(T.must(BUNDLE_PATH)) &&
    !file_path.start_with?(RbConfig::CONFIG["rubylibdir"])
end
```

### Automated Tests

We didn't add any new automated tests as part of this change. However we did update existing ones. 

The reasoning behind that was we thought the change in functionality was small enough this time that updating the existing tests would be ok. If this is not ok definitely let us know.

### Manual Tests

Note as of 2:00pm on January 5th, 2024 this has not been manually tested yet. I will update when it has been.

### DIFF
--- a/lib/ruby_lsp/executor.rb
+++ b/lib/ruby_lsp/executor.rb
@@ -301,7 +301,7 @@ module RubyLsp
       target = parent if target.is_a?(Prism::ConstantReadNode) && parent.is_a?(Prism::ConstantPathNode)
 
       dispatcher = Prism::Dispatcher.new
-      base_listener = Requests::Definition.new(uri, nesting, @index, dispatcher)
+      base_listener = Requests::Definition.new(uri, nesting, @index, dispatcher, document.typechecker_enabled?)
       dispatcher.dispatch_once(target)
       base_listener.response
     end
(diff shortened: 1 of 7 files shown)


jscharf:
I have signed the CLA!

vinistock (approved):
Awesome! Thanks for the contribution

PR #1281, opened 2024-01-05
Add Request class to standardise request interface

st0012:
### Motivation

Instead of having some request classes implementing `run` method as interface, and other listener-based classes implementing `response` instead, we should standardise the interface to just `response`.

Since I made a similar change in #1247 as well, merging this first will also make it easier to review.

### Implementation

We can enforce the interface by making all request classes to inherit an abstract `Request` class.

Note: after doing a major refactor in #1247, the only commonality shared between all request classes is that they all implement a method to return the result. I don't think there's more that we could put into the `Request` class atm.

### Automated Tests

All existing tests should pass.

### DIFF
--- a/lib/ruby_lsp/check_docs.rb
+++ b/lib/ruby_lsp/check_docs.rb
@@ -56,7 +56,7 @@ module RubyLsp
         klass_name = klass.name
         next unless klass_name
 
-        klass if klass_name.match?(/RubyLsp::Requests::\w[^:]+$/) && !klass_name.include?("Support")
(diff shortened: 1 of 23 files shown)


egiurleo (approved):
Really cool how the architecture and patterns in this codebase are consolidating over time 😀

st0012:
> Which at first glance makes me think like the initialize method is the thing running the request.

@vinistock It make sense. But if we change it before completely splitting Listener and Request, it'd be a big change as we'd need rename `Listener#response` too, which would affect non-request listeners as well.

PR #1288, opened 2024-01-08
Rename defined_in_gem? to not_in_dependencies?

vinistock:
### Motivation

I noticed that this method's name was actually mentioning the opposite of its implementation. The method checks if a file path is not in dependencies.

### DIFF
--- a/lib/ruby_lsp/requests/support/common.rb
+++ b/lib/ruby_lsp/requests/support/common.rb
@@ -70,7 +70,7 @@ module RubyLsp
         end
 
         sig { params(file_path: String).returns(T.nilable(T::Boolean)) }
-        def defined_in_gem?(file_path)
+        def not_in_dependencies?(file_path)
           BUNDLE_PATH &&
             !file_path.start_with?(T.must(BUNDLE_PATH)) &&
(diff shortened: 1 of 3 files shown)


lib/ruby_lsp/requests/support/common.rb
```diff
@@ -70,7 +70,7 @@ def create_code_lens(node, title:, command_name:, arguments:, data:)
         end
 
         sig { params(file_path: String).returns(T.nilable(T::Boolean)) }
-        def defined_in_gem?(file_path)
+        def not_in_dependencies?(file_path)
           BUNDLE_PATH &&
             !file_path.start_with?(T.must(BUNDLE_PATH)) &&
             !file_path.start_with?(RbConfig::CONFIG["rubylibdir"])
```
aryan-soni, line 76: Looks pretty straightforward, seems like we're checking whether it's in the project's gems or the standard library.

`not_in_dependencies` does sound cleaner than `not_defined_in_gem`, so the renaming makes sense here.

aryan-soni (approved):

egiurleo (approved):

PR #1290, opened 2024-01-09
Return empty diagnostics result if there are no syntax errors and rubocop isn't installed

catlee:
### Motivation
In https://github.com/neovim/neovim/issues/26640 we discovered that `ruby-lsp` sometimes sends back a nil response to `textDocument/diagnostic`.

I tracked this down to when we have a syntactically valid file, and we don't have `rubocop` available, then the diagonstic code returns `nil`.

According to the LSP spec, we should return an empty array in this case.

### Implementation

Return an empty array if we don't have `Support::RuboCopDiagnosticsRunner`

### Automated Tests

I added some tests here, but not sure if the way I'm manipulating the `Support` module is good.
I also noticed that the number of diagnostics we get back from the tests varies; sometimes we get back just rubocop diagnostics, and sometimes we get back rubocop + sorbet diagnostics.

### DIFF
--- a/test/requests/diagnostics_test.rb
+++ b/test/requests/diagnostics_test.rb
@@ -2,7 +2,6 @@
 # frozen_string_literal: true
 
 require "test_helper"
-require "ruby_lsp/requests/support/rubocop_diagnostics_runner"
 
 class DiagnosticsTest < Minitest::Test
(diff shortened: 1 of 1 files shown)


olimorris:
Amazing work @catlee!

PR #1292, opened 2024-01-10
Remove Gemfile.lock for Ruby head

andyw8:
This will prevent failures such as:

```
nokogiri-1.16.0-x86_64-darwin requires ruby version < 3.4.dev, >= 3.0, which is
incompatible with the current version, 3.4.0.dev
```

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -40,7 +40,8 @@ jobs:
       - uses: actions/checkout@v3
         with:
           submodules: "recursive"
-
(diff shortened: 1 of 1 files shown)


andyw8:
Yeah, I guess not.

KaanOzkan:
It could be useful in the future when a platform specific gem is added to Gemfile.

Also the current CI failure on ruby head seems legitimate https://github.com/Shopify/ruby-lsp/actions/runs/7477978717/job/20351904232#step:4:85

PR #1294, opened 2024-01-10
Show method parameters as completion details

vinistock:
### Motivation

We weren't showing method parameters as part of completion label details. I think it's nice to add them, so that you can see more of the signature as you're selecting items.

This is what it looks like. The addition is only the parameters within the parenthesis.

image

### Implementation

Added the missing value using the parameter for the entry we're completing.

### Automated Tests

Added a test to verify we're presenting this.


st0012 (approved):

harrison-hd:
Awesome! 🙏

PR #1296, opened 2024-01-11
Enable bypassTypechcker by default

st0012:
Because this project has Sorbet as a dependency, this setting is required for testing features like completion. And instead of asking every contributor to manually turn it on before testing, which could easily be missed, we can just turn it on by default.


Morriar:
> instead of asking every contributor to manually turn it on before testing

It's only relevant to contributors working on typing related features though, right?

PR #1297, opened 2024-01-11
Rename Request#response to Request#perform

st0012:
### Motivation

As @vinistock pointed out [here](https://github.com/Shopify/ruby-lsp/pull/1247#discussion_r1447637861), using `#response` seems to imply that the request was already completed before it's called. In ruby-lsp's case, many requests are used in this pattern:

```rb
FooRequest.new(args).response
```

In this context, it feels like the computation is done in `FooRequest#initialize` and `#response` is just a getter, which is a bit misleading.

So this PR renames `#response` to `#perform` to make it more clear.

### Implementation

Renames relevant interfaces and callers.

### DIFF
--- a/lib/ruby_lsp/executor.rb
+++ b/lib/ruby_lsp/executor.rb
@@ -94,7 +94,7 @@ module RubyLsp
         cached_response = document.cache_get(request[:method])
         return cached_response if cached_response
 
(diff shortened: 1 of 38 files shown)


vinistock (approved):
Thanks!

PR #1302, opened 2024-01-12
Require the Sorbet version that started supporting `T.anything`

### LINKED ISSUE #1300: RubyLsp::Requests::Request breaks the Ruby LSP extension on shopify-build
### Ruby version

3.1.2

### Code snippet

_No response_

### Description

I am seeing the following error when using Ruby LSP on https://github.com/Shopify/shopify-build:
```
Formatting error: undefined method `anything' for T:Module sig { abstract.returns(T.anything) } ^^^^^^^^^
```

I am guessing this is coming from the recently added `RubyLsp::Requests::Request` which is using `T.anything`: https://github.com/Shopify/ruby-lsp/blob/4f1012e407ed14cccb66007746217b5a558e9f44/lib/ruby_lsp/requests/request.rb#L13

I think the reason we're seeing this is due to the fact that we're on a version of Sorbet (`0.5.10326`) that does not have the `T.anything` type. We use vendored gems on `shopify-build` and update gems once a year. `T.anything` was added to Sorbet in April, 2023: https://github.com/sorbet/sorbet/commit/67c657ad03cdc8a0993ee2b2552f8a12cf638db5 so we haven't updated our Sorbet to a version after that one yet.

Should this type make extension error out?

### Expected output

`T.anything` should not cause Ruby LSP to fail

st0012:
### Motivation

Without this, projects using older Sorbet versions can hit `NoMethodError` when the `T.anything` signature is hit, as shown in #1300.

Closes #1300

### Implementation

Bump the version requirement of `sorbet-runtime` in `gemspec`.

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -4,7 +4,7 @@ PATH
     ruby-lsp (0.13.3)
       language_server-protocol (~> 3.17.0)
       prism (>= 0.19.0, < 0.20)
-      sorbet-runtime (>= 0.5.5685)
+      sorbet-runtime (>= 0.5.10782)
 
 GEM
   remote: https://rubygems.org/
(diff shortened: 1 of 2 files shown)


vinistock (approved):
Does this version also support `T::Class`? If so, then there are a couple of places where we might be able to improve typing. Not on this PR though.

PR #1305, opened 2024-01-12
Fix newline on Gemfile dependency hover response

bravehager:
### Motivation

https://github.com/Shopify/ruby-lsp/commit/31aa8ef8f38eb802728dc473ef2bb35aed4c3384 stripped a newline from the markdown output when hovering over a dependency in a Gemfile.

Not really opposed to the change in formatting but it does seem like an unintended regression.

Before:
Screenshot 2024-01-12 at 12 15 35 PM

After:
Screenshot 2024-01-12 at 12 15 57 PM

### Implementation

Add back the newline.


vinistock (approved):
That looks better indeed. Thanks for the contribution!

PR #1307, opened 2024-01-12
Clarify the role of `.ruby-lsp/Gemfile` when addons are in use

andyw8:
Following up from [this Slack discussion](https://ruby-dx.slack.com/archives/C050D17S6FP/p1705085627678539).


README.md
```diff
@@ -38,6 +38,8 @@ group :development do
 end
 ```
 
+Note: If using an addon such as `ruby-lsp-rails`, `ruby-lsp` will be declared in the main `Gemfile.lock`, so the `.ruby-lsp/Gemfile` and its lockfile are not used, and can be deleted.
```
st0012, line 41: If they're not used, maybe they should be moved by the extension? 🤔

andyw8, line 41: In the case of the Slack conversation, I suspect they first started with Ruby LSP only, then later added `ruby-lsp-rspec`, which resulted in a redundant `.ruby-lsp/` directory.

jscharf: This is true, I don't remember the exact order of operations, but I know we started with LSP only first. Then we found out the recommended way was to remove it from our Gemfile because it would be managed internally.

I don't remember when exactly we added `ruby-lsp-rspec` but I could dig up the commit if it's helpful.

vinistock (commented):
We're starting to see some of the limitations related to how addons are managed, but I don't believe this is the solution.

This looks like a consequence of how they are installed and how we handle the dependency version constraints on `ruby-lsp`, which I still believe is not ideal.

Having the addons control which version of the main LSP gets installed through version constraints defeats our purpose to have developers always on the latest version and creates external dependencies that are not in our best interest (i.e.: other projects we have no control over may limit the LSP's upgrades).

PR #1311, opened 2024-01-15
Restructure DocumentSymbol request

st0012:
### Motivation

The current response-merging approach means that addons and Ruby LSP can't build symbols on top each others. For example, `ruby-lsp-rails` can't simply insert its test symbols to the class symbols Ruby LSP builds. Similarly, Ruby LSP won't be able to add class symbols to the `describe` group `ruby-lsp-rspec` builds.

So this PR explores a different approach: exposing the stack of nodes to addons.

### Implementation

1. Introduce `RubyDocument::DocumentSymbolStack`, which provides `push`, `pop`, and `peek` interfaces.
2. `Request::DocumentSymbol` now only expect `Listeners::DocumentSymbol` to return a response. Addon listeners should directly insert symbols to the stack or children of the existing symbols.
3. As a result of 2, addons' document symbol implementation don't even need to inherit from `Listener`. Examples:
    - https://github.com/st0012/ruby-lsp-rspec/pull/16
    - https://github.com/Shopify/ruby-lsp-rails/pull/231

### Examples

I implemented document symbol for tests on both `ruby-lsp-rails` and `ruby-lsp-rspec`, which have different requirements. But this architecture seems to satisfy both.

| Requirement | `ruby-lsp-rails` | `ruby-lsp-rspec` |
|--------|--------|--------|
| Insert to existing symbols (e.g. class) | Yes | No |
| Insert new parent nodes | No | Yes |
| *Accept Ruby LSP symbols | No | Yes |

#### Accept Ruby LSP symbols

In RSpec, it's possible to define classes or modules under a `describe`/`context` block. So Ruby LSP should be able to insert symbols to the parent nodes addons create:

```rb
RSpec.describe RubyLsp::RSpec do
  class Foo
    def bar
    end
  end
end
```

Screenshot 2024-01-15 at 12 45 22


lib/ruby_lsp/ruby_document.rb
```diff
@@ -10,5 +10,48 @@ def parse
...
+      sig { returns(T::Array[Interface::DocumentSymbol]) }
+      attr_reader :children
+
+      sig { void }
+      def initialize
+        @children = T.let([], T::Array[Interface::DocumentSymbol])
+      end
+    end
+
+    class DocumentSymbolStack
```
st0012, line 26: I know it's likely that this class probably won't be used outside of the document symbol feature. But I'm hesitant putting it under `Requests::DocumentSymbol` as I don't think addons should reference back to a specific request 🤔

st0012:
@vinistock I added `RubyLsp::Response` as the abstract class for response data structures. Can you give it another look and let me know if you think the design make sense?

aryan-soni:
> @vinistock I added `RubyLsp::Response` as the abstract class for response data structures. Can you give it another look and let me know if you think the design make sense?

FWIW, I've just started restructuring how we handle responses for hovers based on @vinistock's suggested approach ([WIP here](https://github.com/Shopify/ruby-lsp/compare/main...as/response-refactor#diff-32f243d853b45666df68409c5f16b12a6dfe8ef2f07b99e2073f3b140e95dd1cR7-R24)). 

Not necessarily needed in this PR, but ideally we can converge on a parent class that the hovering refactoring can also use. In general, refactoring how we handle responses is useful across the board, so ideally there's a more general abstraction we can create.

st0012:
@aryan-soni I plan to explore that once we also have `SemanticHighlight` restructured as stated here: https://github.com/Shopify/ruby-lsp/pull/1311#discussion_r1452732917

I think we need to first understand what complicated response types would look like before we can decide the general abstraction.

vinistock (approved):
Looks good to me. Let's continue with more requests so that we converge.

I think finding consistent naming for all requests might be helpful too. Something like `HoverResponseBuilder`, `DocumentSymbolResponseBuilder`, etc.

Issue #1312, opened 2024-01-15
Error while indexing: invalid byte sequence in UTF-8


jason-o-matic:
### Ruby version

3.0.6

### Description

I get a notification in VSCode with this error message and "Source: Ruby LSP (Extension)", but no further information. I assume there's some file in my project that has weird characters, but the error message doesn't indicate the file with issues, so I'm not sure where to look.

I think the bug here is that this error message should be more helpful and it should reference whichever files had the encoding issue.

### Expected output

An error message with more information about what exactly the problem was, which file it was in, perhaps a line number and column, and even the problematic string.

vinistock:
Thank you for the bug report! It's possible that we're just surfacing the error messages coming from Prism (the parser).

If run `BUNDLE_GEMFILE=.ruby-lsp/Gemfile bundle exec ruby-lsp-doctor` in the command line, it will run indexing and print the files as it goes. That last file before the error should be the problematic one.

If you can, please share what the piece of problematic code is.

jason-o-matic:
Here's the last file and the error message:
```
indexing: /Users/jason/.rbenv/versions/3.0.6/lib/ruby/gems/3.0.0/bundler/gems/color-tools-53138ca93a20/lib/color/hsl.rb
/Users/jason/.rbenv/versions/3.0.6/lib/ruby/gems/3.0.0/gems/ruby-lsp-0.13.4/lib/ruby_indexer/lib/ruby_indexer/collector.rb:307:in `match?': invalid byte sequence in UTF-8 (ArgumentError)
	from /Users/jason/.rbenv/versions/3.0.6/lib/ruby/gems/3.0.0/gems/ruby-lsp-0.13.4/lib/ruby_indexer/lib/ruby_indexer/collector.rb:307:in `block in collect_comments'
	from /Users/jason/.rbenv/versions/3.0.6/lib/ruby/gems/3.0.0/gems/ruby-lsp-0.13.4/lib/ruby_indexer/lib/ruby_indexer/collector.rb:302:in `downto'
	from /Users/jason/.rbenv/versions/3.0.6/lib/ruby/gems/3.0.0/gems/ruby-lsp-0.13.4/lib/ruby_indexer/lib/ruby_indexer/collector.rb:302:in `collect_comments'
[...]

PR #1313, opened 2024-01-17
Force UTF-8 for indexing

andyw8:
Closes https://github.com/Shopify/ruby-lsp/issues/1312

(need to discuss with @vinistock if this is a reasonable approach).

### DIFF
--- a/lib/ruby_indexer/lib/ruby_indexer/index.rb
+++ b/lib/ruby_indexer/lib/ruby_indexer/index.rb
@@ -184,7 +184,7 @@ module RubyIndexer
 
     sig { params(indexable_path: IndexablePath, source: T.nilable(String)).void }
     def index_single(indexable_path, source = nil)
-      content = source || File.read(indexable_path.full_path)
+      content = source || File.read(indexable_path.full_path).force_encoding("UTF-8")
       result = Prism.parse(content)
       collector = Collector.new(self, result, indexable_path.full_path)
       collector.collect(result.value)


lib/ruby_indexer/lib/ruby_indexer/index.rb
```diff
@@ -184,7 +184,7 @@ def index_all(indexable_paths: RubyIndexer.configuration.indexables, &block)
 
     sig { params(indexable_path: IndexablePath, source: T.nilable(String)).void }
     def index_single(indexable_path, source = nil)
-      content = source || File.read(indexable_path.full_path)
+      content = source || File.read(indexable_path.full_path).force_encoding("UTF-8")
```
st0012, line 187: I somehow remember @nirvdrum advocating not to use this method in his 2023 RubyKaigi talk 😂

vinistock: Yeah, I'm not sure this is the right solution.

Why isn't the file encoding `UTF-8` when we read it? Is the problem that the file is using an incompatible encoding?

And if we force everything to be `UTF-8`, can we find ourselves in a situation where we're converting incompatible data into the encoding? Maybe the solution would be to rescue, warn and ignore files with encoding issues.

In other words, it's not clear to me if this is a mistake in the gem that we should be resilient against or if this is something we need to enforce from our side.

@nirvdrum and @kddnewton do you have thoughts on this one?

PR #1317, opened 2024-01-21
Integrate builder pattern to refactor hover responses

aryan-soni:
### Motivation

Hover response construction and manipulation can be delegated to a builder class to promote extensibility and encapsulation.

Eventually, we want to facilitate deterministic categorization for hover responses. As part of this, we should first encapsulate our responses into a `HoverResponseBuilder` class, which will then handle the ordering of content in the future.

### Implementation

The previous approach for hover responses had some suboptimal aspects:
1. Within `Requests::Hover`, the `perform` method was responsible for constructing the response in a convoluted way.
2. The listener was responsible for instantiating `Hover` objects.
3. We were always passing in the `range` when constructing a `Hover` response, but as per the [LSP spec](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#hover), we weren't using this range in a meaningful fashion.

The new approach ensures:
1. We abstract away response construction to a builder. When we add deterministic categorization for hover responses, this will ensure that we have a concrete separation of concerns.
2. We now only construct a `Hover` object at one time, within `Requests::Hover` instead of the listener. This optimizes runtime as it ensures we limit the construction of objects.
3. We can disregard the `range`.
4. There is a more general `ResponseBuilder` class that we can use to refactor responses for other areas of the LSP as well.

### Automated Tests

In `hover_expectations_test.rb`, we don't want to modify the expectations themselves (as the contract persists), but we can change the initialization work done to facilitate the tests such that it now reflects our changes.

We can also updated our broader expectations to remove `range`, since we are now phasing this out. 

### Manual Tests

We're not adding/removing any functionality, but are simply laying the foundation for improvements later in the future. As a consequence, testing this involves ensuring that our hover functionality is unchanged. 

As we can see, the hover functionality persists:

hover_functionality

### DIFF
--- a/ADDONS.md
+++ b/ADDONS.md
@@ -129,6 +129,7 @@ module RubyLsp
     end
 
     # All listeners have to inherit from ::RubyLsp::Listener
+    # NOTE: Listeners are currently being refactored, but this implementation embodies the current approach for most listeners.
     class Hover < ::RubyLsp::Listener
       extend T::Sig
       extend T::Generic
--- a/lib/ruby_lsp/addon.rb
+++ b/lib/ruby_lsp/addon.rb
@@ -1,6 +1,8 @@
 # typed: strict
 # frozen_string_literal: true
 
+require_relative "response_builder"
+
 module RubyLsp
   # To register an addon, inherit from this class and implement both `name` and `activate`
   #
@@ -118,12 +120,13 @@ module RubyLsp
     # Creates a new Hover listener. This method is invoked on every Hover request
     sig do
       overridable.params(
+        response_builder: RubyLsp::HoverResponseBuilder,
         nesting: T::Array[String],
         index: RubyIndexer::Index,
         dispatcher: Prism::Dispatcher,
       ).returns(T.nilable(Listener[T.nilable(Interface::Hover)]))
     end
-    def create_hover_listener(nesting, index, dispatcher); end
+    def create_hover_listener(response_builder, nesting, index, dispatcher); end
 
     # Creates a new DocumentSymbol listener. This method is invoked on every DocumentSymbol request
     sig do
--- a/lib/ruby_lsp/listeners/hover.rb
+++ b/lib/ruby_lsp/listeners/hover.rb
@@ -1,13 +1,13 @@
 # typed: strict
 # frozen_string_literal: true
 
+require "ruby_lsp/response_builder"
+
 module RubyLsp
   module Listeners
-    class Hover < Listener
+    class Hover
       extend T::Sig
-      extend T::Generic
-
-      ResponseType = type_member { { fixed: T.nilable(Interface::Hover) } }
+      include Requests::Support::Common
 
       ALLOWED_TARGETS = T.let(
         [
@@ -19,11 +19,9 @@ module RubyLsp
         T::Array[T.class_of(Prism::Node)],
       )
 
-      sig { override.returns(ResponseType) }
-      attr_reader :_response
-
       sig do
         params(
+          response_builder: RubyLsp::HoverResponseBuilder,
           uri: URI::Generic,
           nesting: T::Array[String],
           index: RubyIndexer::Index,
@@ -31,14 +29,13 @@ module RubyLsp
           typechecker_enabled: T::Boolean,
         ).void
       end
-      def initialize(uri, nesting, index, dispatcher, typechecker_enabled)
+      def initialize(response_builder, uri, nesting, index, dispatcher, typechecker_enabled) # rubocop:disable Metrics/ParameterLists
+        @response_builder = response_builder
         @path = T.let(uri.to_standardized_path, T.nilable(String))
         @nesting = nesting
         @index = index
         @typechecker_enabled = typechecker_enabled
-        @_response = T.let(nil, ResponseType)
 
-        super(dispatcher)
         dispatcher.register(
           self,
           :on_constant_read_node_enter,
@@ -86,12 +83,7 @@ module RubyLsp
         target_method = @index.resolve_method(message, @nesting.join("::"))
         return unless target_method
 
-        location = target_method.location
-
-        @_response = Interface::Hover.new(
-          range: range_from_location(location),
-          contents: markdown_from_index_entries(message, target_method),
-        )
+        @response_builder << markdown_from_index_entries(message, target_method)
       end
 
       private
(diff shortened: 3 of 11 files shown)


ADDONS.md
```diff
@@ -129,6 +129,7 @@ module RubyLsp
     end
 
     # All listeners have to inherit from ::RubyLsp::Listener
+    # NOTE: Listeners are currently being refactored, but this implementation embodies the current approach for most listeners.
```
aryan-soni, line 132: Thought it would be helpful to add a note here since the code below will no longer be accurate for `Hover`

test/expectations/hover/documented_class.exp.json
```diff
@@ -9,16 +9,6 @@
...
-        },
-        "range": {
-            "start": {
-                "line": 1,
-                "character": 6
-            },
-            "end": {
-                "line": 1,
-                "character": 9
-            }
```
st0012, line 21: I didn't expect the changes to affect the request's behaviour. Do you know what's causing these, and why are they necessary for the task?

aryan-soni: Really good question.

If you take a look at [this excerpt](https://github.com/Shopify/ruby-lsp/pull/1317/files#diff-ffcc875e5a79593a72a08918054a81f51e7410fb0dc451b01db8067c5814a0d5L83-L88), we were previously ignoring the ranges of all other response except the first response.

In essence, we weren't concatenating the ranges in a meaningful fashion. Given that the range itself is optional and that we aren't using it in the intended way, @vinistock and I figured it would make sense to omit it as part of the initial refactoring.

lib/ruby_lsp/response_builder.rb
```diff
@@ -0,0 +1,50 @@
+# typed: strict
+# frozen_string_literal: true
+
+module RubyLsp
+  class OperationNotPermitted < StandardError; end
+
+  class ResponseBuilder
```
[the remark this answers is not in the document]
aryan-soni: Yeah, I agree we should unify the abstract classes. 

I'll make the change, but to your point, I definitely think we should narrow in to a common builder class before we proceed with the release.

lib/ruby_lsp/requests/hover.rb
```diff
@@ -52,19 +53,16 @@ def initialize(document, index, position, dispatcher, typechecker_enabled)
           target = parent
         end
 
-        @listeners = T.let([], T::Array[Listener[ResponseType]])
+        @listeners = T.let([], T::Array[Listener[HoverResponseBuilder]])
```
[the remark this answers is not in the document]
aryan-soni: Looks like we can remove this line since we don't use it elsewhere regardless.

ADDONS.md
vinistock: Yeah, we'll need to update all of the docs. It might be worth waiting until the refactors are complete and then we update them once.

lib/ruby_lsp/addon.rb
```diff
@@ -1,6 +1,8 @@
 # typed: strict
 # frozen_string_literal: true
 
+require_relative "response_builder"
```
vinistock, line 4: This will be necessary for most of the LSP, so I think we can move this require to `ruby_lsp/internal` where we require most parts.

lib/ruby_lsp/listeners/hover.rb
```diff
@@ -1,13 +1,13 @@
 # typed: strict
 # frozen_string_literal: true
 
+require "ruby_lsp/response_builder"
```
vinistock, line 4: If we move to `internal`, then we don't need this one.

lib/ruby_lsp/requests/hover.rb
```diff
@@ -2,6 +2,7 @@
 # frozen_string_literal: true
 
 require "ruby_lsp/listeners/hover"
+require "ruby_lsp/response_builder"
```
vinistock, line 5: This can be removed too.

lib/ruby_lsp/requests/hover.rb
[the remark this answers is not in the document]
vinistock: Agreed with Stan's comment about the future structure of listeners, but indeed we can just remove this line since it's no longer used.

lib/ruby_lsp/response_builder.rb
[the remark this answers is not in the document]
st0012: I decided to adopt the builder concept in #1325 (with a slightly different form of structure). 

But I still think we should not add more common APIs before we've converted most requests. For example, we probably don't want addons to have the `pop` API for semantic highlighting.

lib/ruby_lsp/requests/support/common.rb
```diff
@@ -86,7 +86,7 @@ def self_receiver?(node)
           params(
             title: String,
             entries: T.any(T::Array[RubyIndexer::Entry], RubyIndexer::Entry),
-          ).returns(Interface::MarkupContent)
+          ).returns(String)
```
vinistock, line 89: Can you please double-check if we use this method anywhere else? I kind of expected stuff to break when we changed the type here.

If I'm not mistaken we use it for documentation somewhere. Since it now returns a `String` instead of an `Interface::MarkupContent`, we need to do the same thing we did in hover and instantiate it.

```ruby
Interface::MarkupContent.new(
  kind: "markdown",
  value: markdown_from_index_entries(...),
)
```

test/expectations/hover/documented_class.exp.json
st0012: Thanks for the explanation 👍

vinistock: We were also using `range` incorrectly. The [spec](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#textDocument_hover) says it's an optional range used to visualize a hover, but we were always just passing the location of the declaration, which doesn't sound entirely correct.

Removing it seems to make no difference in the behaviour we see in the editor.

lib/ruby_lsp/response_builder.rb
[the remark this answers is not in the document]
aryan-soni: I think it would be best to finalize on an approach for the builder in this PR, get this merged, then you can rebase in your PR. I can factor in your approach and modify mine accordingly. 

Thoughts? Just thinking of the best way to minimize conflicts.

lib/ruby_lsp/requests/support/common.rb
aryan-soni: Great point — see #1328, where I've modified `signature_help.rb` and `completion.rb` as you've described.

PR #1325, opened 2024-01-25
Restructure semantic highlighting and responses

st0012:
### Motivation

- We want semantic highlighting to ve extensible in a way that addons can access to individual tokens without being able to erase the entire token stack.
- We're exploring ways to standardise responses that expose different APIs.

#### Dedicated `ResponseBuilder` classes

All extensible requests should have a specialised `ResponseBuilder` class, which should:

- Encapsulate the core data and only expose limited APIs to addons
- Be responsible for converting the internal data structure to interface objects.
     - So `SemanticHighlighting#response` should return `Interface::SemanticTokens` instead of the custom `SemanticToken` objects.

And instead of requiring individual response builders, I think it's great to mimic the autoloading configuration of requests.

#### Extensible SemanticHighlighting request

Now `Requests::SemanticHighlighting` also calls `create_semantic_highlighting_listener` on addons to create additional listeners.

### DIFF
--- a/test/requests/semantic_highlighting_expectations_test.rb
+++ b/test/requests/semantic_highlighting_expectations_test.rb
@@ -59,7 +59,7 @@ class SemanticHighlightingExpectationsTest < ExpectationsTestRunner
       )
       # This is the token modified by the addon
       assert_equal(
-        { delta_line: 1, delta_start_char: 2, length: 13, token_type: 13, token_modifiers: 1 },
+        { delta_line: 1, delta_start_char: 2, length: 13, token_type: 15, token_modifiers: 1 },
         decoded_response[3],
       )
     end
@@ -79,8 +79,10 @@ class SemanticHighlightingExpectationsTest < ExpectationsTestRunner
           end
 
           def on_call_node_enter(node)
-            if node.message == "before_create"
-              @response_builder.last.modifier << 0
+            current_token = @response_builder.last
+            if node.message == "before_create" && node.message_loc == current_token.location
+              current_token.replace_type(:keyword)
+              current_token.replace_modifier([:declaration])
             end
           end
         end
--- a/lib/ruby_lsp/response_builders/semantic_highlighting.rb
+++ b/lib/ruby_lsp/response_builders/semantic_highlighting.rb
@@ -4,6 +4,8 @@
 module RubyLsp
   module ResponseBuilders
     class SemanticHighlighting < ResponseBuilder
+      class UndefinedTokenType < StandardError; end
+
       TOKEN_TYPES = T.let(
         {
           namespace: 0,
@@ -104,6 +106,24 @@ module RubyLsp
           @length = length
           @type = type
           @modifier = modifier
+        end
+
+        sig { params(type_symbol: Symbol).void }
+        def replace_type(type_symbol)
+          type_int = TOKEN_TYPES[type_symbol]
+          raise UndefinedTokenType, "Undefined token type: #{type_symbol}" unless type_int
+
+          @type = type_int
+        end
+
+        sig { params(modifier_symbols: T::Array[Symbol]).void }
+        def replace_modifier(modifier_symbols)
+          @modifier = modifier_symbols.filter_map do |modifier_symbol|
+            modifier_index = TOKEN_MODIFIERS[modifier_symbol]
+            raise UndefinedTokenType, "Undefined token modifier: #{modifier_symbol}" unless modifier_index
+
+            modifier_index
+          end
         end
       end
 


lib/ruby_lsp/responses/response.rb
```diff
@@ -0,0 +1,18 @@
...
+    class Response
+      extend T::Sig
+      extend T::Generic
+
+      abstract!
+
+      ResponseType = type_member
+
+      sig { abstract.returns(ResponseType) }
+      def result; end
```
st0012, line 15: Maybe the names should be these instead?

- `ResponseBuilders`
- `ResponseBuilder`
- `ResponseBuilder#response`

lib/ruby_lsp/requests/semantic_highlighting.rb
```diff
@@ -24,8 +24,6 @@ class SemanticHighlighting < Request
       extend T::Sig
       extend T::Generic
 
-      ResponseType = type_member { { fixed: T::Array[Listeners::SemanticHighlighting::SemanticToken] } }
-
```
aryan-soni, line 28: The `extend T::Generic` is no longer necessary if we remove the declaration of the generic type

lib/ruby_lsp/response_builders/response.rb
aryan-soni, line 1: The file name should probably be `response_builder.rb`

lib/ruby_lsp/response_builders/response.rb
```diff
@@ -0,0 +1,18 @@
...
+
+module RubyLsp
+  module ResponseBuilders
+    class ResponseBuilder
+      extend T::Sig
+      extend T::Generic
+
+      abstract!
+
+      ResponseType = type_member
```
aryan-soni, line 12:
```suggestion
      ResponseType = type_member { { upper: Object } }
```

aryan-soni (commented):
Took a look with @vinistock 

He'll follow up with a couple more comments, but looks pretty good, just a couple things we wanted to note

test/requests/semantic_highlighting_expectations_test.rb
```diff
@@ -21,16 +21,75 @@ def run_expectations(source)
...
+
+          def initialize(response_builder, dispatcher)
+            @response_builder = response_builder
+            dispatcher.register(self, :on_call_node_enter)
+          end
+
+          def on_call_node_enter(node)
+            if node.message == "before_create"
+              @response_builder.last.modifier << 0
+            end
```
vinistock, line 84: Let's add a test for an addon that needs to override the token type rather than changing the modifiers and check what happens.

We do this for the Sorbet annotations and it'll be necessary for other addons. The base Ruby LSP pushes the `method` token type for method calls, but we want the annotations to have the `type` token type.

It's important we stick to the types and modifiers from the [spec](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#semanticTokenTypes) because that's what themes use to define the colours they will use. If we deviate from that, then existing themes will highlight code incorrectly.

The main question here is: if the Ruby LSP and an addon push different token types for the same exact location, does VS Code prefer the first or the second token?

If it prefers the second one, then maybe we don't need to do anything and just allow the last addon to have the final say. If it picks the first one, then we'll need to override the type.

st0012: > Is there ever a case where an addon would like to modify a token that isn't the current one (last one we pushed)?

I can't think of any at the moment.

> And how would the addon author know they need to modify an existing token rather than adding a separate one? Is there an API we can provide that will guide them to do the right thing?

I was thinking about not adding `add_token` to the builder and use `instance_variable_get` in our own listener to insert tokens. Then the only exposed APIs are `replace_type` and `replace_modifier`. WDYT?

vinistock: I'm not sure that would be enough. An addon could want to push a token that doesn't already exist.

What do you think about exposing an API and handling this complexity from our side? Something like this

```ruby
# This is pseudo code, but you get the idea
class SemanticHighlightingResponseBuilder
  def add_token(location, ...)
    if last_token.location == location
      # modify last token with type and modifiers
    else
      # push new token
    end
  end
end
```

st0012: > An addon could want to push a token that doesn't already exist.

Do you have a use case for it? If not, maybe we can wait for feedback before introducing new APIs that ourselves would not need?

> What do you think about exposing an API and handling this complexity from our side?

It's worth exploring. But again, without a real use case I can not confidently say it'd be sufficient or necessary. For the ideas I have for `ruby-lsp-rails` and `ruby-lsp-rspec`, being able to replace type or modifier is already enough. Maybe when ERB support is shipped we can have new tokens? But if that's the case, let's design the API when that happens.

Also, I think we can have this discussion in a follow up issue/PR as the builder structure is needed to unblock other requests' refactors.

vinistock (approved):

Issue #1246, opened 2023-12-12
Investigate why `to_spec` is nil for prerelease gems


andyw8:
See https://github.com/Shopify/ruby-lsp/pull/1242

This can be reproduced on the `irb` repo with `gem "reline", "0.5.0.pre.1"`

vinistock:
Amazing, @kimesf! Thank you very much for investigating, fixing the bug and then bringing back the context.

I think we can close this issue then and direct anyone who still has the problem to upgrade `rubygems` and `bundler`. @andyw8 do you agree?

kimesf:
> 👍 We'll keep the fix in #1242, and can remove once we are confident people are on a newer rubygems.

Just to point it out, there were multiple fixes for the issue (#1159, #1242 and #1326).

Issue #1141, opened 2023-10-25
Format on save is not working.


thibgn:
### Operating System

Mac OS 13.0

### Ruby version

3.2.2

### Project has a bundle

- [X] Has bundle

### Ruby version manager being used

rbenv

### Description

I'm having a few issues with Ruby LSP that I can't seem to fix.

First, the indexing is stuck at 0%. 
No matter how long I wait, the settings I use, I never see anything else than `Ruby LSP: indexing files: 0% completed`

But the main issue is the **formatting that doesn't work on save** (it _kinda_ works when I format manually, but it only fixes indentation).

I obviously checked my VSCode configuration and nothing should prevent the format on save (it's working for other languages).

Here's the relevant settings I use for ruby : 
```json
  "rubyLsp.rubyVersionManager": "rbenv",
  "[ruby]": {
	"editor.defaultFormatter": "Shopify.ruby-lsp",
	"editor.formatOnSave": true,
	"editor.formatOnType": true,
	"editor.tabSize": 2,
	"editor.insertSpaces": true,
	"files.trimTrailingWhitespace": true,
	"files.insertFinalNewline": true,
	"files.trimFinalNewlines": true,
	"editor.rulers": [120],
	"editor.semanticHighlighting.enabled": true
  },
```

I noticed I had both `rvm` & `rbenv` installed so I removed rvm and made sure rbenv was [properly configured](https://github.com/rbenv/rbenv#basic-git-checkout).
I also made sure `$SHELL` was available : i'm using `zsh` so its value is `/bin/zsh`

I also removed everything ruby or rails related in my VScode extensions.

Here's the logs from Ruby LSP : 
```
Ruby LSP> Skipping custom bundle setup since /Users/thib/code/myproject/api/.ruby-lsp/Gemfile.lock already exists and is up to date
Ruby LSP> Running bundle install for the custom bundle. This may take a while...
The Gemfile's dependencies are satisfied
Starting Ruby LSP...
```

Also, in the Output tab, `Window` is logging the following which may be related ? 

```
2023-10-25 10:01:00.488 [error] [Extension Host] Received response message without id: Error is: 
{
    "code": -32603,
    "message": "#<NoMethodError: undefined method `dependencies' for nil:NilClass>",
    "data": {
        "errorClass": "NoMethodError",
        "errorMessage": "undefined method `dependencies' for nil:NilClass",
[...]

PR #1326, opened 2024-01-26
handles one more #to_spec returning nil in RubyIndexer::Configuration

kimesf:
### Motivation

Apparently, there is a know issue when `#to_spec` returns `nil` (see #1141 and #1246).

While you are still investigating why exactly it happens, some precautions were made in order to run `ruby-lsp` with no crashes (see PRs [1242](https://github.com/Shopify/ruby-lsp/pull/1242) and [1159](https://github.com/Shopify/ruby-lsp/pull/1159)).

It seems there is one more `#to_spec #=> nil` that needs handling. This PR handles this case.

### Implementation

To make sure the fix would allow me to run `ruby-lsp` with no crashes, I implemented the fix locally by manually changing the `/lib/ruby_indexer/lib/ruby_indexer/configuration.rb` file on my setup.

It worked.

Reviewers, be aware. I'm also not sure why this happens nor possible side effects of this change.

### Manual Tests

Run `ruby-lsp` for gemfile (bundled with 2.5.3):

```Ruby
source 'https://rubygems.org'

git_source(:github) { |repo| "https://github.com/#{repo}.git" }

ruby '3.3.0'

gem 'rails', '~> 7.1.2'
gem 'validates_timeliness', '>= 6.0.1'
```

It crashes with:

```console
.../ruby-lsp-0.13.4/lib/ruby_indexer/lib/ruby_indexer/configuration.rb:216:in `block (3 levels) in initial_excluded_gems': undefined method `dependencies' for nil (NoMethodError)

            d.to_spec.dependencies.include?(transitive_dependency)
```

### DIFF
--- a/lib/ruby_indexer/lib/ruby_indexer/configuration.rb
+++ b/lib/ruby_indexer/lib/ruby_indexer/configuration.rb
@@ -213,7 +213,7 @@ module RubyIndexer
 
           # If the transitive dependency is included as a transitive dependency of a gem outside of the development
           # group, skip it
-          next if others.any? { |d| d.to_spec && d.to_spec.dependencies.include?(transitive_dependency) }
+          next if others.any? { |d| d.to_spec&.dependencies&.include?(transitive_dependency) }
 
           excluded << transitive_dependency
         end


kimesf:
I have signed the CLA!

andyw8:
I notice that gem has an `encoding` magic comment on the first line, I wonder if that could relate to this problem 🤔 

https://github.com/adzap/validates_timeliness/blob/master/validates_timeliness.gemspec#L1

kimesf:
Thanks to @vinistock pointing to what I think is the right code in rubygems, I did some digging, but no final conclusions yet.

[See the opened PR draft in rubygems](https://github.com/rubygems/rubygems/pull/7428).

lib/ruby_indexer/lib/ruby_indexer/configuration.rb
```diff
@@ -213,7 +213,7 @@ def initial_excluded_gems
 
           # If the transitive dependency is included as a transitive dependency of a gem outside of the development
           # group, skip it
-          next if others.any? { |d| d.to_spec.dependencies.include?(transitive_dependency) }
+          next if others.any? { |d| d.to_spec && d.to_spec.dependencies.include?(transitive_dependency) }
```
vinistock, line 216:
```suggestion
          next if others.any? { |d| d&.to_spec.dependencies.include?(transitive_dependency) }
```

st0012: Wouldn't this cause `dependencies` to be called on `nil`?

vinistock: Yeah, sorry, I meant with safe navigation in all chained methods. I'm fine with leaving it was is too.

vinistock:

PR #1328, opened 2024-01-27
Refactor hover request to adhere to builder pattern

aryan-soni:
### Motivation

Hover response construction and manipulation can be delegated to a builder class to promote extensibility and encapsulation.

Eventually, we want to facilitate deterministic categorization for hover responses. As part of this, we should first encapsulate our responses.

### Implementation

The previous approach for hover responses had some suboptimal aspects:
1. Within `Requests::Hover`, the `perform` method was responsible for constructing the response in a convoluted way.
2. The listener was responsible for instantiating `Hover` objects.
3. We were always passing in the `range` when constructing a `Hover` response, but as per the [LSP spec](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#hover), we weren't using this range in a meaningful fashion.

The new approach ensures:
1. We abstract away response construction to a builder. When we add deterministic categorization for hover responses, this will ensure that we have a concrete separation of concerns.
2. We now only construct a `Hover` object at one time, within `Requests::Hover` instead of the listener. This optimizes runtime as it ensures we limit the construction of objects.
3. We can disregard the `range`.

### Automated Tests

In `hover_expectations_test.rb`, we don't want to modify the expectations themselves (as the contract persists), but we can change the initialization work done to facilitate the tests such that it now reflects our changes.

We can also updated our broader expectations to remove `range`, since we are now phasing this out. 

### Manual Tests

We're not adding/removing any functionality, but are simply laying the foundation for improvements later in the future. As a consequence, testing this involves ensuring that our hover functionality is unchanged. 

As we can see, the hover functionality persists:

hover_functionality

### DIFF
--- a/lib/ruby_lsp/addon.rb
+++ b/lib/ruby_lsp/addon.rb
@@ -118,12 +118,13 @@ module RubyLsp
     # Creates a new Hover listener. This method is invoked on every Hover request
     sig do
       overridable.params(
+        response_builder: ResponseBuilders::Hover,
         nesting: T::Array[String],
         index: RubyIndexer::Index,
         dispatcher: Prism::Dispatcher,
       ).returns(T.nilable(Listener[T.nilable(Interface::Hover)]))
     end
-    def create_hover_listener(nesting, index, dispatcher); end
+    def create_hover_listener(response_builder, nesting, index, dispatcher); end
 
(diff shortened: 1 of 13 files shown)


st0012 (approved):
Nice 👍

PR #1337, opened 2024-01-29
Fully adopt the new response builder pattern and remove the Listener class

st0012:
### Motivation

After adopting the response builder pattern in extensible requests, we currently have 3 types of requests:

1. Not listener-based
2. (Legacy) `Listener`-class-based listeners
3. PORO (plain old Ruby object) listeners

While we think it's fine to cut a release first, I found that it's actually not that hard to eliminate 2 completely.

### Implementation

- I adopted the new pattern in the remaining requests with dedicated commits:
    - Completion
    - DocumentHighlight
    - DocumentLink
    - FoldingRanges
    - InlayHints
    - SignatureHelp
- Then I removed the `Listener` class completely
- Finally, I also updated the messages in the `check_docs` task


lib/ruby_lsp/response_builders/completion.rb
```diff
@@ -0,0 +1,28 @@
+# typed: strict
+# frozen_string_literal: true
+
+module RubyLsp
```
vinistock, line 4: We don't need to create these specific response builders for the requests that just need an array. We can instead specify the type parameter when instantiating to get the type safety.

That will allow us to get rid of most of these builder classes and then we only need the ones that truly have different behaviour than the base one.

Here's an [example](https://sorbet.run/#%23%20typed%3A%20true%0A%0Aclass%20ResponseBuilder%0A%20%20extend%20T%3A%3ASig%0A%20%20extend%20T%3A%3AHelpers%0A%20%20extend%20T%3A%3AGeneric%0A%0A%20%20Type%20%3D%20type_member%20%7B%7B%20upper%3A%20Object%20%7D%7D%0A%0A%20%20def%20initialize%0A%20%20%20%20%40items%20%3D%20T.let%28%5B%5D%2C%20T%3A%3AArray%5BType%5D%29%0A%20%20end%0A%0A%20%20sig%20%7B%20params%28item%3A%20Type%29.void%20%7D%0A%20%20def%20%3C%3C%28item%29%0A%20%20%20%20%40items%20%3C%3C%20item%0A%20%20end%0Aend%0A%0Ainteger_builder%20%3D%20ResponseBuilder%5BInteger%5D.new%0A%23%20Can't%20push%20incorrect%20types%20to%20it%0Ainteger_builder%20%3C%3C%20%22hello!%22%0Ainteger_builder%20%3C%3C%20123).

lib/rubocop/cop/ruby_lsp/use_register_with_handler_method.rb
```diff
@@ -13,9 +13,8 @@ module RubyLsp
       # # Register without handler method.
       #
       # # bad
-      # class MyListener < Listener
+      # class MyListener
```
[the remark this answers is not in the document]
st0012: Yeah there are tests covering it: `test/rubocop/cop/ruby_lsp/use_register_with_handler_method_test.rb`.

lib/ruby_lsp/response_builders/collection_response_builder.rb
```diff
@@ -0,0 +1,30 @@
+# typed: strict
+# frozen_string_literal: true
+
+module RubyLsp
+  module ResponseBuilders
+    class CollectionResponseBuilder < ResponseBuilder
+      extend T::Sig
+      extend T::Helpers
```
vinistock, line 8: I know I put this in the example, but I just realized this is not necessary 😅.

vinistock (approved):
Awesome 🚀

PR #1340, opened 2024-01-30
Account for class name conflicts in resolve_method

vinistock:
### Motivation

Ruby allows for method names that may conflict with constant names, such as `def Array(); end`. When we have a scenario like that, then we might fetch constant entries mixed with method entries, which results in `undefined method 'owner' for RubyIndexer::Entry::Class`.

### Implementation

Started filtering in the `find` for `Entry::Member` only, which includes attributes and methods.

As a note, we may want to explore splitting the entries hash to divide constants and methods. In that case, we'd have two smaller hashes and prefix trees rather than one large one. Not sure what the performance and memory implications would be, but it would make it easier to prevent issues like these.

### Automated Tests

Added two tests to prevent regressions for name conflicts or attributes.

### DIFF
--- a/lib/ruby_indexer/lib/ruby_indexer/index.rb
+++ b/lib/ruby_indexer/lib/ruby_indexer/index.rb
@@ -247,8 +247,8 @@ module RubyIndexer
 
       owner_name = T.must(owner_entries.first).name
       T.cast(
-        method_entries.find do |entry|
-          entry.is_a?(Entry::Member) && entry.owner&.name == owner_name
+        method_entries.grep(Entry::Member).find do |entry|
+          T.cast(entry, Entry::Member).owner&.name == owner_name
         end,
         T.nilable(Entry::Member),
       )


lib/ruby_indexer/lib/ruby_indexer/index.rb
```diff
@@ -239,14 +239,19 @@ def follow_aliased_namespace(name)
...
+      method_entries = self[method_name]
       owner_entries = self[receiver_name]
       return unless owner_entries && method_entries
 
       owner_name = T.must(owner_entries.first).name
-      method_entries.find { |entry| entry.owner&.name == owner_name }
+      T.cast(
+        method_entries.find do |entry|
+          entry.is_a?(Entry::Member) && entry.owner&.name == owner_name
+        end,
```
paracycle, line 252:
```suggestion
        method_entries.grep(Entry::Member).find do |entry|
          entry.owner&.name == owner_name
        end,
```

KaanOzkan (approved):

PR #1341, opened 2024-01-30
Gemfile cleanup

andyw8:
Using the `platforms` block means we can avoid the ugly constant.

~And since this is a gem, there's not really any need for a `development` group.~

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -4,25 +4,25 @@ source "https://rubygems.org"
 
 gemspec
 
-# sorbet-static is not available on Windows. We also skip Tapioca since it depends on sorbet-static-and-runtime
-NON_WINDOWS_PLATFORMS = [:ruby] # C Ruby (MRI), Rubinius or TruffleRuby, but NOT Windows
-
 group :development do
   gem "bundler", "~> 2.5"
   gem "debug", "~> 1.9", require: false
-  gem "minitest", "~> 5.21"
   gem "minitest-reporters", "~> 1.6"
+  gem "minitest", "~> 5.21"
   gem "mocha", "~> 2.1"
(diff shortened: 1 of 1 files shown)


andyw8:
Ah, forgot about that, restored the group.

Gemfile
```diff
@@ -4,25 +4,25 @@ source "https://rubygems.org"
 
 gemspec
 
-# sorbet-static is not available on Windows. We also skip Tapioca since it depends on sorbet-static-and-runtime
-NON_WINDOWS_PLATFORMS = [:ruby] # C Ruby (MRI), Rubinius or TruffleRuby, but NOT Windows
-
 group :development do
   gem "bundler", "~> 2.5"
   gem "debug", "~> 1.9", require: false
-  gem "minitest", "~> 5.21"
```
andyw8, line 13: Sorted alphabetically.

Gemfile
```diff
@@ -4,25 +4,25 @@ source "https://rubygems.org"
...
-  gem "rdoc", require: false
-  gem "psych", "~> 5.1", require: false
-
+  gem "rubocop", "~> 1.60"
   gem "syntax_tree", ">= 6.1.1", "< 7"
+
+  platforms :ruby do # C Ruby (MRI), Rubinius or TruffleRuby, but NOT Windows
+    # sorbet-static is not available on Windows. We also skip Tapioca since it depends on sorbet-static-and-runtime
+    gem "tapioca", "~> 0.12", require: false
+    gem "sorbet-static-and-runtime"
```
Morriar, line 26: Shouldn't this also be sorted?

andyw8: I suppose so 😁 
Fixed

vinistock (approved):

PR #1346, opened 2024-02-01
Add a basic RubyLSP dependencies custom request

paracycle:
### Motivation

In order to support the ["Dependencies" view in VSCode](https://github.com/Shopify/vscode-ruby-lsp/pull/1018), we need to add a custom request in Ruby LSP to gather and return a list of dependencies of the current project.

### Implementation

Discover all direct dependencies using `Bundler.definition.locked_deps` and then convert all of the specs in the `Bundler.definition` into a list of spec name, version, path and if the spec is a direct dependency or not. The last property is useful for either sorting the view based on direct dependency vs transitive dependency, and, in the future, will also allow us to show "Remove" buttons next to direct dependencies.

### Automated Tests

No tests yet

### Manual Tests

Can't manually test it except for making the LSP request.

### DIFF
--- a/lib/ruby_lsp/executor.rb
+++ b/lib/ruby_lsp/executor.rb
@@ -223,15 +223,7 @@ module RubyLsp
       when "rubyLsp/textDocument/showSyntaxTree"
         show_syntax_tree(uri, request.dig(:params, :range))
       when "rubyLsp/workspace/dependencies"
-        dep_keys = Bundler.definition.locked_deps.keys.to_set
-        Bundler.definition.specs.map do |spec|
-          {
-            name: spec.name,
-            version: spec.version,
-            path: spec.full_gem_path,
-            dependency: dep_keys.include?(spec.name),
-          }
-        end
+        workspace_dependencies
       else
         VOID
       end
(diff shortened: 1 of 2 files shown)


lib/ruby_lsp/executor.rb
```diff
@@ -222,6 +222,16 @@ def run(request)
         workspace_symbol(request.dig(:params, :query))
       when "rubyLsp/textDocument/showSyntaxTree"
         show_syntax_tree(uri, request.dig(:params, :range))
+      when "rubyLsp/workspace/dependencies"
+        dep_keys = Bundler.definition.locked_deps.keys.to_set
+        Bundler.definition.specs.map do |spec|
```
st0012, line 227: I think we had a few cases where `definition.to_spec` returns `nil` (e.g. #1326). Can `definition.specs` also contain `nil`?

paracycle: I don't think so, since `definition.specs` should map directly to the list of specs in the `GEM` section of a `Gemfile.lock`.

Btw, the issue you mention is a RubyGems issue, not a Bundler issue, whereas the calls we are making here are purely to Bundler.

PR #1350, opened 2024-02-04
Enable hover content to be categorized

aryan-soni:
### Motivation

Enables hover content to be displayed deterministically, based on categories. Add-ons that add content to hovers can specify the category of the content, then the Ruby LSP can render the content in hovers in an order that semantically maps to the category specified.

### Implementation

Updated the response builder to require categorization, with 3 possible categories for hover content (code, links and documentation).

The response is now stored via a hash, instead of an array. The hash contains the categories as its keys, and the associated content as its values.

In `common.rb`, a helper method has been extracted to map each piece of content with its category. This helper is then used when pushing to the response builder.

### Automated Tests

The `hover_expectations_test` has been updated based on these changes. Note that the expectation tests associated with metaprogramming that are used have not been updated, since our expectation of what we expect to be rendered is the same.

### Manual Tests

The goal here is to ensure we aren't disrupting the hover behaviour. Below is an example of the hover functionality in its current state: 

hover_before_changes

This is the hover functionality with the changes in this PR made:

hover_after_change

The hover remains unchanged. This is because the LSP categorizes everything as expected and has not modified the visual output of the hover itself.

### DIFF
--- a/lib/ruby_lsp/response_builders/hover.rb
+++ b/lib/ruby_lsp/response_builders/hover.rb
@@ -16,7 +16,6 @@ module RubyLsp
         @response = T.let(
           {
             title: +"",
-            signature: +"",
             links: +"",
             documentation: +"",
           },
@@ -40,7 +39,6 @@ module RubyLsp
       sig { override.returns(ResponseType) }
       def response
         result = T.must(@response[:title])
-        result << @response[:signature] if @response[:signature]
         result << "\n" << @response[:links] if @response[:links]
         result << "\n" << @response[:documentation] if @response[:documentation]
 
--- a/test/requests/hover_expectations_test.rb
+++ b/test/requests/hover_expectations_test.rb
@@ -288,10 +288,10 @@ class HoverExpectationsTest < ExpectationsTestRunner
               "Links", :links
             )
             @response_builder.push(
-              "Signature", :signature
+              "Title", :title
             )
             @response_builder.push(
-              "Title", :title
+              "Signature", :title
             )
           end
         end


lib/ruby_lsp/response_builders/hover.rb
```diff
@@ -12,27 +12,39 @@ class Hover < ResponseBuilder
...
-        @stack.empty?
+        @response.all? { |_key, value| value.empty? }
       end
 
-      sig { override.returns(String) }
+      sig { override.returns(ResponseType) }
       def response
-        @stack.join("\n\n")
+        result = T.must(@response[:title])
+        result << @response[:signature] if @response[:signature]
```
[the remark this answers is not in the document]
aryan-soni: When you mentioned that this would require us to remove the code blocks where we add it, are you essentially suggesting we swap all instances of `title` with `signature`, then remove the additional `signature` instances?

lib/ruby_lsp/listeners/hover.rb
```diff
@@ -43,7 +43,9 @@ def initialize(response_builder, uri, nesting, index, dispatcher, typechecker_en
         )
       end
 
-      sig { params(node: Prism::ConstantReadNode).void }
+      sig do
+        params(node: Prism::ConstantReadNode).void
+      end
```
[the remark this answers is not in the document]
vinistock: That's odd. It shouldn't change the block syntax from `{}` to `do..end` unless the line is too long (which isn't the case here).

There's no difference in terms of behaviour, but for signatures that fit in a single line, we always aim to use the block oneliners with `{}`.

lib/ruby_lsp/response_builders/hover.rb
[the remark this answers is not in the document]
vinistock: I mean changing `title` to `signature`, then removing the "\`\`\`ruby...```" blocks from places like [here](https://github.com/Shopify/ruby-lsp/blob/9159861f94da242fb49383bca3f7ccaa9022df0f/lib/ruby_lsp/requests/support/common.rb#L92) and adding those blocks on this line I commented on.

That way, all content added to `signature` is automatically assumed to be Ruby code and inserted inside the Ruby block.

st0012:
To be honest this is a bit different than I thought (which I didn't explain clearly):

When I mentioned categorised response, I meant something like:

```
{
  "typechecker" => [...],
  "documentation" => ["<markdown string>"],
  "indexer" => [...]
}
```

And the priority would be:

1. typechecker
2. documentation
3. indexer

As a result, the hover content would be something like:

    [from type checker]
    ```rb
    sig do
      returns(String)
    end
    def foo; end
    ```
    ---
    [from indexer]
    ```
    foo
    Definitions: [foo.rb](link)
    ```

The idea is that we make assumption on the different levels of quality different types of tools can have, thus reduce the effect the duplicated content could have by pushing it down.

WDYT?

st0012:
> If add-ons start pushing types of content in an order based on where they came from, this could lead to suboptimal ordering:

In my proposal there will no concept of signature, title, definitions...etc. It's only the type of source that matters, which has a major benefit: We don't need to create our own specs and concepts for hover's content.

For example, we need to tell users that for `links` type we expect markdown links. We also need to maintain this layout ourselves and tell addon authors we expect each section to have what types of elements (e.g. use headers in title but don't use them for links?)

```
Title

Signature

Links

Documentation
```

And speaking of signature, I think the name is a big confusing.

- If we want to keep this PR's implementation, can we rename it to `code` or `ruby_code`?
- Re @vinistock's suggestion on code block: I feel putting signature content inside an already built code block could be quite confusing to addon authors, especially when we expect markdown link but not markdown code blocks.
    - It's also less flexible because we may want to display C code too?

In the contrary, I think `documentation`, `typechecker`, `indexer`, and `framework` (I just added this for `ruby-lsp-rails`) are easier to grasp. Making addon authors generate complete markdown string should also save both our maintenance work and theirs.

That said, I think with the current number of addons, the result of either approach may not be hugely different. And we can make another breaking-change later as well. So I'm fine going with this approach first too.

vinistock:
That's not super readable though. The Steep signature would be far apart from the method declaration it relates to (especially if there's a lot of documentation). I really believe we should favour readability in the trade off here.

Consider this other example: a type checker addon can accurately detect the right method being invoked, so it wants to contribute hover documentation. It would also push links and declarations way too far below to be useful. E.g.: there are many examples in Rails with [this level of documentation](https://github.com/rails/rails/blob/f0d433bb46ac233ec7fd7fae48f458978908d905/activesupport/lib/active_support/callbacks.rb#L12) (you'd need to scroll a lot to get to the part that shows the declaration and definition links).

Is the concern what we discover new categories we didn't anticipate?

vinistock:
Sure, let's do that.

test/requests/hover_expectations_test.rb
```diff
@@ -247,9 +247,12 @@ class Post
       })
 
       assert_nil(response.error, response.error&.full_message)
-      assert_match(<<~RESPONSE.strip, response.response.contents.value)
+      assert_equal(<<~RESPONSE.strip, response.response.contents.value)
+        ```ruby
+        Post
         Title
         Signature
+        ```
```
aryan-soni, line 255: Because of the restructuring, the test was complaining if "```" was not embedded. I decided to restructure the test to cover the whole hover.

Overall, this test isn't as comprehensive as the hardcoded fixtures anyways, and those have not been changed

lib/ruby_lsp/requests/support/common.rb
```diff
@@ -125,7 +124,7 @@ def markdown_from_index_entries(title, entries)
           categorized_markdown = categorized_markdown_from_index_entries(title, entries)
 
           <<~MARKDOWN.chomp
-            #{categorized_markdown[:title]}
+            #{"```ruby\n#{categorized_markdown[:code]}\n```"}
```
vinistock, line 127: I think this is fine, but the style is a little odd indeed. Can we do this instead?

```suggestion
            ```ruby
            #{categorized_markdown[:code]}
            ```
```

lib/ruby_lsp/listeners/hover.rb
```diff
@@ -123,7 +123,7 @@ def generate_gem_hover(node)
         # Remove leading whitespace if a heredoc was used for the summary or description
         info = info.gsub(/^ +/, "")
 
-        @response_builder.push("**#{spec.name}** (#{spec.version})", :title)
+        @response_builder.push("**#{spec.name}** (#{spec.version})", :code)
```
vinistock, line 126: Here's a good example of how our previous decision doesn't work. The gem's name and version isn't really code. And putting it inside a code block messes up the markdown styling.

Sorry for the trouble, but I think we need to revert this into `title` and not add the code block every time.

aryan-soni: I see your point.

I can undo my latest commit (to keep the 2 categories separate). Or are you suggesting something else?

aryan-soni:
Commit history is getting messy here, so I'm closing in favour of #1368

PR #1361, opened 2024-02-07
Update Slack invite link

andyw8:
This one has no expiry.


vinistock (approved):
We need to do the same in `vscode-ruby-lsp`

andyw8:
Will update vscode-ruby-lsp.

PR #1364, opened 2024-02-07
Fix stability of indexing test

andyw8:
This was occasionally failing, for example [here](https://github.com/Shopify/ruby-lsp/actions/runs/7820959228/job/21336844351):

```
Failure:
ExecutorTest#test_initialized_recovers_from_indexing_failures [/Users/runner/work/ruby-lsp/ruby-lsp/test/executor_test.rb:149]
Minitest::Assertion: not all expectations were satisfied
unsatisfied expectations:
- expected exactly once, invoked twice: #<AnyInstance:RubyIndexer::Index>.index_all(any_parameters)
```
It seems like a race condition which may be tricky to track down, but looking at the test I don't think there's any need for this to be an expectation?


test/executor_test.rb
```diff
@@ -146,7 +146,7 @@ def test_initialized_populates_index
   end
 
   def test_initialized_recovers_from_indexing_failures
-    RubyIndexer::Index.any_instance.expects(:index_all).once.raises(StandardError, "boom!")
```
vinistock, line 149: The idea of the `once` was to ensure that we weren't trying to index the codebase more than once in the `initialized` notification. I suppose the problem is that the thread switches and the stubbing is not thread safe.

Can we stub on the specific instance of the index instead? So that we continue to ensure it never gets indexed twice? Something like

```ruby
@executor.instance_variable_get(:@index).expects(:index_all).once.raises(StandardError, "boom!")
```

andyw8: Yeah we can do that, updated.

vinistock (approved):

PR #1365, opened 2024-02-07
Always print error backtraces

vinistock:
### Motivation

If an error occurs in the server, it's difficult for users to report it. We should always print the backtrace information so that we can more easily diagnose errors.

### Implementation

Started warning the backtrace.

I also removed the `$VERBOSE = nil` from our tests. I understand the idea to remove noise from tests, since most of the warnings come from RuboCop and not us, but we need to be able to assert things against stderr, which is the only pipe we can print to. If verbose is off, the unit test I added can't pass.

### Automated Tests

Added a test to ensure the backtrace is present.

### DIFF
--- a/lib/ruby_lsp/executor.rb
+++ b/lib/ruby_lsp/executor.rb
@@ -25,7 +25,7 @@ module RubyLsp
       begin
         response = run(request)
       rescue StandardError, LoadError => e
-        warn(e.message)
+        $stderr.puts(e.full_message)
         error = e
       end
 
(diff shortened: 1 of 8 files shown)


lib/ruby_lsp/executor.rb
```diff
@@ -26,6 +26,8 @@ def execute(request)
         response = run(request)
       rescue StandardError, LoadError => e
         warn(e.message)
+        backtrace = e.backtrace
+        warn(backtrace.join("\n")) if backtrace
```
paracycle, line 30: 2 things here:

1. By using `Kernel#warn` we are making it so that `$VERBOSE = nil` or setting the verbosity level of the running Ruby interpreter will prevent these from going into the Output tab. I don't think that's what we want, so maybe we should explicitly start writing to `$stderr`.
2. I think `e.full_message` should output the message and the backtrace exactly as Ruby would have reported it: https://ruby-doc.org/3.0.3/Exception.html#full_message-method If we use that, then we don't need to worry about printing these things separately.

PR #1367, opened 2024-02-08
Add `RUBY_LSP_BACKTRACE` option for debugging

andyw8:
### Motivation

While working on an addon, I was encountering a Sorbet type error. Since `TypeError` inherits from `StandardError`, it was being rescued, meaning I couldn't see the full backtrace.

### Implementation

I've added an option to run with `RUBY_LSP_BACKTRACE`. (I originally considered just `BACKTRACE` but didn't want to risk clashing with other tooling).

Interestingly, this has caught a existing failure in the tests that was failing silently. (I can look into the fix for that).

### Automated Tests

Update to always run with `RUBY_LSP_BACKTRACE` in CI, since it can help to catch issues.

### Manual Tests

n/a

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -45,7 +45,7 @@ jobs:
           cache-version: 1
 
       - name: Run tests
-        run: bundle exec rake
+        run: RUBY_LSP_BACKTRACE=1 bundle exec rake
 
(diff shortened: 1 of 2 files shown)


andyw8:
Haven't felt any need for this recently so closing for now.

PR #1368, opened 2024-02-08
Categorize hover content

aryan-soni:
### Motivation

Enables hover content to be displayed deterministically, based on categories. Add-ons that add content to hovers can specify the category of the content, then the Ruby LSP can render the content in hovers in an order that semantically maps to the category specified.

### Implementation

Updated the response builder to require categorization, with 3 possible categories for hover content (title, links and documentation).

The response is now stored via a hash, instead of an array. The hash contains the categories as its keys, and the associated content as its values.

In `common.rb`, a helper method has been extracted to map each piece of content with its category. This helper is then used when pushing to the response builder.

### Automated Tests

The `hover_expectations_test` has been updated based on these changes. Note that the expectation tests associated with metaprogramming that are used have not been updated, since our expectation of what we expect to be rendered is the same.

### Manual Tests

The goal here is to ensure we aren't disrupting the hover behaviour. Below is an example of the hover functionality in its current state: 

hover_before_changes

This is the hover functionality with the changes in this PR made:

hover_after_change

The hover remains unchanged. This is because the LSP categorizes everything as expected and has not modified the visual output of the hover itself.

### DIFF
--- a/lib/ruby_lsp/response_builders/hover.rb
+++ b/lib/ruby_lsp/response_builders/hover.rb
@@ -12,27 +12,37 @@ module RubyLsp
       sig { void }
       def initialize
         super
-        @stack = T.let(
-          [],
-          T::Array[String],
+
+        @response = T.let(
+          {
+            title: +"",
+            links: +"",
+            documentation: +"",
+          },
+          T::Hash[Symbol, String],
         )
       end
 
-      sig { params(hover_response: String).void }
-      def push(hover_response)
-        @stack << hover_response
+      sig { params(content: String, category: Symbol).void }
+      def push(content, category:)
+        hover_content = @response[category]
+        if hover_content
+          hover_content << content + "\n"
+        end
       end
-
-      alias_method(:<<, :push)
 
       sig { returns(T::Boolean) }
       def empty?
-        @stack.empty?
+        @response.values.all?(&:empty?)
       end
 
-      sig { override.returns(String) }
+      sig { override.returns(ResponseType) }
       def response
-        @stack.join("\n\n")
+        result = T.must(@response[:title])
+        result << "\n" << @response[:links] if @response[:links]
+        result << "\n" << @response[:documentation] if @response[:documentation]
+
+        result.strip
       end
     end
   end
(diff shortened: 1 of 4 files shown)


lib/ruby_lsp/response_builders/hover.rb
```diff
@@ -12,27 +12,37 @@ class Hover < ResponseBuilder
...
 
-      sig { params(hover_response: String).void }
-      def push(hover_response)
-        @stack << hover_response
+      sig { params(content: String, category: Symbol).void }
+      def push(content, category)
+        hover_content = @response[category]
+        if hover_content
+          hover_content << content + "\n"
+        end
```
st0012, line 31: Should we raise if developers try to push to an not supported category?

vinistock: I would vote for just ignoring it so that addons don't end up breaking hover.

lib/ruby_lsp/response_builders/hover.rb
```diff
@@ -12,27 +12,37 @@ class Hover < ResponseBuilder
...
+          },
+          T::Hash[Symbol, String],
         )
       end
 
-      sig { params(hover_response: String).void }
-      def push(hover_response)
-        @stack << hover_response
+      sig { params(content: String, category: Symbol).void }
+      def push(content, category)
```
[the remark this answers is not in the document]
vinistock: We can also rename the method if you think that'd be clearer. Maybe `add_content`?

vinistock (approved):

lib/ruby_lsp/response_builders/hover.rb
aryan-soni: Yeah, we ultimately don't want the core Ruby LSP hover to be vulnerable to disruption by addons.

lib/ruby_lsp/response_builders/hover.rb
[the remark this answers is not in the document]
aryan-soni: I'm leaning towards keeping it as `push` just to maintain some level of consistency with the other responses. 

Will add the keyword argument though

PR #1369, opened 2024-02-08
Handle nested method calls in signature help without adjusting positions

vinistock:
### Motivation

Related to https://github.com/Shopify/vscode-ruby-lsp/issues/1025 and https://github.com/Shopify/vscode-ruby-lsp/issues/1021

We've been able to determine that something in signature help is causing the LSP to get stuck and spike in CPU. I haven't been able to reproduce this myself, but looking at our code for signature help the only thing that could get us into that situation is the `locate_node` invocation.

If we try to find a position that doesn't exist, we may get stuck in [this loop](https://github.com/Shopify/ruby-lsp/blob/8c990dd9582bbf6262cf56669c87ae10cb2fb914/lib/ruby_lsp/document.rb#L207), which would explain the server getting stuck and spiking in CPU. I can't think of anything else that would get the server stuck.

This PR removes the position adjustment we had, so that we always use the position we get from the editor, to try to avoid this infinite loop issue.

### Implementation

The main issue to be solved is handling signature help for nested method calls properly. We need to understand for which method to display the help for.

Instead of relying on the position and the previous checks, I reworked the code to check for nested method invocations.

### Automated Tests

Added two new tests.

### DIFF
--- a/lib/ruby_lsp/requests/signature_help.rb
+++ b/lib/ruby_lsp/requests/signature_help.rb
@@ -50,22 +50,12 @@ module RubyLsp
       end
       def initialize(document, index, position, context, dispatcher)
         super()
-        current_signature = context && context[:activeSignatureHelp]
         target, parent, nesting = document.locate_node(
-          { line: position[:line], character: position[:character] - 2 },
+          { line: position[:line], character: position[:character] },
           node_types: [Prism::CallNode],
         )
 
-        # If we're typing a nested method call (e.g.: `foo(bar)`), then we may end up locating `bar` as the target
-        # method call incorrectly. To correct that, we check if there's an active signature with the same name as the
-        # parent node and then replace the target
-        if current_signature && parent.is_a?(Prism::CallNode)
-          active_signature = current_signature[:activeSignature] || 0
-
-          if current_signature.dig(:signatures, active_signature, :label)&.start_with?(parent.message)
-            target = parent
-          end
-        end
+        target = adjust_for_nested_target(target, parent, position)
 
         @target = T.let(target, T.nilable(Prism::Node))
         @dispatcher = dispatcher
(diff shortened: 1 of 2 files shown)


lib/ruby_lsp/requests/signature_help.rb
```diff
@@ -50,22 +51,12 @@ def provider
       end
       def initialize(document, index, position, context, dispatcher)
         super()
-        current_signature = context && context[:activeSignatureHelp]
         target, parent, nesting = document.locate_node(
-          { line: position[:line], character: position[:character] - 2 },
+          { line: position[:line], character: position[:character] },
```
st0012, line 55: Just curious: what was the original `-2` for and why don't we need it anymore?

lib/ruby_lsp/requests/support/common.rb
```diff
@@ -46,6 +46,23 @@ def visible?(node, range)
           range.cover?(loc.start_line - 1) && range.cover?(loc.end_line - 1)
         end
 
+        sig { params(node: Prism::Node, position: T::Hash[Symbol, T.untyped]).returns(T::Boolean) }
+        def node_covers?(node, position)
```
st0012, line 50: Do we need to expose it now? If we only have signature help using it at the moment, I think it can be a private method there instead.

st0012 (commented):
Not related to this PR, but it looks like we don't have integration tests for signature help? 🤔

lib/ruby_lsp/requests/signature_help.rb
vinistock: It was basically trying to point the location at the method call when it was pointing to the empty space inside the parenthesis.

```ruby
foo( )
#   ^ if it pointed here
# ^ we brought it back here to find the call node
```

With the nested checks, we shouldn't need it anymore.

st0012 (approved):

PR #1377, opened 2024-02-12
Add note about finding other community addons

### LINKED ISSUE #1373: Add Shoulda context addon in addons docs

### Use case

Hi!, As you mentioned in [ruby-lsp-rails #244](https://github.com/Shopify/ruby-lsp-rails/issues/244) i wanted to have support for shoulda context in my tests. You said that it should be better to have an addon, so here you have `ruby-lsp-shoulda-context` 👍🏻 .

## :link: [Ruby Lsp Shoulda Context](https://github.com/domingo2000/ruby-lsp-shoulda-context)
The addon permits to run, run in terminal and debug [Shoulda Context](https://github.com/thoughtbot/shoulda-context) like DSL syntax for tests. This improves a lot the development experience when working with this kind of test setups in plain ruby projects or in rails.

It in early stage of development, but it support what i think are a lot of use cases, so it would be nice if you give it a try, and if  you like, you could add it to the ADDONS readme 😄.

### Description

The addon currently parse the contexts blocks and should blocks similarly as Rspec addon does. It reads the workspace to see if its using pure ruby, bundler or rails to run the commands in the correct way. You can see the currently implemented features in the repository.

### Implementation

Just add the addon in your Gemfile :rocket:

Note: The only inconvinience i encounter implementing the addon is that if i need to parse the main class to, i can't access to the previously parsed TestItem from the original `ruby-lsp` so the `CodeLens` got duplicated, but is just a detail.

vinistock:
### Motivation

Closes #1373

Add a link to rubygems to help developers find other Ruby LSP addons.

### DIFF
--- a/README.md
+++ b/README.md
@@ -85,6 +85,9 @@ features. This is the mechanism that powers addons like
 - [Ruby LSP RSpec](https://github.com/st0012/ruby-lsp-rspec)
 - [Ruby LSP rubyfmt](https://github.com/jscharf/ruby-lsp-rubyfmt)
 
+Other community driven addons can be found in [rubygems](https://rubygems.org/search?query=name%3A+ruby-lsp) by
+searching for the `ruby-lsp` prefix.
+
 For instructions on how to create addons, see the [addons documentation](ADDONS.md).
 
 ## Learn More


KaanOzkan (approved):
Results aren't perfect but I think best we can do without changing the search functionality.

Issue #1343, opened 2024-01-31
Handle same method name in multiple files


gottlike:
### Ruby version

3.2.2

### Description

In a project I have multiple folders with basically the same functionality (very modular), so I could have `def do_something` in `folder-one/script.rb` and `folder-two/script.rb`. However, the method description could differ.

Currently Ruby LSP doesn't necessarily display the description of the current one on hover. Also when clicking, it might open another file.

### Expected output

I would expect Ruby LSP to show/use the method definition of the current file only, when there's duplicate methods in the project.

gottlike:
Thanks for your reply. Sure, here is a simple example:

`test1.rb`
```ruby
# frozen_string_literal: true

module Func
  def start
    something('great')
  end

  # Does something interesting
  def something(input)
    puts input
  rescue StandardError => e
    raise("#{name}.#{__method__} > #{e.message}")
  end
end
```

`test2.rb`
```ruby
# frozen_string_literal: true

module Func
  # Does something else entirely
  def something(input)
    puts input
  rescue StandardError => e
    raise("#{name}.#{__method__} > #{e.message}")
  end
end
```

![image](https://github.com/Shopify/ruby-lsp/assets/278766/663bb675-6295-415f-9a7d-fd86e396ef5f)

vinistock:
Thank you for the context, I understand the problem now.

There's indeed a bug there, we should be concatenating the documentation for all definitions of `Fun#something`, since it's being overridden in multiple places.

However, notice that the declarations are not considered file specific. All declarations inside the same workspace are indexed and will show up in the editor.

gottlike:
Well, if both are shown, that's still fine, I guess :). Thanks!

PR #1380, opened 2024-02-13
Fix handle resolve multiple methods

snutij:
### Motivation

Closes: https://github.com/Shopify/ruby-lsp/issues/1343

### Implementation

When we invoke `resolve_method`, do not pick only the first match, bot return an array containing all matches.
- replace `find` by `select`
- return an empty array when nothing match (**feel free to challenge this point**, as the interface change)

Handle in all requests using this resolve following [advices](https://github.com/Shopify/ruby-lsp/issues/1343):
- merge documentation for hover
- push all `Location` for definition 
- merge documentation for signature help

**Question:** On `signature_help` request, we could propose multiple [signatures](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#signatureHelp), with for each [signature information](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#signatureInformation) the related documentation and parameters. I did a try, it changes the interface, to allow user changing signature with arrow. You can find an example for VS Code documentation [here](https://code.visualstudio.com/docs/languages/javascript#_signature-help). Do you think it can be better or bring some unwanted complexity? Idk how it's handled by other IDE.

### Automated Tests

Added tests in all request and indexer, again, if missed some useful cases, just say it 👍 

### Manual Tests

Given this code: 

```
# file 1
module Func
  def start
    something()
  end

  # Does something interesting
  def something(argument_name)
    puts argument_name
  end
end

# file 2
module Func
  # Does something interesting
  def something(argument_with_other_name)
    puts argument_with_other_name
  end
end
```

For hover on `something`: 
image

For go to definition on `something`:
image

For signature_help on `something`:
image

### DIFF
--- a/lib/ruby_indexer/lib/ruby_indexer/index.rb
+++ b/lib/ruby_indexer/lib/ruby_indexer/index.rb
@@ -237,20 +237,20 @@ module RubyIndexer
       real_parts.join("::")
     end
 
-    # Attempts to find a given method for a resolved fully qualified receiver name. Returns `nil` if the method does not
-    # exist on that receiver
-    sig { params(method_name: String, receiver_name: String).returns(T.nilable(Entry::Member)) }
+    # Attempts to find methods for a resolved fully qualified receiver name.
+    # Returns an empty array if the method does not exist on that receiver
+    sig { params(method_name: String, receiver_name: String).returns(T::Array[Entry::Member]) }
     def resolve_method(method_name, receiver_name)
       method_entries = self[method_name]
       owner_entries = self[receiver_name]
-      return unless owner_entries && method_entries
+      return [] unless owner_entries && method_entries
 
       owner_name = T.must(owner_entries.first).name
       T.cast(
-        method_entries.grep(Entry::Member).find do |entry|
+        method_entries.grep(Entry::Member).select do |entry|
           T.cast(entry, Entry::Member).owner&.name == owner_name
         end,
-        T.nilable(Entry::Member),
+        T::Array[Entry::Member],
       )
     end
 
--- a/lib/ruby_indexer/test/index_test.rb
+++ b/lib/ruby_indexer/test/index_test.rb
@@ -226,9 +226,9 @@ module RubyIndexer
         end
       RUBY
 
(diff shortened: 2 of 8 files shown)


lib/ruby_indexer/lib/ruby_indexer/index.rb
```diff
@@ -237,20 +237,20 @@ def follow_aliased_namespace(name)
...
-    # exist on that receiver
-    sig { params(method_name: String, receiver_name: String).returns(T.nilable(Entry::Member)) }
+    # Attempts to find methods for a resolved fully qualified receiver name.
+    # Returns an empty array if the method does not exist on that receiver
+    sig { params(method_name: String, receiver_name: String).returns(T::Array[Entry::Member]) }
     def resolve_method(method_name, receiver_name)
       method_entries = self[method_name]
       owner_entries = self[receiver_name]
-      return unless owner_entries && method_entries
+      return [] unless owner_entries && method_entries
```
vinistock, line 246: I think I would keep the nilable return. The reasoning being that Sorbet will be able to tell us the right type after an early return ([example](https://sorbet.run/#%23%20typed%3A%20true%0Aextend%20T%3A%3ASig%0A%0Asig%20%7Breturns%28T.nilable%28T%3A%3AArray%5BInteger%5D%29%29%7D%0Adef%20foo%0Aend%0A%0Ar%20%3D%20foo%0AT.reveal_type%28r%29%0Areturn%20unless%20r%0A%0AT.reveal_type%28r%29)).

With arrays only, we would need to remember to check if they are empty everywhere.

snutij: TIL. I'm not so familiar with Sorbet. 👍  

Let me know if it's OK for you.

vinistock (approved):
Awesome! 🚀

lib/ruby_indexer/lib/ruby_indexer/index.rb
vinistock: It looks great!

PR #1381, opened 2024-02-14
Automatically install `ruby-lsp-rails` as part of custom bundle

andyw8:
### Motivation

With `ruby-lsp-rails` v0.3.0, we switched away from the mounted Rack application approach. This means `ruby-lsp-rails` can now be used without having to add it to the app's Gemfile.

Also closes #1267

### Implementation

Previously we would install `debug` and `ruby-lsp` into the custom bundle. We now also install `ruby-lsp-rails` if the app uses Rails.

### Automated Tests

Updated

### Manual Tests

Verified.

### DIFF
--- a/lib/ruby_lsp/setup_bundler.rb
+++ b/lib/ruby_lsp/setup_bundler.rb
@@ -57,15 +57,18 @@ module RubyLsp
     def setup!
       raise BundleNotLocked if @gemfile&.exist? && !@lockfile&.exist?
 
-      # Do not setup a custom bundle if both `ruby-lsp` and `debug` are already in the Gemfile
-      if @dependencies["ruby-lsp"] && @dependencies["debug"]
+      # Do not set up a custom bundle if LSP dependencies are already in the Gemfile
+      if @dependencies["ruby-lsp"] &&
+          @dependencies["debug"] &&
+          (@dependencies["rails"] ? @dependencies["ruby-lsp-rails"] : true)
         $stderr.puts(
-          "Ruby LSP> Skipping custom bundle setup since both `ruby-lsp` and `debug` are already in #{@gemfile}",
+          "Ruby LSP> Skipping custom bundle setup since LSP dependencies are already in #{@gemfile}",
         )
 
-        # If the user decided to add the `ruby-lsp` and `debug` to their Gemfile after having already run the Ruby LSP,
-        # then we need to remove the `.ruby-lsp` folder, otherwise we will run `bundle install` for the top level and
-        # try to execute the Ruby LSP using the custom bundle, which will fail since the gems are not installed there
+        # If the user decided to add `ruby-lsp` and `debug` (and potentially `ruby-lsp-rails`) to their Gemfile after
+        # having already run the Ruby LSP, then we need to remove the `.ruby-lsp` folder, otherwise we will run `bundle
+        # install` for the top level and try to execute the Ruby LSP using the custom bundle, which will fail since the
+        # gems are not installed there
         @custom_dir.rmtree if @custom_dir.exist?
         return run_bundle_install
       end
@@ -143,6 +146,10 @@ module RubyLsp
         parts << 'gem "debug", require: false, group: :development, platforms: :mri'
       end
 
+      if @dependencies["rails"] && !@dependencies["ruby-lsp-rails"]
+        parts << 'gem "ruby-lsp-rails", require: false, group: :development'
+      end
+
       content = parts.join("\n")
       @custom_gemfile.write(content) unless @custom_gemfile.exist? && @custom_gemfile.read == content
     end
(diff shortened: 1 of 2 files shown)


vinistock (approved):
Let's write a PR description, but the changes are good

andyw8:
@vinistock I think we need a few more changes, e.g. https://github.com/Shopify/ruby-lsp/blob/c1ae8fdcdb238d128a519fdde02741a011e47c74/lib/ruby_lsp/setup_bundler.rb#L226

vinistock:
Ah, yeah. True. We need to add `ruby-lsp-rails` to that check.

andyw8:
(once this is merged, we can update the ruby-lsp-rails README to indicate that it doesn't need to be added to the Gemfile).

lib/ruby_lsp/setup_bundler.rb
```diff
@@ -143,6 +148,12 @@ def write_custom_gemfile
         parts << 'gem "debug", require: false, group: :development, platforms: :mri'
       end
 
+      unless @dependencies["ruby-lsp-rails"]
+        if @dependencies["rails"]
+          parts << 'gem "ruby-lsp-rails", require: false, group: :development'
+        end
+      end
```
vinistock, line 155:
```suggestion
      if @dependencies["rails"] && !@dependencies["ruby-lsp-rails"]
        parts << 'gem "ruby-lsp-rails", require: false, group: :development'
      end
```

lib/ruby_lsp/setup_bundler.rb
```diff
@@ -183,22 +194,26 @@ def run_bundle_install(bundle_gemfile = @gemfile)
...
-        # If ruby-lsp or debug are not in the Gemfile, try to update them to the latest version
+        # If any of `ruby-lsp`, `ruby-lsp-rails` or `debug` are not in the Gemfile, try to update them to the latest
+        # version
         command.prepend("(")
         command << " && bundle update "
         command << "ruby-lsp " unless @dependencies["ruby-lsp"]
         command << "debug " unless @dependencies["debug"]
+        if @dependencies["rails"]
+          command << "ruby-lsp-rails " unless @dependencies["ruby-lsp-rails"]
+        end
```
vinistock, line 216: Not sure if the line fits, but let's at least merge the conditionals.
```suggestion
          command << "ruby-lsp-rails " if @dependencies["rails"] && !@dependencies["ruby-lsp-rails"]
```

lib/ruby_lsp/setup_bundler.rb
```diff
@@ -221,13 +236,13 @@ def run_bundle_install(bundle_gemfile = @gemfile)
...
+      # If `ruby-lsp`, `ruby-lsp-rails` and `debug` are in the Gemfile, then we shouldn't try to upgrade them or else it
+      # will produce version control changes
+      return false if @dependencies.values_at("ruby-lsp", "ruby-lsp-rails", "debug").all?
 
-      # If the custom lockfile doesn't include either the `ruby-lsp` or `debug`, we need to run bundle install before
-      # updating
-      return false if custom_bundle_dependencies["ruby-lsp"].nil? || custom_bundle_dependencies["debug"].nil?
+      # If the custom lockfile doesn't include `ruby-lsp`, `ruby-lsp-rails` or `debug`, we need to run bundle install
+      # before updating
+      return false if custom_bundle_dependencies.values_at("ruby-lsp", "ruby-lsp-rails", "debug").all?(&:nil?)
```
vinistock, line 245: For this one, we want to check if any of the custom bundle dependencies are missing. This change is checking if all of them are missing, which would make us not run bundle install if only one of them is missing.
```suggestion
      return false if custom_bundle_dependencies["ruby-lsp"].nil? || custom_bundle_dependencies["debug"].nil? || (@dependencies["rails"] && custom_bundle_dependencies["ruby-lsp-rails"].nil?)
```

test/setup_bundler_test.rb
```diff
@@ -22,17 +23,45 @@ def test_removes_ruby_lsp_folder_if_both_gems_were_added_to_the_bundle
     FileUtils.rm_r(".ruby-lsp") if Dir.exist?(".ruby-lsp")
   end
 
+  def test_in_a_rails_app_removes_ruby_lsp_folder_if_all_gems_were_added_to_the_bundle_RAILS
```
vinistock, line 26:
```suggestion
  def test_in_a_rails_app_removes_ruby_lsp_folder_if_all_gems_were_added_to_the_bundle
```

test/setup_bundler_test.rb
```diff
@@ -186,6 +215,7 @@ def test_creates_custom_bundle_if_no_gemfile
         assert_path_exists(".ruby-lsp")
         assert_path_exists(".ruby-lsp/Gemfile")
         assert_match("ruby-lsp", File.read(".ruby-lsp/Gemfile"))
+        # assert_match("ruby-lsp-rails", File.read(".ruby-lsp/Gemfile"))
```
vinistock, line 218: This shouldn't be in the custom bundle if there's no top level Gemfile (because Rails has no way to be there).
✂️

vinistock (commented):
Let's add one test for the update case (which would've caught the logic issue).

lib/ruby_lsp/setup_bundler.rb
andyw8: For this one, I think's actually clear to branch on rails/non-rails, even though there's a bit of duplication (updated pushed).

lib/ruby_lsp/setup_bundler.rb
```diff
@@ -143,6 +147,11 @@ def write_custom_gemfile
         parts << 'gem "debug", require: false, group: :development, platforms: :mri'
       end
 
+      # No `require: false` for ruby-lsp-rails since we need to load the Railtie.
```
vinistock, line 150: This shouldn't matter for the custom bundle. The railtie is only loaded if `ruby-lsp-rails` is a part of the main Gemfile.

vinistock (approved):

PR #1382, opened 2024-02-14
Update setup_bundler.rb

B0rrA:
### Motivation

[https://github.com/Shopify/ruby-lsp/issues/1675](https://github.com/Shopify/ruby-lsp/issues/1675
)
### Implementation

You are using `1>&2` only. This type of redirection is more a tool for managing output rather than a way to silence it entirely. If your goal is to avoid seeing the output at all, redirecting to `/dev/null` (or to a file if you wish to review the output later) would be the correct approach.

### Automated Tests

minor change

### DIFF
--- a/lib/ruby_lsp/setup_bundler.rb
+++ b/lib/ruby_lsp/setup_bundler.rb
@@ -208,6 +208,9 @@ module RubyLsp
 
       command << " || bundle install) "
 
+       # Safely silence output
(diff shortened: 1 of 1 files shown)


paracycle:
But the goal of that `1>&2` is NOT to silence output, instead it sends all output to stderr, so that the output of that command gets logged to the `Output` tab in VSCode.

Issue #1333, opened 2024-01-29
Start keeping track of ancestors


vinistock:
We currently don't keep track of ancestors for modules or classes. We need to start doing it, so that we can properly discover inherited methods or constants.

- [x] Start tracking parent classes https://github.com/Shopify/ruby-lsp/pull/1088
- [x] Start tracking included modules #1383 
- [x] Start tracking extended modules (depends on better representing singleton classes)
- [x] Start tracking prepended modules #1820 
- [x] Provide a way to lazily linearize the ancestors on a given namespace
- [x] Ensure modifications that invalidate ancestor tree clears cache

PR #1383, opened 2024-02-15
Indexed included modules on namespaces

domingo2000:
### Motivation

Resolves second point of #1333 

This changes index included modules of all Namespaces, including `Class` and `Module` nodes.

### Implementation

Each time a class or module is visited by the collector, we instantiate a new visitor that only visits its inmediate nodes (depth 1 of the tree), looking for `CallNode` that match the `include` method. The collector has an index of all the included modules to keep track of all the modules included for a singe entry. Then pass this values to the `Namespaces` nodes. Because its passed by reference and because it's a `Set` data structure, the modules are propagated and not repeated.

### Automated Tests

Tests where added in `lib/ruby_indexer/test/classes_and_modules_test.rb` to cover different cases.

### Manual Tests

Test indexing the same repo and does not crash.

### DIFF
--- a/lib/ruby_indexer/lib/ruby_indexer/collector.rb
+++ b/lib/ruby_indexer/lib/ruby_indexer/collector.rb
@@ -152,6 +152,8 @@ module RubyIndexer
         handle_attribute(node, reader: false, writer: true)
       when :attr_accessor
         handle_attribute(node, reader: true, writer: true)
+      when :include
+        handle_include(node)
       end
     end
 
@@ -349,6 +351,25 @@ module RubyIndexer
         @index << Entry::Accessor.new(name, @file_path, loc, comments, @current_owner) if reader
         @index << Entry::Accessor.new("#{name}=", @file_path, loc, comments, @current_owner) if writer
       end
+    end
+
+    sig { params(node: Prism::CallNode).void }
+    def handle_include(node)
+      return unless @current_owner
+
+      arguments = node.arguments&.arguments
+      return unless arguments
+
+      names = arguments.filter_map do |node|
+        if node.is_a?(Prism::ConstantReadNode) || node.is_a?(Prism::ConstantPathNode)
+          node.full_name
+        end
+      rescue Prism::ConstantPathNode::DynamicPartsInConstantPathError
+        # TO DO: add MissingNodesInConstantPathError when released in Prism
+        # If a constant path reference is dynamic or missing parts, we can't
+        # index it
+      end
+      @current_owner.included_modules.concat(names)
     end
   end
 end
--- a/lib/ruby_indexer/lib/ruby_indexer/entry.rb
+++ b/lib/ruby_indexer/lib/ruby_indexer/entry.rb
@@ -40,6 +40,22 @@ module RubyIndexer
 
       abstract!
 
+      sig { returns(T::Array[String]) }
+      attr_accessor :included_modules
+
+      sig do
+        params(
+          name: String,
+          file_path: String,
+          location: Prism::Location,
+          comments: T::Array[String],
+        ).void
+      end
+      def initialize(name, file_path, location, comments)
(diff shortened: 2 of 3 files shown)


domingo2000:
Also a `include` method could be called from different redefinitions of the class, for example if i write

```
module Foo
end

class Bar
   include Foo
end

class Bar
   include Foo
end
```

The ancestor array will contain the `Foo` entry only once and not twice. That's why i thought a Map data structure is more suitable than an array because the way ruby add included Modules to the ancestors chain

vinistock:
You're right in your assessment. However, we need to limit what we support since Ruby allows you to do crazy stuff that's not easy to detect statically.

For the case of overriding `include`, we'll simply not support that. Overriding such fundamental parts of the language, like `include`, `extend`, `public`, `private` can lead to a lot of headache and it's frankly not good practice.

For the case of dynamic inclusion, like your example of including something from a method, that's simply not possible to support. Imagine you have this scenario

```ruby
module Bar
  def baz; end
end

class Foo
  def dymamic_include
    include Bar
  end
end

foo = Foo.new
foo.baz # This invocation of `baz` is invalid and leads to a `NoMethodError`
foo.dynamic_include
foo.baz # This invocation of `baz` is valid and points to `Bar#baz`
```

This is very difficult to support without a type system. Even with a type system, it would still be really hard to do this in a performant way. Imagine that you could have a much more complicated structure than that one, including multiple ancestors, conditionals where the module is only included in one branch, etc. It becomes too complex to support for little benefit.

We'll move forward with only supporting the non dynamic includes and considering that they are not overridden.

domingo2000:
Ok, with the simplification you provide i can simplify the implementation to support just simple includes to the ancestors chain. I will be working on that if you like.

domingo2000:
@vinistock There is the implementation we talked about, also added a inner test class to check that the `@current_owner` works correctly in the inner classes when using the collector. I will be looking for any comments.

lib/ruby_indexer/lib/ruby_indexer/entry.rb
```diff
@@ -44,6 +44,22 @@ class Namespace < Entry
...
+          name: String,
+          file_path: String,
+          location: Prism::Location,
+          comments: T::Array[String],
+        ).void
+      end
+      def initialize(name, file_path, location, comments)
+        super(name, file_path, location, comments)
+        @included_modules = T.let([], T::Array[String])
+      end
```
vinistock, line 62: Let's move this above the declaration of `short_name` and below `abstract!`.

domingo2000:
Thanks for the feedback!, Just pushed the corrections and added support for ConstantPathNode excluding dynamic Constant Path Node references as you sugested.

lib/ruby_indexer/lib/ruby_indexer/collector.rb
```diff
@@ -350,5 +352,24 @@ def handle_attribute(node, reader:, writer:)
...
+    sig { params(node: Prism::CallNode).void }
+    def handle_include(node)
+      return unless @current_owner
+
+      arguments = node.arguments&.arguments
+      return unless arguments
+
+      names = arguments.filter_map do |node|
+        if node.is_a?(Prism::ConstantReadNode) || node.is_a?(Prism::ConstantPathNode)
+          T.let(node.full_name, String)
```
vinistock, line 365: Do you need this `T.let` here?

vinistock (approved):
Amazing. Thank you for the contribution! I just left a question on the `T.let` because I think we may not need it. Other than that, we're ready to ship

lib/ruby_indexer/lib/ruby_indexer/collector.rb
vinistock: Oh, it shouldn't be an issue that it is untyped. Don't worry about it. Can you please remove the `T.let` and just leave `node.full_name`?

PR #1384, opened 2024-02-15
Use Snapper generator for RDoc documentation

vinistock:
### Motivation

We need to polish the new Snapper RDoc theme before we can propose it more generally. Let's install RDoc from our source so that we can generate our docs using it and dogfood it. That will help us make incremental improvements.


st0012 (approved):
🚀

PR #1393, opened 2024-02-20
Add SimpleCov

andyw8:
### Motivation

I was trying out SimpleCov a while ago, but I figured we may as well merge the branch since it can sometimes by handy to look at the current coverage. I made it opt-in since the overhead from running tests with coverage enabled is:

**Main:** bundle exec rake  50.53s user 9.65s system 82% cpu 1:12.59 total
**This branch:** bundle exec rake  69.23s user 8.92s system 86% cpu 1:30.70 total

(It writes to `coverage/` which is already in `.gitignore`).


test/fixtures/prism
vinistock, line 1: Can we remove these fixture updates?

test/test_helper.rb
```diff
@@ -1,6 +1,16 @@
 # typed: strict
 # frozen_string_literal: true
 
+require "sorbet-runtime"
```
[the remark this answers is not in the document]
vinistock: We get it when we require `ruby_lsp/internal`. Can we move the `SimpleCov.start` below that require?

Or is the recommendation that this needs to be required before anything else?

test/test_helper.rb
```diff
@@ -1,6 +1,17 @@
...
+
+if ENV["COVERAGE"]
+  require "simplecov"
+
+  SimpleCov.start do
+    T.bind(self, SimpleCov::Configuration)
+    enable_coverage :branch
```
vinistock, line 11: Out of curiosity, what does this do?

test/test_helper.rb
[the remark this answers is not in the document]
st0012: IIRC SimpleCov needs to be required and activated before the target (in this case, `ruby_lsp/internal`) is required. So yeah, we probably need to do this.

test/test_helper.rb
st0012: It checks condition coverage too (versus just line coverage), which is particularly useful for inline conditions like `foo if bar?`

Reference: https://github.com/simplecov-ruby/simplecov?tab=readme-ov-file#branch-coverage-ruby--25

st0012 (commented):
Are we able to use any coverage Saas (e.g. https://about.codecov.io/)? Otherwise, I feel we may underutilise this tool because it takes several steps to see it locally.

st0012 (approved):

PR #1399, opened 2024-02-22
Fix method completion on explicit self

bjarosze:
### Motivation

Method completion on explicit self does not work.

### Implementation

Completion didn't show because text edit range is incorrect. When completing method we have to replace `CallNode#message` text, so we have to calculate range from `#message_loc` instead of node location.

### Automated Tests

Added assertion for text edit range start position.

### Manual Tests

```
      class Foo
        def bar(a, b); end
        def baz(c, d); end

        def process
          self.b
        end
      end
```

Invoke completion on `self.b`


bjarosze:
I have signed the CLA!

Issue #1211, opened 2023-11-24
Add completion for require relative


vinistock:
We currently have completion for regular requires, which search the indexed files based on require paths. We can also provide `require_relative` completion, which would be a nice improvement.

I think the implementation would be something like this:

- First, refactor the current completion so that we're handling `CallNode` instead of `StringNode` https://github.com/Shopify/ruby-lsp/blob/e5a2c89b66769a83e218e717fbeeee009d89ed69/lib/ruby_lsp/requests/completion.rb#L58
- Then, check if the method call is `require` and execute the current implementation
- If it's `require_relative`, then we want to
    - Expand the path given as argument. For example, if we have `require_relative "../something/"`, we need to expand it to the absolute path using `Pathname`
    - After we have the absolute path, we need to list all Ruby files available that path to provide as completion suggestions. Likely doing something along the lines of `Dir.glob("#{absolute_path}/**/*.rb")`

nsurve180:
hi @vinistock , can I work on this issue?

namrata18s:
hi @vinistock , can I work on this issue?

PR #1403, opened 2024-02-22
Add `require_relative` completion

snutij:
### Motivation

Closes: https://github.com/Shopify/ruby-lsp/issues/1211

Implement `require_relative` for completion request.

### Implementation

The path is relative to the file where the request is triggered, so I had to drill down `document.uri` to the listener.

Then following advice:
- remove all `StringNode` code related, as we handle `require` and `require_relative` through `CallNode`
- split logic related to different `CallNode` to dedicated methods to avoid too complex handler method 
- Add related tests, some points to note: 
  - `prefix` wasn't use in `path_completion` helper, so I removed it.
  - to have relative path works, I had to pass `tmpdir` to caller and store the document within this specific dir
  - without that (like in other test cases), we store the file in `/foo/fake.rb` where all files created in `with_file_structure` are in `tmp/very/long/path/foo/`

### Automated Tests

Duplicate all existing `require` tests and adapt them for `require_relative`. Syntax tested:
- `foo/` 
- `./`
- `../`
- `foo/ba` partial_path

### Manual Tests

Feb-22-2024 23-49-28


st0012 (approved):
🎩 and it works well. Thanks so much for the feature and refactor 😄

PR #1414, opened 2024-02-26
Adjust monorepo configurations

vinistock:
### Motivation

Complete explanation in https://github.com/Shopify/ruby-lsp/pull/1407.

### Implementation

My approach for this PR has been applying the minimal changes necessary so that we can work in the new monorepo structure.

There are many things we can do after this is merged, like

- Improving code workspace configurations
- Consolidating more of our documentation
- Migrating our integration tests that use a fake client to use the actual VS Code extension

My proposal is that we get a minimally working setup shipped and then iterate on it.

### Manual tests

1. Open the `lsp.code-workspace` workspace in VS Code
2. Verify you can see both workspaces (gem and VS Code extension) at the same level
3. Verify you can use launch tasks, such as starting the VS Code extension in development mode (don't forget to yarn install!)
4. Basically, do a brief verification of basic development and try to find anything that seems to be breaking the experience. If nothing major stands out, we can probably ship and iterate on improving it


.github/workflows/ci.yml
```diff
@@ -24,7 +24,7 @@ jobs:
 
       - name: Lint Ruby files and RBI shims
         run: bundle exec rubocop
-  build:
+  build_ruby:
```
andyw8, line 27: Not a blocker, but we might want to look at skipping the Ruby tests if only the `vscode` part is touched, and vice versa.

CONTRIBUTING.md
```diff
@@ -91,22 +102,22 @@ class MyRequestExpectationsTest < ExpectationsTestRunner
 end
 ```
 
-## Debugging with VS Code
+## Debugging
```
andyw8, line 105: Why remove the "with" part here? (the instructions below are VS Code specific)

.github/workflows/ci.yml
vinistock: I think we don't want to do that. Part of the advantage of having both in the same repository is that we'll be able to have real integration tests that boot make requests from the extension to the server booted from the same repo.

Running all tests will help us catch changes that might break the integration.

PR #1427, opened 2024-02-28
Remove YJIT setting

vinistock:
### Motivation

Starting to migrate https://github.com/Shopify/vscode-ruby-lsp/pull/923 into this repo by parts.

We added the possibility to turn off YJIT (even when Ruby was compiled with support for it) just in case there was a major issue with YJIT that prevented people from using the LSP. It's been a long time and no one has ever reported any YJIT related issues that weren't fixed before Ruby releases, so keeping this setting doesn't add much value.

### Implementation

Removed the setting, the related status item and command.

### DIFF
--- a/vscode/src/ruby.ts
+++ b/vscode/src/ruby.ts
@@ -223,12 +223,7 @@ export class Ruby implements RubyInterface {
       this.yjitEnabled && (major > 3 || (major === 3 && minor >= 2));
 
     // Starting with Ruby 3.3 the server enables YJIT itself
-    const useYjit =
-      vscode.workspace.getConfiguration("rubyLsp").get("yjit") &&
-      major === 3 &&
-      minor === 2;
-
-    if (this.supportsYjit && useYjit) {
+    if (this.supportsYjit && major === 3 && minor === 2) {
       // RUBYOPT may be empty or it may contain bundler paths. In the second case, we must concat to avoid accidentally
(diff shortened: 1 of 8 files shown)


andyw8 (approved):

vscode/src/ruby.ts
```diff
@@ -223,12 +223,7 @@ export class Ruby implements RubyInterface {
       this.yjitEnabled && (major > 3 || (major === 3 && minor >= 2));
 
     // Starting with Ruby 3.3 the server enables YJIT itself
-    const useYjit =
-      vscode.workspace.getConfiguration("rubyLsp").get("yjit") &&
-      major === 3 &&
-      minor === 2;
-
-    if (this.supportsYjit && useYjit) {
+    if (this.supportsYjit && major === 3 && minor === 2) {
```
paracycle, line 226: This is strange. Am I reading this line correctly, that it only enables YJIT on Ruby 3.2 and not on any other version, like 3.3?

paracycle: Oh, I just saw the comment above this line. In that case, can't we just move the exact version check to the definition of `supportsYjit` and make it:
```ts
this.supportsYjit = this.yjitEnabled && major == 3 && minor == 2;
```?

paracycle: And maybe rename that field to reflect the new meaning as well.

PR #1432, opened 2024-02-28
Do not require specifying release note category on chores

vinistock:
### Motivation

This workflow was not doing exactly what we wanted. We don't want to require people to tag chore changes with categories like `server` or `vscode`, since it doesn't really apply to chores, dependencies or documentation tasks.

Also, there was a bug that was causing us to never check the category, because we didn't reset the value of `matched`.

I started exiting early (successfully) if the PR is tagged with a chore related label. And I reset the value of `matched` before we check for the categories.


vinistock:
It might be a little weird because we need to use the `gh` CLI to hit the API. Since the script is currently short, can I suggest we maintain bash for now and if it grows in complexity we re-evaluate?

KaanOzkan (approved):
This lgtm but we could convert rest of the script to ruby and supply the PR labels as CLI arguments to that ruby script. `gh` call would stay in bash.

vinistock:
Okay, I've switched to using a Ruby script. Please take another look.

bin/check_labels
```diff
@@ -0,0 +1,33 @@
...
+unless label_info
+  warn("Missing argument for label information")
+  exit(1)
+end
+
+pr_labels = JSON.parse(label_info, symbolize_names: true)[:labels].map { |label| label[:name] }
+
+# If the PR includes at least one of the labels ignored for release notes, then exit successfully
+ignored_labels = ["documentation", "chore", "dependencies"]
+exit(0) if pr_labels.any? { |label| ignored_labels.include?(label) }
```
st0012, line 17: Small suggestion:

```suggestion
exit(0) if (pr_labels & ignored_labels).any?
```

bin/check_labels
```diff
@@ -0,0 +1,33 @@
...
+
+# Otherwise, the PR must indicate what type of change it is
+type_labels = ["bugfix", "enhancement", "breaking-change", "other"]
+
+if pr_labels.none? { |label| type_labels.include?(label) }
```
st0012, line 22:
```suggestion
if (pr_labels & type_labels).empty?
```

bin/check_labels
```diff
@@ -0,0 +1,33 @@
...
+  warn("PR must be labeled with one of the following change types: #{type_labels.join(", ")}")
+  exit(1)
+end
+
+# And it should label the category to which we need to include the notes
+category_labels = ["server", "vscode"]
+
+if pr_labels.none? { |label| category_labels.include?(label) }
```
st0012, line 30:
```suggestion
if (pr_labels & category_labels).empty?
```

st0012 (commented):
When comparing array overlap in Ruby, `&` could be handy.

PR #1434, opened 2024-02-29
Introduce `.ruby-lsp.yml` config file

### LINKED ISSUE #1295: Correct configuration file from general index.yml in root of project with relocating it to already existing .ruby-lsp folder.

### Use case

It should eliminate confusion and do not block general file name for specific purposes of projects.

### Description

Currently configuration file `.index.yml` has too general name relative to root of a project and thus do not represent purpose properly. Described in this Readme chapter https://github.com/Shopify/ruby-lsp#configuring-code-indexing.  

There is `.ruby-lsp` folder that is created in root of a project. I think this `.index.yml` file should be stored and read from that folder.

### Implementation

_No response_

andyw8:
### Motivation

We want to have a general purpose configuration mechanism for Ruby LSP that not tied to a particular editor.

Also closes #1295 (but with a slightly different approach)

### Implementation

As agreed with @vinistock, we will repurpose the current indexing configuration file, `.index.yml`.

There will be a new top-level key, `indexing`, but within that the structure was the same as before.

### Automated Tests

Included

### Manual Tests

* Check out this branch
* Ensure indexing is working (e.g. the Jump to Definition works for constants)
* Edit `.ruby-lsp.yml` and rename the `indexing` key.
* Restart the LSP and verify that error notification appears (the remaining LSP feature should continue working)
* 
* Copy `.index.yml` from `main`, and delete `.ruby-lsp.yml`.
* Restart the LSP and ensure a deprecation warning about `.index.yml` shows in the output panel.
* Ensure an error message is showing but otherwise everything works (e.g. formatting)
* Delete `.index.yml` and restart the LSP (so now there are no config files)
* Ensure indexing completes successfully (e.g. the Jump to Definition works for constants)

### DIFF
--- a/lib/ruby_lsp/configuration.rb
+++ b/lib/ruby_lsp/configuration.rb
@@ -1,0 +1,32 @@
+# typed: strict
+# frozen_string_literal: true
+
+module RubyLsp
+  class Configuration
+    extend T::Sig
+
+    sig { params(workspace_uri: URI::Generic).void }
+    def initialize(workspace_uri)
+      @workspace_uri = workspace_uri
+    end
+
+    sig { returns(T::Hash[String, T.untyped]) }
+    def indexing
+      # Need to use the workspace URI, otherwise, this will fail for people working on a project that is a symlink.
+      index_path = File.join(@workspace_uri.to_standardized_path, ".index.yml")
+      ruby_lsp_path = File.join(@workspace_uri.to_standardized_path, ".ruby-lsp.yml")
+
+      if File.exist?(index_path)
+        unless ENV["RUBY_LSP_ENV"] == "test"
+          $stderr.puts("The .index.yml configuration file is deprecated. Please rename it to .ruby-lsp.yml and " \
+            "update the structure as described in the README: https://github.com/Shopify/ruby-lsp#configuration")
+        end
+        YAML.parse_file(index_path).to_ruby
+      elsif File.exist?(ruby_lsp_path)
+        YAML.parse_file(ruby_lsp_path).to_ruby.fetch("indexing")
+      else
+        {}
+      end
+    end
+  end
+end
--- a/.index.yml
+++ b/.index.yml
@@ -1,2 +1,0 @@
-excluded_patterns:
-  - "**/test/fixtures/**/*.rb"
--- a/.ruby-lsp.yml
+++ b/.ruby-lsp.yml
@@ -1,0 +1,3 @@
+indexing:
+  excluded_patterns:
+    - "**/test/fixtures/**/*.rb"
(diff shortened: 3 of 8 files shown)


lib/ruby_indexer/lib/ruby_indexer/configuration.rb
```diff
@@ -45,16 +45,22 @@ def initialize
 
     sig { void }
     def load_config
-      return unless File.exist?(".index.yml")
+      config_file = if File.exist?(".ruby-lsp.yml")
+        ".ruby-lsp.yml"
+      elsif File.exist?(".index.yml") # previously used name
```
andyw8, line 50: Should we add a deprecation warning?

lib/ruby_indexer/test/configuration_test.rb
```diff
@@ -20,6 +20,23 @@ def test_load_configuration_executes_configure_block
       assert(indexables.none? { |indexable| indexable.full_path == __FILE__ })
     end
 
+    def test_supports_older_index_configuration
+      FileUtils.mv(".ruby-lsp.yml", ".ruby-lsp.yml.tmp")
```
andyw8, line 24: I could have made this just renamed `.ruby-lsp.yml` to `.index.yml` but in future they will be different structures.

andyw8:
(will need some updates after https://github.com/Shopify/ruby-lsp/pull/1436)

lib/ruby_lsp/configuration.rb
```diff
@@ -0,0 +1,30 @@
...
+    end
+
+    sig { returns(T::Hash[String, T.untyped]) }
+    def indexing
+      # Need to use the workspace URI, otherwise, this will fail for people working on a project that is a symlink.
+      index_path = File.join(@workspace_uri.to_standardized_path, ".index.yml")
+      ruby_lsp_path = File.join(@workspace_uri.to_standardized_path, ".ruby-lsp.yml")
+
+      if File.exist?(index_path)
+        $stderr.puts("The .index.yml configuration file is deprecated. Please rename it to .ruby-lsp.yml and update the
```
andyw8, line 20: I'm wondering if this should display a notification.

andyw8:
(looks like some tests need a slightly adjustment for Windows)

vinistock:
I think the changes are good, but I wonder if having the configuration file isn't deviating from the LSP approach. We might be better off getting rid of the `.index.yml` (with a deprecation warning first) and then start accepting all of the configuration from the editor.

After #1818 is shipped, we could have the indexing configuration be a part of the global state and gather the configuration from the initialization options.

I think it's worth ensuring we're following the LSP approach at this point, because otherwise we will have introduced two configuration files (`.index.yml` and `.ruby-lsp.yml`), both of which may be not following the standard.

Thoughts?

andyw8:
TODO: update `ruby-lsp-doctor` to handle new config

andyw8:
(re-opening for now but really we should create a seperate issue/PR for @vinistock's [suggestion](https://github.com/Shopify/ruby-lsp/pull/1434#issuecomment-2018723331)).

andyw8:
Closing since we are planning a different approach: https://github.com/Shopify/ruby-lsp/issues/2156

PR #1436, opened 2024-02-29
Allow config to be passed into Indexer

andyw8:
### Motivation

We plan to rename `.index.yml` to `.ruby-lsp.yml` and use that for general configuration. As a first step, this PR refactors the indexer to accept a configuration hash, instead of reading the config file directly.

### Implementation

Pass the config as a hash.

### Automated Tests

Updated

### Manual Tests

Ensure the indexing runs successfully.

### DIFF
--- a/test/expectations/expectations_test_runner.rb
+++ b/test/expectations/expectations_test_runner.rb
@@ -34,6 +34,9 @@ class ExpectationsTestRunner < Minitest::Test
         include ExpectationsRunnerMethods
       RB
 
+      # The fixtures tests take some time to run, this allows us to skip them locally if desired
+      return if ENV["RUBY_LSP_SKIP_FIXTURES_TEST"]
+
       Dir.glob(TEST_RUBY_LSP_FIXTURES).each do |path|
         test_name = File.basename(path, ".rb")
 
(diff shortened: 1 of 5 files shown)


test/expectations/expectations_test_runner.rb
```diff
@@ -34,6 +34,9 @@ def default_args
         include ExpectationsRunnerMethods
       RB
 
+      # The fixtures tests take some time to run, this allows us to skip them locally if desired
+      return if ENV["RUBY_LSP_SKIP_FIXTURES_TEST"]
```
andyw8, line 38: Useful? I can move to a separate PR if desired.

st0012: Out of curiosity, what tests made you want to add this? I assume you ran `RUBY_LSP_SKIP_FIXTURES_TEST=true dev test`, but what tests were in your mind when running it?

Asking because it may be a potential opportunity to add ways to run tests. For example, maybe we can make the `Run Tests` option work when right-clicking on a folder? It'll make running all ruby indexer's tests easier.

Screenshot 2024-03-05 at 19 37 14

andyw8:
This will be awkward to rebase so I'm going create a fresh branch.

PR #1441, opened 2024-03-01
Extract `Notification.window_show_error`

andyw8:
We have very similar calls in 7 places in `executor.rb`.

I've only extracted one place as an illustration, but what do folk think of this?


vinistock (approved):
Nice

PR #1443, opened 2024-03-01
Show require and require_relative completions on quotes

vinistock:
### Motivation

Previously, we were showing require completions on `/` and require_relative completions on `./`. However, that's not very useful, we can show all options the moment the developer adds the quotes for the string.

### Implementation

- Added `"` as a trigger character, so that we can match empty require arguments
- Adapted the require relative completion a bit to account for this possibility

### Automated Tests

Added a test showing the behaviour.

### Manual Tests

2. Type `require "` or `require_relative "` into any file
3. Verify that you get completions offered right away


lib/ruby_lsp/requests/completion.rb
```diff
@@ -34,7 +34,7 @@ class << self
         def provider
           Interface::CompletionOptions.new(
             resolve_provider: false,
-            trigger_characters: ["/"],
+            trigger_characters: ["/", "\""],
```
andyw8, line 37: What if the project's styleguide is to use single quotes?

PR #1444, opened 2024-03-02
Add Rake document symbols

### LINKED ISSUE #1412: Support document symbol for Rakefile

### Use case

Document symbol could be very helpful for navigating between tasks inside a large Rakefile.

### Description

Ruby LSP can add task declarations as document symbols so VS Code users can use `Go to Symbol` to quickly search and jump between task definitions.

### Implementation

_No response_

thomasmarshall:
### Motivation

Closes #1412

This commit adds document symbols for the `namespace` and `task` methods in the Rake DSL. This will allow users to navigate large Rake files using the document outline feature in VS Code.

Screenshot 2024-03-02 at 14 29 20

### Implementation

I used the `on_call_node_enter` and `on_call_node_leave` hooks to generate document symbols for call nodes named `namespace` and `task`, using `NAMESPACE` and `METHOD` symbol kinds respectively.

Is there a good way to narrow these document symbols to only Rake files? Otherwise unrelated call nodes named `namespace` and `task` might incorrectly appear in the outline. I noticed this is an issue for `test` in `ruby-lsp-rails` too, so maybe it's not something to be too worried about.

I pushed [another branch](https://github.com/Shopify/ruby-lsp/compare/main...rake-document-symbols-addon) that prototypes pulling this out into a Rake addon (though still in the `ruby-lsp` project). I thought that might be a more suitable solution, as it would give a good opportunity to filter based on the whether the filename is `Rakefile` or `*.rake`. Please take a look at that alternative approach and let me know if it's more suitable (though it breaks the API for `create_document_symbol_listener` to pass through the URI).

### Automated Tests

I added document symbol expectation tests using a Ruby fixture and JSON.

### Manual Tests

- View a Rakefile or other Ruby file using the Rake DSL.
- Expand the document outline panel in VS Code.


lib/ruby_lsp/listeners/document_symbol.rb
```diff
@@ -357,6 +369,66 @@ def handle_alias_method(node)
...
+        name_argument = arguments.arguments.first
+        return unless name_argument
+
+        name = case name_argument
+        when Prism::StringNode then name_argument.content
+        when Prism::SymbolNode then name_argument.value
+        when Prism::KeywordHashNode
+          first_element = name_argument.elements.first
+          case first_element
+          when Prism::AssocNode
```
andyw8, line 414: Would this be better as an `if` since there's only one clause?

andyw8:
Also not a blocker, but we might later want to make this run only on `.rake` files (or `Rakefile`), since there's a small chance that an app or gem could define its own `task` macro for other purposes.

andyw8:
> I pushed [another branch](https://github.com/Shopify/ruby-lsp/compare/main...rake-document-symbols-addon) that prototypes pulling this out into a Rake addon (though still in the ruby-lsp project). I thought that might be a more suitable solution, as it would give a good opportunity to filter based on the whether the filename is Rakefile or *.rake.

I'd say an addon isn't necessary for this since Rake is built-in to Ruby, and will always be available.

thomasmarshall:
> I noticed one small problem when verifying in VS Code. If you have a task with a blank name, e.g. because you're editing the name, it results in a error from the extension

I've fixed this now and added a test, but it only covers the empty string case, not the incomplete symbol case (i.e. `task :`) because another test failed when the fixture had invalid Ruby. Let me know if there's a better approach.

FYI I noticed the same error currently exists for:

- `attr_accessor :`
- `alias :`
- `def`

But they are unrelated to this feature so I left them as-is.

> it would be nice if there was a better way to visually distinguish the macros from methods

Agreed! But I'm not sure which symbols to choose—any thoughts?

> Also not a blocker, but we might later want to make this run only on `.rake` files (or `Rakefile`), since there's a small chance that an app or gem could define its own `task` macro for other purposes.

This is what I was trying to do with the addon—I thought that would be a good opportunity to inspect the URI—but I understand it's not really necessary here. I've added the condition directly to the `DocumentSymbol` listener instead.

andyw8:
Thanks, it was only the blank string case where that I was seeing the extension error. I'm not seeing any error for the `attr_accessor :` and other cases you mention though 🤔

thomasmarshall:
Hmm I can recreate all three cases pretty reliably on latest `main`.

<details>
  <summary>Screenshots</summary>
  Screenshot 2024-03-12 at 18 16 44
  Screenshot 2024-03-12 at 18 24 36
  Screenshot 2024-03-12 at 18 25 02
</details>

`handle_attr_accessor` returns early if `name` is nil, but not if it's empty:

https://github.com/Shopify/ruby-lsp/blob/621644ca6b0bcbbc3baa033ddf85f96c5b0d3345/lib/ruby_lsp/listeners/document_symbol.rb#L306

Same for `handle_alias`:

https://github.com/Shopify/ruby-lsp/blob/621644ca6b0bcbbc3baa033ddf85f96c5b0d3345/lib/ruby_lsp/listeners/document_symbol.rb#L340

But Prism will happily parse this as having a symbol node with an empty value:

```rb
Prism.parse("attr_accessor :")
```

<details>
  <summary>Parse result</summary>

```
  @ ProgramNode (location: (1,0)-(1,15))
  ├── locals: []
  └── statements:
      @ StatementsNode (location: (1,0)-(1,15))
      └── body: (length: 1)
          └── @ CallNode (location: (1,0)-(1,15))
              ├── flags: ignore_visibility
              ├── receiver: ∅
              ├── call_operator_loc: ∅
              ├── name: :attr_accessor
              ├── message_loc: (1,0)-(1,13) = "attr_accessor"
              ├── opening_loc: ∅
              ├── arguments:
              │   @ ArgumentsNode (location: (1,14)-(1,15))
              │   ├── flags: ∅
              │   └── arguments: (length: 1)
              │       └── @ SymbolNode (location: (1,14)-(1,15))
              │           ├── flags: forced_us_ascii_encoding
              │           ├── opening_loc: (1,14)-(1,15) = ":"
              │           ├── value_loc: (1,15)-(1,15) = ""
              │           ├── closing_loc: ∅
              │           └── unescaped: ""
              ├── closing_loc: ∅
              └── block: ∅,
```
</details>

Similar issue with `def`:

```rb
Prism.parse("def")
```


<details>
  <summary>Parse result</summary>

```
  @ ProgramNode (location: (1,0)-(1,3))
  ├── locals: []
  └── statements:
      @ StatementsNode (location: (1,0)-(1,3))
      └── body: (length: 1)
          └── @ DefNode (location: (1,0)-(1,3))
              ├── name: :""
              ├── name_loc: (1,3)-(1,3) = ""
              ├── receiver: ∅
              ├── parameters: ∅
              ├── body:
              │   @ StatementsNode (location: (1,3)-(1,3))
              │   └── body: (length: 1)
              │       └── @ MissingNode (location: (1,3)-(1,3))
              ├── locals: []
              ├── def_keyword_loc: (1,0)-(1,3) = "def"
              ├── operator_loc: ∅
              ├── lparen_loc: ∅
              ├── rparen_loc: ∅
              ├── equal_loc: ∅
              └── end_keyword_loc: (1,3)-(1,3) = "",
```
</details>

andyw8:
Ah, I'm seeing the error now on a separate file. I'll log an issue.

lib/ruby_lsp/listeners/document_symbol.rb
```diff
@@ -83,6 +86,19 @@ def on_call_node_enter(node)
           handle_attr_accessor(node)
         elsif node.name == :alias_method
           handle_alias_method(node)
+        elsif node.name == :namespace
```
vinistock, line 89: Since we now have so many cases doing `node.name`, can we please extract it into a local variable before the `if`?

lib/ruby_lsp/listeners/document_symbol.rb
```diff
@@ -357,6 +373,76 @@ def handle_alias_method(node)
...
+
+        create_document_symbol(
+          name: name,
+          kind: Constant::SymbolKind::METHOD,
+          range_location: name_argument.location,
+          selection_range_location: name_argument.location,
+        )
+      end
+
+      sig { returns(T::Boolean) }
```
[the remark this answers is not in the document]
vinistock: I think it's fine to start here and extract later. We're not using it anywhere else just yet.

andyw8 (approved):

lib/ruby_lsp/listeners/document_symbol.rb
andyw8: I've update an update for that.

PR #1453, opened 2024-03-05
Support RuboCop's LSP mode

### LINKED ISSUE #1449: Support `autocorrect: contextual`

### Use case

Helps to prevent unwanted autocorrections while writing code. This behaviour was recently added to RuboCop.

### Description

See https://github.com/rubocop/rubocop/pull/12657 for details.

### Implementation

TBC

koic:
### Motivation

Resolves #1449.

This PR applies `RuboCop::LSP.enable`, which is the minimal API for RuboCop to handle in the context of LSP: https://docs.rubocop.org/rubocop/1.61/usage/lsp.html#language-server-development

### Implementation

Furthermore, `RuboCop::LSP.enable` method and `RuboCop::LSP` module were introduced in RuboCop 1.61.0. This PR ensures compatibility with existing versions of RuboCop dependencies. However, in older versions of RuboCop prior to 1.60, `AutoCorrect: contextual` behaves the same as `AutoCorrect: true`.

### DIFF
--- a/lib/ruby_lsp/requests/support/rubocop_runner.rb
+++ b/lib/ruby_lsp/requests/support/rubocop_runner.rb
@@ -13,6 +13,10 @@ rescue LoadError
   raise StandardError, "Incompatible RuboCop version. Ruby LSP requires >= 1.4.0"
 end
 
+if RuboCop.const_defined?(:LSP) # This condition will be removed when requiring RuboCop >= 1.61.
+  RuboCop::LSP.enable
+end
+
 module RubyLsp
   module Requests
     module Support


koic:
I have signed the CLA!

vinistock:
Thank you very much for the contribution, @koic! I'll go ahead and merge this, but I wanted to ask a question. I think we may have an opportunity to improve the developer experience even more.

It's great that the auto-correct in this new contextual mode won't fix things while in the middle of typing. However, it would be nice if we could still offer quickfixes through code actions, so that if people really want to fix, they can use those.

Is there a way to get the auto-correction results for an offense even in contextual mode so that we can offer the quickfixes?

koic:
That a good question. RuboCop's own LSP does not yet support quickfix, so there is no method provided to force autocorrect from the LSP context. But, since quickfix is intended for users to explicitly want corrections, it is something I'd like to support in the future.

Technically, if reusing the current RuboCop's API, applying `RuboCop::LSP.disable` just before the quickfix and then switching back to `RuboCop::LSP.enable` after completing the quickfix should work. I'd like to explore this again after supporting quickfix in RuboCop's own LSP.

PR #1455, opened 2024-03-05
Fix missing images on the documentation website

st0012:
### Motivation

Currently, the documentation site's images are all missing: 

Screenshot 2024-03-05 at 23 59 44

### Implementation

It's caused by 2 reasons:

1. The README's image paths don't exist under the `/docs` directory.
2. The Snapper template has a CSP rule that blocks local images.

2 has been fixed via https://github.com/Shopify/rdoc/pull/15

This commit addresses 1 by copying the images that README uses to the `/docs` directory in the doc publishing workflow.

The upside of this approach is that it's simple and we don't need to change the README's image paths, nor move the images around.

The downside is that we still can't see the images locally. But I think it's a reasonable trade-off.

### DIFF
--- a/.github/workflows/publish_docs.yml
+++ b/.github/workflows/publish_docs.yml
@@ -26,6 +26,12 @@ jobs:
       - name: Generate documentation
         run: bundle exec rake rdoc
 
+      - name: Copy README images to the docs folder
+        run: |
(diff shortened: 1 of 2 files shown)


vinistock (approved):
Nice

PR #1459, opened 2024-03-05
Use `highlighted_area` for rubocop diagnostics

### LINKED ISSUE #1119: Look into using RuboCop's highlight range for diagnostics
According to https://github.com/Shopify/ruby-lsp/issues/355#issuecomment-1769036623, RuboCop violations include a highlight range, which can be used by us to report diagnostics in a more editor friendly manner.

Let's investigate if this works as expected and switch to using that range.

Earlopain:
### Motivation

Closes #1119

### Implementation

Like suggested, this uses `highlight_range` on the offense to only mark the first line of multiline offenses.
This method result is not memoized by rubocop. I don't know if this really matters but I've added the method result in a local variable.

### Automated Tests

Yes. Almost all current tests need no changes with this.

### Manual Tests

image

The quick fix still adds the rubocop disable command at the right location.

### DIFF
--- a/lib/ruby_lsp/requests/support/rubocop_diagnostic.rb
+++ b/lib/ruby_lsp/requests/support/rubocop_diagnostic.rb
@@ -40,6 +40,9 @@ module RubyLsp
 
         sig { returns(Interface::Diagnostic) }
         def to_lsp_diagnostic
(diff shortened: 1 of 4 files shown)


vinistock (approved):

PR #1460, opened 2024-03-05
Set value instead of placeholder for textarea in issue template

vinistock:
### Motivation

I think the issue is that the text we selected is too long for a placeholder. In fact, we want it to be the value inside the textarea so that folks can edit it.


andyw8 (approved):

andyw8:
Oh, shouldn't we be using `|` instead of `>`?

> For multiline strings, you can use the vertical bar (|) pipe character or the greater-than (>) character. The difference between the two is that the vertical bar character preserves the new line while the greater-than character folds the new line and converts it into spaces.

vinistock:
> Oh, shouldn't we be using | instead of >?

Yes, good catch!

PR #1464, opened 2024-03-06
Allow addons to register for filewatching events

andyw8:
(writing in progress!)

### DIFF
--- a/lib/ruby_lsp/executor.rb
+++ b/lib/ruby_lsp/executor.rb
@@ -229,6 +229,8 @@ module RubyLsp
         end
       end
 
+      # TODO: document once we are happy with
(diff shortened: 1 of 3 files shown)


lib/ruby_lsp/executor.rb
```diff
@@ -229,6 +229,8 @@ def did_change_watched_files(changes)
         end
       end
 
+      # TODO: document once we are happy with
```
vinistock, line 232: Let's create an issue instead of leaving a todo comment.

vinistock (approved):

lib/ruby_lsp/executor.rb
andyw8: https://github.com/Shopify/ruby-lsp/issues/1465

(will remove todo)

Issue #1466, opened 2024-03-06
Improve how we detect Rails apps


andyw8:
### Description

Currently we look for a dependency on the Rails gem, and if found, we assume a Rails app is present, and add "ruby-lsp-rails" to the custom bundle. But this approach can result in false positives. Instead, we should check the presence of `bin/rails`.

andyw8:
Done in https://github.com/Shopify/ruby-lsp/pull/1896

PR #1467, opened 2024-03-06
Improve Rails app detection

andyw8:
### Motivation

Closes https://github.com/Shopify/ruby-lsp/issues/1466

### Implementation

Check for the presence of `bin/rails` as suggested by Rafael.

### Automated Tests

Included

### Manual Tests

It's slightly awkward as we need to ensure the setup process doesn't end up installing the published version:

- Check out this branch.
- Temporarily change the version to something nonsense value, e.g. `0.14.999` so you can easily identify it.
- Run `gem build`
- Run `gem install <filename>`
- Check out the `rails/rails` repo.
- Delete the current `.ruby-lsp` directory if it exists
- Reload the window
- Ensure the Ruby LSP output window doesn't show a failure for `bin/rails`
- Ensure `.ruby-lsp/Gemfile` doesn't contain `ruby-lsp-rails`
- Remember to uninstall the temporary local gem afterwards, or it may result in confusing failures.

### DIFF
--- a/lib/ruby_lsp/setup_bundler.rb
+++ b/lib/ruby_lsp/setup_bundler.rb
@@ -60,7 +60,7 @@ module RubyLsp
       # Do not set up a custom bundle if LSP dependencies are already in the Gemfile
       if @dependencies["ruby-lsp"] &&
           @dependencies["debug"] &&
-          (@dependencies["rails"] ? @dependencies["ruby-lsp-rails"] : true)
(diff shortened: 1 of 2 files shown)


vinistock:
My two cents is that, as long as the `ruby-lsp-rails` is not failing, we can delay this decision a little bit because we may find value in having the addon present for Ruby apps that have `rails` in the Gemfile, but aren't Rails applications.

PR #1468, opened 2024-03-06
Log if Sorbet is detected

andyw8:
### Motivation

We were investigate an issue on a repo where autocompletion wasn't working. It was caused by Sorbet being detected, even though the app wasn't set up for Sorbet. Logging this in may help to diagnose similar issues in future.

### Implementation

Added a log entry.

### Automated Tests

n/a

### Manual Tests

Verified against the internal Shopify repo where this came up.


andyw8:
Since this output is quite [noisy](https://github.com/Shopify/ruby-lsp/actions/runs/8178947187/job/22363953280?pr=1468), I've changed it to not show when running tests.

lib/ruby_lsp/requests/support/dependency_detector.rb
```diff
@@ -64,7 +64,12 @@ def detect_typechecker
       return false if ENV["RUBY_LSP_BYPASS_TYPECHECKER"]
 
       Bundler.with_original_env do
-        Bundler.locked_gems.specs.any? { |spec| spec.name == "sorbet-static" }
+        sorbet_static_detected = Bundler.locked_gems.specs.any? { |spec| spec.name == "sorbet-static" }
+        # Don't show message while running tests, since it's noisy
+        if sorbet_static_detected && !defined?(Minitest)
```
andyw8, line 69: I wondering if there's a better way... I could do `ENV["CI"]` but it would still show locally.

andyw8: Yeah it's not ideal... any thoughts @st0012?

st0012 (approved):
I think it's valuable to add as it's always hard to see if Sorbet is involved in the issue we investigate. It's also cheap to remove if we found it not working as expected or causing other issues.

st0012:
@vinistock Removing it would result in hundreds of lines of the warning when running the test suite though. I think it or a similar check is necessary.

vinistock:
Can we instead set a variable ourselves then? Something like

```ruby
# test/test_helper.rb

["RUBY_LSP_ENV"] = "test"
```

andyw8:
Yeah that's an option (pushed).

andyw8:
Working now - the problem was with `Bundler.with_original_env`.

vinistock (approved):

PR #1469, opened 2024-03-06
[Experimental] RuboCop with Prism (only for ruby-lsp's own code)

andyw8:
Trying out the new support in RubCop 1.62.0:

https://github.com/rubocop/rubocop/releases/tag/v1.62.0


.rubocop.yml
```diff
@@ -9,9 +9,10 @@ require:
   - ./lib/rubocop/cop/ruby_lsp/use_register_with_handler_method
 
 AllCops:
+  ParserEngine: parser_prism
   NewCops: disable
   SuggestExtensions: false
-  TargetRubyVersion: 3.0
+  TargetRubyVersion: 3.3
```
andyw8, line 15: Needs 3.3 as `TargetRubyVersion`., so we may not be able to adopt yet.

koic: Since support is provided for Ruby 3.0 as a runtime:
https://github.com/Shopify/ruby-lsp/blob/3fffcfb40e2975959a0ccedd90a3dbfceac03c10/ruby-lsp.gemspec#L25

It appears necessary to align with `TargetRubyVersion: 3.0`. `ParserEngine: parser_prism` cannot be utilized unless it is shifted to support Ruby 3.3+ only.

In other words, at present, only Ruby applications or libraries that support Ruby 3.3+ can apply `ParserEngine: parser_prism`. So, still need to use the Parser gem as the ParserEngine in this repo.

vinistock:
Yeah, that's a good idea.

andyw8:
@kddnewton is there somewhere we should report issues for `parser_prism`, or is it too early for that?

bin/check_labels
```diff
@@ -14,20 +14,20 @@ pr_labels = JSON.parse(label_info, symbolize_names: true)[:labels].map { |label|
 
 # If the PR includes at least one of the labels ignored for release notes, then exit successfully
 ignored_labels = ["documentation", "chore", "dependencies"]
-exit(0) if (pr_labels & ignored_labels).any?
+exit(0) if pr_labels.intersect?(ignored_labels)
```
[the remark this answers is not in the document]
andyw8: Ah, this is actually just due to the `TargetRubyVersion` change.

.github/workflows/require_labels.yml
```diff
@@ -11,6 +11,9 @@ jobs:
       GH_TOKEN: ${{ github.token }}
     steps:
       - uses: actions/checkout@v4
+      - uses: ruby/setup-ruby@v1
```
andyw8, line 14: The default Ruby on ubuntu-latest must be older, since it was failing on `Array#intersect`.

bin/check_labels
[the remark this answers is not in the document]
andyw8: And since this isn't part of the shipped code, it doesn't matter that it's not compatible with older Rubies.

.github/workflows/ci.yml
```diff
@@ -24,6 +24,21 @@ jobs:
 
       - name: Lint Ruby files and RBI shims
         run: bundle exec rubocop
+  rubocop_prism_experimental:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+        with:
+          submodules: "recursive"
```
vinistock, line 32: Do we need the submodules here?

.github/workflows/ci.yml
```diff
@@ -24,6 +24,21 @@ jobs:
...
+
+      - name: Set up Ruby
+        uses: ruby/setup-ruby@v1
+        with:
+          ruby-version: "3.3"
+          bundler-cache: true
+
+      - name: Lint Ruby files and RBI shims
+        run: bundle exec rubocop -c  .rubocop-prism-experimental.yml
```
vinistock, line 41: This is probably going to fail on every PR. Should we force this to exit with a zero status so that we avoid all PRs being red (and main)?

andyw8: Do you think there'd be that many? I was planning that we just add them as ignores in the YAML (like what I did for Layout/HeredocIndentation), then remove once resolved upstream.

andyw8:
Fair enough, I'll close this for now.

PR #1502, opened 2024-03-08
Implement Shadowenv as a version manager object

vinistock:
### Motivation

Migrate Shadowenv to the new version manager implementation approach, with its own dedicated class.

### Implementation

1. Added Sinon JS so that we can use stubs, mocks and spys. This makes testing much easier
2. Moved all Shadowenv activation logic inside its own dedicated class.
3. Started out simple with the parent class `VersionManager`, only requiring that subclasses implement the `activate` method

### Automated Tests

One of the main advantages of this new approach is exactly being able to test each version manager in detail. Added tests for all scenarios we cover.

Had to start downloading Shadowenv, so that we can run the tests using the real tool.

### Manual Tests

4. In the extension host window, open a project using Shadowenv
5. Verify that the Ruby LSP boots properly using the declared Ruby version

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -87,7 +87,22 @@ jobs:
       - name: Set up Ruby
         uses: ruby/setup-ruby@v1
         with:
-          ruby-version: "3.3"
+          ruby-version: "3.3.0"
+
+      - name: Download shadowenv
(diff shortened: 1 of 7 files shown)


KaanOzkan (approved):

andyw8 (approved):
Tophatted against code-db

PR #1511, opened 2024-03-08
Fix minitest "Run in Terminal" executing multiple tests when only one is clicked

Earlopain:
### Motivation

Consider the example file at https://github.com/Shopify/ruby-lsp/blob/31395c2a606d1038dcbbc9692e8931fc1cd977a5/test/fixtures/minitest_tests.rb

Clicking on the Code Lens for the first test  at line 6 will execute 5 tests:

* `test_public` which I actually clicked.
* `test_public_vcall` and `test_public_command` because they are in the same group and start with the same name.
* `test_public` and `test_public_2` from `AnotherTest` because the group ends with the same name and the methods start with the same name.

Note that this is also a problem when matching against groups. Clicking the code lens on line 1 will result in 7 tests being run: All tests from `Test` and the tests from `AnotherTest` as well. This is a separate issue which I don't know the solution to and don't plan to address in this PR.

### Implementation

Stop matching with a regex when a specific test is selected to be run. When a test is nested, pass along the whole class chain and construct the name from there.

### Automated Tests

Updated expectations.

### Manual Tests

I added a `require "minitest/autorun"` to `minitest_tests.rb` and `nested_minitest_tests.rb` and confirmed that each code lens for methods only results in exactly one test being run.


lib/ruby_lsp/listeners/code_lens.rb
```diff
@@ -181,14 +181,15 @@ def add_test_code_lens(node, name:, command:, kind:)
         )
       end
 
-      sig { params(group_name: String, method_name: T.nilable(String)).returns(String) }
-      def generate_test_command(group_name:, method_name: nil)
+      sig { params(group_stack: T::Array[String], method_name: T.nilable(String)).returns(String) }
+      def generate_test_command(group_stack:, method_name: nil)
         command = BASE_COMMAND + T.must(@path)
+        group_name = T.must(group_stack.last)
```
vinistock, line 187: We also need to make this more precise for classes. I think that will require adding extra logic for the specs, which will continue to use a regex. Maybe we can introduce a new parameter for `spec_name`, which we use as a regex, and that will allow you to use `group_stack.join("::")` for running test classes.

Also, I we need to verify that it works the same for test-unit and update it as well.

Earlopain: I think I found a pretty neat solution here. This should now work like you suggested but instead of just matching against `group_stack.join("::")` I also added a `\A` start of string matcher. This makes it even more precise in what is being executed. I did manual tests with this and its working as expected.

lib/ruby_lsp/listeners/code_lens.rb
```diff
@@ -181,18 +181,30 @@ def add_test_code_lens(node, name:, command:, kind:)
...
           command += if method_name
-            " --name " + "/#{Shellwords.escape(group_name + "#" + method_name)}/"
+            # Direct match only
+            " --name " + Shellwords.escape(group_stack.join("::") + "#" + method_name)
+          elsif spec_name
+            " --name " + "/#{Shellwords.escape(spec_name)}/"
           else
-            " --name " + "/#{Shellwords.escape(group_name)}/"
+            # Match all test methods that begin with the class path.
+            " --name " + "\"/\\A#{Shellwords.escape(group_stack.join("::"))}/\""
```
vinistock, line 203: Nitpick.
```suggestion
            " --name " + "\"/^#{Shellwords.escape(group_stack.join("::"))}/\""
```

lib/ruby_lsp/listeners/code_lens.rb
```diff
@@ -181,18 +181,30 @@ def add_test_code_lens(node, name:, command:, kind:)
...
           end
         when "test-unit"
+          # TODO: Check if this can use more specific test matching like minitest
```
vinistock, line 206: Please, create an issue for this and remove the todo comment.

vinistock (commented):
Looking great! Just two minor comments and we can ship

lib/ruby_lsp/listeners/code_lens.rb
Earlopain: I tried using `$` and was wondering why it wasn't working 🤦

Earlopain:
Should be good now!

vinistock:
Sorry, I just tested this on the [index_test](https://github.com/Shopify/ruby-lsp/blob/6071b1263a672f4ca9548e6a42de45220ebbd9da/lib/ruby_indexer/test/index_test.rb#L6) and noticed there's one more issue. We're not pushing modules into the group stack, which means names inside modules are incomplete. If you try to run the test class, it runs `/^IndexTest/` which misses the module name and thus doesn't run any tests.

We need to start doing this

```ruby
# Register for the `on_module_node_enter` and `on_module_node_leave` events in initialize
# Then create the handlers to keep track of modules

def on_module_node_enter(node)
  @group_stack.push(node.constant_path.full_name)
rescue DynamicPartsInConstantPathError, MissingNodesInConstantPathError
  # We can't process dynamic module references e.g.: module foo::Something
end

def on_module_node_leave(node)
  @group_stack.pop
end
```

Earlopain:
Yeah, I see what you mean. I will add handling and tests for that tomorrow.

The code sample you've written above can cause the stack to become unbalanced (or out of sync, whatever you call that). Is that a concern? If so, I think the easiest would be to push some dummy value, like the empty string perhaps. Does that sound right?

vinistock:
> The code sample you've written above can cause the stack to become unbalanced

What do you mean by this? Are you talking about cases with dynamic references? In those cases, there's not much we can do anyway. Imagine you have this

```ruby
def foo
  rand > 0.5 ? Foo : Bar
end

module foo
  class MyTest < Minitest::Test
  end
end
```
There's no way to be specific about what we need to run in this case. What we could do is push a special string in the stack that we know is an invalid constant identifier and then we can use that to run the test with a fallback mechanism.

For example, we can push `<dynamic_reference>` into the stack. When we're adding the code lens, if there's a dynamic reference, then we simplify the regex and just use something like `/MyTest$/`. It won't be as precise, but it's likely the best we can do in this scenario.

> Writing tests for this ain't easy. 

Can you elaborate? I can provide some assistance to write tests. Basically, you'd need to create a few new fixtures, which include both the module nesting and a dynamic reference. And then we need to add an expectation file with the expected response.

vinistock:
Ah, good point. That might be another argument for us pushing `<dynamic_reference>` into the stack. It would both allow running `MyOtherTest` with lower precision and wouldn't mess up `CompletlyDifferentTest`.

Earlopain:
My previous comment was just wrong. Docs and types state that `constant_path` will return a plain `Node`. This whole thing stems from me not setting up the sorbet vscode extension. Sorry about all this

Earlopain:
Force-pushed without commiting the changes beforehand. Oops. Can this be reopenend or do I need to make a new PR?

I _think_ I've got all the cases covered now and have the matchers as specific as they can be. Let me know if there's something I've missed.

I've used `(#|::)` to only match against classes with the correct name. So a class named `Test` will not execute tests from `OtherTest` when invoked from the class code lens. I've added a comment for this in the code to make the intention behind this clear. This whole thing turned out more complicated than I initially thought.

For the module node enter thing I had trouble with, I added a new method `module_constant_name` to the Common module. This seemed to fit nicely there because `constant_name` is already there.

lib/ruby_lsp/requests/support/common.rb
```diff
@@ -147,6 +147,15 @@ def constant_name(node)
...
         end
+
+        sig { params(node: Prism::ModuleNode).returns(T.nilable(String)) }
+        def module_constant_name(node)
+          path = node.constant_path
+          case path
+          when Prism::ConstantPathNode, Prism::ConstantReadNode, Prism::ConstantPathTargetNode
+            constant_name(path)
+          end
+        end
```
vinistock, line 158: Nitpick: let's rename to `namespace_constant_name` because this is the same story for classes.

```suggestion
        sig { params(node: T.any(Prism::ModuleNode, Prism::ClassNode)).returns(T.nilable(String)) }
        def namespace_constant_name(node)
          path = node.constant_path
          case path
          when Prism::ConstantPathNode, Prism::ConstantReadNode, Prism::ConstantPathTargetNode
            constant_name(path)
          end
        end
```

vinistock (approved):
I re-opened the PR, no need to open a separate one. This looks great, I just have a minor nitpick about the naming of the new support method

vinistock:

Earlopain:
Thank _you_ for coaching me through this

PR #1529, opened 2024-03-11
Implement Mise as a manager object

vinistock:
### Motivation

Implement Mise activation as a manager object.

### Implementation

Quite similar to Shadowenv, we use the executable to run a Ruby script with the environment activated.

According to the Mise docs, the standard installation location is `~/.local/bin/mise` and you have to use alternative methods to get it installed elsewhere. Because of this, I think it's fine to begin only supporting the default installation path.

Like most version managers, Mise does not support Windows.

### Automated Tests

Added a test just to verify we invoke the command with the expected arguments. I'm going to vote for not installing Mise in CI.

### Manual Tests

1. Start the extension in development mode on this branch
2. In the extension host window, open a project using Mise to manage the Ruby version
3. Verify that the Ruby LSP was booted using the right Ruby version by clicking the language status item (the `{}` button) right next to the language mode (where it says `Ruby` in the bottom right corner of the editor)

### DIFF
--- a/vscode/src/test/suite/ruby/mise.test.ts
+++ b/vscode/src/test/suite/ruby/mise.test.ts
@@ -1,0 +1,57 @@
+import assert from "assert";
+import path from "path";
+import os from "os";
+
+import * as vscode from "vscode";
+import sinon from "sinon";
+
+import { Mise } from "../../../ruby/mise";
+import { WorkspaceChannel } from "../../../workspaceChannel";
+import * as common from "../../../common";
+
+suite("Mise", () => {
+  if (os.platform() === "win32") {
+    // eslint-disable-next-line no-console
+    console.log("Skipping Mise tests on Windows");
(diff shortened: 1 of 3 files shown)


Morriar (approved):

vscode/src/test/suite/ruby/mise.test.ts
```diff
@@ -0,0 +1,57 @@
+import assert from "assert";
+import path from "path";
+import os from "os";
+
+import * as vscode from "vscode";
+import sinon from "sinon";
+
+import { Mise } from "../../../ruby/mise";
+import { WorkspaceChannel } from "../../../workspaceChannel";
+import * as common from "../../../common";
```
andyw8, line 10: (nit) apart from `Mise`, I expect all these `import`s will be same for each manager test... could we centralize them in one place?

PR #1589, opened 2024-03-12
Remove pattern from document selector

### LINKED ISSUE #1349: rubocop stops working at 0.5.6
### Ruby version

3.2.2 and 3.3.0

### Code snippet

_No response_

### Description

When I open folder in vscode on windows with ruby code (Gemfile has ruby-lsp and rubocop) - I get formatting, code suggestions etc. But after upgrading from 0.5.5 to 0.5.6 - i stops working. In OUPUT of ruby-lsp I don't see any activity when 0.5.6 (or newer) is installed. Event more - when I highlight all document and choose "Format Document" - vscode says there are no formatters for ruby and gives me option to install some.

Problem is ONLY on windows - on linux (via WSL, or via SSH) formatting etc. works as expected in verisons of ruby-lsp newer than 0.5.5

### Expected output

It should work like before on windows

vinistock:
### Motivation

Closes #1349

In https://github.com/Shopify/vscode-ruby-lsp/pull/986, we ended up making a mistake using `path.join` for the document selector pattern. In Windows, that's going to result in `workspace\**\*` (using backslashes), which doesn't get recognized properly.

### Implementation

I took another look at it and there's no reason why we actually need the pattern. We want to handle all Ruby files, so we can just remove it.

This is the same that [Rust Analyzer does](https://github.com/rust-lang/rust-analyzer/blob/a2e274142f35d21fd28d28655f4af8e8531ab649/editors/code/src/client.ts#L82).

### Manual Tests

2. Open any file and try to go to the definition of a constant
3. Verify you can do that inside the workspace
4. Jump to the source code of a gem
5. Verify you can also go to definitions inside the gem's source code

### DIFF
--- a/vscode/src/client.ts
+++ b/vscode/src/client.ts
@@ -91,7 +91,6 @@ function collectClientOptions(
   configuration: vscode.WorkspaceConfiguration,
   workspaceFolder: vscode.WorkspaceFolder,
   outputChannel: WorkspaceChannel,
-  ruby: Ruby,
 ): LanguageClientOptions {
   const pullOn: "change" | "save" | "both" =
     configuration.get("pullDiagnosticsOn")!;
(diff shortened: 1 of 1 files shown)


paracycle:
Question: Does this mean that if I open a Ruby file that doesn't belong to the current project in the editor and save it, it will also be formatted using the formatting rules of my project?

vinistock:
It won't because we skip running [formatting](https://github.com/Shopify/ruby-lsp/blob/3242bf714ae4a52d9c1da00d3dbc85772ad94d4f/lib/ruby_lsp/executor.rb#L344) for files that are outside the workspace. Same thing for diagnostics.

You get all other features though.

PR #1590, opened 2024-03-12
Exclude the `tmp` directory from indexing by default

vinistock:
### Motivation

Let's exclude the `tmp` directory by default, since it may include irrelevant Ruby files that end up getting indexed.

### DIFF
--- a/lib/ruby_indexer/lib/ruby_indexer/configuration.rb
+++ b/lib/ruby_indexer/lib/ruby_indexer/configuration.rb
@@ -20,7 +20,7 @@ module RubyIndexer
     def initialize
       @excluded_gems = T.let(initial_excluded_gems, T::Array[String])
       @included_gems = T.let([], T::Array[String])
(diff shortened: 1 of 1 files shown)


andyw8:
It would be more work, but I wonder if we should parse `.gitignore` and use the basis of what to skip?

vinistock:
There are some cases where you do want gitignored files to be indexed though. For example, Prism generates Ruby files using templates. If you're working on Prism itself, you'd miss those declarations.

Do you feel like having gitignored patterns by default and then including what you want manually is better? We can definitely try that.

andyw8:
Let's go with just `tmp` for now, and we can consider that for a future enhancement? It could get quite complex, for example here's what ripgrep does: https://github.com/BurntSushi/ripgrep/blob/master/GUIDE.md#automatic-filtering

andyw8 (approved):

vinistock:
Yeah, that seems quite complex indeed.

PR #1592, opened 2024-03-12
Use original Bundler environment to find workspace dependencies

vinistock:
### Motivation

We need to use `Bundler.with_original_env` when getting the gems for the dependencies view, otherwise we include things from the custom bundle, which are not a part of the application's dependencies.

### Implementation

Just wrapped the code in a call to `Bundler.with_original_env`.


paracycle (approved):
🤦 We knew that this was a problem when we were pairing, but then we complete missed adding this. Thank you!

PR #1600, opened 2024-03-13
Simplify magic comment regex

vinistock:
### Motivation

We actually spend a considerable amount of time checking each comment to see if they are a magic comment to exclude from the documentation while indexing. Since we always run this regex against comments, we can simplify it to achieve better performance.

### Implementation

We don't need the `^#\s*` part of the regex. We know for a fact the line we're checking will be a comment, we just care if the magic comment is there.

This is x1.81 faster than the previous implementation on my machine.

### Automated Tests

Current tests should cover it.


andyw8:
Could this give false positives in rare cases? e.g.:

```
ruby-lsp(main):003> "# Here is a tip for coding: have fun".match?(RubyIndexer.configuration.magic_comment_regex)
=> true
```

PR #1606, opened 2024-03-13
Fix no code lenses for minitest specs with constant paths

Earlopain:
### Motivation

A `describe` with a class path did not show any code lenses. This fixes that.

### Implementation

Add the `ConstantPathNode` as a case to be handled and extract the path with the `constant_name` helper to handle/guard against dynamic paths.

### Automated Tests

Yes

### Manual Tests

See the modified expectation file, its code lenses result in the correct command being executed.

### DIFF
--- a/lib/ruby_lsp/listeners/code_lens.rb
+++ b/lib/ruby_lsp/listeners/code_lens.rb
@@ -258,8 +258,8 @@ module RubyLsp
         name = case first_argument
         when Prism::StringNode
           first_argument.content
-        when Prism::ConstantReadNode
-          first_argument.full_name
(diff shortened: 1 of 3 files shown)


andyw8:
Not a blocker, but I realised we also aren't correctly handling nested defintions like:

```ruby
module Foo
  describe Bar do
    it 'it_class_constant_path'
  end
end
```

Earlopain:
That's a good point. I think the best thing to do here is to also make use of the `group_stack` for specs. That would also help with making the test command more specific when running specs.

I will explore that in a separate PR.

PR #1611, opened 2024-03-14
Add missing expectation for `minitest_with_dynamic_constant_path.rb`

Earlopain:
### Motivation

In #1511, I added a fixture but forgot to add the expectation for it. This adds that missing fixture.

I also added a test to verify that every fixture is used in at least one expectation so that this doesn't happen again. However, there are already a few where this is the case and I'm not certain if they are used elsewhere or if these are genuine mistakes.

I also discovered the rails.js file, for which usage was removed in https://github.com/Shopify/ruby-lsp/commit/670f9048fc7c662d7289e87a85007841a5275834

### Automated Tests

These are the fixtures without expectations:

```rb
[
  "aref_call_aref_assign",
  "aref_on_method_call",
  "call_then_aref_assignment",
  "class_reference_target",
  "constant_reference_target",
  "double_aref_with_assignment",
  "hash_safe_access",
  "keyword_splat_nil",
  "method_def_object_receiver",
  "nested_block_arguments",
  "return_multiple_statements",
]
```

Two others I added based on context clues but for these I'm not sure. I invite someone with more knowledge to build up on this PR and add expectations to make the test pass. If that is not possible since these are special somehow then I'll just remove the test and keep the added expectation I forgot.

### Manual Tests

N/A


vinistock:
I don't think we can check for this. The way our tests are executed, if you don't have an explicit expectation for a fixture, it just verifies that the feature doesn't break when being executed against it.

And we have several cases where we added fixtures just to ensure the basic functionality wasn't breaking on some unconventional Ruby syntax.

We can definitely remove the `rails.js` file, but not sure we can add this verification.

Issue #1613, opened 2024-03-14
Don't try to create document symbols if name is blank


andyw8:
### Description

As reported in https://github.com/Shopify/ruby-lsp/pull/1444#issuecomment-1992297264.

vinistock:
We're seeing this for multiple things indeed. I think your original suggestion of checking if the name is empty in `create_document_symbol` is sensible and we should put that bugfix out.

PR #1614, opened 2024-03-14
Fix blank name for document symbol

andyw8:
### Motivation

Closed https://github.com/Shopify/ruby-lsp/issues/1613


andyw8:
I'll update the other occurrences once @vinistock verifies the approach.

andyw8:
@vinistock saw your comment on the issue, will switch to that.

lib/ruby_lsp/listeners/document_symbol.rb
```diff
@@ -217,6 +217,9 @@ def on_def_node_enter(node)
           kind = name == "initialize" ? Constant::SymbolKind::CONSTRUCTOR : Constant::SymbolKind::METHOD
         end
 
+        # while code is being typed, the name can be blank, but this will cause an error in the extension.
+        return if name.empty?
```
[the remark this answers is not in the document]
andyw8: Oh I think I accidentally committed an older version, lemme fix.

lib/ruby_lsp/listeners/document_symbol.rb
```diff
@@ -47,12 +47,15 @@ def initialize(response_builder, dispatcher)
...
       sig { params(node: Prism::ClassNode).void }
       def on_class_node_enter(node)
-        @response_builder << create_document_symbol(
+        symbol = create_document_symbol(
           name: node.constant_path.location.slice,
           kind: Constant::SymbolKind::CLASS,
           range_location: node.location,
           selection_range_location: node.constant_path.location,
         )
+        return unless symbol
```
andyw8, line 56: These extra nil checks are a little clunky... another option might be to set `name` to some placeholder like `<blank>`.

lib/ruby_lsp/listeners/document_symbol.rb
```diff
@@ -347,7 +360,6 @@ def handle_alias_method(node)
           )
         elsif new_name_argument.is_a?(Prism::StringNode)
           name = new_name_argument.content
-          return if name.empty?
```
andyw8, line 350: tbc

lib/ruby_lsp/listeners/document_symbol.rb
vinistock: Oh, I like that idea. Much nicer than all of these early returns and allows the developer to find attributes or aliases that are missing names.

vinistock (approved):

andyw8:
Not working as expected, setting back to draft for now.

andyw8:
Ah, it's related to the selection range... think I have a fix.

andyw8:
On second thoughts, this is a bit awkward to handle and I don't think it's worth the effort to perfect the behaviour for incomplete syntax in all situations, as long as we're not causing errors.

PR #1648, opened 2024-03-15
Refactor refactor into server and base server

vinistock:
### Motivation

I believe introducing the `Executor` class was not a great pick. In essence, we need the server to have
- A nice separation between the basic machinery (receiving requests, sending responses, queues, etc) and the request execution logic
- A single entity that controls the entire global state of the LSP
- Ease of deciding which messages should be sent to the editor

The second point is an evident issue in our codebase right now. Instead of having global state information such as dependencies, the encoding or the default Gemfile name in an object owned by the server that gets passed to all requests, we are resorting to singletons or constants, which is a bad pattern.

In addition to that, our test are becoming more and more verbose because to run a request you need to first instantiate multiple pieces of state that should really be created and owned by the server.

This PR is a refactor proposal to fix these problems.

### Implementation

I split by commit to make it a bit easier to review. Here are the highlights of the implementation:

1. Created a `BaseServer`, which is an abstract class that implements all of the base machinery of the LSP, like handling queues, dispatching requests and shutting down
2. Moved all of the `Executor` implementation inside `Server`, which now only contains the logic for our requests
3. Another major mistake in our previous implementation is that our `execute` method always returned something. This resulted in the bad pattern of having to return `VOID` when processing notifications since the server is not allowed to return anything for those. It also resulted in other weirdness because requests couldn't easily send notifications in addition to returning a response. In this new approach, every message that the server sends to the editor is simply pushed to a thread queue and gets picked up by a dispatcher. This means that all request methods now return `void`. We need to explicitly invoke `send_message` with a `Result` to return something. And it also means that we can easily push notifications by invoking `send_message` before the request is complete

Desired outcomes:
- Easier to manage and share the global state of the LSP with requests
- Easier testing of the server
- Better structure to match the async nature of LSPs. We need to be able to send as many messages as desired in a request cycle (or none for notifications)

### Automated Tests

This PR massively simplifies our testing. We can now just invoke this helper method to get a test server instance.

The server instance:
- already indexes the given `source`
- already opens the given `source` as a document, so that we can make requests against it
- automatically shuts down even if assertions fail

```ruby
with_server(source) do |server, uri|
  server.text_document_hover(
    id: 1,
    params: { textDocument: { uri: uri }, position: { character: 3, line: 5 } }
  )
  server.pop_response # get the result object
end
```

### Manual Tests

Basically a full smoke test. Verify all features of the LSP are working properly. You might want to use

```json
{
  "rubyLsp.branch": "vs/server_refactor"
}
```

And then just test the LSP in an application.


paracycle (commented):
Am I missing something or is there no `RubyLsp::Server.new` being called anywhere except the `ruby-lsp-check` executable?

test/expectations/expectations_test_runner.rb
```diff
@@ -110,27 +110,33 @@ def ruby_requirement_magic_comment_version(fixture_path)
   private
 
   def test_addon(addon_creation_method, source:)
-    stub_no_typechecker
     message_queue = Thread::Queue.new
+    stub_no_typechecker
```
st0012, line 114: Is this a cosmetic change or there's some order-dependent side-effects?

st0012 (approved):
LGTM 👍 
I like how this cleans up the test setup.

test/expectations/expectations_test_runner.rb
vinistock: Just cosmetic. I actually think I was trying to remove `message_queue`, but then I realized we need it to load the addons.

PR #1649, opened 2024-03-15
Use stubs instead of configs in feature status test

vinistock:
### Motivation

This test is flaky on CI, likely because it keeps updating VS Code settings. We now have `sinon` in the extension, so let's use stubs to get a more predictable outcome with the enabled features status.


egiurleo (approved):
🥳

PR #1650, opened 2024-03-15
Remove path exists and use VS Code's file system

vinistock:
### Motivation

Another step towards https://github.com/Shopify/vscode-ruby-lsp/issues/1044. We're almost done.

This PR removes our `pathExists` helper, which used `fs` instead of VS Code's file system API under the hood.

### Implementation

Started using `vscode.workspace.fs.stat` to check if files exist.

### DIFF
--- a/vscode/src/rubyLsp.ts
+++ b/vscode/src/rubyLsp.ts
@@ -1,12 +1,10 @@
-import path from "path";
-
 import * as vscode from "vscode";
 import { Range } from "vscode-languageclient/node";
 
 import { Telemetry } from "./telemetry";
 import DocumentProvider from "./documentProvider";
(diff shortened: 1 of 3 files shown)


vscode/src/rubyLsp.ts
```diff
@@ -470,4 +466,26 @@ export class RubyLsp {
...
     }
   }
+
+  private async lockfileExists(workspaceUri: vscode.Uri) {
+    try {
+      await vscode.workspace.fs.stat(
+        vscode.Uri.joinPath(workspaceUri, "Gemfile.lock"),
+      );
+      return true;
+    } catch (error: any) {
```
andyw8, line 476: This feels ugly but I guess there's no better way.

PR #1796, opened 2024-03-18
Implement Windows activation as a manager object

vinistock:
### Motivation

Implement automatic Ruby environment activation for Windows as a manager object. 

### Implementation

The majority of version managers actually don't work on Windows. A popular option is `RubyInstaller`, which only install rubies, but doesn't allow switching versions.

The implementation basically uses all of the mechanisms from `chruby` to activate the environment on Windows, with the only difference being where we search for the installations. RubyInstaller places the rubies under `C:/Ruby33-x64` based on version and arch.

### Automated Tests

Added tests.

### DIFF
--- a/vscode/src/ruby.ts
+++ b/vscode/src/ruby.ts
@@ -10,6 +10,7 @@ import { Shadowenv } from "./ruby/shadowenv";
 import { Chruby } from "./ruby/chruby";
 import { VersionManager } from "./ruby/versionManager";
 import { Mise } from "./ruby/mise";
+import { RubyInstaller } from "./ruby/rubyInstaller";
 
 export enum ManagerIdentifier {
   Asdf = "asdf",
@@ -19,6 +20,7 @@ export enum ManagerIdentifier {
   Rvm = "rvm",
   Shadowenv = "shadowenv",
   Mise = "mise",
+  RubyInstaller = "rubyInstaller",
   None = "none",
   Custom = "custom",
 }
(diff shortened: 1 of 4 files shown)


vscode/src/ruby.ts
```diff
@@ -19,6 +20,7 @@ export enum ManagerIdentifier {
   Rvm = "rvm",
   Shadowenv = "shadowenv",
   Mise = "mise",
+  Windows = "windows",
```
andyw8, line 23: I wonder if we should be more specific here, e.g. `RubyInstallerWindows`.

vscode/src/ruby.ts
```diff
@@ -290,6 +297,11 @@ export class Ruby implements RubyInterface {
       // If the Mise binary doesn't exist, then continue checking
     }
 
+    if (os.platform() === "win32") {
```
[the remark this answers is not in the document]
vinistock: My understanding is that VS Code would be connected to the Linux subsystem and then this wouldn't return `win32`. Do you believe that's not the case?

vscode/src/ruby.ts
vinistock: My idea with going less specific is that if there's another popular way of installing Ruby on Windows we discover later, then we can just include it in the same implementation.

But I'm fine with making it more specific too. Do you prefer that?

vscode/src/ruby.ts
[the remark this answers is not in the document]
andyw8: Yeah that's probably right.

andyw8 (approved):

vscode/src/ruby.ts
vinistock: Renamed to `RubyInstaller`.

PR #1801, opened 2024-03-19
No-op Sorbet type assertions

vinistock:
### Motivation

We should no-op the Sorbet type assertions so that we don't pay the performance price of runtime type checks when running the LSP.

### Implementation

Added two modules to patch as much as we can from the Sorbet runtime and reduce the overhead from runtime type checks.

### DIFF
--- a/lib/ruby_lsp/disable_sorbet.rb
+++ b/lib/ruby_lsp/disable_sorbet.rb
@@ -1,0 +1,66 @@
+# typed: true
+# frozen_string_literal: true
+
+module RubyLsp
+  # No-op all inline type assertions defined in T
+  module InlineTypeAssertions
+    def cast(value, type, checked: true)
+      value
+    end
+
+    def let(value, type, checked: true)
+      value
+    end
+
+    def must(arg)
+      arg
+    end
+
+    def absurd(value)
+      value
+    end
+
+    def bind(value, type, checked: true)
+      value
+    end
+
+    def assert_type!(value, type, checked: true)
+      value
+    end
+
+    def any(type_a, type_b, *types)
+      T::Types::Union.new([type_a, type_b, *types])
+    end
+
+    def nilable(type)
+      T::Types::Union.new([type, T::Utils::Nilable::NIL_TYPE])
+    end
+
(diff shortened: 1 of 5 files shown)


lib/ruby_lsp/disable_sorbet.rb
```diff
@@ -0,0 +1,66 @@
...
+        bounds = {}
+        bounds[:fixed] = fixed unless fixed.nil?
+        bounds[:lower] = lower unless lower == T.untyped
+        bounds[:upper] = upper unless upper == BasicObject
+        -> { bounds }
+      end
+    end
+
+    T::Generic.prepend(self)
+  end
```
paracycle, line 65: I don't think we need this at all. This used to be a compatibility patch when Sorbet runtime decided to move from `Foo = type_member(fixed: String)` to `Foo = type_member {{ fixed: String }}`, but it has been a long time now and we really shouldn't need the compat patch anymore.

I even removed it from Tapioca at this point and will be removing it from Shopify Types as well.

lib/ruby_lsp/disable_sorbet.rb
```diff
@@ -0,0 +1,66 @@
...
+    end
+
+    def bind(value, type, checked: true)
+      value
+    end
+
+    def assert_type!(value, type, checked: true)
+      value
+    end
+
```
paracycle, line 30: I think we are missing `unsafe` here (and in Shopify Types):
```suggestion

    def unsafe(value)
      value
    end
```

lib/ruby_lsp/disable_sorbet.rb
vinistock: Oh, I had misunderstood the purpose of the patch indeed. https://github.com/Shopify/ruby-lsp/pull/1803.

lib/ruby_lsp/disable_sorbet.rb
paracycle: Oh ok, I hadn't thought of that. Still no harm in overriding the method in case upstream implementation changes at some point, I guess.

PR #1803, opened 2024-03-20
Remove unnecessary generic patch

vinistock:
### Motivation

I had misunderstood what this patch was doing. It's there for compatibility and not performance, which is not relevant for us.


vinistock:
Already implemented in #1802.

PR #1804, opened 2024-03-20
Fix failing test on main for DocumentSymbol

andyw8:
Was failing due to the change in https://github.com/Shopify/ruby-lsp/pull/1444/files#diff-7059e8ce6c342ec9a646c2f293ce120c65bdb933d70c81da97471b21c6465bdaR100


st0012 (approved):
Why was it not caught in the original PR tho?

PR #1805, opened 2024-03-20
Fix lazy document parsing and error reporting

vinistock:
### Motivation

We identified two issues after the server refactor in main:

1. Error reporting was failing, because the response was not in the expected structure
2. Trying to open an untitled Ruby file was failing because we were trying to run lazy document parsing event on text synchronization notifications (which we didn't do before the refactor)

This PR fixes both.

### Implementation

We never want to parse documents on text synchronization notifications, so I excluded them by checking against `textDocument/did`, which all of the notifications use.

For the error, the issue was simply that the fields have to be nested inside the `error` key.

### Manual Tests

2. Open a new untitled file
3. Switch the language mode to Ruby
4. Verify it does not fail because of lazy parsing

**Note**: I actually identified a bug in RuboCop while trying this out and you will see an `IndexError` when opening the empty file. RuboCop is trying to access some index in the source buffer which fails because the string is empty.

### DIFF
--- a/lib/ruby_lsp/base_server.rb
+++ b/lib/ruby_lsp/base_server.rb
@@ -36,17 +36,22 @@ module RubyLsp
     sig { void }
     def start
       @reader.read do |message|
+        method = message[:method]
+
+        # We must parse the document under a mutex lock or else we might switch threads and accept text edits in the
+        # source. Altering the source reference during parsing will put the parser in an invalid internal state, since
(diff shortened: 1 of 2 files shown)


andyw8 (approved):
All working as expected now.

Earlopain:
Looks like the RuboCop error is surfaced through #1459. I have openend https://github.com/rubocop/rubocop/pull/12802

PR #1806, opened 2024-03-20
Migrate document symbol integration test to use client

vinistock:
### Motivation

One of the main advantages of having the monorepo is that we can now write real integration tests that use the extension to launch the server and verify responses.

This PR migrates document symbol to use that approach, to get us started migrating the tests.

### Implementation

The idea is to do as much setup as possible once and then reuse it for all examples. I also moved installing `ruby-lsp` and `debug` to the CI script, so that we avoid doing it multiple times.


KaanOzkan (approved):

.github/workflows/ci.yml
```diff
@@ -109,6 +109,7 @@ jobs:
         env:
           DISPLAY: ":99.0"
         run: |
+          gem install ruby-lsp debug
```
andyw8, line 112: For contributors, should we document the need to run this locally?

PR #1808, opened 2024-03-20
Prevent RuboCop from rescuing Interrupt

vinistock:
### Motivation

We should apply the changes in https://github.com/Shopify/ruby-lsp/pull/1522 everywhere and prevent RuboCop from rescuing interrupts.

There's no scenario where we would want the runner to rescue that and it could prevent the server from stopping.

### Implementation

Removed the module and just put the implementation in the runner.


lib/ruby_lsp/requests/support/rubocop_runner.rb
```diff
@@ -92,6 +92,7 @@ def run(path, contents)
           @options[:stdin] = contents
 
           super([path])
+          raise Interrupt if aborting?
```
andyw8, line 95: Maybe worth an explanatory comment?

vinistock: Added one.

PR #1815, opened 2024-03-21
Migrate integration tests to use the client

vinistock:
### Motivation

A continuation of the approach in https://github.com/Shopify/ruby-lsp/pull/1806, migrating all of our integration tests to use the client.

I have high hopes that this will improve the flakiness we've been seeing for integration tests on Windows, which are often failing due to concurrent requests. With this new approach, we only consider the client as initialized once we receive the final indexing progress notification, which should give us deterministic results.

### Implementation

I migrated tests one at a time in separate commits. Some tests didn't make sense anymore or simply didn't actually test anything, so I removed a few.


aryan-soni (approved):
Can confirm that everything passes locally for me:

```
  Client
    ✔ document symbol
    ✔ document highlight
    ✔ hover
    ✔ definition
    ✔ semantic highlighting
    ✔ document link (102ms)
    ✔ workspace symbol (297ms)
    ✔ code lens
    ✔ diagnostic (271ms)
    ✔ folding range
    ✔ formatting (96ms)
    ✔ selection range
    ✔ on type formatting
    ✔ code actions
    ✔ code action resolve
```

As seen above, a few of them are indicated to take longer than the others, thought I'd note this

vinistock:
Yeah, some requests are slower than others and the test runner warns about that. Nothing to worry about as long as they don't exceed the timeout.

PR #1816, opened 2024-03-21
Enable add-ons to extend completion

aryan-soni:
### Motivation

Exposes the completion feature to add-ons.

### Implementation

Created a factory method for the feature, and updated the request to ensure the appropriate add-on config is in place.  

### Automated Tests

Created a unit test to ensure add-ons can indeed contribute to this feature.

### Manual Tests

You can test that the completion functionality itself hasn't been disrupted. You can also technically use this branch from an add-on like `ruby-lsp-rails` and leverage the completion feature there to test it, but this would require configuration similar to the other features there.


vinistock (approved):
Awesome!

PR #1818, opened 2024-03-22
Introduce global state

st0012:
### Motivation

By having a `GlobalState` object, we can gradually move scattered global states into a centralized location. For example, this PR moves the workspace uri from `Store` and the all dependency detector states into it.

After this is merged, we can move other `Store`'s states like `formatter`, `features_configuration`...etc. into it as well.

### Implementation

1. Create `RubyLsp::GlobalState` class
2. Move `Store`'s `workspace_uri` into `GlobalState` and expose the processed path as `workspace_path`
3. Remove `DependencyDetector` and move its logic into `GlobalState`
4. Move `Server#index` into `GlobalState` as well
5. We also fixed an issue caused by `SignatureHelp` using insufficient check for typechecker's presence on the file.

### DIFF
--- a/lib/ruby_lsp/test_helper.rb
+++ b/lib/ruby_lsp/test_helper.rb
@@ -18,7 +18,6 @@ module RubyLsp
     end
     def with_server(source = nil, uri = Kernel.URI("file:///fake.rb"), stub_no_typechecker: false, &block)
       server = RubyLsp::Server.new(test_mode: true)
-      server.process_message(method: "initialized")
       server.global_state.stubs(:typechecker).returns(false) if stub_no_typechecker
 
       if source


lib/ruby_lsp/global_state.rb
```diff
@@ -1,27 +1,38 @@
 # typed: strict
 # frozen_string_literal: true
 
-require "singleton"
-
 module RubyLsp
-  class DependencyDetector
-    include Singleton
+  class GlobalState
```
KaanOzkan, line 5: We could use [`Singleton`](https://docs.ruby-lang.org/en/3.3/Singleton.html) to prevent instantiation but we'd need a workaround for tests that are creating new ones.

KaanOzkan (approved):

lib/ruby_lsp/global_state.rb
vinistock: We're trying to move away from using `Singleton` as it makes everything accessible globally.

lib/ruby_lsp/listeners/completion.rb
```diff
@@ -10,16 +10,16 @@ class Completion
...
           typechecker_enabled: T::Boolean,
           dispatcher: Prism::Dispatcher,
           uri: URI::Generic,
         ).void
       end
-      def initialize(response_builder, index, nesting, typechecker_enabled, dispatcher, uri) # rubocop:disable Metrics/ParameterLists
+      def initialize(response_builder, global_state, nesting, typechecker_enabled, dispatcher, uri) # rubocop:disable Metrics/ParameterLists
         @response_builder = response_builder
-        @index = index
+        @global_state = global_state
```
vinistock, line 22: NIT: since we use the index quite a bit, how about we put it in an instance variable for convenience?

```suggestion
        @global_state = global_state
        @index = global_state.index
```

lib/ruby_lsp/listeners/definition.rb
```diff
@@ -10,18 +10,18 @@ class Definition
...
           nesting: T::Array[String],
-          index: RubyIndexer::Index,
           dispatcher: Prism::Dispatcher,
           typechecker_enabled: T::Boolean,
         ).void
       end
-      def initialize(response_builder, uri, nesting, index, dispatcher, typechecker_enabled) # rubocop:disable Metrics/ParameterLists
+      def initialize(response_builder, global_state, uri, nesting, dispatcher, typechecker_enabled) # rubocop:disable Metrics/ParameterLists
         @response_builder = response_builder
+        @global_state = global_state
```
vinistock, line 22: Same here, maybe we can put the index in a variable for convenience?

lib/ruby_lsp/listeners/hover.rb
```diff
@@ -28,18 +28,18 @@ class Hover
...
           nesting: T::Array[String],
-          index: RubyIndexer::Index,
           dispatcher: Prism::Dispatcher,
           typechecker_enabled: T::Boolean,
         ).void
       end
-      def initialize(response_builder, uri, nesting, index, dispatcher, typechecker_enabled) # rubocop:disable Metrics/ParameterLists
+      def initialize(response_builder, global_state, uri, nesting, dispatcher, typechecker_enabled) # rubocop:disable Metrics/ParameterLists
         @response_builder = response_builder
+        @global_state = global_state
```
vinistock, line 40: Same here, maybe we can put the index in a variable for convenience?

lib/ruby_lsp/listeners/signature_help.rb
```diff
@@ -10,27 +10,29 @@ class SignatureHelp
...
-          index: RubyIndexer::Index,
           dispatcher: Prism::Dispatcher,
+          typechecker_enabled: T::Boolean,
         ).void
       end
-      def initialize(response_builder, nesting, index, dispatcher)
+      def initialize(response_builder, global_state, nesting, dispatcher, typechecker_enabled)
+        @typechecker_enabled = typechecker_enabled
         @response_builder = response_builder
+        @global_state = global_state
```
vinistock, line 22: Same here, maybe we can put the index in a variable for convenience?

lib/ruby_lsp/requests/code_lens.rb
```diff
@@ -34,17 +34,18 @@ def provider
...
       end
-      def initialize(uri, dispatcher)
+      def initialize(global_state, uri, dispatcher)
         @response_builder = T.let(
           ResponseBuilders::CollectionResponseBuilder[Interface::CodeLens].new,
           ResponseBuilders::CollectionResponseBuilder[Interface::CodeLens],
         )
         super()
-        Listeners::CodeLens.new(@response_builder, uri, dispatcher)
+        Listeners::CodeLens.new(@response_builder, global_state, uri, dispatcher)
```
vinistock, line 48: We need to pass the global state to code lens listeners so that they can check the test framework.

lib/ruby_lsp/requests/workspace_symbol.rb
```diff
@@ -22,9 +22,10 @@ class WorkspaceSymbol < Request
       extend T::Sig
       include Support::Common
 
-      sig { params(query: T.nilable(String), index: RubyIndexer::Index).void }
-      def initialize(query, index)
+      sig { params(global_state: GlobalState, query: T.nilable(String), index: RubyIndexer::Index).void }
+      def initialize(global_state, query, index)
```
vinistock, line 26: Can we remove the index now that it's inside the global state?

lib/ruby_lsp/server.rb
```diff
@@ -577,12 +582,12 @@ def workspace_did_change_watched_files(message)
 
         case change[:type]
         when Constant::FileChangeType::CREATED
-          @index.index_single(indexable)
+          @global_state.index.index_single(indexable)
```
vinistock, line 585: NIT: we can put the index in a local variable and then avoid doing `@global_state.index` in every branch.

lib/ruby_lsp/server.rb
```diff
@@ -594,7 +599,11 @@ def workspace_symbol(message)
       send_message(
         Result.new(
           id: message[:id],
-          response: Requests::WorkspaceSymbol.new(message.dig(:params, :query), @index).perform,
+          response: Requests::WorkspaceSymbol.new(
+            @global_state,
+            message.dig(:params, :query),
+            @global_state.index,
```
vinistock, line 605: This is no longer necessary.

lib/ruby_lsp/store.rb
```diff
@@ -33,7 +30,6 @@ def initialize
       @formatter = T.let("auto", String)
```
vinistock, line 30: I think the formatter was moved to the global state. This has to be removed.

test/requests/completion_test.rb
```diff
@@ -800,24 +792,6 @@ def test_relative_completion_with_partial_path
     end
   end
 
-  def test_completion_addons
```
vinistock, line 803: Was removing this test intentional?

andyw8:
Since https://github.com/Shopify/ruby-lsp/pull/1833 is now merged, I think should be good to do a release after this to unblock ruby-lsp-rails?

PR #1814, opened 2024-03-21
Allow config to be passed into Indexer

andyw8:
(New version of https://github.com/Shopify/ruby-lsp/pull/1436)

### Motivation

We plan to rename `.index.yml` to `.ruby-lsp.yml` and use that for general configuration. As a first step, this PR refactors the indexer to accept a configuration hash, instead of reading the config file directly.

### Implementation

Pass the config as a hash.

### Automated Tests

Updated

### Manual Tests

* Open this branch
* Ensure indexing is working (e.g. the Jump to Definition works for constants)
* Edit `.index.yml` to make it invalid syntax and restart the LSP
* Ensure an error message is showing but otherwise everything works (e.g. formatting)
* Delete `.index.yml` and restart the LSP
* Ensure indexing completes successfully (e.g. the Jump to Definition works for constants)

PR #1827, opened 2024-03-25
Apply indexing configuration to ruby-lsp-doctor

andyw8:
### Motivation

Due to the changes in https://github.com/Shopify/ruby-lsp/pull/1814, the `ruby-lsp-doctor` tool needs to have the configuration loaded.

### Implementation

Check if an `.index.yml` exists, and load configuration if so.

### Automated Tests

Since this is a troubleshooting tool, I think tests are not critical.

### Manual Tests

- Create a file "foo/bar.rb"
- Run the index and pipe through grep to confirm `foo/bar.rb` is listed
- Update `.index.yml` to exclude `"**/foo**"`
- Run the index and pipe through grep to confirm `foo/bar.rb` is no longer listed

Notes:
- This will need adjusted again for https://github.com/Shopify/ruby-lsp/pull/1434
- I think I found a separate problem relating to relative paths while testing this. I will look into and create an issue if needed.

### DIFF
--- a/exe/ruby-lsp-doctor
+++ b/exe/ruby-lsp-doctor
@@ -5,8 +5,14 @@ $LOAD_PATH.unshift(File.expand_path("../lib", __dir__))
 require "ruby_lsp/internal"
 
 if File.exist?(".index.yml")
-  RubyIndexer.configuration.apply_config(YAML.parse_file(".index.yml").to_ruby)
+  begin
+    config = YAML.parse_file(".index.yml").to_ruby
+  rescue => e
+    abort("Error parsing config: #{e.message}")
+  end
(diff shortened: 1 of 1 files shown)


exe/ruby-lsp-doctor
```diff
@@ -4,6 +4,9 @@
 $LOAD_PATH.unshift(File.expand_path("../lib", __dir__))
 require "ruby_lsp/internal"
 
+if File.exist?(".index.yml")
+  RubyIndexer.configuration.apply_config(YAML.parse_file(".index.yml").to_ruby)
```
vinistock, line 8: This may file to parse. Let's rescue here and warn if the YAML file is invalid.

vinistock (approved):

PR #1828, opened 2024-03-25
Implement RVM as a manager object

vinistock:
### Motivation

Implement RVM support as a manager object.

### Implementation

Similar to Mise, we execute RVM from the default installation path. However, RVM doesn't actually activate anything in the environment variables when we run the script. `ENV.to_h` is basically empty.

Because of that, we need to return the essential parts of the environment ourselves from the script, to get the right `GEM_HOME`, `GEM_PATH` and `PATH`.

### Automated Tests

Added a test.

### DIFF
--- a/vscode/src/ruby/rvm.ts
+++ b/vscode/src/ruby/rvm.ts
@@ -1,0 +1,56 @@
+/* eslint-disable no-process-env */
+import os from "os";
+import path from "path";
+
+import { asyncExec } from "../common";
+
+import { ActivationResult, VersionManager } from "./versionManager";
+
+// Ruby enVironment Manager. It manages Ruby application environments and enables switching between them.
+// Learn more:
+// - https://github.com/rvm/rvm
+// - https://rvm.io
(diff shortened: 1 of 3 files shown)


vscode/src/ruby/rvm.ts
```diff
@@ -0,0 +1,56 @@
+/* eslint-disable no-process-env */
+import os from "os";
+import path from "path";
+
+import { asyncExec } from "../common";
+
+import { ActivationResult, VersionManager } from "./versionManager";
+
+// Ruby enVironment Manager. It manages Ruby application environments and enables switching between them.
```
andyw8, line 9: TIL it's not `Ruby Version Manager`!

andyw8 (approved):
(not tophatted since I don't have RVM setup)

vscode/src/ruby/rvm.ts
vinistock: It used to be called that apparently https://rvm.io/rvm/myths.

Issue #1822, opened 2024-03-25
Support custom `mise` installation paths. 


Jesse-Cameron:
### I have checked that this feature is not already implemented

- [X] This feature does not exist

### Use case

`ruby-lsp` recently [shipped out of the box support](https://github.com/Shopify/ruby-lsp/pull/1529) for the [`mise` version manager](https://mise.jdx.dev/) :partying_face: . however it doesn't support non-default installation locations. it would be helpful if the extensions allowed for non-default paths.

### Description

I would like to be able to override the binary path if they have installed `mise` somewhere other than `~/.local/bin/mise`.

### Implementation

After talking with @vinistock in slack. We have discussed an approach to installation:

 - First, migrate the current version manager configuration. It's currently just a string enum, but we need to allow for other configurations too. So, I would turn the config into an object that looks something like:
  ```
  "rubyLsp.versionManager": {
    "name": "mise",
  }
  ```
 - We can auto-correct people's configuration so that it's automatically migrated from the previous format into the new one
 - Then, we can add `miseExecutablePath` as part of the object, defaulting to `~/.local/bin/mise`

PR #1831, opened 2024-03-26
Enhance version manager config to accept more fields

vinistock:
### Motivation

Step towards https://github.com/Shopify/ruby-lsp/issues/1822

Because each version manager has its own customizations, I believe we need to offer more settings to match the behaviour in the editor. We currently only allow setting the identifier of the manager, but that's not always enough (e.g.: see the issue).

### Implementation

The idea is to change the `versionManager` setting so that it becomes an object and then we nest the identifier inside of that object. After this goes in, then we'll be able to add extra configuration fields, like the version manager executable path.

All of the changes are solely to migrate `versionManager` from a string to an object that looks like `{ identifier: "chruby" }`.

The only other change is the automated migration in `extension.ts`, where we read all of the existing configuration values and update them automatically to match the new format. That code is temporary and can be removed after it has been there for a while.

### Automated Tests

Adapted our existing tests.

### Manual Tests

2. Open your JSON settings
3. Verify that your version manager setting was migrated to the new format
4. verify the server boots properly

### DIFF
--- a/vscode/README.md
+++ b/vscode/README.md
@@ -85,7 +85,12 @@ by clicking `Change version manager` in the language status center or by changin
 // "rbenv"
 // "rvm"
 // "shadowenv"
-"rubyLsp.rubyVersionManager": "chruby"
+// "mise"
+{
+  "rubyLsp.rubyVersionManager": {
+    "identifier": "chruby",
+  },
+}
 ```
 
 To make sure that the Ruby LSP can find the version manager scripts, make sure that they are loaded in the shell's
--- a/vscode/VERSION_MANAGERS.md
+++ b/vscode/VERSION_MANAGERS.md
@@ -10,30 +10,20 @@ If you're using a different version manager that's not supported by this extensi
 executable into the PATH, you will probably need to define custom activation so that the extension can find the correct
 Ruby.
 
-For these cases, set `rubyLsp.rubyVersionManager` to `"custom"` and then set `rubyLsp.customRubyCommand` to a shell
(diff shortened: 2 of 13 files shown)


paracycle:
For reference, the older Ruby extension used that pattern for Ruby linter configuration: https://github.com/rubyide/vscode-ruby/blob/main/packages/vscode-ruby-client/package.json#L283-L343

You can see how `rubocop` is either a `boolean` or `object` with rich fields, so either `"ruby.lint": { "rubocop": true }` or `"ruby.lint": { "rubocop": { "command": "foo" } }` is valid configuration.

andyw8:
So we'd avoid the `identifier` key, and do this?

```
"rubyLsp.rubyVersionManager": { "chruby": { "foo": "bar" } }
```

paracycle:
> @paracycle that style only works because in both cases the configuration `lint` is an `object` type. In our case, one is a `string` type and the other an `object`, there's no way we could have both options at the same time.

Why can't `"rubyLsp.rubyVersionManager"` be of type `["string", "object"]`?

vinistock:
Talked with Ufuk on Slack. Unfortunately, if we use the `string` or `object` types, we lose the ability to provide enum completions.

In fact, you can add the enum and it does show the right completions, but then it considers the `{ identifier: }` version of the setting as invalid, which makes the experience a bit odd.

With the current implementation, old settings get migrated automatically and you get completion every step of the way, so I think this is a nicer approach.

PR #1833, opened 2024-03-26
Extract `RubyLsp::TestHelper` for addons to use

andyw8:
This will allow us to use `with_server` in addons, e.g. https://github.com/Shopify/ruby-lsp-rails/pull/301

### DIFF
--- a/test/test_helper.rb
+++ b/test/test_helper.rb
@@ -18,6 +18,7 @@ $LOAD_PATH.unshift(File.expand_path("../lib", __dir__))
 $VERBOSE = nil unless ENV["VERBOSE"] || ENV["CI"]
 
 require_relative "../lib/ruby_lsp/internal"
+require_relative "../lib/ruby_lsp/test_helper"
 require_relative "../lib/rubocop/cop/ruby_lsp/use_language_server_aliases"
 require_relative "../lib/rubocop/cop/ruby_lsp/use_register_with_handler_method"
 
@@ -40,6 +41,7 @@ Minitest::Reporters.use!(minitest_reporter)
 module Minitest
   class Test
     extend T::Sig
+    include RubyLsp::TestHelper
 
     Minitest::Test.make_my_diffs_pretty!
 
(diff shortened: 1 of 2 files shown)


test/test_case.rb
```diff
@@ -0,0 +1,41 @@
+# typed: true
+# frozen_string_literal: true
+
+# NOTE: This module is intended to be used by addons for writing their own tests, so keep that in mind if changing.
+
+module RubyLsp
+  module TestCase
```
andyw8, line 7: I've made this a module rather than a class so that it can used from addons that don't use minitest, such as `ruby-lsp-rspec.`

test/test_case.rb
```diff
@@ -0,0 +1,41 @@
...
+    extend T::Sig
+
+    sig do
```
andyw8, line 10: Should we remove all the Sorbet aspects from this, so that we're not forcing addon authors to use it?

test/test_case.rb
andyw8: (maybe `RubyLsp::TestHelpers` would be a better name)

lib/ruby_lsp/test_case.rb
```diff
@@ -0,0 +1,41 @@
+# typed: true
+# frozen_string_literal: true
+
+# NOTE: This module is intended to be used by addons for writing their own tests, so keep that in mind if changing.
+
+module RubyLsp
+  module TestCase
```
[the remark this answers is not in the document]
andyw8: But if it's a class then we couldn't use it for https://github.com/st0012/ruby-lsp-rspec/blob/main/spec/ruby_lsp_rspec_spec.rb, right?

test/test_helper.rb
```diff
@@ -18,6 +18,7 @@
 $VERBOSE = nil unless ENV["VERBOSE"] || ENV["CI"]
 
 require_relative "../lib/ruby_lsp/internal"
+require_relative "../lib/ruby_lsp/test_helper"
```
vinistock, line 21: I know it's not related to this PR, but can we please update all of these to use regular requires? The `lib` directory gets inserted into the `$LOAD_PATH` a few lines above, there's no need to use relative requires.

vinistock (approved):
Not strictly necessary to be included in this PR, but we should mention these test helpers in the addon documentation.

PR #1843, opened 2024-03-27
WIP: Move `test_mode` setting to `GlobalState`

andyw8:

### DIFF
--- a/lib/ruby_lsp/base_server.rb
+++ b/lib/ruby_lsp/base_server.rb
@@ -8,9 +8,8 @@ module RubyLsp
 
     abstract!
 
-    sig { params(test_mode: T::Boolean).void }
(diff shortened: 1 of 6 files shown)


lib/ruby_lsp/base_server.rb
```diff
@@ -22,11 +21,11 @@ def initialize(test_mode: false)
       @store = T.let(Store.new, Store)
       @outgoing_dispatcher = T.let(
         Thread.new do
-          unless test_mode
-            while (message = @outgoing_queue.pop)
-              @mutex.synchronize { @writer.write(message.to_hash) }
-            end
+          # unless @global_state.test_mode
```
andyw8, line 24: need to figure out how we access the global state here.

PR #1845, opened 2024-03-27
Implement ASDF as a manager object

vinistock:
### Motivation

Migrate ASDF to the manager object implementation. This is the last version manager to migrate.

The only parts left of the Ruby activation restructuring is handling no version manager, custom Ruby activation and allowing manual selection of the Ruby installation.

### Implementation

ASDF activation is similar to Rbenv. The main difference is that the `asdf.sh` file is not an executable and can only be sourced to allow invoking `asdf`.

In addition to that, the [asdf documentation](https://asdf-vm.com/guide/getting-started.html#_3-install-asdf) lists a number of different installation alternatives, all of which end up with the `asdf.sh` script in a different file path. I listed all possible paths described in the docs and we search for the script in those paths.

Then we source the script and use `asdf exec ruby` to run the activation script in the context of the activated Ruby. Like Rbenv, ASDF doesn't set `GEM_HOME` or `GEM_PATH` so we need to do that ourselves.

I tested this manually on a Spin instance where I installed ASDF.

### Automated Tests

Added a test.

### Manual Tests

2. Open a project with a `.tools-version` file
3. Verify the Ruby LSP boots properly using the specified Ruby version

### DIFF
--- a/vscode/src/ruby.ts
+++ b/vscode/src/ruby.ts
@@ -15,6 +15,7 @@ import { Rbenv } from "./ruby/rbenv";
 import { Rvm } from "./ruby/rvm";
 import { None } from "./ruby/none";
 import { Custom } from "./ruby/custom";
+import { Asdf } from "./ruby/asdf";
 
 export enum ManagerIdentifier {
   Asdf = "asdf",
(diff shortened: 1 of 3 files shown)


andyw8:
I use asdf on my personal machine so can try this later.

vinistock:
> ASDF also supports .ruby-version, will that work here?

It should. We're invoking the `asdf` executable, so it should handle whatever ASDF handles in the terminal too.

> I use asdf on my personal machine so can try this later.

Awesome! Please report back if everything is working.

PR #1849, opened 2024-03-27
Use translated Prism AST to run RuboCop

vinistock:
### Motivation

Use the Prism AST translator to reuse our existing AST to run RuboCop.

### Implementation

RuboCop doesn't have an API where we can simply ask "format this AST", so implementing this requires patching a handful of places so that we can pass the AST around.

The idea is that we instantiate the `ProcessedSource` ourselves and prevent the Prism translator from re-parsing the file if we already have an AST in hand.

### Questions

1. Is this the best way to pass the existing AST around? Could we be doing this in a more elegant way?
2. What do we need to do to prevent this from failing in older RuboCop versions that do not support Prism as a backend?

### Automated Tests

Will add tests.

### Manual Tests

3. Verify that diagnostics and formatting is working properly

### DIFF
--- a/lib/ruby_lsp/document.rb
+++ b/lib/ruby_lsp/document.rb
@@ -36,7 +36,7 @@ module RubyLsp
 
     sig { returns(Prism::ProgramNode) }
     def tree
-      @parse_result.value
+      T.unsafe(@parse_result.value).first
     end
 
     sig { returns(T::Array[Prism::Comment]) }
@@ -113,7 +113,7 @@ module RubyLsp
       ).returns([T.nilable(Prism::Node), T.nilable(Prism::Node), T::Array[String]])
     end
     def locate_node(position, node_types: [])
-      locate(@parse_result.value, create_scanner.find_char_position(position), node_types: node_types)
+      locate(T.unsafe(@parse_result.value).first, create_scanner.find_char_position(position), node_types: node_types)
     end
 
     sig do
--- a/lib/ruby_lsp/requests/support/rubocop_runner.rb
+++ b/lib/ruby_lsp/requests/support/rubocop_runner.rb
@@ -1,4 +1,4 @@
-# typed: strict
+# typed: true
 # frozen_string_literal: true
 
 begin
@@ -16,6 +16,8 @@ end
 if RuboCop.const_defined?(:LSP) # This condition will be removed when requiring RuboCop >= 1.61.
   RuboCop::LSP.enable
 end
+
+require "prism/translation/parser/rubocop"
 
 module RubyLsp
   module Requests
@@ -74,6 +76,7 @@ module RubyLsp
           @offenses = T.let([], T::Array[RuboCop::Cop::Offense])
           @errors = T.let([], T::Array[String])
           @warnings = T.let([], T::Array[String])
+          @parse_result = T.let(nil, T.nilable(Prism::ParseResult))
 
           args += DEFAULT_ARGS
           rubocop_options = ::RuboCop::Options.new.parse(args).first
@@ -82,14 +85,15 @@ module RubyLsp
           super(rubocop_options, config_store)
         end
 
-        sig { params(path: String, contents: String).void }
-        def run(path, contents)
+        sig { params(path: String, contents: String, parse_result: Prism::ParseResult).void }
+        def run(path, contents, parse_result)
           # Clear Runner state between runs since we get a single instance of this class
           # on every use site.
           @errors = []
           @warnings = []
           @offenses = []
           @options[:stdin] = contents
+          @parse_result = parse_result
 
           super([path])
 
@@ -107,6 +111,28 @@ module RubyLsp
         sig { returns(String) }
         def formatted_source
           @options[:stdin]
+        end
+
+        sig { params(file: String).returns(RuboCop::ProcessedSource) }
+        def get_processed_source(file)
+          config = @config_store.for_file(file)
+          parser_engine = config.parser_engine
+          return super unless parser_engine == :parser_prism
+
+          processed_source = T.unsafe(::RuboCop::AST::ProcessedSource).new(
+            @options[:stdin],
+            Prism::Translation::Parser::VERSION_3_3,
+            file,
+            parser_engine: parser_engine,
+            prism_result: @parse_result,
+          )
+          processed_source.config = config
+          processed_source.registry = mobilized_cop_classes(config)
+          # We have to reset the result to nil after returning the processed source the first time. This is needed for
+          # formatting because RuboCop will keep re-parsing the same file until no more auto-corrects can be applied. If
+          # we didn't reset it, we would end up operating in a stale AST
+          @parse_result = nil
+          processed_source
         end
 
         class << self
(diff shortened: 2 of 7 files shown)


vinistock:
Tagging @koic and @kddnewton. I'd love to hear your thoughts on the approach and if there's something that can be upstreamed to Prism or RuboCop to make this integration more seamless.

lib/ruby_lsp/document.rb
```diff
@@ -36,7 +36,7 @@ def initialize(source:, version:, uri:, encoding: Constant::PositionEncodingKind
 
     sig { returns(Prism::ProgramNode) }
     def tree
-      @parse_result.value
+      T.unsafe(@parse_result.value).first
```
kddnewton, line 39: Probably pulling this out into its own instance variable/reader would be good

sorbet/rbi/shims/rubocop.rbi
```diff
@@ -0,0 +1,6 @@
+# typed: true
+
+class Prism::Translation::Parser33; end
```
kddnewton, line 3: Let's get this pulled into prism's RBI's, this is an oversight.

lib/ruby_lsp/requests/support/rubocop_runner.rb
```diff
@@ -138,3 +164,107 @@ def file_finished(_file, offenses)
...
+    ast&.complete!
+    tokens.map! { |t| RuboCop::AST::Token.from_parser_token(t) }
+
+    [ast, comments, tokens]
+  end
+
+  RuboCop::AST::ProcessedSource.prepend(self)
+end
+
+module Prism
```
kddnewton, line 207: This whole patch can probably just move into prism. We can have the parser accept an AST. I don't think we should maintain this patch inside ruby-lsp.

kddnewton (changes requested):
Looks good. In general I think we should provide the ability to pass through a parsed AST to `Prism::Translation::Parser` in prism itself instead of patching it here. One way to do this would be change it in prism to do:

```ruby
def parse(source_buffer)
  ...
  Prism.parse(...)
  ...
end
```

instead to:

```ruby
def parse(source_buffer)
  ....
  prism_parse(...)
  ...
end

def prism_parse(...)
  Prism.parse(...)
end
```

and then subclass it in Ruby LSP and modify that method.

lib/ruby_lsp/requests/support/rubocop_runner.rb
```diff
@@ -109,6 +113,28 @@ def formatted_source
...
+          parser_engine = config.parser_engine
+          return super unless parser_engine == :parser_prism
+
+          processed_source = T.unsafe(::RuboCop::AST::ProcessedSource).new(
+            @options[:stdin],
+            Prism::Translation::Parser::VERSION_3_3,
+            file,
+            parser_engine: parser_engine,
+            prism_result: @parse_result,
+          )
```
vinistock, line 128: We could avoid some patching if there was a way to instantiate the `ProcessedSource` with an existing AST and token list. @koic, would you be open to accepting that?

It would allow us to avoid some patching.

vinistock: All features in the Ruby LSP depend only the Prim's parse result, so we keep that in our representation of documents.

For RuboCop, we just [invoke run with the file contents](https://github.com/Shopify/ruby-lsp/blob/049d0e69d2523b7f54a349b4aa1e5e78b4e6c988/lib/ruby_lsp/requests/support/rubocop_runner.rb#L94), which will let it re-parse the file and do everything it needs to do.

The Ruby LSP supports multiple formatters. We hook into RuboCop if it is available, but it's not a dependency. So we can't depend on RuboCop internals for our basic document representations, like using a `ProcessedSource`. All RuboCop things are limited to the runner here.

Now that RuboCop supports the Prism backend, we're trying to find what the easiest way of reusing the existing AST is, so that we can stop parsing documents twice.

The ideal API would be something like what Syntax Tree supports, which allows you to format a given AST. That would not only simplify the code here a lot, it would enable RuboCop to support range formatting for LSPs, which is currently impossible.

That said, the ideal API may require significant refactors, which is why I suggested something a bit simpler that will still minimize the number of patches. If we can instantiate the `ProcessedSource` in a way that allows us to reuse the existing AST and prevents the second parse, that'll already simplify the code quite a bit.

marcandre: But a ProcessedDocument also holds the tokens...

koic: The use case this PR aims to address appears reasonable to me. When using a Ruby code with multiple tools (e.g., RuboCop, TypeProf, Steep...), parsing the same source code to an AST in each tool becomes a redundant process. In such cases, it makes sense to reuse the one AST across different tools.

In this PR, the only change to the interface for reusing the AST is the addition of the `prism_result` keyword argument. This means that backward compatibility is maintained, and there is no impact on the existing RuboCop (AST).

In summary, I think this proposal can be accepted by RuboCop and RuboCop AST. Considering this is not a specialized requirement but a backend tool for LSP, it would be more rational to provide a published interface rather than having users write monkey patches. Below, I share one concern and an idea related to it.

## NOTE: Naming and the two ASTs (whitequark AST and Prism AST)

A point of caution is that the current AST RuboCop can analyze is only the whitequark/parser AST interface (`Prism::Translation::Parser`). It cannot process the vanilla Prism AST interface.

Therefore, I'm not sure if `prism_result` is the best name for the AST argument (since it's not the Prism AST). The keyword argument name could be `ast`, which is simpler, but naming is difficult and I'm not confident that it's better 😅

Here is a proposed design for the `ast` parameter, assuming future RuboCop can use Prism AST:

- Current situation: When the `ast: ast` parameter is passed along with either `parser_engine: :parser_whitequark` or `parser_engine: parser_prism`, the AST is processed as a whitequark AST.
- Future extension: When the `ast: ast` parameter and `parser_engine: :prism` are passed, the AST would be processed as a Prism AST (note that `parser_engine: prism` is not yet implemented).

This would potentially allow for a future interface where RuboCop can use a direct Prism AST without `Prism::Translation::Parser`, and handle any incompatibilities in ASTs.

koic:
@vinistock 

Ruby LSP currently has over 1.3M+ installations, making this optimization meaningful for users. A key point is that users can benefit from it without any breaking changes.  

To implement this proposal, the following PRs have been opened for Prism, RuboCop AST, and RuboCop:  

- https://github.com/ruby/prism/pull/3478
- https://github.com/rubocop/rubocop-ast/pull/359
- https://github.com/rubocop/rubocop/pull/13899

As described in this PR, the return value of `document.parse_result` must be `Prism::ParseLexResult`. For more details, refer to the sample in https://github.com/rubocop/rubocop/pull/13899.  

Additionally, if `ProcessedSource` is used directly from Ruby LSP, refer to https://github.com/rubocop/rubocop-ast/pull/359. Pay particular attention to the keyword argument `prism_result` and the requirement that its value must be `Prism::ParseLexResult`.

cc @bbatsov @marcandre @kddnewton

bbatsov:
@koic I'm totally in favor of doing the necessary changes to RuboCop and `rubocop-ast`.

koic:
@vinistock RuboCop AST https://github.com/rubocop/rubocop-ast/pull/359 introduced an API for `ProcessedSource` to reuse `PrismLexResult`. This is available in rubocop-ast 1.39 or later. However, actual reuse requires Prism https://github.com/ruby/prism/pull/3478, which has not been released yet. In any case, I hope these changes eliminate the need for individual hacks.

koic:
> However, actual reuse requires Prism https://github.com/ruby/prism/pull/3478, which has not been released yet.

It is included in the released Prism 1.4.
https://rubygems.org/gems/prism/versions/1.4.0

vinistock:
Closing in favour of #3584

PR #1850, opened 2024-03-27
Load addons in with_server by default

st0012:
### Motivation

After the server re-architecture, calling Addon.load_addons outside of server with another message queue doesn't reflect how things really work outside of tests.

And since we want with_server to be a good end-to-end helper, especially for hiding the server initialization details, we should load addons by default so addon authors don't have to worry about it.

### Implementation

1. Add `Server#load_addons` method
2. Call the above method inside `with_server` helper
    - I decided to do this instead of sending `initialized` message because that'll also start indexing and send progress responses
3. Remove `test_addon` test helpers


test/requests/code_lens_expectations_test.rb
```diff
@@ -130,11 +130,13 @@ def test_code_lens_addons
       class Test < Minitest::Test; end
     RUBY
 
-    test_addon(:create_code_lens_addon, source: source) do |server|
+    create_code_lens_addon
```
st0012, line 133: I know it's a bit weird to rely on `with_server`'s ensure block to clear the anonymous addon implicitly. But IMO the weirdness comes these tests' need of 1-time addons. It won't be the case in real addons' tests.

st0012:
I've made the addons loading behaviour skippable after chatting with @vinistock.

andyw8 (approved):

andyw8:
(let's update the PR title?)

vinistock (approved):

st0012:
I see. The reasoning is that if we make it opt-in, addon authors basically need to add `load_addons: true` in every test, which is an odd design IMO. That plus the fact that we haven't had a case to not load addons (either for testing scenario or performance reason), we decided to make it opt-out instead.

Issue #1758, opened 2024-02-28
Switch to using VS Code's file system API instead of fs


vinistock:
We are incorrectly using Node's `fs` module to interact with files. This module will always only consider local files and not take into account VS Code's configuration, if it's being used in a remote setup.

We need to use the file system API provided through [vscode.workspace.fs](https://code.visualstudio.com/api/references/vscode-api#FileSystem) for all of our file interactions.

PR #1853, opened 2024-03-28
Use VS Code's file API to check for workspace permissions

vinistock:
### Motivation

Closes https://github.com/Shopify/ruby-lsp/issues/1758

This PR removes our final usage of `fs` and now we fully favour VS Code's file system API.

### Implementation

Switched to using `workspace.fs.stat` to check the permissions of the workspace URI. The API is a bit odd at the moment.

Permissions is `undefined` if VS Code has full permissions. If it's readonly, then permissions is a bitmask. However, the only value currently included in the bitmask enum is readonly, which means TypeScript will complain if I try to make any bitmask checks saying that it's a useless comparison on literal values.

For now, if permissions is set, we're considering that it will be set to readonly, given that's the only permission available.

### DIFF
--- a/vscode/src/workspace.ts
+++ b/vscode/src/workspace.ts
@@ -1,5 +1,3 @@
-import fs from "fs/promises";
-
 import * as vscode from "vscode";
 import { CodeLens, State } from "vscode-languageclient/node";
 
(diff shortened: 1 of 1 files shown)


vscode/src/workspace.ts
```diff
@@ -56,7 +54,17 @@ export class Workspace implements WorkspaceInterface {
     }
 
     try {
-      await fs.access(this.workspaceFolder.uri.fsPath, fs.constants.W_OK);
+      const stat = await vscode.workspace.fs.stat(this.workspaceFolder.uri);
+      const permissions = stat.permissions;
```
paracycle, line 58: Nitpick: Do we really need this local variable? I guess we could just check for `stat.permissions` inside the `if` clause below, right?

vinistock: Yeah, that's better.

PR #1872, opened 2024-04-03
Deactivate addons automatically in `with_server`

vinistock:
### Motivation

Now that we load addons automatically by default, we need to ensure we call `shutdown` instead of `run_shutdown`, because that method deactivates addons.

Otherwise, in https://github.com/Shopify/ruby-lsp-rails/pull/301, each test will spawn a new runner client and nothing will ensure the addon gets properly deactivated.

### DIFF
--- a/lib/ruby_lsp/test_helper.rb
+++ b/lib/ruby_lsp/test_helper.rb
@@ -46,7 +46,7 @@ module RubyLsp
         RubyLsp::Addon.addons.each(&:deactivate)
         RubyLsp::Addon.addons.clear
       end
-      T.must(server).run_shutdown
+      T.must(server).shutdown
     end
   end
 end
--- a/lib/ruby_lsp/server.rb
+++ b/lib/ruby_lsp/server.rb
@@ -108,6 +108,11 @@ module RubyLsp
       end
     end
 
+    sig { override.void }
+    def shutdown
+      Addon.addons.each(&:deactivate)
+    end
+
     private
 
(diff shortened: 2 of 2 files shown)


lib/ruby_lsp/test_helper.rb
```diff
@@ -46,7 +46,7 @@ def with_server(source = nil, uri = Kernel.URI("file:///fake.rb"), stub_no_typec
         RubyLsp::Addon.addons.each(&:deactivate)
         RubyLsp::Addon.addons.clear
       end
-      T.must(server).run_shutdown
+      T.must(server).shutdown
```
st0012, line 49: Aren't addons already deactivated in the condition above?

st0012 (commented):
What's the difference between `shutdown` and `run_shutdown`? They don't seem to be   exchangeable?

lib/ruby_lsp/test_helper.rb
vinistock: Oh, I missed that we were invoking deactivation manually. In that case, we probably don't need this PR.

vinistock:
> What's the difference between shutdown and run_shutdown? They don't seem to be exchangeable?

The `BaseServer#run_shutdown` handles all of the threads and queues shutting down. Basically, anything that would be generic to any LSP. `Server#shutdown` is invoked first to handle specific logic for the Ruby LSP. The separation is there just to avoid mixing the foundations with our specific logic.

Issue #1251, opened 2023-12-13
Better handle multibyte character locations


vinistock:
The locations returned by Prism are all byte locations. However, editors expect different values depending on the selected encoding.

For UTF-8, we need to consider the length of characters, considering emoji and other 3/4 byte characters as length 1. For UTF-16 like VS Code, we need to consider 3/4 byte characters as length 2.

Currently, we're using Prism locations directly everywhere, which is incorrect for documents containing multibyte characters. We need some sort of abstraction that will return the right column positions taking the encoding into account.

vinistock:
Talk to Kevin about whether we can solve this on Prism.

vinistock:
Paired with Kevin and we proposed a [new code units API in Prism](https://github.com/ruby/prism/pull/2406). That's going to make solving this issue in the entire LSP much more easy.

We can make the encoding selected by the editor available everywhere inside the `RubyLsp` namespace and then always pass that to the code unit column methods, so that we get the positions expected by UTF-16 clients such as VS Code.

vinistock:
Fixed by #2619.

Definition, hover, signature help, completion and workspace symbol should now properly work with multibyte characters thanks to the fixes to the indexer.

We also fixed the locations returned by semantic highlighting in a previous PR.

For now, I'm going to close this issue as complete, but if anyone notices any other location issues for multibyte characters please report.

PR #1879, opened 2024-04-03
Move encoding into global state

vinistock:
### Motivation

First step for #1251

Move the editor's selected encoding to the global state, so that we can start using it in requests to get the code unit locations.

### Implementation

Removed encoding from store and moved it to global state.

### DIFF
--- a/lib/ruby_lsp/document.rb
+++ b/lib/ruby_lsp/document.rb
@@ -20,13 +20,13 @@ module RubyLsp
     sig { returns(URI::Generic) }
     attr_reader :uri
 
-    sig { returns(String) }
+    sig { returns(Encoding) }
     attr_reader :encoding
 
-    sig { params(source: String, version: Integer, uri: URI::Generic, encoding: String).void }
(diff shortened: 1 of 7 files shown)


lib/ruby_lsp/document.rb
```diff
@@ -209,7 +209,7 @@ def find_char_position(position)
         # need to adjust for surrogate pairs
         requested_position = @pos + position[:character]
 
-        if @encoding == Constant::PositionEncodingKind::UTF16
+        if @encoding == Encoding::UTF_16LE
```
andyw8, line 212: Why `Encoding::UTF_16LE` rather than `Encoding::UTF_16`?

vinistock: I honestly can't remember the exact reason, but when Kevin and I were implementing the code units API we realized that using the generic UTF_16 did not work properly. It needed to specify `LE` or `BE` and since VS Code uses `UTF_16LE`, we use that.

andyw8 (approved):

aryan-soni (approved):

st0012 (approved):

PR #1895, opened 2024-04-05
Use selected encoding for computing semantic tokens

### LINKED ISSUE #1162: String with emoji in it throws off semantic highlighting
### Ruby version

3.1.4

### Code snippet

puts "🌱 string with emoji #{Rails.env}" }

### Description

I'm unsure if this is a known issue and/or if this is the LSP's issue. I still thought to report to you because you are the experts here 😅 .

I've tested this error by disabling ligatures and using Nerd Fonts; the results are unaffected.

I'm also using some custom TextMate rules, so I've disabled those, but the results were the same.

When there are no emojis in a string, the highlighting and other information features the LSP provides work just fine.

Code_RIsByfeLKG

When I add an emoji, the text analysis is broken:

Code_94sXNq2hox

Code_277qFtMNCX

No matter what kind of emoji I'm using:

Code_b17C2FNsQp

Also if I'm adding more emojis, it only seems to nudge the parsing to the right direction, only until I delete them or add more 😓 

Code_rn2p2F8Qyo

Code_a0WYtBlmg0

### Expected output

No matter how many emojis I use in a string, the semantic highlighting and other features provided by the LSP works the same way as with text and code without emoji.

vinistock:
### Motivation

Closes #1162

Another step towards #1251

For UTF-16 encoded editors, we need to return semantic token information taking that encoding into consideration. Consider the following example

```ruby
def 👋
  "hello"
end

👋.upcase
```
In the last line of the script, we have two method calls and we should return one semantic token for each. If the editor's encoding is UTF-8, then the expected emoji length is 1. If it's UTF-16, then the expected length is 2, which changes both the length of the emoji token and the delta start offset of the `upcase` token.

### Implementation

1. Updated our minimum Prism requirement to `v0.23`. The new code unit APIs are only available starting on that version
2. Started passing the global state down to semantic highlighting, so that we can access the encoding
3. In the response builder, started using the code unit information to compute the locations of the tokens

### Automated Tests

Added a test with some multibyte characters with some help from @st0012.

### Manual Tests

1. Start the LSP on `main`
2. Create a file
3. Copy the contents of the new multibyte fixture
4. Paste it in the file
5. Verify that the highlighting is totally thrown off by the multibyte characters

Do the same on this branch and verify that the highlighting is now correct.

### DIFF
--- a/lib/ruby_lsp/requests/semantic_highlighting.rb
+++ b/lib/ruby_lsp/requests/semantic_highlighting.rb
@@ -40,10 +40,13 @@ module RubyLsp
         end
       end
 
-      sig { params(dispatcher: Prism::Dispatcher, range: T.nilable(T::Range[Integer])).void }
-      def initialize(dispatcher, range: nil)
+      sig { params(global_state: GlobalState, dispatcher: Prism::Dispatcher, range: T.nilable(T::Range[Integer])).void }
+      def initialize(global_state, dispatcher, range: nil)
         super()
-        @response_builder = T.let(ResponseBuilders::SemanticHighlighting.new, ResponseBuilders::SemanticHighlighting)
+        @response_builder = T.let(
+          ResponseBuilders::SemanticHighlighting.new(global_state.encoding),
+          ResponseBuilders::SemanticHighlighting,
+        )
         Listeners::SemanticHighlighting.new(dispatcher, @response_builder, range: range)
 
         Addon.addons.each do |addon|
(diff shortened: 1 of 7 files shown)


KaanOzkan (approved):
Code lgtm but I couldn't reproduce the bug in VSCode

aryan-soni (approved):
Tested again.

Main branch:
image

This branch:

image

I now notice the difference with the `叫聲` symbol; had initially missed it. Code also LGTM

PR #1896, opened 2024-04-06
Detect rails as the testing framework if direct dependency on railties

Earlopain:
### Motivation

I have an application that depends on Rails in the following way:

```rb
# Gemfile
rails_version = "~> 7.1.0"
gem "actionpack",    rails_version
gem "actionview",    rails_version
gem "activemodel",   rails_version
gem "activerecord",  rails_version
gem "activesupport", rails_version
gem "railties",      rails_version
```

The intent is to not pull in the other framework parts, like activestorage for example. Since the testing framework detection currently does a check for rails I am not getting code lenses for tests with ruby-lsp-rails.

### Implementation

Also check for `railties`. A rails app doesn't boot if `railties` isn't present (`require "rails/commands"` in `bin/rails` is the first failure point).

### Automated Tests

Yup.

### Manual Tests

Create a rails app with the above gemfile, create a test, observe code lense.


andyw8:
Thanks for the PR. I have a slight concern that there could be some gems that depend on `railties` but use plain minitest. 🤔

lib/ruby_lsp/global_state.rb
```diff
@@ -104,8 +104,10 @@ def detect_test_library
       if direct_dependency?(/^rspec/)
         "rspec"
       # A Rails app may have a dependency on minitest, but we would instead want to use the Rails test runner provided
-      # by ruby-lsp-rails.
-      elsif direct_dependency?(/^rails$/)
+      # by ruby-lsp-rails. A Rails app doesn't need to depend on the rails gem itself, individual components like
+      # activestorage may be added to the gemfile so that other components aren't downloaded. Check for the presence
+      #  of bin/rails to support these cases.
+      elsif direct_dependency?(/^rails$/) || File.exist?(File.join(workspace_path, "bin/rails"))
```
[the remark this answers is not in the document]
st0012: Yes Rails engines should NOT depend on `rails` but only `railties`.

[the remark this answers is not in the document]
andyw8: > However, the command being executed will always be bin/rails, so it doesn't work for non-mountable engines right now anyways.

Yes, good point. So this wouldn't break any current behaviour. Let's go with @st0012's suggestion then.

st0012 (approved):

lib/ruby_lsp/global_state.rb
```diff
@@ -104,8 +104,10 @@ def detect_test_library
       if direct_dependency?(/^rspec/)
         "rspec"
       # A Rails app may have a dependency on minitest, but we would instead want to use the Rails test runner provided
-      # by ruby-lsp-rails.
-      elsif direct_dependency?(/^rails$/)
+      # by ruby-lsp-rails. A Rails app doesn't need to depend on the rails gem itself, individual components like
+      # activestorage may be added to the gemfile so that other components aren't downloaded. Check for the presence
+      #  of bin/rails to support these cases.
+      elsif File.exist?(File.join(workspace_path, "bin/rails"))
```
vinistock, line 110:
```suggestion
      elsif File.exist?(File.join(workspace_path, "bin", "rails"))
```

vinistock:
This PR will close #1466, so let's add to the description.

lib/ruby_lsp/global_state.rb
st0012: Ah sorry I clicked merge at the same time 😅

Earlopain:
I don't think it will close this? There are still a few other places that check against a rails dependecy here.

lib/ruby_lsp/global_state.rb
vinistock: Let's apply this change. We can't use forward slashes for Windows.

Earlopain: Hm, is that really the fix then? https://ruby-doc.org/3.3.0/File.html#method-c-join doesn't mention this being platform-aware and seems to always use `/` anyways.

vinistock: Then maybe we should be using `Pathname` instead and need to switch this to

```suggestion
      elsif Pathname.new(workspace_path).join("bin").join("rails").exist?
```

PR #1909, opened 2024-04-10
[VSCode] Change the order of existence checks for Manager. Check 'asdf' last.

dlwr:
### Motivation

It seems many people use asdf for managing environments other than Ruby. Chruby, rbenv, and rvm are mutually exclusive, and it's unlikely for someone to use two of them simultaneously. However, it's possible to use one of them in combination with asdf. For example, using rbenv for Ruby and asdf for bun. In such cases, it would be preferable to detect rbenv first. Therefore, I thought it would be best to check the other three before asdf, leaving asdf to be checked last.

### Implementation

I simply changed the order of checks.

### Automated Tests

I didn't think they were necessary, but please let me know if they are.

### Manual Tests

On an environment with both asdf and rbenv installed, set `"rubyLsp.rubyVersionManager": "auto"` and check if rbenv is detected. That would be considered OK. Below are screenshots from testing in my environment.

| Before | After |
| --- | --- |
| image | image | 

---

If there's anything I might have overlooked or any initial documentation that I should review first, I would appreciate it if you could let me know.

### DIFF
--- a/vscode/src/ruby.ts
+++ b/vscode/src/ruby.ts
@@ -288,10 +288,10 @@ export class Ruby implements RubyInterface {
     }
 
     const managers = [
-      ManagerIdentifier.Asdf,
       ManagerIdentifier.Chruby,
       ManagerIdentifier.Rbenv,
       ManagerIdentifier.Rvm,
+      ManagerIdentifier.Asdf,
     ];
 
     for (const tool of managers) {


KaanOzkan (approved):

vinistock (approved):
Thank you for the contribution! It makes sense to check for Ruby specific tools first indeed

PR #1912, opened 2024-04-10
Ignore non-ruby files in file watcher

Earlopain:
### Motivation

Currently there is only a listener for ruby files but that might changein the future. Addons are also able to register additional file watcher events, config files as an example.

I do not think that having non-rb files being passed to the indexer has any disadvantage, it just won't be parsed down to anything meaningful. You can also just paste giberish into a file and call it `.rb`. Still, this guard enforces my expectations of what the filepath is actually supposed to look like.

### Implementation

Simply bail out if the file doesn't end with `.rb`.

### Automated Tests

Validate that an update event for `.rubocop.yml` doesn't end up going to the indexer.

### Manual Tests

Make some changes to a file and see that the index is still being updated.

### DIFF
--- a/lib/ruby_lsp/server.rb
+++ b/lib/ruby_lsp/server.rb
@@ -610,6 +610,7 @@ module RubyLsp
         uri = URI(change[:uri])
         file_path = uri.to_standardized_path
         next if file_path.nil? || File.directory?(file_path)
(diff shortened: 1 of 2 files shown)


vinistock (approved):
Good catch. Thanks for the contribution!

PR #1914, opened 2024-04-11
Allow configuring Mise executable path

### LINKED ISSUE #1822: Support custom `mise` installation paths.

vinistock:
### Motivation

Closes #1822

Allow configuring the Mise executable path if it's not installed in the default location.

### Implementation

Added a new setting under version manager. If the value is set to something, we use that as the Mise path, otherwise we fallback to the default installation path.

### Automated Tests

Added a test to verify that we use the configured value to invoke Mise.

### DIFF
--- a/vscode/package.json
+++ b/vscode/package.json
@@ -249,6 +249,10 @@
                 "custom"
               ],
               "default": "auto"
+            },
+            "miseExecutablePath": {
+              "description": "The path to the Mise executable, if not installed in ~/.local/bin/mise",
+              "type": "string"
             }
           },
           "default": {
--- a/vscode/src/ruby/mise.ts
+++ b/vscode/src/ruby/mise.ts
@@ -34,15 +34,22 @@ export class Mise extends VersionManager {
   }
 
   async findMiseUri(): Promise<vscode.Uri> {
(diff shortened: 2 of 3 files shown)


egiurleo (approved):
Nice!

Issue #1009, opened 2023-09-14
Explore caching index for gems


vinistock:
Caching the index for gems is easier than caching the entire index because we know the files will only change if the version of the gem has changed.

I think we can get a considerable performance improvement on indexing if we cache the entries for gems. Something like

```
.ruby-lsp/
  Gemfile
  Gemfile.lock
  cache/
    rails-7.1.0
    yarp-0.12.0
```
Where each gem cache is basically a Marshal dump of the index. This may require defining a way to merge different indices.

For the long term, it would be amazing if rubygems could generate the index cache during packaging, so that all gems are exported with an index by default. By doing the work ahead of time, we would be guaranteed to always have cached indices for all gems, significantly speeding up indexing.

vinistock:
We would depend on every gem maintainer to use the bundler plugin, which wouldn't really scale.

andyw8:
It could be distributed as a gem which Ruby LSP installs in the `.ruby-lsp` bundle.

vinistock:
Would bundler pick up the plugin automatically? Even from a different Gemfile?

PR #1917, opened 2024-04-11
Refactor global usage of `Prism::Location` to minimize memory usage

aryan-soni:
### Motivation

Currently, we store `Prism::Location` objects globally to keep track of certain location specifications for indexed entries.

These location specifications are the Start Line, End Line, Start Column, and End Column, all of which are integers.

However, `Prism::Location` objects themselves store a lot of information that we don't need to globally store. Consequently, we can refactor our usage of `Prism::Location` objects in indexed entries to only store what we need.

### Impact on Performance

Depending on the repository, memory usage of the indexer decreased as much as 21%, which is non-trivial, especially with larger repositories. 

### Implementation

All `RubyIndexer::Entry` objects can be changed to receive one of two arguments for its `location`:
1. A `Prism::Location` object
2. An array storing the Start Line, End Line, Start Column, and End Column

Regardless of which of the two arguments is received, we only persistently store the Start Line, End Line, Start Column, and End Column. It should be noted that the addition of Option 2 for arguments will make it much easier for us to cache indexed gems (#1009), as we will no longer have to create `Prism::Location` objects when deserializing the cache (we can directly send the integers).

This is a breaking change, as repositories like `ruby-lsp-rails` which directly access `RubyIndexer::Entry` objects can no longer expect `entry.location` to yield anything, they must directly call `entry.start_line` and so on.

### Automated Tests

Automated tests have been updated to not rely on `entry.location`, but on `entry.start_line` ... instead. Apart from that, everything should operate as expected.

### Manual Tests

Nothing specific here, we're just changing the layer at which we abstract away the `Prism::Location` object, so everything should function as normal.

### DIFF
--- a/lib/ruby_indexer/lib/ruby_indexer/entry.rb
+++ b/lib/ruby_indexer/lib/ruby_indexer/entry.rb
@@ -11,8 +11,8 @@ module RubyIndexer
     sig { returns(String) }
     attr_reader :file_path
 
-    sig { returns(Prism::Location) }
-    attr_reader :location
+    sig { returns(Integer) }
+    attr_reader :start_line, :end_line, :start_column, :end_column
 
     sig { returns(T::Array[String]) }
     attr_reader :comments
@@ -20,13 +20,33 @@ module RubyIndexer
     sig { returns(Symbol) }
     attr_accessor :visibility
 
-    sig { params(name: String, file_path: String, location: Prism::Location, comments: T::Array[String]).void }
+    sig do
+      params(
+        name: String,
+        file_path: String,
+        location: T.any(Prism::Location, [Integer, Integer, Integer, Integer]),
+        comments: T::Array[String],
+      ).void
+    end
     def initialize(name, file_path, location, comments)
       @name = name
       @file_path = file_path
-      @location = location
       @comments = comments
       @visibility = T.let(:public, Symbol)
+
+      if location.is_a?(Prism::Location)
+        start_line = location.start_line
+        end_line = location.end_line
+        start_column = location.start_column
+        end_column = location.end_column
+      else
+        start_line, start_column, end_line, end_column = location
+      end
+
+      @start_line = T.let(start_line, Integer)
+      @end_line = T.let(end_line, Integer)
+      @start_column = T.let(start_column, Integer)
+      @end_column = T.let(end_column, Integer)
     end
 
     sig { returns(String) }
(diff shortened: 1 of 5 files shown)


aryan-soni:
> > Is ruby-lsp currently storing the locations with their attached comments?
> 
> No, it's just the plain location objects from parsing files. Even if the `Prism::Source` objects are shared, they aren't used by the index, we only need the line and column information.
> 
> So we end up maintaining the full text source, the offset arrays, etc, for all files in both the project and every dependency.

A secondary benefit apart from the memory reduction is also the process of serialization and deserialization.

I'm looking into caching entries to improve performance, and we want to avoid having to construct and deconstruct `Prism::Location` objects because we don't leverage all the benefits they provide.

We're still benefitting from the core information we need, but abstracting away the extraneous state.

kddnewton:
I see. In that case I can start to get onboard with this, but I still think you're going to face problems dropping the location objects because you're not handling column information correctly. Right now you're calling `start_column` and `end_column`, which return byte offsets. This is incorrect, and we have methods on the `Location` object to handle returning code units to you that we wrote specifically for ruby-lsp. Before you drop the location objects, you should be sure that you're pulling the right offsets otherwise you won't be able to recalculate it later.

vinistock:
Indeed. That work is being tracked in https://github.com/Shopify/ruby-lsp/issues/1251, but it will require some extra refactors. We have to pass the encoding down to the index, so that it can be used by locations to get the code unit locations.

But I don't think there's a need to hold this PR, we can definitely still deliver that refactor after these changes.

kddnewton:
I suppose. In that case the only other thing I will add is that you may want an object to represent these things, to make it easier to pass them around. Passing around raw arrays can make things difficult in the future. Having something like an `EntryLocation` could go a long way toward managing this.

aryan-soni:
> I suppose. In that case the only other thing I will add is that you may want an object to represent these things, to make it easier to pass them around. Passing around raw arrays can make things difficult in the future. Having something like an `EntryLocation` could go a long way toward managing this.

Not opposed to this; we can change the constructor to accept either a `Prism::Location` object or a `EntryLocation` object. 
However, the only cases where the alternative param (e.g., `EntryLocation`) would be used are the following:
1. Within the `Alias` constructor.
2. For the upcoming caching implementation, when we deserialize the cache and construct an `Entry` object.

For the first case, instantiating an `EntryLocation` is certainly a cleaner abstraction. 

However, for the second use case, instead of deserializing a JSON array with `JSON.parse` out the box, we would have to define separate serialization and deserialization methods for `EntryLocation`.

We can refactor this down the road, but the current scope is fairly limited and a pre-emptive abstraction might add complexity to the caching.

That being said, I'm open to further discussion if there's an angle I'm missing.

aryan-soni:
> I think it makes sense to create our own location object. We may discover there's more shared functionality we want to put in it eventually. I suggest calling it `RubyIndexer::Location`.

Took a look at this. There looks to be no notable difference in memory usage, so I'm aligned with moving forward with this approach — I've pushed the change. This also ensures we don't need to modify a lot of pre-existing code.

E.g., Indexing in `ruby-lsp` consumed `35.79 MB` of memory, vs. `35.68 MB` with the array approach.

Issue #1457, opened 2024-03-05
Migrate `.rubocop.yml` watching to the server


vinistock:
We currently watch for changes in `.rubocop.yml` from the [extension](https://github.com/Shopify/ruby-lsp/blob/c27ef7f053e74cf87753e7b7646d2650268c147f/vscode/src/workspace.ts#L233) and simply restart the server in case the configuration was modified.

We should migrate the file watching so that it's initiated from the server, making it available in other editors too. Furthermore, we don't need to restart the server, but rather just re-initialize the RuboCop runner object to use the new configs.

### Implementation

1. We need to send another request for file watching like [this one](https://github.com/Shopify/ruby-lsp/blob/c27ef7f053e74cf87753e7b7646d2650268c147f/lib/ruby_lsp/executor.rb#L539), but instead requesting to watch the `.rubocop.yml` file.
2. When we receive the notification that a file has been modified, we need to add a specific check for `.rubocop.yml` [here](https://github.com/Shopify/ruby-lsp/blob/c27ef7f053e74cf87753e7b7646d2650268c147f/lib/ruby_lsp/executor.rb#L211)
3. We need to somehow update the configurations used by the RuboCop runner, which are initialized [here](https://github.com/Shopify/ruby-lsp/blob/c27ef7f053e74cf87753e7b7646d2650268c147f/lib/ruby_lsp/requests/support/rubocop_runner.rb#L68). Alternatively, if there's no way to reconfigure the same runner instance, we might be able to re-initialize the singleton, which will do the right thing

Earlopain:
@vinistock I notice that this was accidentally closed by the attached PR

PR #1921, opened 2024-04-12
Move rubocop file watching into the server

Earlopain:
### Motivation

For #1457. Need to figure out how to clear old diagnostics after the config has been changed. Otherwise the user can see stale diagnostics.


vinistock:
I wonder if this isn't a bug in the language server node package. There's no mention in the spec that pull diagnostics can't be cleared by a publish notification. And the documentation for publish explicitly mentions that pushing an empty array will clear previously existing diagnostics. It might be worth asking in https://github.com/microsoft/vscode-languageserver-node.

Another option, which might work (hopefully), is to collect diagnostics after re-creating the RuboCop instance and try to push the violations using the new configuration. Although, we may find ourselves in a similar situation if no diagnostics are collected after the configuration change.

Earlopain:
I'll try to check that out sometime. Gonna dig around a bit beforehand. There are a few references to `publishDiagnostics`, I might find something

PR #1926, opened 2024-04-12
Avoid memoizing dependencies

vinistock:
### Motivation

I noticed that we were keeping the project's dependencies in memory forever, but we only need that during boot. Let's save some memory and allow the array to be garbage collected.

### Implementation

I turned the array into a local variable that gets used during `apply_options`.

### Automated Tests

Adapted our tests.

### DIFF
--- a/lib/ruby_lsp/global_state.rb
+++ b/lib/ruby_lsp/global_state.rb
@@ -30,8 +30,8 @@ module RubyLsp
 
       @formatter = T.let("auto", String)
       @linters = T.let([], T::Array[String])
-      @test_library = T.let(detect_test_library, String)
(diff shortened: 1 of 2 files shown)


st0012 (commented):
I think it's fine to not storing the dependencies, but do we also need to change the timing test library and typechecker are detected? To me it's quite odd that we now need to rely on `apply_options` to detect those attributes even though they actually don't care about the options.
IMO we can just also call `gather_dependencies` in `initialize` too and keep the same behaviour.

vinistock:
Why do you find it odd? To me it's a bit more organized to have default values in the initialize and then all automatic detection happens once we apply user options.

PR #1929, opened 2024-04-13
Allow request ids to be strings

Earlopain:
### Motivation

The LSP spec allows this: https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#requestMessage

Addons that want to sent requests must provide an id but since the counter is internal to the server that is not easily possible at the moment.

See the conversation starting at https://github.com/Shopify/ruby-lsp-rails/issues/326#issuecomment-2050336058 for more context


vinistock (approved):

PR #1930, opened 2024-04-13
Move `supports_progress` to the global state

Earlopain:
### Motivation

Consolidate global state into the GlobalState object. It looks like the store was previously used like some sort of global state, it just holds this variable and doesn't reference it itself.

### Automated Tests

There were not tests for this, I added some basic ones. Not sending this property apparently equals support? That at least is how the current implementation handles that but I don't think that is correct:

https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#clientCapabilities
> A missing property should be interpreted as an absence of the capability.

I tweakted the implementation to adhere to the spec

### Manual Tests

Start the LSP with the VSCode extension and check for the index progress indicator

### DIFF
--- a/lib/ruby_lsp/global_state.rb
+++ b/lib/ruby_lsp/global_state.rb
@@ -23,6 +23,9 @@ module RubyLsp
     sig { returns(T::Boolean) }
     attr_reader :supports_watching_files
 
+    sig { returns(T::Boolean) }
+    attr_reader :supports_progress
+
     sig { void }
     def initialize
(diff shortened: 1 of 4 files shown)


Earlopain:
There's something not-so-obivous (to me at least) thing going on where some tests now hang because the progress notifications aren't being sent anymore. Indexing also seems to bail during a glob call but I don't really find a exception for that. I could just stub progress in for the tests but since I don't understand what's going on I'll just close this.

PR #1936, opened 2024-04-15
neovim docs: Rename ruby_ls to ruby_lsp

nithinbekal:
Follow up to [this discussion](https://github.com/Shopify/ruby-lsp/pull/1877/files#r1559571577) about renaming `ruby_ls` to `ruby_lsp` in nvim-lspconfig.

The PR on nvim-lspconfig has shipped: https://github.com/neovim/nvim-lspconfig/pull/3105

Waiting on mason-lspconfig PR to ship: https://github.com/williamboman/mason-lspconfig.nvim/pull/395

Note: While I was renaming the LSP here, I also noticed that there's a reference to `ruby_ls` in a large snippet for enabling pull diagnostics. This has been extracted to a [pull_diags.nvim](https://github.com/catlee/pull_diags.nvim) plugin, so I recommended that to make it more approachable.

PR #1937, opened 2024-04-15
docs: Recommend pull_diags.nvim instead of manual config for neovim

nithinbekal:
When updating the ruby LSP name for neovim [here](https://github.com/Shopify/ruby-lsp/pull/1936), I noticed that there's a reference to `ruby_ls` in a large snippet for enabling pull diagnostics. This has been extracted to a [pull_diags.nvim](https://github.com/catlee/pull_diags.nvim) plugin, so I recommended that to make it more approachable. (cc @catlee)

I've extracted that commit from the PR, so that it can be discussed separately.

### DIFF
--- a/EDITORS.md
+++ b/EDITORS.md
@@ -61,102 +61,10 @@ mason_lspconfig.setup_handlers {
 }
 ```
 
-### Neovim Limitations
+### Limitations
 
 Ruby LSP only supports pull diagnostics, and neovim versions prior to v0.10.0-dev-695+g58f948614 only support [publishDiagnostics](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#textDocument_publishDiagnostics).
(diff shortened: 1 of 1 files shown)


EDITORS.md
```diff
@@ -61,102 +61,10 @@ mason_lspconfig.setup_handlers {
 }
 ```
 
-### Neovim Limitations
+### Limitations
 
 Ruby LSP only supports pull diagnostics, and neovim versions prior to v0.10.0-dev-695+g58f948614 only support [publishDiagnostics](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#textDocument_publishDiagnostics).
```
nithinbekal, line 66: Discussed with @vinistock and decided that recommending a plugin for something temporary might be overkill.

This line and the comment in the snippet below already makes it clear that v0.10 is still in dev, so I'll close this PR, and update the snippet in the [other PR](https://github.com/Shopify/ruby-lsp/pull/1936).

PR #1950, opened 2024-04-17
Vs/refactor entries

vinistock:
### Motivation

Do not review. Will fill details soon.


vinistock:
We decided to drop this approach and go with #2024.

PR #1953, opened 2024-04-18
Use prism v0.26.0

### LINKED ISSUE #1906: Bump prism version
Prism 0.25 fixes an [issue](https://github.com/ruby/prism/issues/2674) with its rbi files (Sorbet type definitions). Without this fix, ruby-lsp is unusable in projects using Sorbet.

ruby-lsp currently depends on prism `(>= 0.22.0, < 0.25`. Please change to  `< 0.26` or similar.

andyw8:
### Motivation

Closes #1906


Gemfile.lock
```diff
@@ -122,8 +122,6 @@ GEM
 
 PLATFORMS
   arm64-darwin
-  x64-mingw-ucrt
-  x64-mingw32
```
st0012, line 126: These need to be dropped.

test/requests/code_actions_formatting_test.rb
```diff
@@ -83,7 +83,9 @@ def assert_corrects_to_expected(diagnostic_code, code_action_title, source, expe
     )
 
     diagnostics = RubyLsp::Requests::Diagnostics.new(global_state, document).perform
-    diagnostic = T.must(T.must(diagnostics).find { |d| d.code == diagnostic_code })
+    # The source of the returned attributes may be RuboCop or Prism. Prism diagnostics don't have a code.
```
andyw8, line 86: I'm not yet sure if this impacts any other Ruby LSP behaviour.

lib/ruby_lsp/listeners/document_link.rb
```diff
@@ -124,7 +124,7 @@ def extract_document_link(node)
         match = comment.location.slice.match(%r{source://.*#\d+$})
         return unless match
 
-        uri = T.cast(URI(T.must(match[0])), URI::Source)
+        uri = T.cast(URI(match[0]), URI::Source)
```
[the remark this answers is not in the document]
st0012: I wonder if we should cast `location.slice` as `String` instead?

andyw8:
I think we can merge this and update again after the next Prism release?

Issue #1619, opened 2024-03-15
Contribute our recommended settings as defaults


vinistock:
We have a set of [recommended settings](https://github.com/Shopify/ruby-lsp/blob/main/vscode/README.md#vs-code-configurations) that ensure the LSP will have all of its features enabled.

I think we should contribute those as [configuration defaults](https://github.com/Shopify/ruby-lsp/blob/94d2c9a713e10a1946e5630b7bb4fc99de19173b/vscode/package.json#L439), so that people automatically get the complete experience. And then they can turn off if they don't want something.

**Note**: let's first double-check that using the configuration default does indeed turn on features like `formatOnSave` and `formatOnType`.

vinistock:
Sounds good. Feel free to put up a PR.

PR #1958, opened 2024-04-22
Add default ruby configuration for visual studio code

jenenliu:
### Motivation

I trying to fix this issue https://github.com/Shopify/ruby-lsp/issues/1619. Add recommand default configurations for Visual Studio code.

### Implementation

Add [ruby] configuration to `ConfigurationDefault` in `vscode/package.json` 

### Automated Tests

Not yet

### Manual Tests

Not yet

### DIFF
--- a/vscode/package.json
+++ b/vscode/package.json
@@ -497,6 +497,14 @@
         "**/bower_components": true,
         "**/*.code-search": true,
         "**/.ruby-lsp": true
+      },
+      "[ruby]": {
+        "editor.defaultFormatter": "Shopify.ruby-lsp", 
+        "editor.formatOnSave": true,
+        "editor.tabSize": 2,
+        "editor.insertSpaces": true,
+        "editor.semanticHighlighting.enabled": true,
+        "editor.formatOnType": true
       }
     }
   },


jenenliu:
Hi @vinistock , I am trying to fix issue https://github.com/Shopify/ruby-lsp/issues/1619, could you please help to guide how can I test ? I try to run Ruby-LSP: restart, but seems new configuration is not working. `formatOnSave` is not working at least. Thanks a lot.

vscode/package.json
```diff
@@ -497,6 +497,14 @@
         "**/bower_components": true,
         "**/*.code-search": true,
         "**/.ruby-lsp": true
+      },
+      "[ruby]": {
+        "editor.defaultFormatter": "Shopify.ruby-lsp", 
+        "editor.formatOnSave": true,
```
brunoprietog, line 503: Sorry, but it doesn't seem to me that this should be the default value. When saving, I want to save what I am seeing now, unless I have explicitly defined it that way. This can be problematic especially when Rubocop is very aggressive in its corrections.

andyw8: We do would want this to be `formatOnSave`, as it the [current recommendation](https://github.com/Shopify/ruby-lsp/blob/main/vscode/README.md#vs-code-configurations). We believe this is the best approach for the vast majority of developers. Saving only runs the autocorrects that RuboCop treats as safe, so it's very unlikely that this would result in incorrect code.

vinistock: Yeah, we do want this value to be `true`. The idea of the recommended defaults is for a user of the Ruby LSP to get all features out of the box, even the ones that might be turned off by default by VS Code.

For example, `formatOnSave` and `formatOnType` are both turned off by default, which leads to many people not realizing that the Ruby LSP can auto-complete `end` tokens for you as you type without the need to add any other extensions.

And remember, this is just a default recommendation. You can very easily turn this off.

jenenliu:
> > could you please help to guide how can I test ?
> 
> @jenenliu yes, absolutely. Here are the steps to test this:
> 
> 1. Before starting, ensure you don't have any of the recommended settings already configured in your VS Code. We want to verify that you get the recommended defaults automatically. If you have those settings, comment them out temporarily
> 2. Now, we need to run the development version of the extension, so that you can verify the changes. First, ensure node dependencies are installed: `cd vscode && yarn install`
> 3. In the `run and debug` panel, select the `Run extension` task and click start/run. This will open a second VS Code window, where the development version of the extension is running. It's in that window that you will be able to verify the changes
> 4. In that second window, open a Ruby project
> 5. Now test to see if the changes are taking effect. Are you seeing auto-formatting? When you type code, is the indentation level 2? If you are seeing those, then the changes work as expected

@vinistock   I've tested, it works as expected, thanks a lot. Now if `formatOnSave` should be included, then I'll update the code, and please help to approve.

vscode/package.json
andyw8: @brunoprietog if you can reproduce, then please create an issue for this so that we can address.

andyw8 (approved):

vscode/package.json
brunoprietog: It wasn't a problem with Ruby LSP, it was a problem with VS Code in general that happened to me with several file types, not just Ruby.

jenenliu:
@aryan-soni could you please help to review ? thx

vscode/package.json
vinistock: We discussed this among the team and agreed to move forward with turning on as is. Our current approach with our [extension pack for Ruby](https://marketplace.visualstudio.com/items?itemName=Shopify.ruby-extensions-pack) is even more aggressive than this, offering to mutate your settings and change the values.

Turning off format on save is as simple as setting it to false for Ruby.

```json
{
  "[ruby]": {
    "editor.formatOnSave": false
  }
}
```

vinistock (approved):
Thank you for the contribution! I tested and this works as expected

vinistock:
@jenenliu there's a linting violation in the package.json. You can automatically fix this by

1. Go into the `vscode` directory `cd vscode`
2. Ensure dependencies are installed `yarn install`
3. Run auto format `yarn run format`

jenenliu:
@vinistock Thanks, code updated.

PR #1960, opened 2024-04-22
Increase timeout for tests

andyw8:
### Motivation

We're seeing timeouts on Windows in CI, e.g. https://github.com/Shopify/ruby-lsp/actions/runs/8789287667/job/24118685238


andyw8:
(ideally we would have time to dig i to better understand why this happening, so consider this a temporary measure).

PR #1961, opened 2024-04-23
Support go to definition for unknown receiver calls

st0012:
### Motivation

Instead of providing no candidates for possible definitions on unknown receiver calls, we now provide 10 candidates (the exact number is subject to change in the future).

This will allow the user to navigate to the definition of more methods without overwhelming them with too many candidates in some cases.

### Implementation

When we can't derive the receiver of the target method, instead of return early we get 10 methods of the same name from the index.

### Automated Tests

A new test case was added for this new feature.


vinistock (approved):
Awesome!!!

PR #1963, opened 2024-04-24
Use Prism v0.27.0

andyw8:

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -122,6 +122,8 @@ GEM
 
 PLATFORMS
   arm64-darwin
+  x64-mingw-ucrt
+  x64-mingw32
   x86_64-darwin
   x86_64-linux
 


lib/ruby_lsp/internal.rb
```diff
@@ -19,14 +19,6 @@
 require "prism/visitor"
 require "language_server-protocol"
 
-# Prism v0.26.0 introduced generics for ParseResult, but it causes some problems so the intention is to remove it.
-# Once that is done, we can remove this patch.
-module Prism
-  class ParseResult
-    extend T::Generic
-  end
-end
```
andyw8, line 28: Actually we will need to conditionally include this... will think how to handle that.

ruby-lsp.gemspec
```diff
@@ -19,7 +19,7 @@ Gem::Specification.new do |s|
   s.require_paths = ["lib"]
 
   s.add_dependency("language_server-protocol", "~> 3.17.0")
-  s.add_dependency("prism", ">= 0.23.0", "< 0.27")
+  s.add_dependency("prism", ">= 0.23.0", "< 0.28")
```
andyw8, line 22: @vinistock should we bump the minimum to `>= 0.27.0` now?

lib/ruby_lsp/listeners/document_link.rb
```diff
@@ -124,7 +124,7 @@ def extract_document_link(node)
         match = comment.location.slice.match(%r{source://.*#\d+$})
         return unless match
 
-        uri = T.cast(URI(match[0]), URI::Source)
+        uri = T.cast(URI(T.must(match[0])), URI::Source)
```
st0012, line 127: Could this `T.must` cause problem when used with Prism `0.26`?

st0012 (approved):
Just have one question, otherwise all good 👍

lib/ruby_lsp/listeners/document_link.rb
vinistock: I don't believe there are runtime errors for calling `T.must` on an untyped value. It's just a static check, which shouldn't impact users of the LSP.

PR #1972, opened 2024-04-25
Document how addons can listen for file changes

Earlopain:
### Motivation

Addons may want to listen for changes to files and do something about it. `ruby-lsp-rails` for example listens for `structure.sql` to restart the rails runner for updated column information, after https://github.com/Shopify/ruby-lsp-rails/pull/341 is merged.

`ruby-lsp` itself wants to check for changes to the rubocop config in the server to resolve https://github.com/Shopify/ruby-lsp/issues/1457

`rubyfmt` for `ruby-lsp-rubyfmt` appears to change behaviour depending on `.gitignore` and `.rubyfmtignore`.

----

I noticed that some activation methods didn't yet take the global state in the docs here . I've updated these in a separate commit here.

### DIFF
--- a/ADDONS.md
+++ b/ADDONS.md
@@ -301,6 +301,62 @@ module RubyLsp
 end
 ```
 
+### Registering for file update events
+
+By default, the Ruby LSP listens for changes to files ending in `.rb` to continuously update its index when Ruby source
+code is modified. If your addon uses a tool that is configured through a file (like RuboCop and its `.rubocop.yml`)
+you can register for changes to these files and react when the configuration changes.
+
+**Note**: you will receive events from `ruby-lsp` and other addons as well, in addition to your own registered ones.
+
+
+```ruby
+module RubyLsp
+  module MyGem
+    class Addon < ::RubyLsp::Addon
+      def activate(global_state, message_queue)
+        register_additional_file_watchers(global_state, message_queue)
+      end
+
+      def deactivate; end
+
+      def register_additional_file_watchers(global_state, message_queue)
+        # Clients are not required to implement this capability
+        return unless global_state.supports_watching_files
+
+        message_queue << Request.new(
+          id: "ruby-lsp-my-gem-file-watcher",
+          method: "client/registerCapability",
+          params: Interface::RegistrationParams.new(
+            registrations: [
+              Interface::Registration.new(
+                id: "workspace/didChangeWatchedFilesMyGem",
+                method: "workspace/didChangeWatchedFiles",
+                register_options: Interface::DidChangeWatchedFilesRegistrationOptions.new(
+                  watchers: [
+                    Interface::FileSystemWatcher.new(
+                      glob_pattern: "**/.my-config.yml",
+                      kind: Constant::WatchKind::CREATE | Constant::WatchKind::CHANGE | Constant::WatchKind::DELETE,
+                    ),
+                  ],
+                ),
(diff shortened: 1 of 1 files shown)


ADDONS.md
```diff
@@ -301,6 +301,66 @@ module RubyLsp
 end
 ```
 
+### Registering for file update events
+
+By default, the Ruby LSP listens for changes to files ending in `.rb` to continuously update its index when Ruby source
+code is modified. If your your addon uses a tool that is configured through a file (like RuboCop and its `.rubocop.yml`)
```
andyw8, line 307:
```suggestion
code is modified. If your addon uses a tool that is configured through a file (like RuboCop and its `.rubocop.yml`)
```

ADDONS.md
```diff
@@ -301,6 +301,66 @@ module RubyLsp
...
+  module MyGem
+    class Addon < ::RubyLsp::Addon
+      def activate(global_state, message_queue)
+        register_additional_file_watchers(global_state, message_queue)
+      end
+
+      def deactivate; end
+
+      def register_additional_file_watchers(global_state, message_queue)
+      # Clients are not required to implement this capability
```
andyw8, line 324: Let's fix the indentation here.

ADDONS.md
```diff
@@ -301,6 +301,66 @@ module RubyLsp
...
+      return unless global_state.supports_watching_files
+
+        message_queue << Request.new(
```
andyw8, line 327: This is good to document but illustrates that Ruby LSP should probably provide a better API for this.

ADDONS.md
```diff
@@ -301,6 +301,66 @@ module RubyLsp
...
+you can register for changes to these files and react when the configuration changes.
+
+**Note**: you will receive events from `ruby-lsp` and other addons as well, in addition to your own registered ones.
+
+
+```ruby
+module RubyLsp
+  module MyGem
+    class Addon < ::RubyLsp::Addon
+      def activate(global_state, message_queue)
```
andyw8, line 317: In the other code samples we include the Sorbet signatures, so I'm wondering if we should do the same here for consistency.

andyw8 (approved):
Looks good to me but I'll like @vinistock to check it over too.

ADDONS.md
```diff
@@ -301,6 +301,66 @@ module RubyLsp
...
+      end
+
+
+      def workspace_did_change_watched_files(changes)
+        if changes.any? { |change| change[:uri].end_with?(".my-config.yml") }
+          # Do something to reload the config here
+        end
+      end
+
+      def name
```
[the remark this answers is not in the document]
Earlopain: All the other examples have it as well. But I can drop it here, sure. It doesn't add much

ADDONS.md
Earlopain: Good catch. Would you be open to add [rubocop-md](https://github.com/rubocop/rubocop-md) in a separate PR?

andyw8: Sure, I haven't tried it before but it looks useful.

ADDONS.md
vinistock: I agree that the API is not the friendliest, but let's move forward with the way it is now and see if we start seeing more use cases of addons that need to watch files.

vinistock (approved):

vinistock:

Issue #1969, opened 2024-04-24
VSCode extension + chruby custom paths via RUBIES = cannot find installation directory for Ruby


swrobel:
### Reproduction steps
1. Start VSCode in a ruby project dir. Error notification pops up.
<img width="456" alt="Screenshot 2024-04-24 at 2 01 32 PM" src="https://github.com/Shopify/ruby-lsp/assets/157270/4e21cf5a-9cfa-413a-8ac2-da3b048b763d">

```
Failed to activate [object Object] environment: Cannot find installation directory for Ruby version 3.3.1 or ruby-3.3.1
```

The only output that I see in the VSCode output tab for Ruby LSP is:
```
2024-04-24 14:10:27.958 [info] (my-app) Discovered Ruby version 3.3.1 from file:///Users/swrobel/Code/my-app/.ruby-version
```

I don't see an activation command that I can attempt to replicate, as suggested in the [troubleshooting steps](https://github.com/Shopify/ruby-lsp/blob/main/TROUBLESHOOTING.md#try-to-run-the-ruby-activation-manually), but running `chruby 3.3.1` in the VSCode terminal works without error.

### Versions / Environment
* VSCode 1.88.1
* Ruby LSP VSCode version v0.5.19
* Ruby LSP Gem version unknown, but I've run the `Ruby LSP: Update language server gem` command

**VSCode settings.json**
```json
  "rubyLsp.rubyVersionManager": {
    "identifier": "chruby"
  },
```

**.ruby-version file contents:**
```
3.3.1
```

A perhaps unique aspect of my setup is that I install rubies via Homebrew, and make them available in chruby by setting the `RUBIES` env var in my `.zprofile`. This seems to be picked up properly by VSCode, regardless of whether it's launched from the dock icon or a terminal command.
```
$ echo $RUBIES
/opt/homebrew/Cellar/ruby/3.3.1 /opt/homebrew/Cellar/ruby@3.2/3.2.4
```

### Another bug?
As an aside, this extension produces a confusing/inaccurate error message if using a .ruby-version file that doesn't specify a patch version, even though that should be acceptable for `chruby`, ex:
```
3.3
```

The error shown is:
```
Failed to activate [object Object] environment: No .ruby-version file was found
```

swrobel:
[...]

PR #1976, opened 2024-04-25
Allow configuring extra chruby paths

vinistock:
### Motivation

Addressed the first part of #1969

Chruby allows configuring extra directories to search for Ruby installations. We'll need to offer a similar setting to match the behaviour.

### Implementation

Added the setting and started pushing the configured URIs into the list of places to search for.

### Automated Tests

Added a test.

### DIFF
--- a/vscode/package.json
+++ b/vscode/package.json
@@ -253,6 +253,10 @@
             "miseExecutablePath": {
               "description": "The path to the Mise executable, if not installed in ~/.local/bin/mise",
               "type": "string"
+            },
+            "chrubyRubies": {
+              "description": "An array of extra directories to search for Ruby installations when using chruby. Equivalent to the RUBIES environment variable",
+              "type": "array"
             }
           },
           "default": {
--- a/vscode/src/ruby/chruby.ts
+++ b/vscode/src/ruby/chruby.ts
@@ -5,6 +5,7 @@ import path from "path";
 import * as vscode from "vscode";
 
 import { asyncExec } from "../common";
+import { WorkspaceChannel } from "../workspaceChannel";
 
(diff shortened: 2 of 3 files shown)


swrobel:
Is it not possible to just pull this from the environment variable? That seems like a cleaner solution, if possible.

vinistock:
Not in a scalable and clean way. For example:

1. How would we know where the environment variable is defined? The VS Code process doesn't source your `~/.zshrc`, `~/.bashrc` or equivalent. Your shell does, but not VS Code. Which means we would need to source those files, which has caused us a number of issues in the past
2. What if the environment variable is not defined directly in the rc file? For example, what if `RUBIES` is only defined when there's a prompt or with some conditional behaviour. In many situations, invoking the shell from NodeJS doesn't match the expected conditionals, which may not set the variable

Trying to integrate with shells (our previous approach) basically resulted in a variety of problems. People can configure their environments in unexpected ways that don't integrate well with the extension. As examples, we had reports of users with shell plugins that refused to activate if they weren't in a TTY, breaking our integration.

In other cases, shell plugins hijacked the stderr pipe, redirecting it to something else and only restoring the original file descriptor if the user was about to write a command, which never gets detected when we invoke the shell from NodeJS - again breaking the editor/server communication.

In yet another case, some shell plugin maxed the user's CPU when we invoke their shell from the NodeJS process. Probably because it did not account for the case of having the shell be invoked as a script.

Basically, integrating with shells proved to be extremely brittle and lead to a lot of frustration. We invested significant effort to decouple from shells to prevent all of those. I understand that having to set `RUBIES` twice is a small inconvenience, but we had to accept that tradeoff in favour of more stable Ruby activation.

andyw8 (approved):

PR #1978, opened 2024-04-25
Make module tracking lazy

vinistock:
### Motivation

This is a small optimization to how we're tracking included and prepended modules. There's no need to eagerly create the arrays to store them, since there might be many modules that don't include or prepend anything.

### Implementation

Made the modules initialization lazy, so that the arrays are only created if we actually found an include or prepend.

### DIFF
--- a/lib/ruby_indexer/lib/ruby_indexer/entry.rb
+++ b/lib/ruby_indexer/lib/ruby_indexer/entry.rb
@@ -60,29 +60,19 @@ module RubyIndexer
 
       abstract!
 
-      sig { returns(T::Array[String]) }
-      attr_accessor :included_modules
-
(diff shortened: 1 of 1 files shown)


aryan-soni (approved):
Looks good!

lib/ruby_indexer/lib/ruby_indexer/entry.rb
```diff
@@ -60,29 +60,19 @@ class Namespace < Entry
 
       abstract!
 
-      sig { returns(T::Array[String]) }
-      attr_accessor :included_modules
+      sig { returns(String) }
+      def short_name
+        T.must(@name.split("::").last)
```
Morriar, line 65: Should this be memoized?

PR #1981, opened 2024-04-26
Discard method call target if position doesn't cover identifier

vinistock:
### Motivation

With the current experience, clicking to go to the definition of a method (or hovering) over any part of the `CallNode` will always succeed. This is not what we want, since people should be able to go to definition or hover separately for method arguments.

It also leads to incorrect behaviour. For example:

```ruby
[].each(&:foo)
#         ^ go to definition here is currently taking you to `each`, which is not correct
```

### Implementation

We still need to locate the method calls based on the entire node, but if we're processing a method call that has no special handling (i.e.: not requires), then we need to ensure that the position being clicked is covered by method's identifier.

Otherwise, it means the user is clicking on the arguments node or on the receiver (neither of which should take you to the method declaration).

### Automated Tests

Added tests demonstrating the behaviour.

### DIFF
--- a/lib/ruby_lsp/requests/definition.rb
+++ b/lib/ruby_lsp/requests/definition.rb
@@ -43,6 +43,7 @@ module RubyLsp
           ResponseBuilders::CollectionResponseBuilder[Interface::Location].new,
           ResponseBuilders::CollectionResponseBuilder[Interface::Location],
         )
+        @dispatcher = dispatcher
 
         target, parent, nesting = document.locate_node(
           position,
@@ -50,33 +51,41 @@ module RubyLsp
         )
 
         if target.is_a?(Prism::ConstantReadNode) && parent.is_a?(Prism::ConstantPathNode)
+          # If the target is part of a constant path node, we need to find the exact portion of the constant that the
+          # user is requesting to go to definition for
           target = determine_target(
             target,
             parent,
             position,
           )
+        elsif target.is_a?(Prism::CallNode) && target.name != :require && target.name != :require_relative &&
+            !covers_position?(target.message_loc, position)
+          # If the target is a method call, we need to ensure that the requested position is exactly on top of the
+          # method identifier. Otherwise, we risk showing definitions for unrelated things
+          target = nil
         end
 
-        Listeners::Definition.new(
-          @response_builder,
-          global_state,
-          document.uri,
-          nesting,
-          dispatcher,
-          typechecker_enabled,
-        )
+        if target
+          Listeners::Definition.new(
+            @response_builder,
+            global_state,
+            document.uri,
+            nesting,
+            dispatcher,
+            typechecker_enabled,
+          )
 
-        Addon.addons.each do |addon|
-          addon.create_definition_listener(@response_builder, document.uri, nesting, dispatcher)
+          Addon.addons.each do |addon|
+            addon.create_definition_listener(@response_builder, document.uri, nesting, dispatcher)
+          end
         end
 
         @target = T.let(target, T.nilable(Prism::Node))
-        @dispatcher = dispatcher
       end
 
       sig { override.returns(T::Array[Interface::Location]) }
       def perform
-        @dispatcher.dispatch_once(@target)
+        @dispatcher.dispatch_once(@target) if @target
         @response_builder.response
       end
     end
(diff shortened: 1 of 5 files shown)


lib/ruby_lsp/requests/definition.rb
```diff
@@ -43,40 +43,49 @@ def initialize(document, global_state, position, dispatcher, typechecker_enabled
...
 
-        Listeners::Definition.new(
-          @response_builder,
-          global_state,
-          document.uri,
-          nesting,
-          dispatcher,
-          typechecker_enabled,
-        )
+        if target
```
andyw8, line 68: (nit) could this be an early return?

paracycle:
I would like to see how this behaves with clicking `foo` in the example below, though:
```ruby
bar(a: foo, 42)
```

It would be smart to add a test case to make sure that we still jump to `foo` and not ignore it because it was in the arguments to `bar`.

lib/ruby_lsp/requests/request.rb
```diff
@@ -65,6 +65,22 @@ def determine_target(target, parent, position)
...
+
+        start_line = location.start_line - 1
+        end_line = location.end_line - 1
+
+        start_covered = start_line < position[:line] ||
+          (start_line == position[:line] && location.start_column <= position[:character])
+        end_covered = end_line > position[:line] ||
+          (end_line == position[:line] && location.end_column >= position[:character])
+
+        start_covered && end_covered
```
paracycle, line 82: I know this makes the code more readable, but we are missing an optimization here that we can get with an early return if we inline both branches.

In other words, if we fail `start_covered` check, we shouldn't have to make extra work to calculate `end_covered`.

Also, it would be good to access `position` fields once and not multiple times.

paracycle:
```suggestion
        position_line = position[:line]
        position_char = position[:character]
        
        (start_line < position_line ||
          (start_line == position_line && location.start_column <= position_char)) &&
        (end_line > position_line ||
          (end_line == position_line && location.end_column >= position_char))
```

lib/ruby_lsp/requests/definition.rb
vinistock: I'll be honest, I'm not a fan of early returns in the `initialize` method. If you feel strongly about it, I can change, but I wouldn't normally.

lib/ruby_lsp/requests/request.rb
vinistock: Good points. I got rid of the `start_covered` and `end_covered` variables, so that the conditional short circuits when failing.

Also, putting the position parts into variables made it more readable, so we're not missing out on anything.

vinistock:
> I would like to see how this behaves with clicking foo in the example below, though:

Added a test to ensure we don't regress, but that already worked from the implementation of `locate_node`.

lib/ruby_lsp/requests/request.rb
```diff
@@ -65,6 +65,22 @@ def determine_target(target, parent, position)
 
         target
       end
+
+      # Checks if a given location covers the position requested
+      sig { params(location: T.nilable(Prism::Location), position: T::Hash[Symbol, T.untyped]).returns(T::Boolean) }
+      def covers_position?(location, position)
```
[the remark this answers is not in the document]
kddnewton: Yeah I think there might be a couple of objects hiding here. The fact that a raw hash is being passed around here makes me really uncomfortable. I think the following objects would be useful instead:

```ruby
Prism::Position::ByteOffset[:offset, :byte_length]
Prism::Position::LineByteColumn[:start_line, :start_byte_column, :end_line, :end_byte_column]
Prism::Position::LineCharacterColumn[:start_line, :start_character_column, :end_line, :end_character_column]
Prism::Position::LineCodeUnitColumn[:start_line, :start_code_unit_column, :end_line, :end_code_unit_column]
```

with APIs that create those various positions from a location and can compare against those positions for things like `covers?`. Unless we're explicit about which kind of column counts we're using, I wouldn't want to make an assumption (because people get burned by this too often with ripper).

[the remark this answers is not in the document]
vinistock: Yeah, indeed it needs to account for all possible ways that you can refer to a location.

PR #1982, opened 2024-04-26
Support symbol procs in definition

vinistock:
### Motivation

Allow jumping to the definition of method calls passed as symbol proc arguments (e.g.: `map(&:to_i)`).

### Implementation

1. Started accepting `Prism::BlockArgumentNode` as a possible target to locate
2. Started handling that node type in the definition listener
3. Made a small refactor to make the method definition helper more reusable

### Automated Tests

Changed the example from #1981 to assert the right positions.


andyw8 (approved):
I verified it also works when the symbol-to-proc isn't the only argument, e.g. `values.inject(0, &:foo)`.

PR #1992, opened 2024-04-29
Improve version manager troubleshooting

st0012:
### Motivation

I think we can mention some version manager specific tips as part of the troubleshooting guideline to help users address them.

### Implementation

1. Move `vscode/VERSION_MANAGERS.md` to top-level so they are more discoverable.
2. Add `asdf` section to `VERSION_MANAGERS.md`

### DIFF
--- a/VERSION_MANAGERS.md
+++ b/VERSION_MANAGERS.md
@@ -4,7 +4,7 @@ This document contains information and tips to help Ruby LSP's VS Code extension
 
 ## asdf
 
-If you use `asdf` and the VS Code extension fails to activate the environment, try updating it to the latest version with `asdf update` and restart VS Code.
+If you use `asdf` and the VS Code extension fails to activate the environment (as described in [this issue](https://github.com/Shopify/ruby-lsp/issues/1985)), you may resolve it by updating `asdf` to the latest version with `asdf update`, and then restart VS Code.
 
 ## Chruby
 


TROUBLESHOOTING.md
```diff
@@ -131,11 +131,15 @@ Did it select your preferred version manager? You can define which version manag
 You can enable logging to the VS Code output tab,
 [as described in the CONTRIBUTING](CONTRIBUTING.md#tracing-lsp-requests-and-responses) docs.
 
+### Environment activation failed
+
+We listed a few version manager related information and tips in this [documentation](VERSION_MANAGERS.md).
```
andyw8, line 136:
```suggestion
We listed version manager related information and tips in this [documentation](VERSION_MANAGERS.md).
```

VERSION_MANAGERS.md
```diff
@@ -1,3 +1,7 @@
+## asdf
+
+If you use `asdf` and the VS Code extension fails to activate the environment, try running `asdf update` and restart VS Code.
```
vinistock, line 3: How about we mention the type of error that we've received reports for (the JSON parse error)?

VERSION_MANAGERS.md
vinistock, line 1: Why are we moving this file?

VERSION_MANAGERS.md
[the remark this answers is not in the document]
andyw8: (since if someone isn't familiar with `asdf`, they may be unclear if that is updating Ruby, or `asdf` itself).

st0012: I thought we plan to improve the message as well?

vinistock: We do, but it felt weird to me to say that if anything fails with ASDF try to update it. I would include a link to one of the reported issues at least so that people can dig more into it and figure out if they are facing the same scenario.

vinistock (approved):

Issue #1893, opened 2024-04-05
Add go to definition support for autoloaded constants


vinistock:
Basically, the same feature as go to definition for [require](https://github.com/Shopify/ruby-lsp/blob/ee5f44d316de3e3f13163344e7a3b8c65c2d9429/lib/ruby_lsp/listeners/definition.rb#L91), but in this case for autoloads.

```ruby
module Foo
  autoload :Bar, "bar" # <- jumping to bar should take you to the right file
end
```

We can probably add the same for the constant too, since we know that it must resolve to `Foo::Bar`, which makes it easy to search in the index.

Super-Xray:
Hi, @vinistock I'm interested in this feature, can I handle this issue?

Super-Xray:
Hello @vinistock , I'm interested in this issue, can I handle it?

Super-Xray:
Thank you very much! Since this is my first OSS contribution, I didb't know this before.

Super-Xray:
However，I have found that the structure of autoload statement is 
![image](https://github.com/Shopify/ruby-lsp/assets/110341228/df209f73-6ab6-42b0-b61f-9ae96a60c4b9)
there aren't any `ConstantNode` here so I think it's difficult to use find_in_index straightly (becuase `constant_name` method requires `ConstantPathNode`/`ConstantReadNode`/`ConstantTargetNode` input). 
In order to find the correct line, I think maybe we can refer to the mechanism of searching class or method in a file (input '@' + class_name/method name) because we can get the class's name from `SymbolNode`. Does this idea work？

vinistock:
The `find_in_index` method accepts the name of the constant you're looking for as a string and already takes care of the namespacing for you.

For example

```ruby
module Foo
  autoload :Bar, "some/file"
end
```

In this case, you need to invoke `find_in_index` with the name `Bar` and the existing functionality will automatically discover that the referenced constant is actually `Foo::Bar`. Basically, what you need to do is extract the name from the `autoload` arguments and pass that name to `find_in_index`.

This will already make the LSP jump to the exact location where that constant is defined. You won't need the `@` based search (document symbol), to land on the right place in the file.

PR #1995, opened 2024-04-30
Definition support for autoloaded constants

Super-Xray:
### Motivation

There are an issue calling for complete this function (https://github.com/Shopify/ruby-lsp/issues/1893).

### Implementation
`autoload` argument is a `Prism::CallNode` which has name `:autoload`. I create `handle_autoload_definition` method and add it to `on_call_node_enter` method in 'lib/ruby_lsp/listeners/definition.rb'. Then in `handle_autoload_definition` method, we can use the name of the constant in the `Prism::SymbolNode` under the `Prism::CallNode` and pass the name to `find_in_index`, which can create the response of jumping.

### Automated Tests
Completed automated tests in `test/requests/definition_expectations_test.rb`, including `test_jumping_to_autoload_definition_when_declaration_exists`, `test_jumping_to_autoload_definition_with_two_definition` and `test_does_nothing_when_autoload_declaration_does_not_exist`

### Manual Tests
2. open an `.rb` file which has `autoload` arguments.
3. move the cursor above the `autoload` arguments, press ctrl & click it
4. sometimes it doesn't work in original file of ruby-lsp since it's a sorbet project


lib/ruby_lsp/listeners/definition.rb
```diff
@@ -145,6 +147,22 @@ def handle_require_definition(node)
         end
       end
 
+      sig { params(node: Prism::CallNode).void }
+      def handle_autoload_definition(node)
+        name = node.name # should be ":autoload"
+        if name != :autoload
+          return
+        end
```
vinistock, line 155: This is already being checked where the method is invoked, so this can be removed.

lib/ruby_lsp/listeners/definition.rb
```diff
@@ -145,6 +147,22 @@ def handle_require_definition(node)
...
+
+        arguments = node.arguments # should be 'ArgumentsNode'
+        return unless arguments.is_a?(Prism::ArgumentsNode)
+
+        symbol = arguments.arguments[0] # should be 'SymbolNode'
+        return unless symbol.is_a?(Prism::SymbolNode)
+
+        find_in_index(symbol.value)
```
vinistock, line 163:
```suggestion
        arguments = node.arguments&.arguments
        return unless arguments

        value = arguments.first&.value
        return unless value

        find_in_index(value)
```

test/requests/definition_expectations_test.rb
```diff
@@ -344,6 +344,91 @@ def bar
...
+      server.process_message(
+        id: 1,
+        method: "textDocument/definition",
+        params: { textDocument: { uri: uri }, position: { character: 3, line: 3 } },
+      )
+      assert_equal(uri.to_s, server.pop_response.response.first.attributes[:uri])
+    end
+  end
+
+  def test_jumping_to_autoload_definition_with_two_definition
```
vinistock, line 368: Multiple declarations for the same constant is already handled and tested by the index, so I'd remove this test.

vinistock (commented):
This is looking great! There's only one piece missing. To improve the precision of where you can cmd click, you need to add one extra check to [this condition](https://github.com/Shopify/ruby-lsp/blob/94194d8f8ee3feff23d2f3d99bebdb5f3a733074/lib/ruby_lsp/requests/definition.rb#L63)

```ruby
target.name != :autoload
```

lib/ruby_lsp/listeners/definition.rb
Super-Xray: got it

Super-Xray:
I have signed the CLA!

vinistock (approved):

Super-Xray:
> There is one typecheck failure, it may indicate a missing nil check.
image
Did you mean these error？

lib/ruby_lsp/listeners/definition.rb
```diff
@@ -145,6 +147,19 @@ def handle_require_definition(node)
         end
       end
 
+      sig { params(node: Prism::CallNode).void }
+      def handle_autoload_definition(node)
+        arguments = node.arguments&.arguments
+        return unless arguments
```
vinistock, line 153: To fix the type checking error.
```suggestion
        arguments = node.arguments&.arguments&.first
        return unless arguments.is_a?(Prism::SymbolNode)
```

lib/ruby_lsp/listeners/definition.rb
```diff
@@ -145,6 +147,19 @@ def handle_require_definition(node)
...
+
+        return unless arguments.first
+
+        value = arguments.first&.value
```
vinistock, line 157:
```suggestion
        value = argument.value
```

test/requests/definition_expectations_test.rb
```diff
@@ -344,6 +344,46 @@ def bar
...
+
+        class Bar; end
+      end
+    RUBY
+
+    with_server(source) do |server, uri|
+      server.process_message(
+        id: 1,
+        method: "textDocument/definition",
+        params: { textDocument: { uri: uri }, position: { character: 3, line: 3 } },
```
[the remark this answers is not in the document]
Super-Xray: > Let's not hold the PR. This behaviour is a long standing bug in the LSP, we can fix it all later.

ok, what should I do now? Closing the PR?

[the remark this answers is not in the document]
vinistock: No, no. We're going to merge it.

Super-Xray:
excuse me, I'm not sure how to solve this conflict now

vinistock:
Sorry, the development on the LSP is moving fast and I feel like you keep chasing a moving target.

We made changes to allow for more precise jump to definition based on a method call and its arguments. You can see an example for [require definition](https://github.com/Shopify/ruby-lsp/blob/a3791faeb4f3c2c6e16135b06b7e378074409777/lib/ruby_lsp/listeners/definition.rb#L59).

The idea is that the event now matches `on_string_node_enter`, then checks if the enclosing call node is a `require`.

You will need to do the exact same thing, but for `on_symbol_node_enter`. Then you check the enclosing call node to verify if it's an `autoload`.

st0012:
@Super-Xray Thanks for the PR! Because we want to have autoload's definition support to be released soon-ish to match [other code navigation features](https://railsatscale.com/2024-07-18-mastering-ruby-code-navigation-major-enhancements-in-ruby-lsp-2024/) we made recently, I took the liberty to open https://github.com/Shopify/ruby-lsp/pull/2395 with your commits and built some small improvements on top of it.

st0012:
Superseded by #2395

PR #1998, opened 2024-05-01
Remove Sorbet sigs from docs

andyw8:
As agreed with @vinistock and @st0012


andyw8:
I also removed `# frozen_string_literal: true` since that seems unnecessary for code samples.

PR #2005, opened 2024-05-02
Add --time-index flag for timing index

st0012:
### Motivation

When working on indexing changes, we often care a lot about its performance impact. But currently there's no built-in utility for us to easily measure that.

### Implementation

Add a new `--time-index` flag to the `ruby-lsp` executable, which when applied will index the full project, print the result, and exit the process.

#### Example usage

1. Open a Ruby project
2. The VS Code extension should install Ruby LSP automatically. If not, run `gem install ruby-lsp` manually.
3. Run `ruby-lsp --time-index` and record the baseline result.
4. Update Gemfile to use the target version of Ruby LSP and run `bundle install`
5. Run `bundle exec ruby-lsp --time-index` and record the target result.
6. Compare the 2 results.

### DIFF
--- a/exe/ruby-lsp
+++ b/exe/ruby-lsp
@@ -18,6 +18,10 @@ parser = OptionParser.new do |opts|
     options[:debug] = true
   end
 
+  opts.on("--time-index", "Measure the time it takes to index the project") do
+    options[:time_index] = true
+  end
+
   opts.on(
     "--branch [BRANCH]",
     "Launch the Ruby LSP using the specified branch rather than the release version",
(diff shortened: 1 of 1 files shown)


exe/ruby-lsp
```diff
@@ -90,4 +94,14 @@ if options[:debug]
   end
 end
 
+if options[:time_index]
+  require "benchmark"
+
+  index = RubyIndexer::Index.new
+
+  result = Benchmark.realtime { index.index_all }
+  puts "Indexing took #{result.round(2)} seconds"
```
st0012, line 103: I feel if we put the version into this message, we can more easily use it in issue reports too?

st0012: For small projects, I feel the smaller decimals won't really be useful?
LIke these are the 3 consecutive runs on Ruby LSP itself:

```
Ruby LSP 0.16.7: Indexing took 0.83 seconds
Ruby LSP 0.16.7: Indexing took 0.86 seconds
Ruby LSP 0.16.7: Indexing took 0.81 seconds
```

Anything after the 2nd decimal feel more like noise?

st0012: I increased it to 5 anyway, let's see how it works first.

PR #2009, opened 2024-05-03
Avoid mutating completion item in resolve

### LINKED ISSUE #1983: Don't suggest fully qualified constants when unnecessary

### Use case

I use the auto-completion feature. Unfortunately, it proposes constants like `MyApp::Foo::Bar` rather than `Foo::Bar` when I am inside the `MyApp` module. This is pointless/not preferred in our coding style.

You can get these suggestions in some cases but I predominantly get the fully qualified constant name which greatly reduces the usability of this feature for me.

### Description

Prefer the shortened version of the constant when inside the class/module hierarchy.

### Implementation

I am guessing it would be possible to track current scope based on how the breadcrumbs in my editor show me which modules/methods i'm inside--it would need to take this context into account.

vinistock:
### Motivation

Closes #1983

Like the author of the issue, I had also noticed that we stopped using short names for constants in completion. Which puzzled me a lot given that we [explicitly handle that](https://github.com/Shopify/ruby-lsp/blob/55e36dc3f1368263c24a29cd4e031e44ebf82f8b/lib/ruby_lsp/listeners/completion.rb#L257) and have unit tests to verify the behaviour.

After reading the [specification](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#textDocument_completion) again, I finally realized the issue.

When we split completion into completion + resolve, there was a mistake in the original PR. The resolve request should never mutate any existing fields of the completion item other than label details and documentation. But instead, we were creating a new completion item from scratch without including all of the original fields.

That means that the `insertText` we were specifying in completion was getting erased during resolve, which forces the editor to use the `label` instead, resulting in the behaviour of showing the fully qualified name.

### Implementation

We never want to mutate anything in resolve that isn't label details and documentation. We should take the original item as a hash, change only those fields and then return it.

This fixes the issue of using the wrong sorting text and the wrong insertion text, which we're currently seeing in `main`.

### Automated Tests

Improved the resolve test a bit.


test/requests/completion_resolve_test.rb
```diff
@@ -11,29 +11,32 @@ class CompletionResolveTest < Minitest::Test
...
+        labelDetails: Interface::CompletionItemLabelDetails.new(
           description: "fake.rb",
         ),
         documentation: Interface::MarkupContent.new(
           kind: "markdown",
-          value: markdown_from_index_entries("Foo", T.must(server.global_state.index["Foo"])),
+          value: markdown_from_index_entries("Foo::Bar", T.must(server.global_state.index["Foo::Bar"])),
         ),
       )
-      assert_match(/This is a class that does things/, result.documentation.value)
```
st0012, line 36: Nit: can we keep this static value check?

PR #2012, opened 2024-05-03
Fix false positive in gem list matching

vinistock:
### Motivation

Our condition may incorrectly match if a gem like `ruby-lsp-rails` is installed, but `ruby-lsp` isn't. We need to ensure that the list includes exactly `ruby-lsp`.

### Implementation

Started using a regex to match the gem name.


vscode/src/workspace.ts
```diff
@@ -193,7 +193,7 @@ export class Workspace implements WorkspaceInterface {
     });
 
     // If the gem is not yet installed, install it
-    if (!stdout.includes("ruby-lsp")) {
+    if (!/\bruby-lsp\b/.exec(stdout)) {
```
andyw8, line 196: 🤔 is this right? `_` would be considered part of the word, but `-` isn't:

```
irb(main):011> "ruby-lsp-rails".match?(/\bruby-lsp\b/)
=> true
```

vinistock: It wasn't indeed. The hyphen also matches as a word boundary. I changed the regex to ensure we match properly.

dirceu (approved):

PR #2017, opened 2024-05-06
Update Prism requirement to v0.28

vinistock:
### Motivation

Update our Prism requirement to `v0.28` or higher. Version `v0.28` includes the new `ASCIISource` with performance optimization for sources that do not contain multibyte characters, which fixes the performance regression we introduced in semantic highlighting.

### Implementation

1. Updated Prism and RBI
2. Re-generated gem RBIs
3. Fixed errors related to the restructuring of constant paths

Before, each part of a constant path was also a constant read. After this version, only the parent parts of the constant path may be constant reads, the main name of the constant is embedded inside the constant path node.

```ruby
# Before
# This would be 3 constant read nodes to compose the constant path

# After
# This is now 2 constant read nodes and the name `C` is a part of the overall constant path
A::B::C
```

### Automated Tests

Fixed a few tests.

### DIFF
--- a/vscode/src/ruby/chruby.ts
+++ b/vscode/src/ruby/chruby.ts
@@ -43,10 +43,6 @@ export class Chruby extends VersionManager {
   async activate(): Promise<ActivationResult> {
     const versionInfo = await this.discoverRubyVersion();
     const rubyUri = await this.findRubyUri(versionInfo);
-    this.outputChannel.info(
-      `Discovered Ruby installation at ${rubyUri.fsPath}`,
-    );
-
     const { defaultGems, gemHome, yjit, version } =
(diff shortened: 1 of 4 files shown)


andyw8:
> The upgrade ended up uncovering some other Windows errors and I was trying to get CI to pass. If you prefer, I can split that into a separate PR, but the changes are necessary.

Ok, yeah I'd say we should split them since there's quite a lot of changes. I'll limit my review to the Ruby part.

vinistock:
Split the other part, which needs to be merged first https://github.com/Shopify/ruby-lsp/pull/2033.

andyw8 (approved):

PR #2019, opened 2024-05-07
Only pop namespace stack when the class or module is pushed onto it

st0012:
### Motivation

Found this when working on visibility tracking: Currently, the indexer doesn't index classes and modules that have dynamic namespace. Similarly, those classes and modules' names are not pushed onto the namespace stack. However, the stack is currently popped on every class or module exit, which could lead to an incorrect stack popping.

### Implementation

This commit fixes the issue by only popping the stack when the class or module name is considered indexable.

Another option would be to always push the class/module onto the stack and always pop the stack after leaving. But I'm not sure if that was intentionally avoided @vinistock?


vinistock:
I think we actually need to always push and pop the stack. If we don't, we will end up pushing methods into the wrong owner. Consider this

```ruby
module Foo
  var = Bar
  module var::Baz
    def qux; end
  end
end
```

If we don't push something into the stack, the `qux` method will be associated with the `Foo` module, which is incorrect. I think we can take a similar strategy that we used for code lens. 

If we find a dynamic declaration, we can either push something into the stack that we can easily identify like `<dynamic declaration>` or use some boolean to remember we're inside a dynamic context. Then we need to discard every method definition we encounter while we're in a dynamic declaration.

Issue #1347, opened 2024-02-03
Definition jumps are not possible with files containing Japanese characters.


rokudone:
### Ruby version

3.1.4, 3.2.2, 3.3.0

### Code snippet

Gemfile
```
# frozen_string_literal: true

source 'https://rubygems.org'

ruby '3.3.0'

gem 'debug', platforms: %i[mri windows]
gem 'rubocop'
gem 'ruby-lsp'
```

vscode/settings.json
```
{
  "[ruby]": {
    "editor.defaultFormatter": "Shopify.ruby-lsp",
    "editor.formatOnSave": true,
    "editor.tabSize": 2,
    "editor.insertSpaces": true,
    "editor.semanticHighlighting.enabled": true,
    "editor.formatOnType": true
  },
  "rubyLsp.rubyVersionManager": "rbenv",
  "rubyLsp.formatter": "rubocop",
  "rubyLsp.enabledFeatures": {
    "codeActions": true,
    "diagnostics": true,
    "documentHighlights": true,
    "documentLink": true,
    "documentSymbols": true,
    "foldingRanges": true,
    "formatting": true,
    "hover": true,
    "inlayHint": true,
    "onTypeFormatting": true,
    "selectionRanges": true,
    "semanticHighlighting": true,
    "completion": true,
    "codeLens": true,
    "definition": true,
    "workspaceSymbol": true,
    "signatureHelp": true
  }
}
```

success file
```ruby
# frozen_string_literal: true

# Test Class
class Test
  TEST = 'test'

  def method1
    # test
    pp TEST
  end
end
```

failed file
```ruby
# frozen_string_literal: true

# Test Class
class Test
  TEST = 'test'

  def method1
    # テスト
    pp TEST
  end
end
```

success file 2
```ruby
# frozen_string_literal: true

# Test Class
class Test
  TEST = 'test'

  def method1
    pp TEST
    # テスト
  end
end
```

VSCode
```
Version: 1.86.0 (user setup)
Commit: 05047486b6df5eb8d44b2ecd70ea3bdf775fd937
Date: 2024-01-31T10:28:19.990Z
Electron: 27.2.3
ElectronBuildId: 26495564
Chromium: 118.0.5993.159
Node.js: 18.17.1
V8: 11.8.172.18-electron.0
OS: Windows_NT x64 10.0.22621
```
However, I have the same issue on my Mac

### Description

[...]

PR #2021, opened 2024-05-08
Fix go to definition and hover for files containing multibyte characters

### LINKED ISSUE #1251: Better handle multibyte character locations

NotFounds:
### Motivation

Closes #1251

When a Ruby file contains multibyte characters (like Japanese, Chinese, emoji, etc), the go to definition and hover features do not work correctly. The definition location or hover documentation will be incorrect.
ref:
- https://github.com/Shopify/ruby-lsp/issues/1347

This is because the current implementation assumes single-byte characters when calculating offsets during index building and document referencing. We need to properly handle multibyte characters to ensure these features work reliably for all users.

### Implementation

1. Modified `﻿RubyLsp::Document` and `﻿RubyLsp::Requests::Request` to calculate locations considering multibyte characters. This change utilizes the API implemented in Prism by https://github.com/ruby/prism/pull/2406.
2. Added ﻿encoding to ﻿Entry and updated the logic to handle multibyte characters when storing locations. Additionally, since the location object has been changed from `﻿Prism::Location` to ﻿`RubyIndexer::Location` by https://github.com/Shopify/ruby-lsp/pull/1917, a new field has been added to ﻿`RubyIndexer::Location`.
3. By passing the ﻿encoding when creating the ﻿Index, the index will be built taking multibyte characters into account.

### Automated Tests

I added a simple expectation test for the definition.


NotFounds:
@andyw8 
Thank you for your review!
As you pointed out, the changes can be broken down into multiple tasks.
I'll close this Pull Request for now and focus on fixing the handling of multibyte characters first.

NotFounds:
I opened a new PullRequest.
https://github.com/Shopify/ruby-lsp/pull/2051

PR #2026, opened 2024-05-09
Mention editor names in the 'other editors' section

st0012:
### Motivation

This will allow users to `cmd+f` search for the editor names, and then link to the `EDITORS.md`, instead of going through readme section by section to spot it.


st0012:
I adopted @andyw8's suggestion 👍

PR #2028, opened 2024-05-09
Support inherited methods in definition, hover and completion

vinistock:
### Motivation

Step towards #899

Use the linearized ancestor information from #2024 to support inherited methods for definition, hover and completion.

### Implementation

1. Changed `resolve_method` to linearize ancestors lazily. Then we search ancestors in order and return the first methods we return, which ensures we jump to the right method
2. Adapted completion to consider all ancestors when filtering possible methods being invoked

### Automated Tests

Added tests.

https://github.com/Shopify/ruby-lsp/assets/18742907/ad6ff4e5-33b2-4f79-ac88-93f0846bbdf1

### DIFF
--- a/lib/ruby_indexer/lib/ruby_indexer/index.rb
+++ b/lib/ruby_indexer/lib/ruby_indexer/index.rb
@@ -249,14 +249,17 @@ module RubyIndexer
     sig { params(method_name: String, receiver_name: String).returns(T.nilable(T::Array[Entry::Member])) }
     def resolve_method(method_name, receiver_name)
       method_entries = self[method_name]
-      owner_entries = self[receiver_name]
-      return unless owner_entries && method_entries
+      ancestors = linearized_ancestors_of(receiver_name.delete_prefix("::"))
+      return unless ancestors && method_entries
 
-      owner_name = T.must(owner_entries.first).name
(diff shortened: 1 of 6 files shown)


vinistock:
Yes. And this will happen automatically, by making `resolve_method` smarter, it will return more accurate results without any changes needed in the Rails addon.

andyw8 (approved):

lib/ruby_indexer/lib/ruby_indexer/index.rb
```diff
@@ -248,16 +248,20 @@ def follow_aliased_namespace(name)
...
-
-      owner_name = T.must(owner_entries.first).name
-      T.cast(
-        method_entries.grep(Entry::Member).select do |entry|
-          T.cast(entry, Entry::Member).owner&.name == owner_name
-        end,
-        T::Array[Entry::Member],
-      )
+      ancestors = linearized_ancestors_of(receiver_name.delete_prefix("::"))
+      return unless method_entries && ancestors.any?
```
Morriar, line 252:
```suggestion
      return unless ancestors&.any?
```

Morriar (commented):
Can't we also handle go to and hover to `super` now?

PR #2029, opened 2024-05-10
vscode/asdf: support configuration environment variables

### LINKED ISSUE #2023: vs code: asdf environment variables ignored
## VS Code Ruby LSP
The extension detects `asdf` but fails when searching for the data directory.

The [code](https://github.com/Shopify/ruby-lsp/blob/main/vscode/src/ruby/asdf.ts) does not appear to use the asdf environment variables:
* `ASDF_DIR`
* `ASDF_DATA_DIR`

Instead, it attempts to detect those values by looking at predetermined locations. This fails when a those values are set to alternative locations. 

### Reproduction steps

1. set `ASDF_DATA_DIR` to ~/asdf-data
2. install/setup asdf to that location (see [docs](https://asdf-vm.com/manage/configuration.html#environment-variables))
3. Open VS Code
4. Open a Ruby file
5. See error:

```
Automatic Ruby environment activation with asdf failed:
Could not find ASDF data dir. Searched in
file:///Users/zach.gianos/.asdf, 
file:///opt/asdf-vm, 
file:///opt/homebrew/opt/asdf/libexec, 
file:///usr/local/opt/asdf/libexec
```

### Note

A recent update broke this customization of the `asdf` setup.

I presume that it was this commit: [30bc0001](https://github.com/Shopify/ruby-lsp/commit/30bc0001b1085463d7bb420756c2de662c34202b).

I'm simply comparing the timeframe of the commit and the functionality change. (I use Ruby LSP every day.)

gi:
This adds support for detecting asdf configuration via its environment variables:
* ASDF_DIR: installation location
* ASDF_DATA_DIR: data directory

### Motivation

Fixes #2023 

### Implementation

If the environment variable is set, include it as the first option for where to search to validate the value.

### DIFF
--- a/vscode/src/ruby/asdf.ts
+++ b/vscode/src/ruby/asdf.ts
@@ -14,37 +14,71 @@ import { VersionManager, ActivationResult } from "./versionManager";
 // Learn more: https://github.com/asdf-vm/asdf
 export class Asdf extends VersionManager {
   async activate(): Promise<ActivationResult> {
-    const asdfUri = await this.findAsdfInstallation();
-    const asdfDaraDirUri = await this.findAsdfDataDir();
     const activationScript =
       "STDERR.print({env: ENV.to_h,yjit:!!defined?(RubyVM::YJIT),version:RUBY_VERSION}.to_json)";
 
     const result = await asyncExec(
-      `. ${asdfUri.fsPath} && asdf exec ruby -W0 -rjson -e '${activationScript}'`,
+      `asdf exec ruby -W0 -rjson -e '${activationScript}'`,
       {
         cwd: this.bundleUri.fsPath,
-        env: {
-          ASDF_DIR: path.dirname(asdfUri.fsPath),
-          ASDF_DATA_DIR: asdfDaraDirUri.fsPath,
-        },
       },
     );
 
     const parsedResult = this.parseWithErrorHandling(result.stderr);
 
+    const asdfInstallDir = await this.getAsdfInstallDir(parsedResult.env);
+    const asdfDataDir = await this.getAsdfDataDir(parsedResult.env);
+
     // ASDF does not set GEM_HOME or GEM_PATH. It also does not add the gem bin directories to the PATH. Instead, it
     // adds its shims directory to the PATH, where all gems have a shim that invokes the gem's executable with the right
     // version
     parsedResult.env.PATH = [
-      vscode.Uri.joinPath(asdfDaraDirUri, "shims").fsPath,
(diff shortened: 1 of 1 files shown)


gi:
I have signed the CLA!

vinistock (changes requested):

This approach would only work if the user opens VS Code through the terminal so that it can inherit the environment of the current shell.

Otherwise, VS Code's environment wouldn't include these environment variables - especially if they are set using shell scripts (which are not sourced automatically by the VS Code process, they are only source by the terminal).

If `ASDF_DIR` and `ASDF_DATA_DIR` can be override and that's a common thing to do, then the answer is to provide new settings so that users can configure that themselves.

Here's an example of how to do it https://github.com/Shopify/ruby-lsp/pull/1976.

gi:
@vinistock Thanks for the feedback.

FYI, a recent update broke this customization of the `asdf` setup.

I presume that it was this commit: [30bc0001](https://github.com/Shopify/ruby-lsp/commit/30bc0001b1085463d7bb420756c2de662c34202b).

I'm simply comparing the timeframe of the commit and the functionality change. (I use Ruby LSP every day.)

gi:
@vinistock Shouldn't the shell be able to determine the values of the `ASDF_DIR` and `ASDF_DATA_DIR` values, so why are they manually set?

vinistock:
We have decoupled all version manager integrations from shells. I explained in more detail [here](https://github.com/Shopify/ruby-lsp/pull/1976#issuecomment-2086120032). The summary is that trying to automatically discover the user's shell and source their configurations has lead to several problems and it's generally a brittle approach. The NodeJS process where extensions run is not the same as your shell/terminal. Nothing you configure in shell scripts is available there.

We're trying to push forward version manager integrations that do not connect to the shell in any way, so that we can achieve more reliable environment activation.

Compared to other version managers, ASDF is proving to be extremely coupled with shells (multiple users have raised issues). It seems that when we source the `asdf.sh` script from the NodeJS `exec` invocation (which uses `/bin/sh`), ASDF is not able to properly determine `ASDF_DIR` and `ASDF_DATA_DIR` automatically, which I believe is the source of many of the errors we're seeing.

If you have better ideas on how to make the integration smoother, please do tell us.

gi:
@vinistock I pushed an update with an alternative approach. I'm not sure this will work either based on the complexities you mentioned, but I'm curious on your insights.

gi:
I created an alternative and simpler update which simply includes the XDG base data directory as it is a common pattern: https://github.com/Shopify/ruby-lsp/pull/2045.

PR #2032, opened 2024-05-10
Synchronize ancestor changes

vinistock:
### Motivation

Another step towards https://github.com/Shopify/ruby-lsp/issues/1333.

We need to be able to synchronize the index when the user changes the ancestor tree. For example, by including a module that wasn't included before or changing a parent class.

### Implementation

I added an `ancestors_hash` method to namespace entries, so that we can more easily discover if changes to the tracked ancestors have been made.

If we identify that a change was made to any module/parent class, then we completely clear the linearization cache and start everything from scratch.

This might seem wasteful, but it's actually quite expensive to figure out exactly which ancestors need to have their linearization re-computed. Consider the following example:

1. The user removes an include from inside a module. That module's ancestor linearization is invalidated
2. However, every other class/module that includes/prepends/extends that module has also had their linearization invalidated, so we must bust the cache for those
3. But then after that, other classes and modules that depended on the ones we operate on step 2 have also had their chains invalidated. And this keeps going on until we found all possible namespaces that may be impacted by an ancestor change

All of that is way too expensive to run while the user is editing. It's easier to always clear the entire cache and do the work to linearize from scratch.

### Automated Tests

Added tests.


lib/ruby_indexer/lib/ruby_indexer/index.rb
```diff
@@ -356,6 +356,36 @@ def linearized_ancestors_of(name)
...
+        original_entries.select { |e| e.is_a?(Entry::Namespace) },
+        T::Array[Entry::Namespace],
+      ).to_h { |e| [e.name, e.ancestor_hash] }
+
+      updated_map = T.cast(
+        updated_entries.select { |e| e.is_a?(Entry::Namespace) },
+        T::Array[Entry::Namespace],
+      ).to_h { |e| [e.name, e.ancestor_hash] }
+
+      @ancestors.clear if original_map.any? { |name, hash| updated_map[name] != hash }
```
Morriar, line 386: I wonder if we actually need to clear the whole cache? Can't we just clear the entries for which their linearization contains an impacted name?

Let's say I have

```rb
class A; end
class B < A; end
class C; end
```

And then I change if to

```rb
class A < C; end
class B < A; end
class C; end
```

I only need to discard the linearizations for `A` and `B` right? `C` is not impacted.

On a small example this may not save much, but for a fully linearized index for a huge codebase, maybe being smarter can save a lot?

Morriar: Let's go with clearing the cache for now and see if it becomes a burden as we use the feature.

The good news is we should be able to GC most of the old linearization and reduce the server memory footprint 😅

Morriar (approved):

PR #2033, opened 2024-05-10
Improve client integration tests

vinistock:
### Motivation

While working on #2017, I realized that our client integration test was not activating Ruby (it just used the one available on CI), which was actually masking some issues.

This PR improves our tests and fixes the related issues to make tests pass.

### Implementation

The PR is split by commit:

1. Environment variables are case sensitive and Windows uses the name `Path` instead of `PATH`, which was causing us to miss the general environment on the RubyInstaller integration
2. This commit includes the improvements to the test, which now displays more information about failures and ensures the right Ruby is activated to run integrations

### Automated Tests

Added tests.

### DIFF
--- a/vscode/src/ruby/chruby.ts
+++ b/vscode/src/ruby/chruby.ts
@@ -43,6 +43,10 @@ export class Chruby extends VersionManager {
   async activate(): Promise<ActivationResult> {
     const versionInfo = await this.discoverRubyVersion();
     const rubyUri = await this.findRubyUri(versionInfo);
+    this.outputChannel.info(
+      `Discovered Ruby installation at ${rubyUri.fsPath}`,
+    );
+
     const { defaultGems, gemHome, yjit, version } =
       await this.runActivationScript(rubyUri);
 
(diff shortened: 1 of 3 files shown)


lib/ruby_lsp/listeners/document_link.rb
```diff
@@ -30,7 +30,7 @@ def gem_paths
                 lookup[spec.name] = {}
                 lookup[spec.name][spec.version.to_s] = {}
 
-                Dir.glob("**/*.rb", base: "#{spec.full_gem_path}/").each do |path|
+                Dir.glob("**/*.rb", base: "#{spec.full_gem_path.delete_prefix("//?/")}/").each do |path|
```
andyw8, line 33: In isolation, the `//?/` may be confusing. We could either add a comment, or extract it to give it a name, e.g. `WINDOWS_LONG_PATH_PREFIX`.

lib/core_ext/uri.rb
```diff
@@ -11,6 +11,9 @@ def from_path(path:, fragment: nil, scheme: "file")
         # On Windows, if the path begins with the disk name, we need to add a leading slash to make it a valid URI
         escaped_path = if /^[A-Z]:/i.match?(path)
           DEFAULT_PARSER.escape("/#{path}")
+        elsif path.start_with?("//?/")
+          # Some paths on Windows start with "//?/". This is a special prefix that allows for long file paths
```
andyw8, line 15: 🤯

andyw8 (approved):

PR #2044, opened 2024-05-13
Add method completion API

vinistock:
### Motivation

Just a tiny improvement over the sequence of changes for handling ancestors. We should offer an API from the index so that addons can get the right completion candidates for a known type instead of implementing the logic in our completion request.

### Implementation

Just extracted it into a method in the index.

### DIFF
--- a/lib/ruby_indexer/lib/ruby_indexer/index.rb
+++ b/lib/ruby_indexer/lib/ruby_indexer/index.rb
@@ -130,6 +130,16 @@ module RubyIndexer
       results.flat_map(&:first)
     end
 
+    sig { params(name: String, receiver_name: String).returns(T::Array[Entry]) }
+    def method_completion_candidates(name, receiver_name)
+      ancestors = linearized_ancestors_of(receiver_name)
+      candidates = prefix_search(name).flatten
+      candidates.select! do |entry|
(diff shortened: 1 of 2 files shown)


lib/ruby_indexer/lib/ruby_indexer/index.rb
```diff
@@ -129,6 +129,17 @@ def fuzzy_search(query)
       results.flat_map(&:first)
     end
 
+    sig { params(name: String, receiver_name: String).returns(T::Array[Entry]) }
+    def method_completion_candidates(name, receiver_name)
+      ancestors = linearized_ancestors_of(receiver_name)
+      candidates = prefix_search(name)
+      candidates = candidates.flatten
```
andyw8, line 136:
```suggestion
      candidates.flatten!
```

andyw8 (approved):

lib/ruby_indexer/lib/ruby_indexer/index.rb
vinistock: Unfortunately, Sorbet doesn't understand that `flatten!` will change the type of the `candidates` variable from `T::Array[T::Array[String]]` to `T::Array[String]`.

I don't like the style either, but it's better than having to use `T.unsafe`.

PR #2045, opened 2024-05-13
vscode/asdf: search the default XDG base data directory

### LINKED ISSUE #2023: vs code: asdf environment variables ignored

gi:
This updates the detection of the asdf data directory for setting the environment variable `ASDF_DATA_DIR` during activation.

Fixes #2023

### Motivation

Fixes #2023 

The XDG base directory specification is becoming more popular in outside of Linux. The default base data directory is `~/.local/share` where applications store their user data within a directory under the base: e.g. `~/.local/share/asdf`.

### Implementation

This update includes the XDG base data directory in the list when detecting the location for activation.

### Automated Tests

Yes.

### Manual Tests

1. Create the directory `~/.local/share/asdf/shims`
2. Start VS Code
3. Open Ruby file

### DIFF
--- a/vscode/src/ruby/asdf.ts
+++ b/vscode/src/ruby/asdf.ts
@@ -51,8 +51,19 @@ export class Asdf extends VersionManager {
   // file, but that may not be the case for a Homebrew installation, in which case the we'd have
   // `/opt/homebrew/opt/asdf/libexec/asdf.sh`, but the data directory might be `~/.asdf`
   async findAsdfDataDir(): Promise<vscode.Uri> {
+    // In order, the data locations are:
+    // 1. Default
+    // 2. XDG Base Directory
+    // 3. Homebrew M series
+    // 4. Homebrew Intel series
     const possiblePaths = [
       vscode.Uri.joinPath(vscode.Uri.file(os.homedir()), ".asdf"),
+      vscode.Uri.joinPath(
+        vscode.Uri.file(os.homedir()),
+        ".local",
+        "share",
+        "asdf",
+      ),
       vscode.Uri.joinPath(vscode.Uri.file("/"), "opt", "asdf-vm"),
       vscode.Uri.joinPath(
         vscode.Uri.file("/"),
--- a/vscode/src/ruby/asdf.ts
+++ b/vscode/src/ruby/asdf.ts
@@ -5,7 +5,7 @@ import os from "os";
 
 import * as vscode from "vscode";
 
-import { asyncExec } from "../common";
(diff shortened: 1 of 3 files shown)


vscode/src/ruby/asdf.ts
```diff
@@ -51,8 +51,19 @@ export class Asdf extends VersionManager {
   // file, but that may not be the case for a Homebrew installation, in which case the we'd have
   // `/opt/homebrew/opt/asdf/libexec/asdf.sh`, but the data directory might be `~/.asdf`
   async findAsdfDataDir(): Promise<vscode.Uri> {
+    // In order, the data locations are:
+    // 1. Default
```
andyw8, line 55: (nit) If we assign each possible path to variable, then we could avoid need for this comment.

andyw8:
(the rest of the team is at RubyKaigi so there might be a little delay until we ship this)

vinistock:
@gi I shipped a few improvements to the ASDF implementation which should fix many of the issues we were seeing before. I think this PR can just add the `.local/share` directory to the list and that should be good 👍.

gi:
@vinistock I just tested the latest version, and the changes work on my machine. Thanks for updating. I'm going to close this as the changes are no longer needed.

The issue was fixed by https://github.com/Shopify/ruby-lsp/pull/2006

PR #2047, opened 2024-05-14
Update README about method definition support

sathishmanohar:
I was under the impression that jump to definitions feature isn't available based on the README. I found out that there were recent improvements in this regard.

https://github.com/Shopify/ruby-lsp/releases/tag/v0.16.5

So, made an edit to remove the part in README indicating that jump to definitions feature isn't working.

### DIFF
--- a/README.md
+++ b/README.md
@@ -33,7 +33,7 @@ The Ruby LSP features include
 - Completion for classes, modules, constants and require paths
 - Fuzzy search classes, modules and constants anywhere in the project and its dependencies (workspace symbol)
 
-Adding method support for completion, hover and workspace symbol is planned, but not yet completed.
+Adding method support for definition, completion, hover and workspace symbol is partially supported, but not yet complete. Follow progress in https://github.com/Shopify/ruby-lsp/issues/899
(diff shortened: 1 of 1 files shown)


sathishmanohar:
I have signed the CLA!

README.md
```diff
@@ -33,7 +33,7 @@ The Ruby LSP features include
 - Completion for classes, modules, constants and require paths
 - Fuzzy search classes, modules and constants anywhere in the project and its dependencies (workspace symbol)
 
-Adding method support for definition, completion, hover and workspace symbol is planned, but not yet completed.
+Adding method support for completion, hover and workspace symbol is planned, but not yet completed.
```
andyw8, line 36: We're close, but definition isn't fully there just yet:

```suggestion
Method support for definition, completion, hover and workspace symbol is partially supported, but not yet complete. Follow progress in https://github.com/Shopify/ruby-lsp/issues/899
```

andyw8 (changes requested):

andyw8 (approved):

PR #2051, opened 2024-05-15
Fix the handling of multibyte characters

NotFounds:
### Motivation
refs: https://github.com/Shopify/ruby-lsp/pull/2021#issuecomment-2110394331

When a Ruby file contains multibyte characters (like Japanese, Chinese, emoji, etc), the go to definition and hover features do not work correctly. Because the document referencing logic does not properly handle multibyte characters when calculating offsets.

This commit fixes the issue by:
* Modifying document referencing to properly handle multibyte characters when mapping between positions and offsets
* Adding test cases to verify go to definition work with multibyte characters

### Implementation

1. Modified ﻿RubyLsp::Document and ﻿RubyLsp::Requests::Request to calculate locations considering multibyte characters. This change utilizes the API implemented in Prism by https://github.com/ruby/prism/pull/2406.
2. By passing the ﻿encoding when creating the ﻿Index, the index will be built taking multibyte characters into account.

### Automated Tests

I added some tests for the definition and `RubyLsp::RubyDocument`.

### Manual Tests

The definition jump for the following code snippet from https://github.com/shopify/ruby-lsp/issues/1347 works in this branch, but does not function correctly in the main branch.

```rb
class Test
  TEST = 'test'

  def method1
    # テスト
    pp TEST
  end
end
```


andyw8:
cc @kddnewton in case you have any views on the Prism usage here.

lib/ruby_lsp/requests/request.rb
```diff
@@ -18,19 +18,19 @@ def perform; end
       private
 
       # Checks if a location covers a position
-      sig { params(location: Prism::Location, position: T.untyped).returns(T::Boolean) }
-      def cover?(location, position)
+      sig { params(location: Prism::Location, position: T.untyped, encoding: Encoding).returns(T::Boolean) }
+      def cover?(location, position, encoding)
```
andyw8, line 22: Could we pass the encoding to the `Request` initializer, and save it an instance variable, so that don't need to pass it around in methods so much?

lib/ruby_lsp/document.rb
```diff
@@ -142,16 +142,20 @@ def locate(node, char_position, node_types: [])
 
         # Skip if the current node doesn't cover the desired position
         loc = candidate.location
-        next unless (loc.start_offset...loc.end_offset).cover?(char_position)
```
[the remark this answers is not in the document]
NotFounds: Thanks for your comments.
In my understanding, the char_position is provided by the editor, so we can't convert it into a byte offset :(
Alternatively, we can memorize the loc.start_code_units_offset and others to reduce the calculation.

lib/ruby_lsp/requests/request.rb
NotFounds: @andyw8 Since the position is now pre-calculated using byte offsets, there is no longer a need to pass the encoding to each method.

ref: https://github.com/Shopify/ruby-lsp/pull/2051/commits/be648a6ea20bb6a9c4fa31302a5e400a011e022b

lib/ruby_lsp/document.rb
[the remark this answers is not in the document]
NotFounds: @kddnewton 
Apologies for the delayed response. I have fixed the issue to convert byte offsets in this commit: https://github.com/Shopify/ruby-lsp/pull/2051/commits/be648a6ea20bb6a9c4fa31302a5e400a011e022b. Please review it.

lib/ruby_lsp/document.rb
```diff
@@ -180,6 +181,18 @@ def sorbet_sigil_is_true_or_higher
       end
     end
 
+    sig { params(char_position: Integer).returns(Integer) }
+    def convert_to_byte_offset(char_position)
+      case encoding
+      when Encoding::UTF_8
+        source.slice(0, char_position).bytesize
+      when Encoding::UTF_16, Encoding::UTF_16LE, Encoding::UTF_16BE
+        char_position * 2
```
NotFounds, line 190: Do we need to consider surrogate pairs?

NotFounds: Thank you for answering. I will work on making the logic common and adding tests.
I'm not very familiar with the directory structure and rules of ruby-lsp, but is it okay if I add the implementation to `lib/ruby_lsp/utils.rb`?

lib/ruby_lsp/utils.rb
```diff
@@ -26,6 +28,22 @@ module RubyLsp
     String,
   )
 
+  module_function
+
+  sig { params(source: String, char_position: Integer, encoding: Encoding).returns(Integer) }
+  def convert_to_byte_offset_position(source, char_position, encoding)
```
NotFounds, line 34: I was unsure whether to implement this in `utils.rb`.
Please let me know if you think there is a more appropriate place.

lib/ruby_lsp/test_helper.rb
```diff
@@ -21,7 +21,13 @@ def with_server(source = nil, uri = Kernel.URI("file:///fake.rb"), stub_no_typec
       &block)
       server = RubyLsp::Server.new(test_mode: true)
       server.global_state.stubs(:typechecker).returns(false) if stub_no_typechecker
-      server.global_state.apply_options({})
+      server.global_state.apply_options({
+        capabilities: {
+          general: {
+            positionEncodings: [LanguageServer::Protocol::Constant::PositionEncodingKind::UTF8],
```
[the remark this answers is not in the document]
NotFounds: In the previous implementation, I couldn't find any place where UTF-8 was explicitly specified.
However, given that tests have been run so far, it seems that it was used the UTF-8.

NotFounds:
It appears that from 'be648a6' onwards, it no longer works correctly on VSCode.
I'll try to fix it. I would appreciate any advice you can provide.

lib/ruby_lsp/requests/request.rb
vinistock, line 1: These changes look style related only. Can we remove them to make it easier to review?

lib/ruby_lsp/test_helper.rb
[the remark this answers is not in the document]
vinistock: The encoding was set in the `initialize` method to UTF-8, but indeed if we invoke `apply_options` with no arguments, it will change to UTF-16.

vinistock (commented):
The idea to convert the editor's position to a byte offset is interesting from a performance perspective, but I'm not sure it's the right approach. It seems reversed in my opinion.

For example, if we fix the locations in the indexer to use code units (which is required for us to return selection ranges properly), then implementing features like find references and rename could prove a bit weird because we would need to compare a byte offset with a code unit.

Or we would need to not apply the same conversion only in those cases.

I think I'd rather favour consistency of using the code unit APIs everywhere for now and if we find performance issues we can try to address those.

It's also relevant to mention that the code unit APIs are only less performant when there are unicode characters, since they require more complicated handling. For ASCII only sources, the performance should be pretty much the same.

NotFounds:
@vinistock
I would appreciate it if you could kindly provide your comments on this PR. Should we wait for the implementation of ResourceUri?

PR #2056, opened 2024-05-16
Feat: Add Extract Method to CodeAction

### LINKED ISSUE #1056: Add extract to method code action
We currently only support the extract to variable refactor. Extracting a piece of code into a method is another useful feature we can provide.

### Notes

- Since there can be many different refactors, organize the code better. The `CodeAction` request returns all available refactors, but resolution should be split into multiple classes. That is, we should create an `ExtractToMethod` request class and then later we can rename the current `CodeActionResolve` into `ExtractToVariable`
- To implement extract to method, we can start simple. Let's just:
    - Find the parent `DefNode` that covers the selected code
    - Place the selected code inside a `def new_method` ... `end` block below the parent `DefNode` we found
    - Switch the originally selected code by an invocation to `new_method`
- After the simple version is working, we can take it a step further by
    - Placing the new method after the `private` invocation in a class/module
    - Discovering which local variables the selected code depends on, so that we can create the new method with parameters

okuramasafumi:
### Motivation

Close #1056 

At Code Party, a side event for RubyKaigi 2024, @vinistock helped me to find a good first issue and I tried to implement this.

### Implementation

We started to accepts multiple kinds of action. We have corresponding method to deal with actions and if it doesn't match, it raises an error.

### Automated Tests

Tests added.

### Manual Tests

Select some texts in a method, click "Extract Method" action and find a new method with the same body.

### DIFF
--- a/lib/ruby_lsp/requests/code_action_resolve.rb
+++ b/lib/ruby_lsp/requests/code_action_resolve.rb
@@ -24,6 +24,7 @@ module RubyLsp
     class CodeActionResolve < Request
       extend T::Sig
       NEW_VARIABLE_NAME = "new_variable"
+      NEW_METHOD_NAME = "new_method"
 
       class CodeActionError < StandardError; end
 
@@ -31,6 +32,7 @@ module RubyLsp
         enums do
           EmptySelection = new
           InvalidTargetRange = new
+          UnknownCodeAction = new
         end
       end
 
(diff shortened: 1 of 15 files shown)


lib/ruby_lsp/requests/code_action_resolve.rb
```diff
@@ -135,6 +149,63 @@ def perform
...
+        source_range = @code_action.dig(:data, :range)
+        return Error::EmptySelection if source_range[:start] == source_range[:end]
+
+        scanner = @document.create_scanner
+        start_index = scanner.find_char_position(source_range[:start])
+        end_index = scanner.find_char_position(source_range[:end])
+        extracted_source = T.must(@document.source[start_index...end_index])
+
+        # Find the closest statements node, so that we place the refactor in a valid position
+        closest_statements, _ = T.cast(
```
vinistock, line 165:
```suggestion
        # Find the closest method declaration node, so that we place the refactor in a valid position
        closest_def, _ = T.cast(
```

vinistock (commented):
Just a renaming suggestion to make the names more accurate. Also, there are a few tests failing because now there are new code actions available for the same existing fixtures. Just adding the extra part of the response to the `.exp` files should be enough to fix it.

vinistock (approved):

PR #2067, opened 2024-05-16
Don't generate code lens for nested definitions

thomasmarshall:
### Motivation

When testing a [test](https://github.com/test-unit/test-unit/blob/f80ede80df1211e2756faea4835f9771b8258faf/test/test-test-case.rb#L15-L16) [framework](https://github.com/minitest/minitest/blob/f60c6f2e29082b404fb6933abb6a1813b2c09c5d/test/minitest/test_minitest_test.rb#L92-L94) or [plugin](https://github.com/thomasmarshall/minitest-verify/blob/bbe0fe5593e04ddcd2cb8d5b8b3f6153e6559222/test/minitest/verify_test.rb#L8-L12), it's common to have test methods inside test methods. This change ensures that the code lens is only generated for tests that will actually be run by the test runner, not for nested definitions.

| Before | After |
| - | - |
|  |  |

### Implementation

It tracks the depth of nested method definition nodes and returns early if the depth is greater than one level.

### Automated Tests

I've added a test that ensures

### Manual Tests

1. Write a test with a another test method inside

    ```rb
    def test_foo
      def test_bar
      end
    end
    ```
2. Confirm that code lens appears for `test_foo` but not `test_bar`


lib/ruby_lsp/listeners/code_lens.rb
```diff
@@ -42,6 +42,7 @@ def initialize(response_builder, global_state, uri, dispatcher)
         @group_stack = T.let([], T::Array[String])
         @group_id = T.let(1, Integer)
         @group_id_stack = T.let([], T::Array[Integer])
+        @def_depth = T.let(0, Integer)
```
andyw8, line 45: For the benefit of future maintainers, I think this is worth a brief explanation:
```suggestion
        # We want to avoid adding code lenses for nested definitions
        @def_depth = T.let(0, Integer)
```

PR #2071, opened 2024-05-19
Fix the hash keys of enabledFeatures

zhengpd:
The enabledFeatures has symbolic keys when configured with hash, which need to be transformed to string keys to work.

bug environment:
- nvim v0.10.0
- ruby-lsp 0.16.6

### DIFF
--- a/lib/ruby_lsp/server.rb
+++ b/lib/ruby_lsp/server.rb
@@ -134,7 +134,7 @@ module RubyLsp
       when Hash
         # If the configuration is already a hash, merge it with a default value of `true`. That way clients don't have
         # to opt-in to every single feature
-        Hash.new(true).merge!(configured_features)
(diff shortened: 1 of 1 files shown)


zhengpd:
I have signed the CLA!

zhengpd:
@andyw8 Here's my neovim config:

```lua
      local init_options = {
        enabledFeatures = {
          diagnostics = false,
        },
      }
      require("lspconfig").ruby_lsp.setup({
        init_options = init_options,
      })
```

I guess it's because `Transport::Stdio::Reader` is using `request = JSON.parse(message, symbolize_names: true)` to parse options, making the keys symbolized.

vinistock (approved):
Thank you for the contribution! Indeed, the initialization parameters are parsed with symbolized keys.

The reason we need them as strings is because of the legacy way of configuring the server, which uses an array of strings (which then becomes a hash with string keys).

Issue #2022, opened 2024-05-08
No RuboCop doc links for extensions


Earlopain:
### Description

Here is how it looks like for buildin cops:
![image](https://github.com/Shopify/ruby-lsp/assets/14981592/0f7f7300-4cfd-4c12-a89f-ca8c81cfe009)

RuboCop Rails (and all other extensions):
![image](https://github.com/Shopify/ruby-lsp/assets/14981592/92880269-1a05-4e6f-8d2e-51470f55ef5b)

This may be an upstream issue https://github.com/rubocop/rubocop/issues/12884

However I think it makes sense to report this here as well since I don't think there are many consumers of this API in the first place.

### Context

`documentation_url` will only return a result for base RuboCop cops. Per the methods documentation, extensions are encouraged to implement this themselves but none I know of do, including all the official ones.

https://github.com/rubocop/rubocop/blob/7f9595aa3ca3524f888b1464c4eeb7e15aabfc22/lib/rubocop/cop/base.rb#L63-L69

`rubocop --show-docs-url Rails/Exit` prints the correct result, however it relies on private API.

PR #2076, opened 2024-05-20
Show doc urls for extension cops

Earlopain:
Requires #2075 and a new RuboCop release.

### Motivation

#2022

### Implementation

https://github.com/rubocop/rubocop/pull/12907 added what was needed to make this work. Doc urls are dependant on `DocumentationBaseURL` from the config, which a class method simply doesn't have access to.

So, pass in the config we have from the runner to construct these. Check the method params to stay compatible with both old and new versions.

### Automated Tests

Yup

### Manual Tests

Open some test file and add a new test case with the following:

```rb
  def test_public
    assert_equal([], foo)
  end 
```

Hover over the diagnostic and check out the hyperlink:
image

Old RuboCop versions will continue to not show these links.

### DIFF
--- a/lib/ruby_lsp/requests/support/rubocop_diagnostic.rb
+++ b/lib/ruby_lsp/requests/support/rubocop_diagnostic.rb
@@ -19,6 +19,16 @@ module RubyLsp
           T::Hash[Symbol, Integer],
         )
 
+        ENHANCED_DOC_URL = T.let(
+          begin
+            gem("rubocop", ">= 1.64.0")
+            true
+          rescue LoadError
+            false
+          end,
+          T::Boolean,
+        )
+
         # TODO: avoid passing document once we have alternative ways to get at
         # encoding and file source
         sig { params(document: Document, offense: RuboCop::Cop::Offense, uri: URI::Generic).void }
@@ -38,8 +48,8 @@ module RubyLsp
           code_actions
         end
 
-        sig { returns(Interface::Diagnostic) }
-        def to_lsp_diagnostic
+        sig { params(config: RuboCop::Config).returns(Interface::Diagnostic) }
+        def to_lsp_diagnostic(config)
           # highlighted_area contains the begin and end position of the first line
           # This ensures that multiline offenses don't clutter the editor
           highlighted = @offense.highlighted_area
@@ -47,7 +57,7 @@ module RubyLsp
             message: message,
             source: "RuboCop",
             code: @offense.cop_name,
-            code_description: code_description,
+            code_description: code_description(config),
             severity: severity,
             range: Interface::Range.new(
               start: Interface::Position.new(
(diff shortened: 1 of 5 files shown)


lib/ruby_lsp/requests/support/rubocop_runner.rb
```diff
@@ -50,6 +50,9 @@ class ConfigurationError < StandardError; end
         sig { returns(T::Array[RuboCop::Cop::Offense]) }
         attr_reader :offenses
 
+        sig { returns(::RuboCop::Config) }
+        attr_reader :config_for_pwd
```
andyw8, line 54: Could we say `working_directory` rather than `pwd`?

Earlopain: Sure, I'll do this after a new RuboCop release

lib/ruby_lsp/requests/support/rubocop_diagnostic.rb
```diff
@@ -80,9 +80,17 @@ def severity
           RUBOCOP_TO_LSP_SEVERITY[@offense.severity.name]
         end
 
-        sig { returns(T.nilable(Interface::CodeDescription)) }
-        def code_description
-          doc_url = RuboCopRunner.find_cop_by_name(@offense.cop_name)&.documentation_url
+        sig { params(config: RuboCop::Config).returns(T.nilable(Interface::CodeDescription)) }
+        def code_description(config)
```
vinistock, line 84: What do you think about overriding this method for older versions to avoid the check on every diagnostic? Something like this

```ruby
def code_description(config)
  cop = RuboCopRunner.find_cop_by_name(@offense.cop_name)
  return unless cop

  doc_url = cop.documentation_url
  Interface::CodeDescription.new(href: doc_url) if doc_url
end

begin
  gem("rubocop", ">= 1.64.0")

  def code_description(config)
    cop = RuboCopRunner.find_cop_by_name(@offense.cop_name)
    return unless cop

    doc_url = cop.documentation_url(config)
    Interface::CodeDescription.new(href: doc_url) if doc_url
  end
rescue LoadError
  # do nothing
end
```

Earlopain: All these extension will have something returned, the department (`Rails`, `Minitest`, etc.) need to have `DocumentationBaseUrl` set. This is also the reason why you need to pass in the config. All (or at least most of) the official extensions do this properly: https://github.com/search?q=org%3Arubocop+DocumentationBaseUrl+-repo%3Arubocop%2Frubocop+-repo%3Arubocop%2Fdocs.rubocop.org&type=code. `rubocop-sorbet` doesn't set this which is why the expectation diff is so small.

In the end you land in this implementation: https://github.com/rubocop/rubocop/blob/54421d6df7fb245f836fe8074690171c02b3d7b6/lib/rubocop/cop/documentation.rb
Previously this only worked for buildin cops, which I changed. As you noticed asking others to add support explicitly didn't work quite out so well and adding generic support like that in rubocop was just easier than changing all the other extensions.

PR #2082, opened 2024-05-22
Improve performance of finding indexables

natematykiewicz:
Currently, all folders and files in the current tree are turned into IndexablePath, and then excluded files are filtered out after.

When there are large file trees that are meant to be excluded, this results in a lot of unnecessary work.

ActiveStorage stores files in the `tmp` directory in many many small folders. So does Bootsnap. Ruby LSP has to traverse all of these files, even though the entire directory should just be ignored.

Rubocop has solved this by breaking the `includes` patterns up into many patterns, applying the exclusions *before* the `Dir.glob`, so I followed in their footsteps. This works great for exclusions that end in "**/*". We still need to loop through all IndexablePath objects and see if they're excluded, in the case that an extension was provided on the excluded path, but this can cut down load time dramatically.

Before this PR in my Rails app, `indexables` took 76 seconds to run. Now it takes, 0.19 seconds. Before and after code both return the same exact file list.

Additionally, I added `node_modules` to the list of excluded trees, since that can be very large and never includes Ruby files.

I also removed the `*.rb` from the bundler path. Having a file extension on that means we need to scan all files. But we simply want to ignore the entire bundler path tree. I kind of wonder if we should always replace `*.rb` with `*`, to help people improve performance. Excluding an actual file name or partial file name makes sense. But excluding the only file extension we scan means we can do it faster by excluding the whole folder.

### Motivation

Opening a ruby file caused my LSP server to print "Ruby LSP: indexing files" for 76 seconds at 0% before the progress bar starts moving.

### Implementation

I knew that Rubocop has solved this problem before, so I looked at [this file](https://github.com/rubocop/rubocop/blob/aee64e2d5d6615ef64803af52c1b82bf95b566ef/lib/rubocop/target_finder.rb#L41) and followed what they did.

### Automated Tests

I added tests for the new pattern `exclude_pattern` that gets used with `fnmatch`, while ensuring I didn't break any existing tests.

### Manual Tests

I made a file that has both implementations of `indexables` on my computer. Then ran this in the Rails Console:

```ruby
Benchmark.measure { RubyIndexer::Configuration.new.indexables }
=>
#<Benchmark::Tms:0x000000012a2dead0
 @cstime=0.0,
 @cutime=0.0,
 @label="",
 @real=0.19083199999295175,
 @stime=0.0784180000000001,
 @total=0.18055299999999974,
 @utime=0.10213499999999964>


Benchmark.measure { RubyIndexer::Configuration.new.old_indexables }
=>
#<Benchmark::Tms:0x000000012a374b20
 @cstime=0.0,
 @cutime=0.0,
 @label="",
 @real=76.83584400010295,
 @stime=13.666566,
 @total=14.943883,
 @utime=1.277317>

old_indexables = RubyIndexer::Configuration.new.old_indexables
indexables = RubyIndexer::Configuration.new.indexables

indexables.size == old_indexables.size
=> true

indexables.map(&:full_path).sort == old_indexables.map(&:full_path).sort
=> true

indexables.size
=> 13933
```

So here you can see that it finds the same ~14k files in 0.25% of the time.

### DIFF
--- a/lib/ruby_indexer/lib/ruby_indexer/configuration.rb
+++ b/lib/ruby_indexer/lib/ruby_indexer/configuration.rb
@@ -31,6 +31,7 @@ module RubyIndexer
 
       @excluded_patterns = T.let(
         [
+          File.join("**", "*_test.rb"),
           File.join("node_modules", "**", "*"),
           File.join("spec", "**", "*"),
           File.join("test", "**", "*"),


lib/ruby_indexer/lib/ruby_indexer/configuration.rb
```diff
@@ -43,6 +50,15 @@ def initialize
...
     end
 
+    sig { returns(String) }
+    def exclude_pattern
+      relative_patterns = @excluded_patterns
+        .select { |p| p.end_with?("/**/*") }
+        .map { |p| p.delete_prefix("#{Dir.pwd}/") }
+
+      "#{Dir.pwd}/{#{relative_patterns.join(",")}}"
+    end
```
vinistock, line 60: Why do we need this method? Could we instead change `@excluded_patterns` to always begin with `Dir.pwd`? Or is the idea to guarantee that user added patterns also start with `Dir.pwd`?

lib/ruby_indexer/lib/ruby_indexer/configuration.rb
```diff
@@ -52,21 +68,25 @@ def indexables
       # having duplicates if BUNDLE_PATH is set to a folder inside the project structure
 
       # Add user specified patterns
-      indexables = @included_patterns.flat_map do |pattern|
+      patterns = indexable_included_file_patterns.sort.map! { |dir| File.join(dir, "*.rb") }
```
vinistock, line 71: Why do we need to sort?

lib/ruby_indexer/lib/ruby_indexer/configuration.rb
```diff
@@ -158,6 +178,37 @@ def apply_config(config)
...
+      flags = File::FNM_PATHNAME | File::FNM_EXTGLOB
+
+      # Escape File.fnmatch? wildcards in the directory
+      base_directory = base_directory.gsub(/[\\\{\}\[\]\*\?]/) do |reserved_glob_character|
+        "\\#{reserved_glob_character}"
+      end
+
+      dirs = Dir.glob(File.join(base_directory, "*/"), flags)
+        .reject do |dir|
+        next true if dir.end_with?("/./", "/../")
```
vinistock, line 192: Will these work on Windows?

lib/ruby_indexer/lib/ruby_indexer/configuration.rb
```diff
@@ -158,6 +178,37 @@ def apply_config(config)
 
     private
 
+    sig { params(base_directory: String, excluded_pattern: String).returns(T::Array[String]) }
+    def indexable_included_file_patterns(base_directory = Dir.pwd, excluded_pattern = exclude_pattern)
```
vinistock, line 182: Do we need the second argument if it's always just the default?

lib/ruby_indexer/lib/ruby_indexer/configuration.rb
```diff
@@ -158,6 +178,37 @@ def apply_config(config)
...
+
+      dirs
+        .flat_map { |dir| indexable_included_file_patterns(dir, excluded_pattern) }
+        .unshift(base_directory)
+    end
+
+    sig { params(base_dir: String, current_dir: String, exclude_pattern: String, flags: Integer).returns(T::Boolean) }
+    def symlink_excluded_or_infinite_loop?(base_dir, current_dir, exclude_pattern, flags)
+      dir_realpath = File.realpath(current_dir)
+      File.symlink?(current_dir.chomp("/")) && (
```
vinistock, line 206: The chomp may not work for Windows paths. Or does it?

vinistock (commented):
Great improvements 🚀

lib/ruby_indexer/lib/ruby_indexer/configuration.rb
natematykiewicz: I was trying to avoid computing it again. I suppose I could memoize it into the exclude_pattern method, but I'd need to invalidate that memoization any time users adjust their `@excluded_patterns` list.

lib/ruby_indexer/lib/ruby_indexer/configuration.rb
natematykiewicz: This method returns a single string. I have a test for it based on the `@excluded_patterns` in line 23. Perhaps it could use a comment. Check this test to see the computed output:

https://github.com/Shopify/ruby-lsp/pull/2082/files#diff-fd9cbb60b18cc9e588f3863d60dfcca34ff90e6cb488169105c973fbaf8eaba8R55

lib/ruby_indexer/lib/ruby_indexer/configuration.rb
natematykiewicz: Also, I'm feeling like `excluded_pattern` isn't as descriptive of a method name as it should be. Rubocop calls it `combined_exclude_glob_patterns`.

lib/ruby_indexer/lib/ruby_indexer/configuration.rb
natematykiewicz: Rubocop has had this code for 3 years, so I assume it works on Windows.

lib/ruby_indexer/lib/ruby_indexer/configuration.rb
natematykiewicz: https://github.com/Shopify/ruby-lsp/pull/2082/commits/244d64df10a77e3bb6ab59d201c481dc4699fa08

lib/ruby_indexer/lib/ruby_indexer/configuration.rb
natematykiewicz: Good call. I was considering using a bsearch at one point, but ended up not. Removed in https://github.com/Shopify/ruby-lsp/pull/2082/commits/228b84fe4aa336ebce632054499bbce407ab2f40

lib/ruby_indexer/lib/ruby_indexer/configuration.rb
natematykiewicz: Also, I realized the code would read more simply if I removed this optional argument and just made a call site pass in `Dir.pwd`.

https://github.com/Shopify/ruby-lsp/pull/2082/commits/b6abcc5ed334feace8150c4ef682f992f14b4272

natematykiewicz: I just added a test to ensure that calling `apply_config` with an excluded_pattern actually does recompute the fnmatch exclude pattern.

https://github.com/Shopify/ruby-lsp/pull/2082/commits/68bb45391d2a2529fa09ecb5e4673abb68179928

lib/ruby_indexer/lib/ruby_indexer/configuration.rb
natematykiewicz: So I asked a windows user what `Dir.pwd`, `File.join(Dir.pwd, 'foo', 'bar')`, and `File::ALT_SEPARATOR` return.

I expected the pwd to use back slashes, but it uses forward slashes. The file join also uses forward slashes. The alt separator is a backslash. So, the code we have here should be fine, since that's the same thing I see on my Mac.

Does Shopify have any developers running Windows that you could ask to run those same commands? I can read some Ruby C code, if we need. I do see that `File.join`'s documentation specifically says it joins using `"/"`. https://rubyapi.org/3.3/o/file#method-c-join

natematykiewicz:
@vinistock I believe I've addressed all of your comments

natematykiewicz:
@vinistock do you need anything else from me on this?

lib/ruby_indexer/lib/ruby_indexer/configuration.rb
natematykiewicz: Nope. It can't have a trailing slash.

```sh
$ ln -s test/fixtures/ test_symlink
```

```ruby
File.symlink?('test_symlink')
=> true

File.symlink?('test_symlink/')
=> false

File.symlink?('test_symlink/'.chomp('/'))
=> true
```

lib/ruby_indexer/lib/ruby_indexer/configuration.rb
```diff
@@ -52,21 +60,25 @@ def indexables
...
-        Dir.glob(pattern, File::FNM_PATHNAME | File::FNM_EXTGLOB).map! do |path|
-          path = File.expand_path(path)
-          # All entries for the same pattern match the same $LOAD_PATH entry. Since searching the $LOAD_PATH for every
-          # entry is expensive, we memoize it until we find a path that doesn't belong to that $LOAD_PATH. This happens
-          # on repositories that define multiple gems, like Rails. All frameworks are defined inside the Dir.pwd, but
-          # each one of them belongs to a different $LOAD_PATH entry
-          if load_path_entry.nil? || !path.start_with?(load_path_entry)
-            load_path_entry = $LOAD_PATH.find { |load_path| path.start_with?(load_path) }
+        Dir.glob(pattern, File::FNM_PATHNAME | File::FNM_EXTGLOB)
+          .map! do |path|
```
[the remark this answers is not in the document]
natematykiewicz: I don't even remember why I did that. I think I had more chained in at one point, so the line got too long and I needed to multiline the chain, then reverted whatever I added, and left it all tabbed over on accident. Anyways, I've reverted that in [this commit](https://github.com/Shopify/ruby-lsp/pull/2082/commits/1651c8a0c9a7172d433da297b6f4418ad4a9ef25), so now this section is untouched.

lib/ruby_indexer/lib/ruby_indexer/configuration.rb
```diff
@@ -154,10 +166,49 @@ def apply_config(config)
       @excluded_patterns.concat(config["excluded_patterns"]) if config["excluded_patterns"]
       @included_patterns.concat(config["included_patterns"]) if config["included_patterns"]
       @excluded_magic_comments.concat(config["excluded_magic_comments"]) if config["excluded_magic_comments"]
+      @fnmatch_exclude_pattern = compute_fnmatch_exclude_pattern
```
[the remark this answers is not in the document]
natematykiewicz: https://github.com/Shopify/ruby-lsp/pull/2082/commits/a29ec5432dddde5977dfeca5f511e9b8493064fb

lib/ruby_indexer/lib/ruby_indexer/configuration.rb
```diff
@@ -154,10 +166,49 @@ def apply_config(config)
...
     end
 
     private
 
+    sig { returns(String) }
+    def compute_fnmatch_exclude_pattern
```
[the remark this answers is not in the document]
natematykiewicz: https://github.com/Shopify/ruby-lsp/pull/2082/commits/a29ec5432dddde5977dfeca5f511e9b8493064fb

lib/ruby_indexer/lib/ruby_indexer/configuration.rb
```diff
@@ -154,10 +166,49 @@ def apply_config(config)
...
+      relative_patterns = @excluded_patterns
+        .select { |p| p.end_with?("/**/*") }
+        .map { |p| p.delete_prefix("#{Dir.pwd}/") }
```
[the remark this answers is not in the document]
natematykiewicz: That makes sense, though I don't think we want the `!` on delete_prefix. We're allocating a new array, but the strings are still the ones in `@excluded_patterns`, which means we'd be mutating the content of `@excluded_patterns`.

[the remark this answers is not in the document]
natematykiewicz: I added that change to the `select` line, but kept what I had on the `map` line, since I don't think we want to mutate `@excluded_patterns`' strings. Let me know if you think otherwise.
https://github.com/Shopify/ruby-lsp/pull/2082/commits/3a8c7fe6c2e42373f9a2508fa3dddfc36fd66d80

lib/ruby_indexer/lib/ruby_indexer/configuration.rb
```diff
@@ -52,7 +59,10 @@ def indexables
       # having duplicates if BUNDLE_PATH is set to a folder inside the project structure
 
       # Add user specified patterns
-      indexables = @included_patterns.flat_map do |pattern|
+      patterns = indexable_included_file_patterns(Dir.pwd, merge_exclude_patterns).map! { |dir| File.join(dir, "*.rb") }
+      patterns = [File.join(Dir.pwd, "**/*.rb")] if patterns.empty?
```
natematykiewicz, line 63: I'm realizing I've totally forgotten about `@included_patterns` and just treated it as hardcoded.

lib/ruby_indexer/lib/ruby_indexer/configuration.rb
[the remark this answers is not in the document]
natematykiewicz: Just to clarify, `included_patterns` should be relative to `Dir.pwd`, right? `apply_config` doesn't automatically prefix it with `Dir.pwd`, and I don't see how users would realistically set `Dir.pwd` in the YAML, considering they'd all have the repo checked out to different folders. Feels like we need to be prepending `Dir.pwd` to every entry in `included_patterns`. It also feels like `excluded_patterns` should similarly be expected to be within `Dir.pwd`. I'm struggling to understand what files someone would be including / excluding that aren't in a repo or a gem. But from what I can tell, excluded_patterns refers solely to application code, not gems. You exclude entire gems, not paths within a gem.

Right now I'm feeling like `include_patterns` doesn't even work, unless the user provides an absolute path to the folder on disk, which isn't very portable. What's the use-case for this? Understanding will help me figure out how to best optimize this, because the goal is to avoid calling `Dir.glob` on folders that are excluded, which means we need to take the included patterns, apply the exclude patterns that end with wildcards against them, so cut down on which folders `Dir.glob` scans.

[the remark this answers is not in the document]
vinistock: That's good point. Now that I think of it, all patterns have to be relative to something, since you can clone a project in any directory.

I'll bring this up with the team, but I think we should indeed make everything relative to `Dir.pwd`.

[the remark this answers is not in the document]
natematykiewicz: If we're including `File.join(Dir.pwd, "**", "*.rb")` by default, all I can really think that people might be including is non-`.rb` ruby files, like a Gemfile or Guardfile. So that leaves me really wondering what people even use `include_patterns` for. When people specify `include_patterns`, it doesn't seem to remove the default pattern. So anything they're including is likely already in the existing pattern, unless it doesn't end with `.rb` or is outside of Dir.pwd.

natematykiewicz:
2nd question @vinistock. Where'd we land on treating included_patterns and excluded_patterns as relative to `Dir.pwd`? It doesn't make much sense to me that someone would put an absolute path in there, because that would only work on 1 computer. This decision heavily affects how I go about solving this.

vinistock:
For the first question: I'm not sure we can always just exclude the entire `vendor` directory. Theoretically, there could be vendored dependencies not managed by Bundler in there. I would say let's start simple, with only the first level of directories and we can then follow up with an improvement. Maybe we can treat the `BUNDLE_PATH` inside the workspace as a special case.

For the second one, yeah, let's make all patterns relative to the workspace.

natematykiewicz:
Sounds good, thanks!

andyw8:
I'll set this to draft until @natematykiewicz has chance to return to it.

natematykiewicz:
I just saw that https://github.com/Shopify/ruby-lsp/pull/2424 was merged. That should make this PR much simpler.

natematykiewicz:
Note, since we're only excluding the very top-level directories, `vendor/bundle` still gets traversed even though it's my BUNDLE_PATH. As Vini alluded to, things get a lot more complicated if we try skip multiple layers to top-level directories (like `vendor/bundle/**/*`). I left a comment calling that out as well.

lib/ruby_indexer/lib/ruby_indexer/configuration.rb
```diff
@@ -60,21 +80,46 @@ def indexables
...
+      included_paths = Dir.glob(File.join(@workspace_path, "*/"), flags)
+        .reject { |dir| File.fnmatch?(merged_excluded_file_pattern, dir, flags) }
+        .map do |included_path|
+          relative_path = included_path
+            .delete_prefix(@workspace_path)
+            .tap { |path| path.delete_prefix!("/") }
+            .tap { |path| path.delete_suffix!("/") }
+
+          [included_path, relative_path]
+        end
```
vinistock, line 98: Two things about this section:

1. We can avoid looping twice by using `filter_map`
2. We should invoke `merged_excluded_file_pattern` only once. It's currently invoking it once per directory, which is more expensive than necessary

Also, why do we need to delete the `/` suffix twice?

```suggestion
      excluded_pattern = merged_excluded_file_pattern
      included_paths = Dir.glob(File.join(@workspace_path, "*/"), flags)
        .filter_map do |included_path|
          next if File.fnmatch?(excluded_pattern, dir, flags)
          
          relative_path = included_path
            .delete_prefix(@workspace_path)
            .tap { |path| path.delete_prefix!("/") }
            .tap { |path| path.delete_suffix!("/") }

          [included_path, relative_path]
        end
```

lib/ruby_indexer/lib/ruby_indexer/configuration.rb
```diff
@@ -52,6 +60,18 @@ def initialize
...
+    sig { returns(String) }
+    def merged_excluded_file_pattern
+      # This regex looks for @excluded_patterns that follow the format of "something/**/*", where
+      # "something" is one or more non-"/"
+      #
+      # Returns "/path/to/workspace/{tmp,node_modules}/**/*"
+      @excluded_patterns
+        .reject { |pattern| File.absolute_path?(pattern) }
+        .filter_map { |pattern| pattern.match(%r{\A([^/]+)/\*\*/\*\z})&.captures&.first }
+        .then { |dirs| File.join(@workspace_path, "{#{dirs.join(",")}}/**/*") }
```
vinistock, line 72:
```suggestion
      @excluded_patterns
        .filter_map do |pattern|
          next if File.absolute_path?(pattern)
          
          pattern.match(%r{\A([^/]+)/\*\*/\*\z})&.captures&.first 
        end
        .then { |dirs| File.join(@workspace_path, "{#{dirs.join(",")}}/**/*") }
```

lib/ruby_indexer/lib/ruby_indexer/configuration.rb
```diff
@@ -24,7 +24,15 @@ def initialize
       @workspace_path = T.let(Dir.pwd, String)
       @excluded_gems = T.let(initial_excluded_gems, T::Array[String])
       @included_gems = T.let([], T::Array[String])
-      @excluded_patterns = T.let([File.join("**", "*_test.rb"), File.join("tmp", "**", "*")], T::Array[String])
+
+      @excluded_patterns = T.let(
+        [
+          File.join("**", "*_test.rb"),
```
vinistock, line 30: Let's take advantage of the change and exclude the `test` and `spec` directories completely.
```suggestion
          File.join("test", "**", "*"),
          File.join("spec", "**", "*"),
```

vinistock (commented):
This is looking great

lib/ruby_indexer/lib/ruby_indexer/configuration.rb
natematykiewicz: Good call https://github.com/Shopify/ruby-lsp/pull/2082/commits/c16b3627aafa16ea9ad48c799646bd387cb4e286

lib/ruby_indexer/lib/ruby_indexer/configuration.rb
natematykiewicz: Yeah, I just removed that line. We want prefixing `/` gone, but we want to keep the suffix ones.
https://github.com/Shopify/ruby-lsp/pull/2082/commits/14495ade396c2772f29e8a97bee1eb9f1aca9558

natematykiewicz: The only reason we didn't have any tests failing is because we weren't including a path whose top-level directory started with an excluded directory name.

lib/ruby_indexer/lib/ruby_indexer/configuration.rb
natematykiewicz: Should we exclude `sorbet/**/*` while we're at it? At my job, `sorbet` directory is 1331 files. `tmp` is 146k files, which is why this change helps so much. Our `spec` directory is 558 files.

natematykiewicz:
@vinistock just so you know, I think I've handled all of your feedback

vinistock:
The changes look good to me, but could you please confirm if this produced any significant speed ups? I benchmarked these changes in two ways and saw no difference for our app, so I just want to confirm that it will actually achieve the desired outcome.

Benchmark 1

- On main, run the script 5 times. Calculate the average of the time spent indexing
- Switch to your branch and repeat
- Is there a noticeable difference?

```ruby
# frozen_string_literal: true

require "bundler/setup"
require "benchmark"
require "ruby_lsp/load_sorbet"
require "ruby_lsp/internal"

T::Utils.run_all_sig_blocks

index = RubyIndexer::Index.new
RubyVM::YJIT.enable

r = Benchmark.realtime do
  index.index_all
end

puts r
```

Benchmark 2 (IPS)

This needs `benchmark-ips`.

- On main, run the script once
- Switch to your branch
- Run the script again
- What is the reported result?

```ruby
# frozen_string_literal: true

require "bundler/setup"
require "benchmark/ips"
require "ruby_lsp/load_sorbet"
require "ruby_lsp/internal"

T::Utils.run_all_sig_blocks

index = RubyIndexer::Index.new
RubyVM::YJIT.enable

Benchmark.ips do |x|
  x.report("old") { index.configuration.indexables }
  x.report("new") { index.configuration.indexables }
  x.hold!("tmp_results")
  x.compare!
end
```

natematykiewicz:
So, how much this change helps entirely depends on how big the excluded directories are. I recently wiped my tmp/storage because it became unbearable to use RubyLsp with how large it was.

I just rebased my branch. On `main`, `RubyIndexer::Index.new.index_all` takes 12 seconds. On my branch it takes 8 seconds. At the beginning of this branch (back in May) it was taking 76 seconds to find the indexables.


Looking at just the `indexables` part right now, this is a ~4 second improvement. But again, the "before" was taking 76 seconds before. I'm just not sure how much of that ~72 second difference between my old "before" and me current "before" is a smaller `tmp` tree, and how much is the hundreds of commits that have happened between now and then.

```ruby
# "main" branch code
Benchmark.measure { RubyIndexer::Configuration.new.indexables }
=>
#<Benchmark::Tms:0x00000001197b3ee0
 @cstime=0.0,
 @cutime=0.0,
 @label="",
 @real=4.37744399998337,
 @stime=1.801075,
 @total=2.530652,
 @utime=0.7295769999999999>

# my changes
Benchmark.measure { RubyIndexer::Configuration2.new.indexables }
=>
#<Benchmark::Tms:0x000000011b9da8c0
 @cstime=0.0,
 @cutime=0.0,
 @label="",
 @real=0.7563209999352694,
 @stime=0.19453800000000054,
 @total=0.754276,
 @utime=0.5597379999999994>
```

vinistock (approved):
Thank you for the contribution! Let's move forward with this since you are seeing gains

natematykiewicz:
Thank you @vinistock!

PR #2083, opened 2024-05-22
Prevent test suite from modifying contributors' Bundler settings

natematykiewicz:
### Motivation

I install my gems in `vendor/bundle`. While working on #2082, I had to re-configure my bundle path every time I ran the test suite. I realized it was because the test suite was deleting my local bundler config.

### Implementation

I switched the tests to use `Bundler.settings.temporary`. You can read more about it here:
https://medium.com/@0xcolby/how-does-bundle-install-work-part-1-87a4349c192a

### Automated Tests

Current tests still work.

### Manual Tests

I can repeatedly run the test suite without having to set my bundle path to `vendor/bundle` every time.

### DIFF
--- a/lib/ruby_indexer/test/configuration_test.rb
+++ b/lib/ruby_indexer/test/configuration_test.rb
@@ -71,7 +71,7 @@ module RubyIndexer
       Bundler.settings.temporary(path: "vendor/bundle") do
         config = Configuration.new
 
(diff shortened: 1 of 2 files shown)


natematykiewicz:
I just pushed a similar commit over to #2082, since that PR is the one that needed to drop the extension, not this one.

PR #2085, opened 2024-05-23
Expand addon error information

andyw8:
As noted by @st0012 in https://github.com/standardrb/standard/pull/630#issuecomment-2127127463

### DIFF
--- a/lib/ruby_lsp/addon.rb
+++ b/lib/ruby_lsp/addon.rb
@@ -99,8 +99,10 @@ module RubyLsp
     end
 
     sig { returns(String) }
-    def backtraces
-      @errors.filter_map(&:backtrace).join("\n\n")
+    def errors_details
+      @errors.map do |error|
+        T.must(error.class.name) + ": " + error.message + "\n" + Array(error.backtrace).join("\n")
+      end.join("\n\n")
     end
 
     # Each addon should implement `MyAddon#activate` and use to perform any sort of initialization, such as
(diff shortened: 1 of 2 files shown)


lib/ruby_lsp/server.rb
```diff
@@ -104,7 +104,7 @@ def load_addons
           ),
         )
 
-        $stderr.puts(errored_addons.map(&:backtraces).join("\n\n"))
+        $stderr.puts(errored_addons.map(&:errors_details).join("\n\n"))
```
andyw8, line 107: When testing with a `raise` on the first like of `activate` in an addon, I found that the Notification isn't displaying. I'm not entirely sure why that is yet. But I think it's worth writing the error details alongside the backtrace in stderr.

lib/ruby_lsp/addon.rb
```diff
@@ -99,8 +99,10 @@ def formatted_errors
     end
 
     sig { returns(String) }
-    def backtraces
-      @errors.filter_map(&:backtrace).join("\n\n")
+    def errors_details
+      @errors.map do |error|
+        T.must(error.class.name) + ": " + error.message + "\n" + Array(error.backtrace).join("\n")
+      end.join("\n\n")
```
vinistock, line 105: I think this would include everything
```suggestion
      @errors.map do |error|
        error.full_message
      end.join("\n\n")
```

vinistock (approved):

PR #2086, opened 2024-05-23
Fix auto indent after end

thomasmarshall:
### Motivation

I noticed a few small issues since enabling on-type formatting recently. In particular, it adds incorrect indentation in some cases after typing the `end` keyword. It's not a big deal really because the code will be auto-formatted when saving the file, but I thought I would try to fix the easier ones to make the editing experience a bit nicer.

#### Complex body indentation

| Before | After |
| - | - |
|  |  |

#### Existing indentation

| Before | After |
| - | - |
|  |  |

### Implementation

- For the complex body indentation, it loops through each line of each node to add indentation as nodes can span multiple lines.
- For existing indentation, it finds the indentation correctly for comparison.

### Automated Tests

I've added a test case for each fix to `OnTypeFormattingTest`.

### Manual Tests

```rb
if true
  if true
    true
  end
en
```

Then add the final `d`.

```rb
if true
    true
en
```

Then add the final `d`.

### DIFF
--- a/lib/ruby_lsp/requests/on_type_formatting.rb
+++ b/lib/ruby_lsp/requests/on_type_formatting.rb
@@ -230,7 +230,9 @@ module RubyLsp
           loc = node.location
           next unless loc.start_column == @indentation
 
-          add_edit_with_text("  ", { line: loc.start_line - 1, character: 0 })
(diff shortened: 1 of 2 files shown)


vinistock (approved):
Thank you for the fix!

Issue #1775, opened 2023-11-21
Allow non-standard values for rubyLsp.formatter option


jscharf:
### This feature is VS Code specific

- [X] VS Code specific

### Use case

While working on a new addon last week, I came across this warning

```
Value is not accepted. Valid values: "auto", "rubocop", "syntax_tree", "none".(1)
```

<img width="941" alt="image" src="https://github.com/Shopify/vscode-ruby-lsp/assets/1317293/70004df0-80d5-4922-999e-1613b3874468">

I think we should be able to add new values in there without having a warning. This will make it less confusing for potential addon creators that their new value is acceptable and not going to cause a problem.

### Description

I heard that the value is currently an Enum, so this feature could potentially be just removing the Enum. 

However there may be an opportunity to do something more. 

### Implementation

If it's possible (I don't know if it is), we could have the list be the current list of `"auto", "rubocop", "syntax_tree", "none"` but then also include any known registered formatters. 

In the activate method of the addon, we run this type of code
```ruby
RubyLsp::Requests::Formatting.register_formatter("rubyfmt", RubyFmtFormatterRunner.instance)
```
So is it possible that we can detect which formatters have been registered to the system? 

If so, we could make the list equal the list of known registered formatters. 

Let me know what you think!

PR #2092, opened 2024-05-27
Add configuration options for third-party formatters

andyw8:
Closes https://github.com/Shopify/ruby-lsp/issues/1775

To properly support additional formatters, we need to change how formatters are selected when using VS Code.

Currently we use an `enum` with fixed set of choices. We _could_ change this to a text field, but that result in a poorer user experience.

Another option is to add the additional formatters to the enum list.

Although we generally try keep ruby-lsp itself unaware of specific addons, in this case it seems a worthwhile trade-off.

With the new choices being added, there is some additional information we want to show the user. There are two ways we could do this:

Using `enumDescriptions`:

Screenshot 2024-05-27 at 1 12 32 PM

Using `enumLabels`:

Screenshot 2024-05-27 at 1 11 30 PM

### DIFF
--- a/vscode/package.json
+++ b/vscode/package.json
@@ -295,7 +295,7 @@
             "Syntax Tree",
             "Standard (need v0.37 or newer of the standard gem)",
             "Rubyfmt (needs the ruby-lsp-rubyfmt gem)",
-            "None"
+            "Do not use a formatter"
           ],
           "default": "auto"
         },
--- a/vscode/package.json
+++ b/vscode/package.json
@@ -285,7 +285,17 @@
             "auto",
             "rubocop",
             "syntax_tree",
+            "standard",
+            "rubyfmt",
             "none"
+          ],
+          "enumDescriptions": [
+            "Auto",
+            "RuboCop",
+            "Syntax Tree",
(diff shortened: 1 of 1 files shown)


vscode/package.json
```diff
@@ -274,8 +274,18 @@
...
             "syntax_tree",
+            "standard",
+            "rubyfmt",
             "none"
           ],
+          "enumDescriptions": [
+            "Auto",
+            "RuboCop",
+            "Syntax Tree",
+            "Standard (need v0.37 or newer of the standard gem)",
```
andyw8, line 285: We'll co-ordinate with `@searls` to determine the release version for the work being done in https://github.com/standardrb/standard/pull/630.

st0012:
I'm not sure about maintaining this list while we don't have any control on the community addons. For example, `ruby-lsp-rubyfmt` is likely not working with the recent Ruby LSP releases as it doesn't use the latest addon interface. Can we instead ask users to choose `auto` and focus on letting the server pick the right formatter?

vinistock:
Yeah, I'm good with that 👍

Issue #2095, opened 2024-05-29
Stack level too deep error on indexing project with docusign_esign gem dependency


Galathius:
### Description

Recently tried to switch from my previous IDE to vscode using `ruby-lsp` extension. On one of my project, I got the indexing process stuck with the message `Ruby LSP: indexing files: 33% completed`.

Checking `Output` tab led me to `Stack level too deep` error:

<details>
<summary>Logs</summary>
2024-05-29 14:59:19.596 [info] (hireglobal-backend) #<Thread:0x000000010607e930 /Users/galathius/.asdf/installs/ruby/3.2.2/lib/ruby/gems/3.2.0/gems/ruby-lsp-0.16.7/lib/ruby_lsp/server.rb:691 run> terminated with exception (report_on_exception is true):

2024-05-29 14:59:19.601 [info] (hireglobal-backend) /Users/galathius/.asdf/installs/ruby/3.2.2/lib/ruby/gems/3.2.0/gems/prism-0.29.0/lib/prism/visitor.rb:31:in `each': stack level too deep (SystemStackError)
	from /Users/galathius/.asdf/installs/ruby/3.2.2/lib/ruby/gems/3.2.0/gems/prism-0.29.0/lib/prism/visitor.rb:31:in `visit_child_nodes'
	from /Users/galathius/.asdf/installs/ruby/3.2.2/lib/ruby/gems/3.2.0/gems/prism-0.29.0/lib/prism/dispatcher.rb:97:in `visit_and_node'
	from /Users/galathius/.asdf/installs/ruby/3.2.2/lib/ruby/gems/3.2.0/gems/prism-0.29.0/lib/prism/node.rb:537:in `accept'
	from /Users/galathius/.asdf/installs/ruby/3.2.2/lib/ruby/gems/3.2.0/gems/prism-0.29.0/lib/prism/visitor.rb:31:in `block in visit_child_nodes'
	from /Users/galathius/.asdf/installs/ruby/3.2.2/lib/ruby/gems/3.2.0/gems/prism-0.29.0/lib/prism/visitor.rb:31:in `each'
	from /Users/galathius/.asdf/installs/ruby/3.2.2/lib/ruby/gems/3.2.0/gems/prism-0.29.0/lib/prism/visitor.rb:31:in `visit_child_nodes'
[...]

PR #2097, opened 2024-05-29
Rescue SystemStackError while indexing

andyw8:
As noted in https://github.com/Shopify/ruby-lsp/issues/2095, there are some circumstance which can result in a stack overflow while indexing.


vinistock (changes requested):
I'm afraid of rescuing stack errors. It could be masking other issues in our own implementation.

Let's first understand if this issue is solvable with tweaks in the Prism AST. We faced the exact same issue before for string concatenations and it has since been fixed in Prism itself.

PR #2098, opened 2024-05-29
Resolve inherited instance variables

vinistock:
### Motivation

This PR adds a new method to the index to resolve instance variables using the ancestor linearization algorithm. This allows us to serve all features for inherited instance variables.

### Implementation

I split the implementation by commit:

1. Added a method to resolve instance variables using the linearized ancestors
2. Started using the method to support inherited variables in definition
3. Started using the method to support inherited variables in hover
4. Adapted the completion implementation to take ancestors into account

### Automated Tests

Added tests.

### Manual Tests

1. Launch the LSP on this branch
2. Create some code structure with inherited instance variables
3. Test out definition, hover and completion
4. Verify the inherited ivars are presented as expected

https://github.com/Shopify/ruby-lsp/assets/18742907/e5ecaa57-bf27-4829-90e1-5d311d7a2f6d

### DIFF
--- a/lib/ruby_indexer/lib/ruby_indexer/index.rb
+++ b/lib/ruby_indexer/lib/ruby_indexer/index.rb
@@ -376,6 +376,18 @@ module RubyIndexer
       entries.select { |e| ancestors.include?(e.owner&.name) }
     end
 
+    # Returns a list of possible candidates for completion of instance variables for a given owner name. The name must
+    # include the `@` prefix
+    sig { params(name: String, owner_name: String).returns(T::Array[Entry::InstanceVariable]) }
+    def instance_variable_completion_candidates(name, owner_name)
+      entries = T.cast(prefix_search(name).flatten, T::Array[Entry::InstanceVariable])
+      ancestors = linearized_ancestors_of(owner_name)
+
+      variables = entries.uniq(&:name)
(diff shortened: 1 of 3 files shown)


lib/ruby_indexer/lib/ruby_indexer/index.rb
```diff
@@ -365,6 +365,29 @@ def linearized_ancestors_of(fully_qualified_name)
       ancestors
     end
 
+    # Resolves an instance variable name for a given owner name. This method will linearize the ancestors of the owner
+    # and find inherited instance variables as well
+    sig { params(variable_name: String, owner_name: String).returns(T.nilable(T::Array[Entry::InstanceVariable])) }
+    def resolve_instance_variable(variable_name, owner_name)
+      entries = T.cast(self[variable_name], T.nilable(T::Array[Entry::InstanceVariable]))
+      ancestors = linearized_ancestors_of(owner_name)
+      return unless entries && ancestors.any?
```
st0012, line 374: Should we `return unless entries` first before we call `linearized_ancestors_of` to potentially save the computation in some cases?

vinistock: Good point! I'll create another PR.

Issue #2036, opened 2024-05-10
Re-think the way we locate nodes as targets and how addons declare the ones they are interested in.


andyw8:
### Description

As first mentioned [here](https://github.com/Shopify/ruby-lsp-rails/issues/371#issuecomment-2096006646).

Additional comment from @vinistock:
We need to think of a way for listener to say something like this:
```
method_call(:symbol)
             ^ attached behaviour when this is clicked
^ but include the method call so that we can verify that it's the method we're interested in
```

PR #2099, opened 2024-05-29
Introduce `NodeContext` for listeners

andyw8:
### Motivation

Current we make the (namespace) nesting available to listeners, but in some situations we require more information.

### Implementation

Changed `locate_node` return a `NodeContext`, and used Sorbet and the tests to identify everywhere that needed to be updated.

Note: This will be a breaking change for add-ons.

In a follow-up PR, we will add an additional field to `NodeContext` to track the current method. This will allow us to solve https://github.com/Shopify/ruby-lsp/issues/2036.

Co-authored with @vinistock 

### Automated Tests

Updated

### Manual Tests

n/a


lib/ruby_lsp/document.rb
```diff
@@ -170,7 +170,8 @@ def locate(node, char_position, node_types: [])
         end
       end
 
-      [closest, parent, nesting.map { |n| n.constant_path.location.slice }]
+      TargetContext.new(closest, parent, nesting.map { |n| n.constant_path.location.slice })
+      # [closest, parent, nesting.map { |n| n.constant_path.location.slice }]
```
vinistock, line 174: ✂️

lib/ruby_lsp/listeners/completion.rb
```diff
@@ -406,8 +408,8 @@ def build_entry_completion(real_name, incomplete_name, range, entries, top_level
       # ```
       sig { params(entry_name: String).returns(T::Boolean) }
       def top_level?(entry_name)
-        @nesting.length.downto(0).each do |i|
-          prefix = T.must(@nesting[0...i]).join("::")
+        @target_context.nesting.length.downto(0).each do |i|
+          prefix = T.must(@target_context.nesting[0...i]).join("::")
```
vinistock, line 412: This is invoked in a loop, so let's avoid the double method dispatch.
```suggestion
        nesting = @target_context.nesting
        nesting.length.downto(0).each do |i|
          prefix = T.must(nesting[0...i]).join("::")
```

lib/ruby_lsp/listeners/hover.rb
```diff
@@ -105,7 +105,7 @@ def on_call_node_enter(node)
         message = node.message
         return unless message
 
-        methods = @index.resolve_method(message, @nesting.join("::"))
+        methods = @index.resolve_method(message, @target_context.nesting.join("::"))
```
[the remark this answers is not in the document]
vinistock: Great idea, we should definitely do that.

lib/ruby_lsp/requests/code_action_resolve.rb
```diff
@@ -162,10 +164,9 @@ def refactor_method
         extracted_source = T.must(@document.source[start_index...end_index])
 
         # Find the closest method declaration node, so that we place the refactor in a valid position
-        closest_def, _ = T.cast(
-          @document.locate(@document.tree, start_index, node_types: [Prism::DefNode]),
-          [T.nilable(Prism::DefNode), T.nilable(Prism::Node), T::Array[String]],
-        )
+        # closest_def, _ = T.cast(
```
vinistock, line 167: ✂️

lib/ruby_lsp/requests/definition.rb
```diff
@@ -63,6 +63,10 @@ def initialize(document, global_state, position, dispatcher, typechecker_enabled
           ],
         )
 
+        target = target_context.closest
+        parent = target_context.parent
+        # nesting = target_context.nesting
```
vinistock, line 68: ✂️

test/ruby_document_test.rb
```diff
@@ -433,27 +433,26 @@ class Post < ActiveRecord::Base
     RUBY
 
     # Locate the `ActiveRecord` module
-    found, parent = document.locate_node({ line: 0, character: 19 })
-    assert_instance_of(Prism::ConstantReadNode, found)
-    assert_equal("ActiveRecord", T.cast(found, Prism::ConstantReadNode).location.slice)
+    # found, parent = document.locate_node({ line: 0, character: 19 })
```
vinistock, line 436: ✂️

lib/ruby_lsp/target_context.rb
```diff
@@ -0,0 +1,21 @@
+# typed: strict
+# frozen_string_literal: true
+
+module RubyLsp
+  class TargetContext
```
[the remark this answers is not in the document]
Morriar: Let's add some documentation as well to explain what is a `target`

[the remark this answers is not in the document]
Morriar: We have a similar concept in Spoom: [`NodeContext`](https://github.com/Shopify/spoom/blob/main/lib/spoom/deadcode/remover.rb#L372) which holds all the useful methods we need to call relatively to a node.

lib/ruby_lsp/target_context.rb
```diff
@@ -0,0 +1,21 @@
...
+    extend T::Sig
+
+    sig { returns(T.nilable(Prism::Node)) }
+    attr_reader :closest, :parent
```
st0012, line 9: nit: I prefer adding `_node` to them so it's immediately clear about what they returns, especially for addons that don't use Sorbet.

st0012: Also, should we simply call `closest_node` `target_node` instead 🤔

st0012 (commented):
I love the idea. Just some suggestions around naming.

lib/ruby_lsp/target_context.rb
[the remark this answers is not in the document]
andyw8: Thanks @Morriar, we like that name.

Issue #2052, opened 2024-05-15
Duplicate test actions


zrzka:
### Reproduction steps

VS Code: 1.89.1
Ruby LSP: v0.5.21
macOS: Sonoma 14.4.1 (23E224)
asdf to manage Ruby (not just) versions

I have a "monorepo" with "three" RoR applications, and I'm using Multi-root workspace. My repo structure is like this:

```
- engines/licensor (RoR engine)
- frontend (RoR app)
  (using the engines/licensor engine)
- backend (RoR app)
  (using the engines/licensor engine)
```

My workspace file contains:

```
{
  "folders": [
    {
      "name": "Root",
      "path": "."
    },
    {
      "name": "Backend",
      "path": "backend"
    },
    {
      "name": "Frontend",
      "path": "frontend"
    },
    {
      "name": "Engine",
      "path": "engines/licensor"
    }
  ],
  "settings": {
    "files.exclude": {
      // Folders from `folders` to remove them from the Root workspace
      "backend/**": true,
      "frontend/**": true,
      "engines/**": true  
    }
  }
}
```

Looking at the list of processes, I have three Ruby LSP servers running. Now, the fun begins.

<details>
<summary>Ignore, will create another issue ...</summary>
Go to Symbol in Editor... shows every symbol three times:

![CleanShot 2024-05-15 at 23 26 41@2x](https://github.com/Shopify/ruby-lsp/assets/1084172/27ebf0a1-aa5d-402f-8235-ceb8762cd2f9)
</details>

Run, Run In Terminal, Debug ... actions are there 3x:

![CleanShot 2024-05-15 at 23 29 30@2x](https://github.com/Shopify/ruby-lsp/assets/1084172/a8a37cd9-03f9-46a3-8f59-47ef88da147b)

It is not relevant which file is opened (e.g. Backend/Frontend/Engine root), everything is there 3x.

I'm wondering whether:

- I'm doing something terribly wrong
- It is a feature request to deduplicate all the things
- Or you have another recommendation how to use Ruby LSP in this kind of setup

andyw8:
:wave: I tried to replicate this by creating a new repo containing two Rails apps, but I'm not seeing the issue you reported:

https://github.com/andyw8/multiroot_monorepo

Is there a way that you can share a minimal repo structure to help diagnose the problem?

andyw8:
Nevermind, I got it:

![Uploading Screenshot 2024-05-16 at 1.38.20 PM.png…]()

[...]

PR #2101, opened 2024-05-29
Fix duplicate LSP features appearing in multiroot workspaces

andyw8:
Closes https://github.com/Shopify/ruby-lsp/issues/2052

Co-authored with @vinistock 

(~TODO: Log an issue for [vscode-languageserver-node](https://github.com/microsoft/vscode-languageserver-node) about seeing unexpected requests for semanticToken for a workspace that isn't even open~) https://github.com/microsoft/vscode-languageserver-node/issues/1487

Screenshot 2024-05-29 at 12 06 39 PM

### DIFF
--- a/vscode/src/client.ts
+++ b/vscode/src/client.ts
@@ -91,6 +91,7 @@ function collectClientOptions(
   configuration: vscode.WorkspaceConfiguration,
   workspaceFolder: vscode.WorkspaceFolder,
   outputChannel: WorkspaceChannel,
+  ruby: Ruby,
 ): LanguageClientOptions {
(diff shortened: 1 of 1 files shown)


vinistock (approved):

andyw8:
For each app in the multiroot workspace there is a seperate LSP instance. If we have only `documentSelector: [{ language: "ruby" }]` then it causes each LSP instance to respond to the requests for Ruby files, causing the duplication. So we also need to limit each to the `pattern` corresponding to the workspace.

st0012:
Ah that makes sense. Thx 👍

PR #2105, opened 2024-05-30
Andyw8/detect missing fixtures submodule

andyw8:
We have this check already in `bin/test` but if you're running the test some other way then it will be missed and you'll get a confusing failure.

### DIFF
--- a/lib/ruby_indexer/test/index_test.rb
+++ b/lib/ruby_indexer/test/index_test.rb
@@ -299,6 +299,10 @@ module RubyIndexer
     end
 
     def test_indexing_prism_fixtures_succeeds
+      unless Dir.exist?("test/fixtures/prism/test/prism/fixtures")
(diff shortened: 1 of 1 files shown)


vinistock (commented):
Two things:

1. Do we still need the code in `bin/test` after this change?
2. Can we go one step further and offer to run this for the user + continue with the test suite? I'm thinking something really scrappy like this

```ruby
if fixtures_are_missing
  puts "Fixtures are missing...update submodule? (Y/n)"
  answer = gets
  if answer.downcase != "y"
    abort("Prism fixtures are required to run tests")
  end

  system("git submodule update --init")
end
```

andyw8:
We _could_, but if someone uses the VS Code test runner UI then it would hang. 🤔

andyw8:
(updated `bin/test`)

PR #2107, opened 2024-05-30
Introduce NodeContext#fully_qualified_name

andyw8:
As discussed in https://github.com/Shopify/ruby-lsp/pull/2099


lib/ruby_lsp/node_context.rb
```diff
@@ -17,5 +17,10 @@ def initialize(closest, parent, nesting)
       @parent = parent
       @nesting = nesting
     end
+
+    sig { returns(String) }
+    def fully_qualified_name
+      @nesting.join("::")
```
vinistock, line 23: I wonder if we should set this in the `initialize` and add an `attr_reader` for it, since the nesting doesn't change after instantiation. Thoughts?

vinistock (approved):

Issue #1996, opened 2024-04-30
Generated instructions in automated test can't execute correctly on Windows


Super-Xray:
### Reproduction steps

1. Start the Ruby LSP using a certain editor
2. Open the ruby test file(for example test\requests\definition_expectations_test.rb)
3. click `run in terminal` attached to a test function
4. the instruction is like `bundle exec ruby -Itest xxx/test/requests/definition_expectations_test.rb --name DefinitionExpectationsTest\#test_jumping_to_method_definitions_when_declaration_exists`
5. output is not correct:

**Code snippet or error message**
```
# Running tests with run options --name DefinitionExpectationsTest\#test_jumping_to_method_definitions_when_declaration_exists --seed 868:



Finished tests in 0.001863s, 0.0000 tests/s, 0.0000 assertions/s.


0 tests, 0 assertions, 0 failures, 0 errors, 0 skips
```

but If I modifed the instruction into `bundle exec ruby -Itest xxx/test/requests/definition_expectations_test.rb --name DefinitionExpectationsTest#test_jumping_to_method_definitions_when_declaration_exists` (just delete a '\\')
and then, the ouput is better: 

```
# Running tests with run options --name DefinitionExpectationsTest#test_jumping_to_method_definitions_when_declaration_exists --seed 47762:

.

Finished tests in 0.050074s, 19.9706 tests/s, 19.9706 assertions/s.


1 tests, 1 assertions, 0 failures, 0 errors, 0 skips
```

The os is win11, I run it on VSCode 1.88.1. The extension edition is v0.5.20. But the starting info in the output window is `Initializing Ruby LSP v0.16.7`，I'm not very sure the role of edition in this phenomenon. I hope somebody can check it. Thanks!

andyw8:
@Super-Xray Can you try this branch please? https://github.com/Shopify/ruby-lsp/pull/2109

Super-Xray:
Sure!

Super-Xray:
I think this problem has been solved. Thanks! @andyw8

PR #2109, opened 2024-05-30
Fix test escaping on Windows

andyw8:
Closes https://github.com/Shopify/ruby-lsp/issues/1996

This isn't a perfect solution, since characters may still be wrongly escaped by `ShellWords` for Windows, but I think it's a worthwhile improvement for now.


vinistock (approved):
Smart solution! Let's wait for the issue reporter to confirm it does indeed work on Windows though.

I will also encourage you to submit a Redmine ticket about Shellwords as it might be a bug they want to fix there.

st0012 (approved):

PR #2115, opened 2024-05-31
Add CallNode field to NodeContext

andyw8:
When we listen for node events, there are some situations where we are interested in not just the node, but the surrounding context. For cases such as a DSL call `has_many :users`, we make a definition request for the `:users` symbol, but we also want to know the surrounding method (`has_many` in this case). This will allow features such as https://github.com/Shopify/ruby-lsp-rails/pull/397

The implementation is fairly simple: While iterating over the queue, if the `parent` is a `CallNode`, then we keep track of that. So at the end of the iteration, we have the innermost `CallNode`, which we can then pass to `NodeContext.new`.

### DIFF
--- a/lib/ruby_lsp/document.rb
+++ b/lib/ruby_lsp/document.rb
@@ -128,6 +128,7 @@ module RubyLsp
       closest = node
       parent = T.let(nil, T.nilable(Prism::Node))
       nesting = T.let([], T::Array[T.any(Prism::ClassNode, Prism::ModuleNode)])
+      call_node = T.let(nil, T.nilable(Prism::CallNode))
 
       until queue.empty?
         candidate = queue.shift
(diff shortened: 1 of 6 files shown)


lib/ruby_lsp/document.rb
```diff
@@ -159,6 +160,10 @@ def locate(node, char_position, node_types: [])
           nesting << candidate
         end
 
+        if candidate.is_a?(Prism::CallNode)
+          call_node = candidate
```
andyw8, line 164: I feel there may be situations that aren't properly handled with this approach, but haven't found a concrete one yet. 🤔

st0012:
@andyw8 could you add some description in the PR(s)? Neither PR describes what feature/problem we're trying to solve and it's not easy to review with this little information 🙂

lib/ruby_lsp/document.rb
andyw8: Added a test and fixed the behaviour.

andyw8:
I've expanded the PR description.

st0012 (commented):
What do we expect in this cases:

```rb
foo do
  123 # this is the target
end
```

```rb
foo(
  if bar
    123 # this is the target
  end
end
```

Are we expecting `foo` being the call node in both cases?

andyw8:
@st0012 good question. This doesn't handle integers, but if we imagine those are symbols or string instead, then I'd say `NodeContext#call_node` should be `nil` for each, i.e. we only care about the `call_node` if it's the _immediate_ parent. I've pushed an update to handle that case.

andyw8:
> Not a blocker, but what are the criteria for the types of node to have this field?

It's hard to say since we can't anticipate all the possible things an addon author might want to build. Currently I just have the Rails DSLs in mind. We can consider others case-by-case.

> 
This doesn't handle integers, but if we imagine those are symbols or string instead

Not a blocker, but what are the criteria for the types of node to have this field?

so that we do match non-immediate parents.

> In that case, I think it should be named `closest_call_node` instead?

I'm open to changing it to that. @vinistock ?

PR #2123, opened 2024-06-03
Run labels check only after PR approval

andyw8:
Addresses part of https://github.com/Shopify/ruby-lsp/issues/2081

As per https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#running-a-workflow-when-a-pull-request-is-approved

### DIFF
--- a/.github/workflows/require_labels.yml
+++ b/.github/workflows/require_labels.yml
@@ -1,11 +1,12 @@
 name: Pull Request Labels
 
 on:
-  pull_request:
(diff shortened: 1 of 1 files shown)


vinistock (approved):

st0012 (approved):

vinistock:
Based on the documentation you linked https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#running-a-workflow-when-a-pull-request-is-approved, it seems that the trigger should be on `approve` and not `submit`.

PR #2124, opened 2024-06-03
Handle unsaved files in main language server client

vinistock:
### Motivation

With the changes in #2101, we regressed with an interesting bug. Because of the narrower document selector, we were simply ignoring unsaved files.

Theoretically, that shouldn't case any issues since we're just ignoring them. However, received reports of https://github.com/Shopify/ruby-lsp/issues/1897#issuecomment-2144452992 happening again and indeed it's easy to reproduce on main.

That issue only happens when we receive an unexpected request for a document before a `textDocument/didOpen` notification gets sent for it (which should never happen).

The reason it's happening is the bug we identified https://github.com/microsoft/vscode-languageserver-node/issues/1487, where the LSP package is sending unexpected semantic token requests without respecting the document selector.

So, the order of the events were:

1. You open an unsaved Ruby file
2. We didn't receive a `textDocument/didOpen` notification for it, because we were accidentally ignoring it in the document selector
3. The LSP package mistakenly sends a semantic token request anyway
4. Since we never got notified that the file was opened in the first place, it breaks the server

Quite an interesting interaction of different issues.

### Implementation

For unsaved files, it's not possible to know to which workspace they belong to. That information only exists after the user chooses where to save the file.

What we need to do is have the main language client (the first one) handle all files while they are unsaved. If the user decides to save the file in a workspace that does not belong to that main client, that is completely fine. VS Code will actually send `didClose` notifications for the `untitled:Untitled-1` uri and then send a `didOpen` notification for the file that was just saved to the correct language client.

The implementation is not the prettiest, but I don't know how else we could go about it.

### DIFF
--- a/vscode/src/client.ts
+++ b/vscode/src/client.ts
@@ -12,6 +12,7 @@ import {
   ExecutableOptions,
   ServerOptions,
   MessageSignature,
+  DocumentSelector,
 } from "vscode-languageclient/node";
 
 import { LSP_NAME, ClientInterface } from "./common";
@@ -92,6 +93,7 @@ function collectClientOptions(
   workspaceFolder: vscode.WorkspaceFolder,
   outputChannel: WorkspaceChannel,
   ruby: Ruby,
+  isMainWorkspace: boolean,
 ): LanguageClientOptions {
   const pullOn: "change" | "save" | "both" =
     configuration.get("pullDiagnosticsOn")!;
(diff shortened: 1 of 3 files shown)


vscode/src/client.ts
```diff
@@ -105,17 +107,27 @@ function collectClientOptions(
...
+      language: "ruby",
+      scheme: "untitled",
+    });
+  }
+
   // For each workspace, the language client is responsible for handling requests for:
   // 1. Files inside of the workspace itself
   // 2. Bundled gems
   // 3. Default gems
+  // Only the first language server we spawn should handle unsaved files, otherwise requests will be duplicated across
```
st0012, line 128: I think this comment is better suited for line 117?

vscode/src/client.ts
```diff
@@ -171,6 +183,7 @@ export default class Client extends LanguageClient implements ClientInterface {
     createTestItems: (response: CodeLens[]) => void,
     workspaceFolder: vscode.WorkspaceFolder,
     outputChannel: WorkspaceChannel,
+    mainWorkspace = false,
```
st0012, line 186: Can we prefix this with `is`, or make the name indicate better that it's a boolean?
Asking because in other places, we have variables like `firstWorkspace` or `workspace` that actually present `workspaceFolder` objects. I think it'll help readability if we can make a clearer distinction between this variable and those.

PR #2125, opened 2024-06-03
Better clarify multi-root workspace documentation

### LINKED ISSUE #1761: Please document what configuration multi-root workspaces actually expect
### Operating System

Ubuntu 23.10

### Ruby version

ruby 3.1.2p20 (2022-04-12 revision 4491bb740a) [x86_64-linux-gnu]

### Project has a bundle

- [X] Has bundle

no manager

### Description

I am attempting to use the LSP on a rails project within a subfolder of a multi-root vscode workspace (configured with a foo.code-workspace file). 

```
# /home/matt/foo/foo.code-workspace
{
	"folders": [
		{
			"path": "../../web/oncall"
		},
		{
			"path": "foo-fe"
		},
		{
			"path": "foo-be"
		},
		{
			"path": "../x"
		}
	],
	"settings": {}
}
```

The rails project is located at `/home/matt/foo/foo-be/front` (e.g. `/home/matt/foo/foo-be/front/Gemfile{,.lock}`. There is no other ruby code within the workspace. 

AFAIK the workspace is correctly configured (VS code works fine, other extensions work fine).

Whenever I start and try to use the LSP I receive an error stating:
```
Activating the Ruby LSP in /home/matt/foo/foo-be, but no lockfile was found. Are you using a monorepo setup?
```

Trying to debug this leads to a maze of documentation which doesn't actually state anywhere what the LSP is expecting.
* The error itself points to https://github.com/Shopify/vscode-ruby-lsp?tab=readme-ov-file#multi-root-workspaces which simply says to follow the VSCode guidelines and links to 
* https://code.visualstudio.com/docs/editor/workspaces#_multiroot-workspaces which documents only that you can add multiple folders and the format for doing so in the .code-workspace file, there is a small link to a more detailed page at
* https://code.visualstudio.com/docs/editor/multi-root-workspaces but still this page does not appear to describe any further requirements that I've missed.

So what's wrong? There's no way to tell because the linked docs only point to generic VSCode docs that don't specify any particular requirements other than the .code-workspace file containing a list of folders.

**Suggestion**
The sentence: "The Ruby LSP supports this use case out of the box as long as my_project contains configuration that follows [VS Code's guidelines](https://code.visualstudio.com/docs/editor/workspaces#_multiroot-workspaces) for multi-root workspaces."

specifically the "guidelines" reference, should be replaced wiht a more specific description of what guideline or expectation the LSP has that the reader will learn from the following the link.

vinistock:
### Motivation

Closes #1761

Our current documentation on mulit-root workspaces is a bit limited and we defer a lot to VS Code's own documentation. I think we can provide more examples and try to explain in more detail what problem multi-root workspaces solve.


vscode/README.md
```diff
@@ -254,28 +254,117 @@ These are the settings that may impact the Ruby LSP's behavior and their explana
...
+The advantage of adopting this configuration is that VS Code and all extensions are informed about which directories
+should be considered as possible workspace roots. Instead of having to configure each extension or tool individually so
+that are aware of your project structure, you only have to do that once for the entire repository.
+
+Some examples of functionality that benefits from multi-root workspaces:
+
+- Extensions that have to integrate with project dependencies (Gemfile, package.json), such as debuggers, language
+  servers, formatters and other tools, are informed about where to search for these files (allowing for automatic
+  detection)
+- If launch.json configurations are placed inside a workspace, VS Code will know to launch them from the appropriate
```
andyw8, line 271:
```suggestion
- If `launch.json` configurations are placed inside a workspace, VS Code will know to launch them from the appropriate
```

andyw8 (approved):
Do we want to re-iterate that you need to click the Open Workspace button?

vscode/README.md
```diff
@@ -254,28 +254,117 @@ These are the settings that may impact the Ruby LSP's behavior and their explana
 
 ### Multi-root workspaces
 
+Multi-root workspaces are VS Code's way to allow users to organize a single repository into multiple distinct concerns.
```
st0012, line 257: Since we already decided to stick with VS Code's multi-root workspace approach, I think it's worth emphasize this here. Maybe with something like:

```md
> ![NOTE]
> To make sure Ruby LSP works well with your multi-root workspace project, please
> read through the instructions below and configure it following the examples
```

andyw8 (approved):

PR #2129, opened 2024-06-04
Add `case` statement to VS Code snippets

SimonBrazell:
### Motivation

Adds a VScode snippet for the Ruby `case` statement.

A pretty trivial one but I've been missing this snippet lately so thought I'd add it here 😁

### Implementation

Pretty straightforward, just adds a snippet that outputs the following:

```ruby
case 
when  
  
when  
  
else
  
end
```

### Automated Tests

No tests, as it's just a snippet, see manual test steps.

### Manual Tests

Type `case` in a ruby file and choose the snippet presented in the context menu.

### DIFF
--- a/vscode/snippets.json
+++ b/vscode/snippets.json
@@ -189,7 +189,16 @@
   },
   "Case": {
     "prefix": ["case"],
-    "body": ["case $0", "when  $1", "  $2", "when  $3", "  $4", "else", "  $5", "end"],
+    "body": [
+      "case $0",
+      "when  $1",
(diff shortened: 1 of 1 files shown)


andyw8 (changes requested):
Thanks, all good except for the tabs issue mentioned.

(reminder for other maintainers - to verify this, you'll need to temporary disable the Sorbet extension since since it has its own `case` snippet, and then run it through the extension host).

andyw8 (approved):

PR #2131, opened 2024-06-05
Exclude dependencies from workspace symbols

### LINKED ISSUE #1277: Exclude gem symbols from appearing in global symbol search

### Use case

When I use VS Code's "Go to file..." with a symbol prefix, it lists all gems as well as my code. I'd rather not jump to symbols in my gems.  I discussed with @vinistock on Twitter and he asked me to raise an issue https://twitter.com/vinistock/status/1742528125135651264

### Description

A configuration option to disable indexing all gems, leaving just our application code in the index.

### Implementation

_No response_

vinistock:
### Motivation

Closes #1277

This PR excludes dependency symbols from the workspace symbol fuzzy search.

### Implementation

Just swapped the conditionals and moved the typechecker to an early return (since we'll do nothing in that case).

### Automated Tests

Adjusted the tests.


test/requests/workspace_symbol_test.rb
```diff
@@ -52,7 +52,7 @@ module Bar; end
     assert_equal(RubyLsp::Constant::SymbolKind::CONSTANT, T.must(result).kind)
   end
 
-  def test_matches_only_gem_symbols_if_typechecker_is_present
+  def test_returns_empty_if_typechecker_is_present
```
andyw8, line 55:
```suggestion
  def test_returns_empty_array_if_typechecker_is_present
```

st0012 (approved):

Issue #1335, opened 2024-01-29
Develop strategy for indexing declarations made in C / native extensions


vinistock:
We currently index all Ruby files bundled with the Ruby installation, but that doesn't cover all declarations available. Core classes like `Array` or `String` are defined in C and thus we miss those declarations. We also miss gem declarations that are made in native extensions.

We need to consider the performance, accuracy and maintenance costs of a few different approaches to decide how we're going to handle this:

- RDoc uses regexes to extract declarations from C files directly. The problem with that is that the C files aren't always included in Ruby installations, so those files may not be there
- We could read the RDoc index objects, but that is also dependent on the installation since Ruby may be installed without documentation
- We can read and process RBS files. However, we'll miss gems that don't include RBS signatures
- We could also bundle empty declarations for the Core library in Ruby files, which is essentially what RBS signatures are. This would involve some manual work to create these files and we would need to ensure they are up to date based on Ruby versions
- Some other idea?

---

```[tasklist]
### Tasks
- [x] [Index Core classes using RBS](https://github.com/Shopify/ruby-lsp/pull/2132)
- [x] [Extract includes, prepends and extends and insert those as well](https://github.com/Shopify/ruby-lsp/pull/2148)
- [x] Index Core methods using RBS https://github.com/Shopify/ruby-lsp/pull/2157
- [ ] https://github.com/Shopify/ruby-lsp/issues/2328
- [ ] https://github.com/Shopify/ruby-lsp/issues/2327
- [ ] https://github.com/Shopify/ruby-lsp/issues/2326
- [ ] https://github.com/Shopify/ruby-lsp/issues/2329
```

andyw8:
As per the team discussion earlier today, we will move ahead with using RBS declarations for indexing core class (String, Integer, etc).

vinistock:
[This branch](https://github.com/Shopify/ruby-lsp/tree/vs/index_rbs) includes a super simple script for how to retrieve declarations which I paired on with Soutaro during RubyKaigi.

My understanding of it is that we can basically just iterate over the declarations that the code returns and then build our index's entries and push them.

PR #2132, opened 2024-06-05
Index core classes using RBS

andyw8:
Step towards https://github.com/Shopify/ruby-lsp/issues/1335

To test:

- Disable the Sorbet extension and hover over a core class name, e.g. String.
- Verify you see the docs in the hover
- Use Go To Definition and verify it takes you to the RBS file with the String definitions.

After this we can move on to methods.


lib/ruby_lsp/server.rb
```diff
@@ -690,6 +690,7 @@ def shutdown
 
     sig { params(config_hash: T::Hash[String, T.untyped]).void }
     def perform_initial_indexing(config_hash)
+      RubyIndexer::IndexRbs.new(@global_state.index).index_core_classes
```
andyw8, line 693: Currently this is fast, so I don't think need to have any logging or progress indicator. Maybe we will once it's expanded to methods, etc.

andyw8:
(seeing CI failures for older Rubies, looking into...)

st0012 (approved):
LGTM 👍 
Once this is merged, I have a branch that will introduce `Object`, `Kernel`, and `BasicObject` to the ancestors list.

lib/ruby_indexer/lib/ruby_indexer/index_rbs.rb
```diff
@@ -0,0 +1,45 @@
+# typed: strict
+# frozen_string_literal: true
+
+module RubyIndexer
+  class IndexRbs
```
vinistock, line 5: The name `IndexRbs` makes it sound like there are two indices: one for RBS and the other one.

Can we call this indexer, collector or something like that?

lib/ruby_indexer/lib/ruby_indexer/index_rbs.rb
```diff
@@ -0,0 +1,45 @@
...
+      end
+    end
+
+    private
+
+    sig { params(source: T.untyped, pathname: Pathname, declarations: T::Array[RBS::AST::Declarations::Base]).void }
+    def process_signature(source, pathname, declarations)
+      declarations.each do |declaration|
+        process_declaration(declaration, pathname)
+      end
```
vinistock, line 29: Should we inline this inside the loop? Is there a need for the `process_signature` method?

lib/ruby_indexer/lib/ruby_indexer/index_rbs.rb
```diff
@@ -0,0 +1,45 @@
...
+      return unless declaration.is_a?(RBS::AST::Declarations::Class)
+
+      nesting = [declaration.name.name.to_s]
+      file_path = pathname.to_s
+      location = RubyIndexer::Location.from_rbs_location(declaration.location)
+      comments = Array(declaration.comment&.string&.lines)
+      parent_class = nil
+      class_entry = Entry::Class.new(nesting, file_path, location, comments, parent_class)
+      @index << class_entry
+    end
```
vinistock, line 43: Since we're going to process multiple declaration types coming from RBS, can we already setup a case statement that only processes classes?

For example

```ruby
declarations.each do |declaration|
  case declaration
  when RBS::AST::Declarations::Class
    handle_class(declaration)
  else
    # Unhandled RBS declaration. Do nothing
  end
end
```

lib/ruby_indexer/lib/ruby_indexer/index_rbs.rb
```diff
@@ -0,0 +1,45 @@
...
+
+    sig { params(declaration: RBS::AST::Declarations::Base, pathname: Pathname).void }
+    def process_declaration(declaration, pathname)
+      return unless declaration.is_a?(RBS::AST::Declarations::Class)
+
+      nesting = [declaration.name.name.to_s]
+      file_path = pathname.to_s
+      location = RubyIndexer::Location.from_rbs_location(declaration.location)
+      comments = Array(declaration.comment&.string&.lines)
+      parent_class = nil
```
vinistock, line 40: Is this `nil` variable needed? Can we not get the parent class?

lib/ruby_indexer/test/configuration_test.rb
```diff
@@ -54,7 +54,8 @@ def test_indexables_includes_default_gems
 
       assert_includes(indexables, "#{RbConfig::CONFIG["rubylibdir"]}/pathname.rb")
       assert_includes(indexables, "#{RbConfig::CONFIG["rubylibdir"]}/ipaddr.rb")
-      assert_includes(indexables, "#{RbConfig::CONFIG["rubylibdir"]}/abbrev.rb")
+      # TODO: understand why this fails for older Rubies
+      # assert_includes(indexables, "#{RbConfig::CONFIG["rubylibdir"]}/abbrev.rb")
```
vinistock, line 58: I assume this is failing because now `abbrev` is a bundled dependency pulled by `rbs` rather than a default gem.

Just switch `abbrev` for some other default gem.

lib/ruby_lsp/server.rb
vinistock: How fast is it on a Core Spin instance?

vinistock (commented):
This can be handled in the next PR, but we need to extract includes, prepends and extends and insert those as well.

lib/ruby_indexer/lib/ruby_indexer/index_rbs.rb
Morriar: Let's use this occasion to rectify the naming, I think we should have one `Index`, the data structure, and a bunch of `XIndexer`, the things collecting data to populate the index: `RubyIndexer`, `RBSIndexer`, `RBIIndexer`, etc.

lib/ruby_indexer/lib/ruby_indexer/index_rbs.rb
Morriar: You should be able to access it from the declaration: https://github.com/ruby/rbs/blob/64a96b134a9c2e4441d389f3895e3dcdba9f8938/lib/rbs/ast/declarations.rb#L92

lib/ruby_indexer/lib/ruby_indexer/location.rb
```diff
@@ -8,6 +8,19 @@ class Location
     sig { returns(Integer) }
     attr_reader :start_line, :end_line, :start_column, :end_column
 
+    class << self
+      extend T::Sig
+      sig { params(rbs_location: RBS::Location).returns(Location) }
+      def from_rbs_location(rbs_location)
```
Morriar, line 14: This may be better suited as a method of the RBS indexer to reduce coupling

lib/ruby_indexer/lib/ruby_indexer/index_rbs.rb
andyw8: Hmm, I prefer it this way for readability.

lib/ruby_indexer/lib/ruby_indexer/index_rbs.rb
andyw8: I've renamed it to `RubyIndexer`.

@Morriar I agree with your suggestion, but we can do that in a follow-up?

ruby-lsp.gemspec
```diff
@@ -20,6 +20,7 @@ Gem::Specification.new do |s|
 
   s.add_dependency("language_server-protocol", "~> 3.17.0")
   s.add_dependency("prism", ">= 0.29.0", "< 0.30")
+  s.add_dependency("rbs", ">= 3", "< 4")
```
andyw8, line 23: I verified it on 3.0.0. With 2.0.0 it fails because "each_signature" isn't defined.

andyw8: (I'm open to removing the upper constraint)

lib/ruby_indexer/lib/ruby_indexer/index_rbs.rb
andyw8: Thanks, updated.

lib/ruby_indexer/lib/ruby_indexer/rbs_indexer.rb
```diff
@@ -0,0 +1,67 @@
...
+
+      sig { params(rbs_location: RBS::Location).returns(RubyIndexer::Location) }
+      def to_ruby_indexer_location(rbs_location)
+        RubyIndexer::Location.new(
+          rbs_location.start_line,
+          rbs_location.end_line,
+          rbs_location.start_column,
+          rbs_location.end_column,
+        )
+      end
```
vinistock, line 19: Do we need this method outside of this class? If not, it should be a private instance method.

ruby-lsp.gemspec
vinistock: Let's keep the upper constraint. We can't guarantee that a major version won't break the LSP.

lib/ruby_lsp/server.rb
```diff
@@ -690,6 +690,20 @@ def shutdown
...
+      index_ruby_code(config_hash)
+    end
+
+    sig { void }
+    def index_core_classes
+      start_time = Process.clock_gettime(Process::CLOCK_MONOTONIC)
+      RubyIndexer::RBSIndexer.new(@global_state.index).index_core_classes
+      end_time = Process.clock_gettime(Process::CLOCK_MONOTONIC)
+      $stderr.puts("Indexed Ruby core classes in #{(end_time - start_time).round(2)} seconds")
+    end
```
vinistock, line 703: We intentionally moved any sort of benchmarking to the client to avoid doing more work on the server. Let's not benchmark the execution time here please.

We can add benchmark scripts if necessary, but I honestly think it should be fine.

lib/ruby_indexer/lib/ruby_indexer/index_rbs.rb
Morriar: Yep 👍

vinistock (approved):
Great work! 🚀

st0012 (approved):

hsw15192617273:
hi, jruby maybe not be able to install rbs. 

```log
❯ gem install rbs                                                                                                                                                                                                                        (base)
Building native extensions. This could take a while...
ERROR:  Error installing rbs:
	ERROR: Failed to build gem native extension.

    current directory: /foo/bar/.rbenv/versions/jruby-9.4.7.0/lib/ruby/gems/shared/gems/rbs-3.5.1/ext/rbs_extension
/foo/bar/.rbenv/versions/jruby-9.4.7.0/bin/jruby -I /foo/bar/.rbenv/versions/jruby-9.4.7.0/lib/ruby/stdlib extconf.rb
checking for whether -std=gnu99 is accepted as CFLAGS... *** extconf.rb failed ***
Could not create Makefile due to some reason, probably lack of necessary
libraries and/or headers.  Check the mkmf.log file for more details.  You may
need configuration options.

Provided configuration options:
	--with-opt-dir
	--without-opt-dir
	--with-opt-include
	--without-opt-include=${opt-dir}/include
	--with-opt-lib
	--without-opt-lib=${opt-dir}/lib
	--with-make-prog
	--without-make-prog
	--srcdir=.
	--curdir
	--ruby=/foo/bar/.rbenv/versions/jruby-9.4.7.0/bin/jruby
RuntimeError: The compiler failed to generate an executable file.
You have to install development tools first.

         try_do at /foo/bar/.rbenv/versions/jruby-9.4.7.0/lib/ruby/stdlib/mkmf.rb:456
    try_compile at /foo/bar/.rbenv/versions/jruby-9.4.7.0/lib/ruby/stdlib/mkmf.rb:571
    with_werror at /foo/bar/.rbenv/versions/jruby-9.4.7.0/lib/ruby/stdlib/mkmf.rb:522
    try_compile at /foo/bar/.rbenv/versions/jruby-9.4.7.0/lib/ruby/stdlib/mkmf.rb:571
     try_cflags at /foo/bar/.rbenv/versions/jruby-9.4.7.0/lib/ruby/stdlib/mkmf.rb:635
  append_cflags at /foo/bar/.rbenv/versions/jruby-9.4.7.0/lib/ruby/stdlib/mkmf.rb:641
   checking_for at /foo/bar/.rbenv/versions/jruby-9.4.7.0/lib/ruby/stdlib/mkmf.rb:942
       postpone at /foo/bar/.rbenv/versions/jruby-9.4.7.0/lib/ruby/stdlib/mkmf.rb:350
           open at /foo/bar/.rbenv/versions/jruby-9.4.7.0/lib/ruby/stdlib/mkmf.rb:320
       postpone at /foo/bar/.rbenv/versions/jruby-9.4.7.0/lib/ruby/stdlib/mkmf.rb:350
           open at /foo/bar/.rbenv/versions/jruby-9.4.7.0/lib/ruby/stdlib/mkmf.rb:320
       postpone at /foo/bar/.rbenv/versions/jruby-9.4.7.0/lib/ruby/stdlib/mkmf.rb:346
   checking_for at /foo/bar/.rbenv/versions/jruby-9.4.7.0/lib/ruby/stdlib/mkmf.rb:941
  append_cflags at /foo/bar/.rbenv/versions/jruby-9.4.7.0/lib/ruby/stdlib/mkmf.rb:640
           each at org/jruby/RubyArray.java:1981
  append_cflags at /foo/bar/.rbenv/versions/jruby-9.4.7.0/lib/ruby/stdlib/mkmf.rb:639
         <main> at extconf.rb:3

To see why this extension failed to compile, please check the mkmf.log which can be found here:

  /foo/bar/.rbenv/versions/jruby-9.4.7.0/lib/ruby/gems/shared/extensions/universal-java-17/3.1.0/rbs-3.5.1/mkmf.log

extconf failed, exit code 1

Gem files will remain installed in /foo/bar/.rbenv/versions/jruby-9.4.7.0/lib/ruby/gems/shared/gems/rbs-3.5.1 for inspection.
Results logged to /foo/bar/.rbenv/versions/jruby-9.4.7.0/lib/ruby/gems/shared/extensions/universal-java-17/3.1.0/rbs-3.5.1/gem_make.out
```

andyw8:
@hsw15192617273 we don't currently support JRuby, but feel free to note this on https://github.com/Shopify/ruby-lsp/pull/1257

vinistock:
@andyw8 sorry, I don't think that's accurate. The Ruby LSP does work on JRuby, we just didn't add it to CI.

Is it possible to add a gemspec constraint to only use `rbs` on non JRuby platforms?

st0012:
> Is it possible to add a gemspec constraint to only use rbs on non JRuby platforms?

I don't think it's possible

> The Ruby LSP does work on JRuby, we just didn't add it to CI.

Wasn't it practically impossible because Sorbet can't run on it?

andyw8:
@@hsw15192617273 was Ruby LSP working for you previously? (i.e. before RBS was introduced).

hsw15192617273:
> @@hsw15192617273 was Ruby LSP working for you previously? (i.e. before RBS was introduced).

year, if install `ruby-lsp 0.16.7`, it's working fine. (I started using JRuby today)

hsw15192617273:
but i not use sorbet in ruby code. just run ruby-lsp for vscode extension. JRuby install sorbet maybe not raise exception, I will confirm it tomorrow.

soutaro:
I planned to implement a pure Ruby RBS parser to make it compatible to other Ruby implementations, but not finished yet...

vinistock:
@andyw8 for the time being, to unblock our JRuby friends, let's remove the gemspec dependency from RBS and check for the presence of the gem instead.

```ruby
begin
  gem("rbs", ">= 3", "< 4")
rescue LoadError
  return
end

module RubyLsp
  class RBSIndexer
    # ...
```
And then where we use it

```ruby
index_rbs if defined?(RubyLsp::RBSIndexer)
```

PR #2136, opened 2024-06-05
Implement complete constant resolution algorithm

vinistock:
### Motivation

Now that we have ancestor linearization, we can finally resolve constants properly. This PR implements the algorithm for constant resolution in Ruby - with a few twists due to the nature of lazy static analysis.

### Implementation

We tried to document the code as much as possible. I'll describe the algorithm, which has essentially 3 distinct paths.

#### Top level references

```ruby
module First
  module Second
    ::CONST
  end
end
```

Top level references are the simplest ones. The nesting is completely ignored. All we have to do is try to find the name in the index. We do need to account for aliases too since the top level reference can include a portion of the name that points to an alias.

#### Qualified references

```ruby
module First
  module Second
    Third::CONST
  end
end
```

For qualified references, this is the order of resolution:

1. First, try to find the declaration within the exact namespace where we found the reference `First::Second::Third::CONST`
2. Then, try to find the declaration in ancestors of the namespace where we found the reference. We search all ancestors of `First::Second`
3. Then, try to find it in enclosing lexical scopes excluding the top level (which is always considered only as a fallback). We search `First::Third::CONST`
4. Finally, since we didn't find anything, search for `Third::CONST` in the top level

#### Unqualified references

```ruby
module First
  module Second
    CONST
  end
end
```

Unqualified references have the order of operations swapped:

1. First, try to find the declaration within the exact namespace where we found the reference `First::Second::CONST`
2. Then, try to find it in enclosing lexical scopes **before searching ancestors**. This also excludes the top level. We search `First::CONST`
3. Then, try to find the declaration in ancestors of the namespace where we found the reference. We search all ancestors of `First::Second`
4. Finally, since we didn't find anything, search for `CONST` in the top level

#### Notes

1. Ruby never searches the ancestors of any enclosing lexical scope except the most specific one
2. We may want to create a cache for resolutions. Maybe a hash that maps the concatenated `name + nesting` and points to the resolved name. However, knowing when to invalidate the cache would be quite tricky. I recommend we avoid caching for now
3. Fun fact: thanks to the lazy nature of our static analysis, we are actually able to "properly" resolve circular references despite them being invalid Ruby code. See `IndexTest#test_resolving_circular_alias and IndexTesttest_resolving_circular_alias_three_levels`. This is why we need that `seen_names` array, which not only protects the index from infinite recursion, but also allows us to resolve these cases

### Automated Tests

Added a bunch of tests for various constant reference scenarios, but if you can think of anything that we missed, please propose more tests.

### DIFF
--- a/lib/ruby_indexer/lib/ruby_indexer/index.rb
+++ b/lib/ruby_indexer/lib/ruby_indexer/index.rb
@@ -140,35 +140,53 @@ module RubyIndexer
       candidates
     end
 
-    # Try to find the entry based on the nesting from the most specific to the least specific. For example, if we have
-    # the nesting as ["Foo", "Bar"] and the name as "Baz", we will try to find it in this order:
-    # 1. Foo::Bar::Baz
-    # 2. Foo::Baz
-    # 3. Baz
-    sig { params(name: String, nesting: T::Array[String]).returns(T.nilable(T::Array[Entry])) }
-    def resolve(name, nesting)
+    # Resolve a constant to its declaration based on its name and the nesting where the reference was found. Parameter
+    # documentation:
+    #
+    # name: the name of the reference how it was found in the source code (qualified or not)
+    # nesting: the nesting structure where the reference was found (e.g.: ["Foo", "Bar"])
+    # seen_names: this parameter should not be used by consumers of the api. It is used to avoid infinite recursion when
+    # resolving circular references
+    sig do
+      params(
+        name: String,
+        nesting: T::Array[String],
+        seen_names: T::Array[String],
+      ).returns(T.nilable(T::Array[Entry]))
+    end
+    def resolve(name, nesting, seen_names = [])
+      # If we have a top level reference, then we just search for it straight away ignoring the nesting
       if name.start_with?("::")
-        name = name.delete_prefix("::")
-        results = @entries[name] || @entries[follow_aliased_namespace(name)]
-        return results&.map { |e| e.is_a?(Entry::UnresolvedAlias) ? resolve_alias(e) : e }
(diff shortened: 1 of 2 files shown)


lib/ruby_indexer/test/index_test.rb
```diff
@@ -46,6 +46,7 @@ class Bar
           end
 
           class Baz
+            puts Foo::Bar
```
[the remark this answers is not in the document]
vinistock: Oops. Leftover testing.

lib/ruby_indexer/lib/ruby_indexer/index.rb
```diff
@@ -434,22 +452,147 @@ def handle_change(indexable)
...
+      else
+        # The name includes some parts of the nesting. We need to remove the redundant parts
+        "#{namespace}::#{T.must(name_parts[index..-1]).join("::")}"
+      end
+    end
+
+    sig { params(full_name: String, seen_names: T::Array[String]).returns(T.nilable(T::Array[Entry])) }
+    def direct_or_aliased_constant(full_name, seen_names)
+      entries = @entries[full_name] || @entries[follow_aliased_namespace(full_name)]
+      entries&.map { |e| e.is_a?(Entry::UnresolvedAlias) ? resolve_alias(e, seen_names) : e }
```
Morriar, line 595: I was thinking of extracting just this part so it can be better reused, example lines 189 and 509

vinistock: I extracted the specific usage of lines 189 and 509 instead. This method is used more often than the usage of lines 189 and 509, it is used 5 times.

Morriar (approved):

st0012 (approved):

Issue #1939, opened 2024-04-15
Support singleton constants resolution (and/or indexing)


st0012:
Example:

```rb
class Foo
  BAR = 10

  class << self
    B # suggest BAR
    CAT = 20
    C # suggest CAT

    def bar
      B # suggest BAR
      C # suggest CAT
    end
  end
end
```

```rb
module Qux
  BAZ = 321
end

module Baz
  prepend Qux
end

class Foo
  BAR = 123
  extend Baz

  class << self
    B # suggest BAR
  end
end
```

PR #2142, opened 2024-06-06
Associate singleton methods with the right owner objects

vinistock:
### Motivation

First step for #1939

This PR starts the work to handle singleton contexts properly. It creates the singleton class object and starts associating methods to it.

The next steps are handling constants defined inside singleton contexts and properly handling instance variables.

### Implementation

First commit creates the `SingletonClass` object. Since they have ancestors of their own, I made it a `Namespace`. Note that modules also have singleton classes.

Also, note that singleton classes also have their own singleton class. Ruby allows you to keep chaining it and creates more levels lazily, which is what we're doing too.

The second commit gets rid of the `InstanceMethod` and `SingletonMethod` classes. There are only methods and the distinction is made by their owner.

### Automated Tests

Added a few tests and corrected others.

### DIFF
--- a/lib/ruby_indexer/lib/ruby_indexer/declaration_listener.rb
+++ b/lib/ruby_indexer/lib/ruby_indexer/declaration_listener.rb
@@ -33,6 +33,8 @@ module RubyIndexer
         :on_class_node_leave,
         :on_module_node_enter,
         :on_module_node_leave,
+        :on_singleton_class_node_enter,
+        :on_singleton_class_node_leave,
         :on_def_node_enter,
         :on_def_node_leave,
         :on_call_node_enter,
@@ -108,6 +110,21 @@ module RubyIndexer
 
     sig { params(node: Prism::ModuleNode).void }
     def on_module_node_leave(node)
+      @stack.pop
+      @owner_stack.pop
+      @visibility_stack.pop
+    end
+
+    sig { params(node: Prism::SingletonClassNode).void }
+    def on_singleton_class_node_enter(node)
+      @visibility_stack.push(Entry::Visibility::PUBLIC)
+
+      current_owner = @owner_stack.last
+      @owner_stack << current_owner.singleton_klass(node.location, collect_comments(node)) if current_owner
+    end
+
+    sig { params(node: Prism::SingletonClassNode).void }
+    def on_singleton_class_node_leave(node)
       @stack.pop
       @owner_stack.pop
       @visibility_stack.pop
@@ -244,7 +261,7 @@ module RubyIndexer
 
       case node.receiver
       when nil
-        @index << Entry::InstanceMethod.new(
+        @index << Entry::Method.new(
           method_name,
           @file_path,
           node.location,
@@ -254,14 +271,14 @@ module RubyIndexer
           @owner_stack.last,
         )
       when Prism::SelfNode
-        @index << Entry::SingletonMethod.new(
+        @index << Entry::Method.new(
           method_name,
           @file_path,
           node.location,
           comments,
           node.parameters,
           current_visibility,
-          @owner_stack.last,
+          @owner_stack.last&.singleton_klass,
         )
       end
     end
--- a/lib/ruby_indexer/lib/ruby_indexer/entry.rb
+++ b/lib/ruby_indexer/lib/ruby_indexer/entry.rb
@@ -128,6 +128,36 @@ module RubyIndexer
       def ancestor_hash
         mixin_operation_module_names.hash
       end
+
+      sig do
+        params(
+          location: T.nilable(Prism::Location),
+          comments: T.nilable(T::Array[String]),
+        ).returns(SingletonClass)
+      end
+      def singleton_klass(location = nil, comments = nil)
+        # Ruby allows for an unlimited number of singleton classes. Every time `singleton_class` is invoked, a new
+        # singleton class is created.
+        #
+        # Note: we name this method `singleton_klass` to avoid conflicting with Ruby's own `singleton_class` method,
+        # which would cause confusion
+
+        # If we created the singleton class to add methods, but later found a singleton class block, then we can update
+        # the declaration location to be more specific and add extra comments to the documentation of the singleton
+        singleton.update_location(location) if location
+        singleton.add_comments(comments) if comments
+        singleton
+      end
+
+      private
+
+      sig { returns(SingletonClass) }
+      def singleton
+        @singleton_klass ||= T.let(
+          SingletonClass.new(@nesting, @file_path, @location, [], nil),
+          T.nilable(SingletonClass),
+        )
+      end
     end
 
     class Module < Namespace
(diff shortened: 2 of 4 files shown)


lib/ruby_indexer/test/classes_and_modules_test.rb
```diff
@@ -469,5 +469,52 @@ class ConstantPathReferences
...
+      index(<<~RUBY)
+        class Foo
+          # Some extra comments
+          class << bar
+          end
+        end
+      RUBY
+
+      foo = T.must(@index["Foo"].first)
+      singleton = foo.singleton_klass
```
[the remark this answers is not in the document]
vinistock: No, there would be only an entry for `Foo`. We access the singleton by invoking `singleton_klass`.

In my head, the way I'm thinking about it is that we would send the nesting with information about whether we're inside single contexts or not. So we look up the names as we normally do.

lib/ruby_indexer/lib/ruby_indexer/entry.rb
```diff
@@ -128,6 +128,36 @@ def mixin_operations
...
         mixin_operation_module_names.hash
       end
+
+      sig do
+        params(
+          location: T.nilable(Prism::Location),
+          comments: T.nilable(T::Array[String]),
+        ).returns(SingletonClass)
+      end
+      def singleton_klass(location = nil, comments = nil)
```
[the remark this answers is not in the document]
Morriar: Maybe.

Whatever approach you choose, I think this is a quirk that should be documented 👍

lib/ruby_indexer/test/classes_and_modules_test.rb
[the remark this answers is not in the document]
Morriar: Wouldn't this show the behavior?

```rb
    def test_dynamic_singleton_class_blocks
      index(<<~RUBY)
        class Foo; end
        class Foo
          # Some extra comments
          class << bar
          end
        end
      RUBY

      foo_1 = T.must(@index["Foo"][0])
      singleton = foo_1.singleton_klass
      assert_empty(singleton.comments)

      foo_2 = T.must(@index["Foo"][1])
      singleton = foo_2.singleton_klass
      assert_equal("Some extra comments", singleton.comments.join("\n"))
    end

File: lib/ruby_indexer/lib/ruby_indexer/declaration_listener.rb [hunk 3]
```diff
@@ -359,6 +346,28 @@ def on_alias_method_node_enter(node)
...
+      ).void
+    end
+    def handle_instance_variable(node, loc)
+      name = node.name.to_s
+      return if name == "@"
+
+      # When instance variables are declared inside the class body, they turn into class instance variables rather than
+      # regular instance variables
+      owner = @inside_def ? @owner_stack.last : singleton_klass
+      @index << Entry::InstanceVariable.new(name, @file_path, loc, collect_comments(node), owner)
```
st0012, line 368: Just to confirm if my understanding is correct: the only case where `owner` is `nil` is when the ivars are defined at the top-level, which becomes ivars of the `main` object?

lib/ruby_indexer/lib/ruby_indexer/declaration_listener.rb
```diff
@@ -558,5 +567,24 @@ def handle_module_operation(node, operation)
...
+      # Return the existing singleton class if available
+      owner = T.cast(
+        @index["#{attached_class.name}::<Class:#{attached_class.name}>"],
+        T.nilable(T::Array[Entry::SingletonClass]),
+      )
+      return owner.first if owner
+
+      # If not available, create the singleton class lazily
+      nesting = @stack + ["<Class:#{@stack.last}>"]
+      entry = Entry::SingletonClass.new(nesting, @file_path, attached_class.location, [], nil)
```
st0012, line 585: Will there be a case that we need a parent class for `SingletonClass`? If not, maybe we can override its `initialize` method to remove the last parameter?

lib/ruby_indexer/test/classes_and_modules_test.rb
```diff
@@ -470,5 +470,51 @@ class ConstantPathReferences
...
+    def test_dynamic_singleton_class_blocks
+      index(<<~RUBY)
+        class Foo
+          # Some extra comments
+          class << bar
+          end
+        end
+      RUBY
+
+      singleton = T.must(@index["Foo::<Class:Foo>"].first)
```
st0012, line 498: Maybe this case is should be `Foo::<Class::bar>`? Otherwise, wouldn't its methods get mixed with the actual `Foo::<Class:Foo>` class's?

lib/ruby_indexer/lib/ruby_indexer/declaration_listener.rb
vinistock: As Ufuk said, we will need it for the next PR to handle proper linearization of singleton classes.

They always inherit from the singleton class of the parent class of the attached class (lol). This:

```ruby
class Foo < Bar
end

Foo.singleton_class.ancestors
# => [<Class:Foo>, <Class:Bar>, <Class:Object>, <Class:BasicObject>, Class, Module, Object, Kernel, BasicObject]
```

lib/ruby_indexer/lib/ruby_indexer/declaration_listener.rb
```diff
@@ -359,6 +346,28 @@ def on_alias_method_node_enter(node)
...
+        ),
+        loc: Prism::Location,
+      ).void
+    end
+    def handle_instance_variable(node, loc)
+      name = node.name.to_s
+      return if name == "@"
+
+      # When instance variables are declared inside the class body, they turn into class instance variables rather than
+      # regular instance variables
```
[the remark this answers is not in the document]
vinistock: It does! In that case, the instance variable gets defined in the singleton class' singleton class.

lib/ruby_indexer/lib/ruby_indexer/entry.rb
```diff
@@ -162,6 +162,21 @@ def ancestor_hash
...
+      extend T::Sig
+
+      sig { params(location: Prism::Location, comments: T::Array[String]).void }
+      def update_singleton_information(location, comments)
+        @location = Location.new(
+          location.start_line,
+          location.end_line,
+          location.start_column,
+          location.end_column,
+        )
```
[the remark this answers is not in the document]
vinistock: Yes. The `Prism::Location` has more things than `RubyIndexer::Location`. We created the latter to minimize memory usage and only store what we absolutely need.

lib/ruby_indexer/test/classes_and_modules_test.rb
vinistock: They will indeed. We essentially have two options: mix it with the regular singleton class methods or store it using `Class::bar` (but you wouldn't be able to access those in any way).

I don't mind either approach.

paracycle: I think we should do the latter, i.e. we should mark this as `Foo::<Class:bar>`. However, since we won't be able to track and address single instances, maybe we should have a short-hand for all singleton class of object instances, like `<Class:<instance>>` or something.

vinistock: Done!

PR #2145, opened 2024-06-07
Ensure update server gem command updates the locked server

vinistock:
### Motivation

Our current implementation of the update language server gem command only runs `gem update ruby-lsp`. But that's actually not relevant for the version of the server that actually gets booted since that's controlled by the version locked in `.ruby-lsp/Gemfile.lock`.

This PR ensures that both the global installation and the locked versions are updated.

### Implementation

We use the method to install or update the server gem in two different places, so I introduced a parameter to indicate when the user manually requested an update.

In those cases, we now also delete the entire `.ruby-lsp` directory, so that the locked server of the gem is updated.

I also started re-booting the server after the update since it doesn't make sense to request an update, but then continue running the old version.

### DIFF
--- a/vscode/src/workspace.ts
+++ b/vscode/src/workspace.ts
@@ -80,7 +80,7 @@ export class Workspace implements WorkspaceInterface {
     }
 
     try {
-      await this.installOrUpdateServer();
+      await this.installOrUpdateServer(false);
     } catch (error: any) {
       this.error = true;
       await vscode.window.showErrorMessage(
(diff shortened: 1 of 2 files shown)


vscode/src/workspace.ts
```diff
@@ -210,8 +210,24 @@ export class Workspace implements WorkspaceInterface {
       return;
     }
 
-    // If we haven't updated the gem in the last 24 hours, update it
+    // In addition to updating the global installation of the ruby-lsp gem, if the user manually requested an udpate, we
```
KaanOzkan, line 213:
```suggestion
    // In addition to updating the global installation of the ruby-lsp gem, if the user manually requested an update, we
```

PR #2148, opened 2024-06-07
Build complete ancestors list with RBS classes and modules

st0012:
### Motivation

Now that we use `rbs` to provide core classes, we can also index core modules through it, and build the entire ancestors list for Ruby classes (mostly: `Object`, `Kernel`, and `BasicObject`).

### Implementation

1. Start indexing modules through RBS as well
2. Make `Object` the default parent for indexed classes

### Automated Tests

Most tests in `index_test.rb` should take those core classes and modules into account.

### DIFF
--- a/lib/ruby_indexer/lib/ruby_indexer/declaration_listener.rb
+++ b/lib/ruby_indexer/lib/ruby_indexer/declaration_listener.rb
@@ -67,6 +67,8 @@ module RubyIndexer
       parent_class = case superclass
       when Prism::ConstantReadNode, Prism::ConstantPathNode
         superclass.slice
+      else
+        "::Object"
       end
 
       nesting = name.start_with?("::") ? [name.delete_prefix("::")] : @stack + [name.delete_prefix("::")]
--- a/lib/ruby_indexer/lib/ruby_indexer/rbs_indexer.rb
+++ b/lib/ruby_indexer/lib/ruby_indexer/rbs_indexer.rb
@@ -11,7 +11,7 @@ module RubyIndexer
     end
 
     sig { void }
-    def index_core_classes
+    def index_ruby_core
       loader = RBS::EnvironmentLoader.new
       RBS::Environment.from_loader(loader).resolve_type_names
 
@@ -34,6 +34,8 @@ module RubyIndexer
       case declaration
       when RBS::AST::Declarations::Class
         handle_class_declaration(declaration, pathname)
+      when RBS::AST::Declarations::Module
+        handle_module_declaration(declaration, pathname)
       else # rubocop:disable Style/EmptyElse
         # Other kinds not yet handled
       end
@@ -47,7 +49,19 @@ module RubyIndexer
       comments = Array(declaration.comment&.string)
       parent_class = declaration.super_class&.name&.name&.to_s
       class_entry = Entry::Class.new(nesting, file_path, location, comments, parent_class)
+      add_declaration_mixins_to_entry(declaration, class_entry)
       @index << class_entry
+    end
+
+    sig { params(declaration: RBS::AST::Declarations::Module, pathname: Pathname).void }
+    def handle_module_declaration(declaration, pathname)
+      nesting = [declaration.name.name.to_s]
+      file_path = pathname.to_s
+      location = to_ruby_indexer_location(declaration.location)
+      comments = Array(declaration.comment&.string)
+      module_entry = Entry::Module.new(nesting, file_path, location, comments)
+      add_declaration_mixins_to_entry(declaration, module_entry)
+      @index << module_entry
     end
 
     sig { params(rbs_location: RBS::Location).returns(RubyIndexer::Location) }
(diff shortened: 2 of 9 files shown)


lib/ruby_lsp/server.rb
```diff
@@ -690,13 +690,13 @@ def shutdown
...
     def perform_initial_indexing(config_hash)
-      index_core_classes
+      index_core_classes_and_modules
       index_ruby_code(config_hash)
     end
 
     sig { void }
-    def index_core_classes
-      RubyIndexer::RBSIndexer.new(@global_state.index).index_core_classes
+    def index_core_classes_and_modules
```
andyw8, line 698: The RBS definitions also cover constants, e.g. `ARGV`, that we may want to add, so perhaps we should generalize this name to something like `index_core`?

st0012: That's fair. Since we have `index_ruby_code`, I named this `index_ruby_core` to be consistent.

lib/ruby_indexer/lib/ruby_indexer/declaration_listener.rb
```diff
@@ -67,6 +67,8 @@ def on_class_node_enter(node)
       parent_class = case superclass
       when Prism::ConstantReadNode, Prism::ConstantPathNode
         superclass.slice
+      else
+        "Object"
```
vinistock, line 71: I'm not sure we can hard code `Object` in this place exactly because it may produce false positives depending on the nesting. I think this needs to be a part of the constant resolution.

Take this scenario as an example (and let's add a unit test for it):

```ruby
module Foo
  class Object
  end

  # This `Bar` class inherits from `::Object`. But if we hard code it here to
  # `Object`, then it will be resolved to `Foo::Object`, which is a different
  # class
  class Bar
  end
end
```

lib/ruby_indexer/lib/ruby_indexer/rbs_indexer.rb
```diff
@@ -47,6 +49,32 @@ def handle_class_declaration(declaration, pathname)
...
+            Entry::Prepend.new(name)
+          end
+        class_entry.mixin_operations << mixin_operation if mixin_operation
+      end
+
+      @index << class_entry
+    end
+
+    sig { params(declaration: RBS::AST::Declarations::Module, pathname: Pathname).void }
+    def handle_module_declaration(declaration, pathname)
```
vinistock, line 72: We need the exact same code that was added to `handle_class_declaration` here, since modules themselves can `include`, `prepend` or `extend` other modules.

lib/ruby_indexer/lib/ruby_indexer/declaration_listener.rb
st0012: I think changing it to `::Object` actually works because we don't clear`::` for parent classes.

```rb
@index.index_single(IndexablePath.new(nil, "/fake/path/foo.rb"), <<~RUBY)
  module Foo
    class Object; end

    class Bar; end
    class Baz < Object; end
  end
RUBY

assert_equal(["Foo::Bar", "Object", "Kernel", "BasicObject"], @index.linearized_ancestors_of("Foo::Bar"))
assert_equal(
  ["Foo::Baz", "Foo::Object", "Object", "Kernel", "BasicObject"],
  @index.linearized_ancestors_of("Foo::Baz"),
)
```

vinistock: Good idea, but if we follow this approach, there are other things we should optimize too.

We can do it in a follow up PR, but it'd be nice to get it out shortly after.

1. Entry owners should no longer be nilable. Every method/attribute/instance variable must always have an owner because top level declarations are inserted into the `Object` class. You'll want to put `Object` as the initial entry in the owner stack and then get rid of the nilable aspect of owner everywhere
2. [This](https://github.com/Shopify/ruby-lsp/blob/bed467e9cdb842e3af6c7dc82d3ee2c9b99123ae/lib/ruby_indexer/lib/ruby_indexer/index.rb#L378) should be simplified, since every class will always have a parent class (careful with `BasicObject` since it needs to be treated as a special case). The nil checks below should also be removed

lib/ruby_indexer/test/index_test.rb
```diff
@@ -5,6 +5,11 @@
 
 module RubyIndexer
   class IndexTest < TestCase
+    def setup
+      super
+      RBSIndexer.new(@index).index_ruby_core
```
vinistock, line 10: Any reason for this not to go in the parent class?

lib/ruby_indexer/lib/ruby_indexer/rbs_indexer.rb
st0012: I added `add_declaration_mixins_to_entry` for it. But atm we can't write test for it yet as no core modules have mixin themselves.

lib/ruby_indexer/test/index_test.rb
vinistock: Talked to Stan. We can perform these test changes in a separate PR.

andyw8 (approved):

PR #2157, opened 2024-06-11
WIP: Use RBS to index methods

andyw8:
TODO:
- [ ] rebase after https://github.com/Shopify/ruby-lsp/pull/2148 is merged


lib/ruby_indexer/test/index_test.rb
```diff
@@ -92,6 +92,7 @@ class Something
     end
 
     def test_fuzzy_search
+      @index = Index.new
```
andyw8, line 95: This is the same approach as @st0012 took in https://github.com/Shopify/ruby-lsp/pull/2148 but we should really better separate the different indexing behaviours.

lib/ruby_indexer/lib/ruby_indexer/rbs_indexer.rb
```diff
@@ -95,5 +105,38 @@ def add_declaration_mixins_to_entry(declaration, entry)
         entry.mixin_operations << mixin_operation if mixin_operation
       end
     end
+
+    sig { params(member: RBS::AST::Members::MethodDefinition, class_entry: T.any(Entry::Class, Entry::Module)).void }
```
st0012, line 109: - The type should be `Entry::Namespace`
- `class_entry` should be renamed if we're taking module too

lib/ruby_indexer/lib/ruby_indexer/rbs_indexer.rb
```diff
@@ -95,5 +105,38 @@ def add_declaration_mixins_to_entry(declaration, entry)
...
+        Entry::Visibility::PRIVATE
+      when :protected
+        Entry::Visibility::PROTECTED
+      else
+        Entry::Visibility::PUBLIC
+      end
+
+      owner = class_entry
+
+      klass = member.instance? ? Entry::InstanceMethod : Entry::SingletonMethod
```
st0012, line 128: Headsup that these classes will be merged after #2142

lib/ruby_indexer/test/index_test.rb
st0012: I don't think it's a good idea to separate different indexing behaviours because then we'll be deviating from what these features are actually going to be used. It'll also make tests harder to maintain in the long term.
 
Now that we also have method names in the PR, we should use this opportunity to decide a few constant/method names that can better avoid false-positive in such tests.

lib/ruby_indexer/lib/ruby_indexer/rbs_indexer.rb
```diff
@@ -95,5 +105,38 @@ def add_declaration_mixins_to_entry(declaration, entry)
         entry.mixin_operations << mixin_operation if mixin_operation
       end
     end
+
+    sig { params(member: RBS::AST::Members::MethodDefinition, entry: Entry::Namespace).void }
+    def handle_member(member, entry)
```
vinistock, line 110: Is the idea for this method to handle both method and attributes? If not, let's rename it to `handle_method`.

vinistock (approved):
After the parameters follow up, let's ensure we're also handling attributes

st0012 (approved):
Not a blocker as it's probably an edge case anyway: Is it necessary to call `handle_member` in `process_declaration` as well?

andyw8:
@st0012 do you mean in case of methods which are defined outside of any module/class?

andyw8:
The typecheck fails:

```
lib/ruby_indexer/lib/ruby_indexer/rbs_indexer.rb:39: This condition was always falsy (FalseClass)
    39 |      when RBS::AST::Members::MethodDefinition
```

Issue #2096, opened 2024-05-29
Support Ruby on Rails detection based on Rails individual gems


louim:
### I have checked that this feature is not already implemented

- [X] This feature does not exist

### Use case

Currently, `ruby-lsp` detects if a [project uses Rails](https://github.com/Shopify/ruby-lsp/blob/410d58453a20089c131eed81a0d5485e0a6b65a7/lib/ruby_lsp/setup_bundler.rb#L151-L153) and will automatically bundle the `ruby-lsp-rails` gem.

It would be useful if this detection was not only done for the `rails` gem, but also on a combination of sub gems commonly used to run Rails. 

### Description

Having the the `ruby-lsp-rails` automatically bundled when a subset of gems are use to run Rails would be useful.

### Implementation

It is possible to run Rails without requiring the Rails gem directly. Here is [one such example in Discourse](https://github.com/discourse/discourse/blob/96305ec9a6a2b37a2eda001a9fddb67bbf1de751/Gemfile#L9-L16).
The current detection:
https://github.com/Shopify/ruby-lsp/blob/410d58453a20089c131eed81a0d5485e0a6b65a7/lib/ruby_lsp/setup_bundler.rb#L151-L153
Could also check if a combination of gem is included in the gemfile and in this case also bundle `ruby-lsp-rails`. I think checking for:
- `actionpack`
- `activemodel`
- `activerecord`
- `activesupport`
- `railties`

Would be enough to ensure that the app is actually running Rails. The other gems included by the Rails gem are not strictly required to run Rails and will sometimes be absent:

- `actioncable`
- `actionmailbox`
- `actionmailer`
- `actiontext`
- `activejob`
- `activestorage`

rafaelfranca:
Yes

louim:
@andyw8 I'll have a look!

PR #2161, opened 2024-06-11
Improve detection for rails as the testing framework

Earlopain:
### Motivation

In some preparation of #2096

Note that this is only for detection of the testing framework, the issue is about installing `ruby-lsp-rails`. It may be relatively straight-forward to do this as well but I'm more familiar with this part of the code.

So, just do it here to talk roughtly about the implementation direction. I do plan to look properly at the issue after this and that should hopefully only involve moving a few things around. 

Perhaps the `GlobalState` can even be created earlier so that `SetupBundler` has access to this method. Though the location of `config/application.rb` depend on `workspace_uri` which does get modified in `apply_options` .

### Implementation

I created a listener for this which checks for `Rails::Application` inheritance.

This does parse the file twice, once on startup and once indexing actually starts. I don't see an easy way around this since proper indexing only starts once the server received `initialized`.

### Automated Tests

One for the happy path, one for the other.

### Manual Tests

I tested this out on my app which doesn't depend on rails but some subcompontents and that works fine.

### DIFF
--- a/lib/ruby_lsp/global_state.rb
+++ b/lib/ruby_lsp/global_state.rb
@@ -1,6 +1,8 @@
 # typed: strict
 # frozen_string_literal: true
 
+require "ruby_lsp/listeners/rails_app"
+
 module RubyLsp
   class GlobalState
     extend T::Sig
@@ -133,9 +135,9 @@ module RubyLsp
         "rspec"
       # A Rails app may have a dependency on minitest, but we would instead want to use the Rails test runner provided
       # by ruby-lsp-rails. A Rails app doesn't need to depend on the rails gem itself, individual components like
-      # activestorage may be added to the gemfile so that other components aren't downloaded. Check for the presence
-      #  of bin/rails to support these cases.
-      elsif bin_rails_present
+      # activestorage may be added to the gemfile so that other components aren't downloaded. Check for inheritance of
+      # `Rails::Application` in config/application.rb to support these cases.
+      elsif rails_app?
         "rails"
       # NOTE: Intentionally ends with $ to avoid mis-matching minitest-reporters, etc. in a Rails app.
       elsif dependencies.any?(/^minitest$/)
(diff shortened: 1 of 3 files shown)


lib/ruby_lsp/global_state.rb
```diff
@@ -166,8 +168,24 @@ def detect_typechecker(dependencies)
...
+      dispatcher = Prism::Dispatcher.new
+      listener = Listeners::RailsApp.new(dispatcher)
+      dispatcher.dispatch(result.value)
+
+      listener.rails_app
+    end
+
+    sig { returns(T.nilable(String)) }
+    def rails_app_config
+      path = File.join(workspace_path, "config/application.rb")
```
louim, line 187: I'm wondering if it's possible that the file is not right there, but would be nested deeper? I'm not too familiar with the different way people configure their project in relation with the workspace. At the same time, it might be best to stay with the simple solution.

lib/ruby_lsp/global_state.rb
```diff
@@ -166,8 +168,24 @@ def detect_typechecker(dependencies)
...
+      return false unless config
+
+      result = Prism.parse(config)
+      return false unless result
+
+      dispatcher = Prism::Dispatcher.new
+      listener = Listeners::RailsApp.new(dispatcher)
+      dispatcher.dispatch(result.value)
+
+      listener.rails_app
```
louim, line 182: I would probably have done it with a simple regex, but your solution looks much cleaner 👌🏼

louim (commented):
This is nice! It will only require a small addition to implement  https://github.com/Shopify/ruby-lsp/issues/2096. I can do so in a subsequent PR to keep things separate.

lib/ruby_lsp/global_state.rb
andyw8: (and we already require that the Rails app is at the root of a workspace, so that's not a concern).

Earlopain:
Right, if the testing framework detection is going to use some other logic for this, how about I just switch this over to `SetupBundler` to really fix the issue instead?

My main reason for not doing that in the first place was that I wasn't sure how to share the implementation between these two but if that isn't a concern then I can just move it over.

Earlopain:
Continuing in #2218

PR #2171, opened 2024-06-12
Offer RuboCop autocorrection for contextual offenses

### LINKED ISSUE #2168: No RuboCop autocorrect lens if autocorrect is contextual
### Description

This method returns `:unsupported`: https://github.com/rubocop/rubocop/blob/967fe88135188c9d74494cbd4355180c140c4a1a/lib/rubocop/cop/base.rb#L424-L432

In line 427:
* A `corrector` is present
* `always_autocorrect?` is false
* `contextual_autocorrect?` is true
* `LSP.enabled?` is true

So, no autocorrect lense for you.

The following snippet reproduces:

```rb
1.times do
  do_something(t)
end
```

You can only disable this, even though autocorrect is possible:
grafik

I haven't yet thought about how to fix this. Maybe it is enough to somehow disable LSP mode for the `diagnostic_runner` only, though I believe that state is global so maybe not that easy

Earlopain:
### Motivation

Fixes #2168

### Implementation

When `RuboCop::LSP.enable` is called, contextual autocorrect will not offer itself as `correctable?` to prevent annoying changes while typing. Instead check if a corrector is present. If it is, then that means some code transformation can be applied.

I believe this is a reasonable change to make. Since `RuboCop::LSP` contains global state, disabling it for the duration of a run isn't feasable because the other runner used for editor autocorrection can pick this up.

### Manual Tests

See the added fixture and check out the diagnostics for it.

### DIFF
--- a/lib/ruby_lsp/requests/support/rubocop_diagnostic.rb
+++ b/lib/ruby_lsp/requests/support/rubocop_diagnostic.rb
@@ -42,7 +42,7 @@ module RubyLsp
         def to_lsp_code_actions
           code_actions = []
 
-          code_actions << autocorrect_action if @offense.correctable?
+          code_actions << autocorrect_action if correctable?
           code_actions << disable_line_action
 
           code_actions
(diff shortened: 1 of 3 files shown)


vinistock (approved):
Thank you for looking into this! It's great that now these contextual cases are not corrected during formatting, but offered in quickfixes 🚀

Issue #2170, opened 2024-06-12
Automatic Ruby environment activation with mise failed: Unexpected token 'm', "mise missi"... is not valid JSON


Peredery:
### Description

Doesn't work with mise

### Reproduction steps

1. Mise installed
2. Select mise as version manager for ruby
3. See error

**Code snippet or error message**
```sh
Automatic Ruby environment activation with mise failed: Unexpected token 'm', "mise missi"... is not valid JSON
```

andyw8:
Thanks, I've opened a PR to add a note about that.

https://github.com/Shopify/ruby-lsp/pull/2172

louim:
Are you sure it was the update to `mise` version that fixed it? Usually it's related to mise not finding the current ruby version defined by the app and printing `mise missing: ruby@version`.

I can reproduce the same problem on the latest version if I put a bogus version of ruby in my `.ruby-version`. eg:
<img width="1624" alt="Screenshot 2024-06-12 at 10 26 42 AM" src="https://github.com/Shopify/ruby-lsp/assets/923718/64dde95c-1b62-4e79-8478-8236519261f0">

The solution would be to `mise i` to try to install the correct version but I feel like this would be reaching out too far.

vinistock:
It would be a great experience to detect that the selected Ruby is missing and offer to install it.

We can't have that for every version manager (e.g.: `chruby` allows you to install Ruby through any means), but it would work for a few of them (asdf, mise, rvm, rbenv).

I created an issue to look into that https://github.com/Shopify/ruby-lsp/issues/2173.

PR #2172, opened 2024-06-12
Add note about Mise version manager

andyw8:
See https://github.com/Shopify/ruby-lsp/issues/2170#issuecomment-2162983705


andyw8:
I've updated it to point to the Mise docs instead. I think the locality of the advice is important - people tend to skim docs, so having this next to the Mise topics makes it more discoverable, rather than as a general statement further up.

vinistock (approved):

PR #2174, opened 2024-06-12
Add code lens support for tldr gem

ianks:
This PR adds basic support for the [tldr](https://github.com/tendersearls/tldr) gem. The only real difference from minitest being we use `bundle exec tldr` instead of the `ruby -Itest` stuff. Happy to make any changes deemed neccesary!


andyw8:
As discussed with @ianks on Slack, this would be better suited as an addon.

Issue #1046, opened 2023-09-28
Add type hierarchy support


vinistock:
Depends on #1333

After the indexer starts keeping track of class and module ancestors, it should be straightforward to add support for the type hierarchy requests.

Documentation: https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#textDocument_prepareTypeHierarchy

PR #2175, opened 2024-06-12
Fix incorrect parent class assignment for Object and BasicObject

st0012:
### Motivation

When the Object and BasicObject classes are reopened, their parent class should remain `BasicObject` and `nil` respectively. This change fixes the current behaviour where the parent class is incorrectly set to `Object`.

The issue was discovered when pairing with @Morriar on #1046 as `Object` are displayed in `Object`'s supertypes list too.

### Implementation

We should check these 2 special cases before deciding the parent class.

### Automated Tests

I added 4 new test cases for it.

### DIFF
--- a/lib/ruby_indexer/lib/ruby_indexer/entry.rb
+++ b/lib/ruby_indexer/lib/ruby_indexer/entry.rb
@@ -152,8 +152,7 @@ module RubyIndexer
       end
       def initialize(nesting, file_path, location, comments, parent_class)
         super(nesting, file_path, location, comments)
-
-        @parent_class = T.let(parent_class, T.nilable(String))
+        @parent_class = parent_class
       end
 
       sig { override.returns(Integer) }
--- a/lib/ruby_indexer/test/index_test.rb
+++ b/lib/ruby_indexer/test/index_test.rb
@@ -1143,8 +1143,32 @@ module RubyIndexer
       assert_nil(@index.resolve("RSpec", []))
     end
 
+    def test_object_superclass_indexing_and_resolution_with_reopened_object_class
+      index(<<~RUBY)
+        class Object; end
+      RUBY
+
+      entries = @index["Object"]
+      assert_equal(2, entries.length)
+      reopened_entry = entries.last
+      assert_equal("::BasicObject", reopened_entry.parent_class)
+      assert_equal(["Object", "Kernel", "BasicObject"], @index.linearized_ancestors_of("Object"))
+    end
+
+    def test_object_superclass_indexing_and_resolution_with_reopened_basic_object_class
+      index(<<~RUBY)
+        class BasicObject; end
+      RUBY
+
+      entries = @index["BasicObject"]
+      assert_equal(2, entries.length)
+      reopened_entry = entries.last
+      assert_nil(reopened_entry.parent_class)
+      assert_equal(["BasicObject"], @index.linearized_ancestors_of("BasicObject"))
+    end
+
     def test_object_superclass_resolution
-      @index.index_single(IndexablePath.new(nil, "/fake/path/foo.rb"), <<~RUBY)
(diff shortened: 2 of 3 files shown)


lib/ruby_indexer/lib/ruby_indexer/entry.rb
```diff
@@ -136,6 +136,9 @@ class Module < Namespace
     class Class < Namespace
       extend T::Sig
 
+      OBJECT_NESTING = T.let(["Object"].freeze, T::Array[String])
```
st0012, line 139: TIL: using arrays in case statement will allocate a new array each time the case statement is hit.

Script:

```rb
a = ["Object"]
b = ["Object"]
c = []

CONST = ["Object"]

def match_with_const(x)
  case x
  when CONST
    puts "Matched"
  end
end

def match_with_array(x)
  case x
  when ["Object"]
    puts "Matched"
  end
end

puts "Arrays count before match_with_const: #{ObjectSpace.count_objects[:T_ARRAY]}"
match_with_const a
match_with_const b
match_with_const c
puts "Arrays count after match_with_const: #{ObjectSpace.count_objects[:T_ARRAY]}"

puts "Arrays count before match_with_const: #{ObjectSpace.count_objects[:T_ARRAY]}"
match_with_array a
match_with_array b
match_with_array c
puts "Arrays count after match_with_const: #{ObjectSpace.count_objects[:T_ARRAY]}"
```

Output:

```
Arrays count before match_with_const: 1804
Matched
Matched
Arrays count after match_with_const: 1804
Arrays count before match_with_const: 1804
Matched
Matched
Arrays count after match_with_const: 1807
```

lib/ruby_indexer/lib/ruby_indexer/entry.rb
```diff
@@ -153,7 +156,20 @@ class Class < Namespace
       def initialize(nesting, file_path, location, comments, parent_class)
         super(nesting, file_path, location, comments)
 
-        @parent_class = T.let(parent_class, T.nilable(String))
+        parent_class ||=
+          case nesting
+          when OBJECT_NESTING
+            # When Object is reopened, its parent class should still be the top-level BasicObject
+            "::BasicObject"
+          when BAISC_OBJECT_NESTING
```
andyw8, line 164:
```suggestion
          when BASIC_OBJECT_NESTING
```

st0012: Good catch, updated 👍

lib/ruby_indexer/lib/ruby_indexer/entry.rb
```diff
@@ -153,7 +156,20 @@ class Class < Namespace
       def initialize(nesting, file_path, location, comments, parent_class)
         super(nesting, file_path, location, comments)
 
-        @parent_class = T.let(parent_class, T.nilable(String))
+        parent_class ||=
```
andyw8, line 159: Wouldn't this work?

`@parent_class = parent_class || case...`

lib/ruby_indexer/test/index_test.rb
```diff
@@ -1145,8 +1145,32 @@ def test_resolving_unindexed_constant_with_no_nesting
...
     end
 
+    def test_object_superclass_indexing_and_resolution_with_reopened_object_class
+      index(<<~RUBY)
+        class Object; end
+      RUBY
+
+      entries = @index["Object"]
+      assert_equal(2, entries.length)
+      reopen_entry = entries.last
```
andyw8, line 1155: reopened_entry

andyw8, line 1155:
```suggestion
      reopened_entry = entries.last
```

andyw8 (approved):

vinistock (approved):

PR #2179, opened 2024-06-12
Show method parameters on hover

vinistock:
### Motivation

I was pretty sure we had this at some point, but I noticed that we were not showing method parameters on hover, which we definitely want to show.

### Implementation

1. Changed the decorated names for optional positional and keyword arguments to show a fake `DEFAULT`. We don't store the actual default values to save memory, but I think that at least showing it's a default is worth it
2. I also noticed that we were inserting the rest argument in an unexpected order. Rest arguments always come before keywords
3. Started showing the parameter decorated names in hover

### Automated Tests

Added a test.


Earlopain:
I don't really like how the presence of a default value is presented, it makes it look like thats supposed to be a constant. But I also don't have any better alternatives to offer.

PR #2180, opened 2024-06-13
Display addons status in the control panel

st0012:
### Motivation

This will give users a quick overview of the addons that are enabled in their workspace.

We plan to expand this feature to display more information about addons in the future, such as:

- The version of the addon
- The gem name of the addon
- Addons that weren't activated due to errors

But some of these features will require changes to the addon API, so we will plan them for a future release.

### Implementation

The feature is implemented on both the server and the extension:

- Server:
  - The server now has an experimental field in the capabilities object, which currently only has the `addon_detection` field.
  - The server now supports a custom `rubyLsp/workspace/addons` request that returns the list of addons that are enabled in the workspace. At this iteration, each addon only has name and errored attributes.
- Extension:
  - In the client.afterStart callback, the extension now sends a `rubyLsp/workspace/addons` request to the server to fetch and store the list of addons.
  - A new `AddonsStatus` status item is added to display addon's status.
    - If the server doesn't have the capability, the status will mention that server 0.17.4 or later is required.
      Screenshot 2024-06-13 at 13 42 47
    - If the server supports the capability but the workspace has no addons, the status will mention that no addons are enabled.
      Screenshot 2024-06-13 at 13 29 26
    - If the workspace has addons, the status will display the names of the addons that are enabled. If the addon failed to activate, `(errored)` will be added after its name. 
      Screenshot 2024-06-13 at 13 29 55

I plan to explore better UI in later iterations as simply having the names listed will already make addon development easier.

### Automated Tests

Added some

### DIFF
--- a/lib/ruby_lsp/server.rb
+++ b/lib/ruby_lsp/server.rb
@@ -76,6 +76,16 @@ module RubyLsp
         text_document_show_syntax_tree(message)
       when "rubyLsp/workspace/dependencies"
         workspace_dependencies(message)
+      when "rubyLsp/workspace/addons"
+        send_message(
+          Result.new(
+            id: message[:id],
+            response:
+              Addon.addons.map do |addon|
+                { name: addon.name, errored: addon.error? }
+              end,
+          ),
+        )
       when "$/cancelRequest"
         @mutex.synchronize { @cancelled_requests << message[:params][:id] }
       end
@@ -104,7 +114,7 @@ module RubyLsp
           ),
         )
 
-        $stderr.puts(errored_addons.map(&:errors_details).join("\n\n"))
+        $stderr.puts(errored_addons.map(&:errors_details).join("\n\n")) unless @test_mode
       end
     end
 
@@ -177,6 +187,9 @@ module RubyLsp
           definition_provider: enabled_features["definition"],
           workspace_symbol_provider: enabled_features["workspaceSymbol"] && !@global_state.typechecker,
           signature_help_provider: signature_help_provider,
+          experimental: {
+            addon_detection: true,
+          },
         ),
         serverInfo: {
           name: "Ruby LSP",
--- a/vscode/src/status.ts
+++ b/vscode/src/status.ts
@@ -178,6 +178,32 @@ export class FormatterStatus extends StatusItem {
   }
 }
 
+export class AddonsStatus extends StatusItem {
+  constructor() {
+    super("addons");
+
+    this.item.name = "Ruby LSP Addons";
+    this.item.text = "Fetching addon information";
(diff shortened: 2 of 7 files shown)


lib/ruby_lsp/server.rb
```diff
@@ -76,6 +76,16 @@ def process_message(message)
         text_document_show_syntax_tree(message)
       when "rubyLsp/workspace/dependencies"
         workspace_dependencies(message)
+      when "rubyLsp/workspace/getAddons"
+        send_message(
```
vinistock, line 80: Let's add a test for this.

vscode/src/status.ts
```diff
@@ -178,6 +178,31 @@ export class FormatterStatus extends StatusItem {
...
+export class AddonsStatus extends StatusItem {
+  constructor() {
+    super("addons");
+
+    this.item.name = "Ruby LSP Addons";
+    this.item.text = "Fetching addon information";
+  }
+
+  refresh(workspace: WorkspaceInterface): void {
+    if (workspace.lspClient) {
```
vinistock, line 190: Can we put this one with an early return so that it's easier to read?

```typescript
if (!workspace.lspClient) {
  return
}
```

vscode/src/status.ts
```diff
@@ -178,6 +178,31 @@ export class FormatterStatus extends StatusItem {
...
+      if (workspace.lspClient.addons === undefined) {
+        this.item.text =
+          "Addons: requires server to be v0.17.4 or higher to display this field";
+      } else if (workspace.lspClient.addons.length === 0) {
+        this.item.text = "Addons: none";
+      } else {
+        const addonNames = workspace.lspClient.addons.map((addon) =>
+          addon.errored ? `${addon.name} (errored)` : `${addon.name}`,
+        );
+        this.item.text = `Addons: ${addonNames.join(", ")}`;
```
vinistock, line 200: Instead of showing all names concatenated by commas, could we started with a quick pick that does nothing instead? The language status items will all expand to fit the widest entry and two addon names is already quite long `Ruby LSP Rails, Ruby LSP RSpec`.

It will also allow you to show the information about whether it errored or not in a [quick pick detail](https://code.visualstudio.com/api/references/vscode-api#QuickPickOptions), which will look nicer.

vinistock: That's okay.

lib/ruby_lsp/server.rb
st0012: I added a test and renamed it to `rubyLsp/workspace/addons` instead, which is consistent with `dependencies`.

vinistock (approved):

PR #2185, opened 2024-06-14
Allow `Entry::Method` to be initialized with `Entry::Parameter`s

andyw8:
For indexing methods using RBS, we need to refactor `Entry::Method` so that we can pass an array of `Entry::Parameter`, rather than a `Prism::ArgumentsNode`. The moved code in `list_params`, is a direct cut and paste, the only changes are to add the necessary namespace prefixes.


vinistock (approved):
Nice. I like that having different indexing mechanisms is already surfacing the coupling between the index and the indexers. We should ensure that the index is fully agnostic to the mechanism inserting entries into it.

Issue #1938, opened 2024-04-15
Support singleton method indexing and resolution


st0012:
Example:

```rb
class Foo
  BAR = 10
  class << self
    def bar; end

    def baz
      b # suggest bar
    end
  end
end

Foo.b # suggest bar and baz
```

PR #2186, opened 2024-06-14
Remember singleton nesting and method nesting when locating nodes

vinistock:
### Motivation

Step towards #1938 

In order to properly resolve singleton contexts, we need to remember that we are inside of one when we locate nodes.

This PR includes both the singleton nesting and the surrounding method declaration in the node context. This information will allow us to properly resolve constants, methods and ivars inside singleton contexts. In addition to that, it also enables us to provide go to definition for `super` calls.

### Implementation

Started pushing more nodes to our nesting tracker and stored everything in the node context.

### Automated Tests

Added a test demonstrating the expected information on multiple situations.

### DIFF
--- a/lib/ruby_lsp/document.rb
+++ b/lib/ruby_lsp/document.rb
@@ -127,7 +127,10 @@ module RubyLsp
       queue = T.let(node.child_nodes.compact, T::Array[T.nilable(Prism::Node)])
       closest = node
       parent = T.let(nil, T.nilable(Prism::Node))
-      nesting = T.let([], T::Array[T.any(Prism::ClassNode, Prism::ModuleNode)])
+      nesting_nodes = T.let(
+        [],
+        T::Array[T.any(Prism::ClassNode, Prism::ModuleNode, Prism::SingletonClassNode, Prism::DefNode)],
+      )
       call_node = T.let(nil, T.nilable(Prism::CallNode))
 
       until queue.empty?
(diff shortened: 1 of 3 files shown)


lib/ruby_lsp/document.rb
```diff
@@ -190,10 +198,31 @@ def locate(node, char_position, node_types: [])
...
+        when Prism::DefNode
+          surrounding_method = node.name.to_s
+          next unless node.receiver.is_a?(Prism::SelfNode)
+
+          complete_nesting << "<Class:#{complete_nesting.last}>"
+        end
       end
 
-      NodeContext.new(closest, parent, nesting.map { |n| n.constant_path.location.slice }, call_node)
+      NodeContext.new(closest, parent, complete_nesting, call_node, surrounding_method)
```
st0012, line 225: Can we rename `nesting` to `nesting_nodes`, and `complete_nesting` can simply be `nesting`.

lib/ruby_lsp/document.rb
```diff
@@ -190,10 +198,31 @@ def locate(node, char_position, node_types: [])
...
+
+      nesting.each do |node|
+        case node
+        when Prism::ClassNode, Prism::ModuleNode
+          complete_nesting << node.constant_path.slice
+        when Prism::SingletonClassNode
+          complete_nesting << "<Class:#{complete_nesting.last}>"
+        when Prism::DefNode
+          surrounding_method = node.name.to_s
+          next unless node.receiver.is_a?(Prism::SelfNode)
```
st0012, line 219: nit: this `next` can be called first.

vinistock: It can't because we want to remember the surrounding instance methods as well. That will be necessary to provide hover, signature help and go to definition on `super` calls.

```ruby
class Foo < Bar
  def something
    # When we find this `super` call, we need to know that we're inside
    # `something` because that's what we will be searching for in the ancestors
    # of `Foo`
    super
  end
end
```

st0012: Oh sorry I misread the scope of the local.

st0012 (approved):

PR #2187, opened 2024-06-14
Add receiver type handling

vinistock:
### Motivation

Step towards #1938 

We are getting to a point where we need a better abstraction to encapsulate the logic of figuring out the correct type to use for resolution depending on the current lexical scope and whatever the target is.

We are going to need this for proper instance variable and method resolution (e.g.: is this an instance variable or a class instance variable? Should I search for this method in the attached or singleton class?).

Because of this I'm proposing the addition of this "type checker" class. Of course, it's not a full fledged type checker, but I think the name is still fitting even if the functionality is partial. And this can house more inference work in the future.

### Implementation

1. Created a `TypeChecker` class that is instantiated with the index. The idea is for only one instance of it to exist in the language server
2. Added a method to determine the receiver type for method calls and instance variables, depending on the current lexical scope
3. Started creating an instance of the type checker in the global state. It's currently not used in this PR, but I will use it to provide proper support for go to definition, signature help, hover and completion in singleton contexts

I also renamed `GlobalState#typechecker` to `has_type_checker` to avoid confusion.

### Automated Tests

Added tests for every scenario covered by the receiver type method.


lib/ruby_lsp/type_checker.rb
```diff
@@ -0,0 +1,79 @@
+# typed: strict
+# frozen_string_literal: true
+
+module RubyLsp
+  # A minimalistic type checker to try to resolve types that can be inferred without requiring a type system or
+  # annotations
+  class TypeChecker
```
Morriar, line 7:
```suggestion
  class TypeInferencer
```

lib/ruby_lsp/type_checker.rb
```diff
@@ -0,0 +1,79 @@
...
+    sig { params(node_context: NodeContext).returns(T.nilable(String)) }
+    def infer_receiver_type(node_context)
+      node = node_context.node
+
+      case node
+      when Prism::CallNode
+        infer_receiver_for_call_node(node, node_context)
+      when Prism::InstanceVariableReadNode, Prism::InstanceVariableAndWriteNode, Prism::InstanceVariableWriteNode,
+        Prism::InstanceVariableOperatorWriteNode, Prism::InstanceVariableOrWriteNode, Prism::InstanceVariableTargetNode
+
```
Morriar, line 24:
```suggestion
```

lib/ruby_lsp/type_checker.rb
```diff
@@ -0,0 +1,79 @@
...
+  # annotations
+  class TypeChecker
+    extend T::Sig
+
+    sig { params(index: RubyIndexer::Index).void }
+    def initialize(index)
+      @index = index
+    end
+
+    sig { params(node_context: NodeContext).returns(T.nilable(String)) }
```
Morriar, line 15: Would it make sense to return the resolved entry rather than a string here?

test/type_checker_test.rb
```diff
@@ -0,0 +1,171 @@
...
+      node_context = index_and_locate({ line: 2, character: 4 }, <<~RUBY)
+        class Foo
+          def bar
+            @hello1
+          end
+        end
+      RUBY
+
+      assert_equal("Foo", @type_checker.infer_receiver_type(node_context))
+    end
```
Morriar, line 157: Should we add tests for top level definitions such as:

```rb
foo
```

and

```rb
def foo
  bar
end
```

lib/ruby_lsp/type_checker.rb
```diff
@@ -0,0 +1,79 @@
...
+        resolved_receiver = @index.resolve(receiver_name, node_context.nesting)
+        name = resolved_receiver&.first&.name
+        return unless name
+
+        *parts, last = name.split("::")
+        return "#{last}::<Class:#{last}>" if T.must(parts).empty?
+
+        "#{T.must(parts).join("::")}::#{last}::<Class:#{last}>"
+      end
+    end
```
Morriar, line 62: We may need to look past the surrounding method to discard cases such has this:

```rb
class Foo
  def foo
    Class.new do
      bar
    end
  end
end
```

which I think would wrongly resolve to `Foo` at the moment

lib/ruby_lsp/type_checker.rb
vinistock: I switched to `TypeInferrer` 👍

lib/ruby_lsp/type_inferrer.rb
```diff
@@ -0,0 +1,85 @@
...
+      when Prism::InstanceVariableReadNode, Prism::InstanceVariableAndWriteNode, Prism::InstanceVariableWriteNode,
+        Prism::InstanceVariableOperatorWriteNode, Prism::InstanceVariableOrWriteNode, Prism::InstanceVariableTargetNode
+        nesting = node_context.nesting
+        # If we're at the top level, then the invocation is happening on `<main>`, which is a special singleton that
+        # inherits from Object
+        return "Object" if nesting.empty?
+        return node_context.fully_qualified_name if node_context.surrounding_method
+
+        nesting = node_context.nesting
+        "#{nesting.join("::")}::<Class:#{nesting.last}>"
```
andyw8, line 31:
```suggestion
        "#{nesting.fully_qualified_name}::<Class:#{nesting.last}>"
```

Morriar (approved):

PR #2188, opened 2024-06-14
Provide features on singletons

vinistock:
### Motivation

Second to last step for #1938. The only part missing is proper linearization of singleton ancestors, which I'm working on.

This PR uses the type checker introduced in #2187 to provide go to definition, hover, completion and signature help inside singleton contexts.

**Note**: I discovered a bug in how we're indexing class instance variables that are defined inside a `def self.foo` method. The completion expectation in the test is wrong intentionally, I will fix that in a different PR.

### Implementation

I split each feature by commit:

1. Definition
2. Hover
3. Completion
4. Signature help

The only notable part of the implementation is completion. When invoking methods on a singleton, we know the receiver type, which means we can show all available methods immediately when the user types the dot.

To support that, we needed a few extra changes:

1. Started considering dot a trigger character
2. When we type in the dot, Prism will consider whatever is the next node to be name of the incomplete method call. We need to know that completion is being triggered by a dot character so that we can search for all available methods instead of trying to extract the name from the AST
3. Finally, when providing completion with the dot, the range that we're substituting in the text edit is slightly different, which we need to ensure is correct or else the editor rejects all completion candidates

### Automated Tests

Added tests for all the features.

### Manual Tests

Launch the LSP on this branch and start testing features for things invoked directly on singletons (remember that ancestors are still missing for now!).


lib/ruby_lsp/requests/completion_resolve.rb
```diff
@@ -56,8 +56,16 @@ def perform
           end
         end
 
+        first_entry = T.must(entries.first)
+
+        if first_entry.is_a?(RubyIndexer::Entry::Member)
+          detail = "(#{first_entry.parameters.map(&:decorated_name).join(", ")})"
```
andyw8, line 62: Worth extracting `.parameters.map(&:decorated_name).join(", ")` be a method on `Entry`?

vinistock: Definitely! We used the same code in other places.

lib/ruby_lsp/listeners/completion.rb
```diff
@@ -15,15 +15,26 @@ class Completion
...
+        dispatcher,
+        uri,
+        trigger_character
+      )
         @response_builder = response_builder
         @global_state = global_state
         @index = T.let(global_state.index, RubyIndexer::Index)
+        @type_inferrer = T.let(global_state.type_inferrer, TypeInferrer)
         @node_context = node_context
         @typechecker_enabled = typechecker_enabled
```
andyw8, line 35: Not for this Pr, but I wonder if we can avoid the need to pass `typechecker_enabled` by instead determining this from `type_inferrer`, i.e. if `type_inferrer` is nil that means a typechecker is in use?

st0012 (approved):

lib/ruby_lsp/listeners/completion.rb
vinistock: For now, that would work. However, long term we would like to integrate more deeply with Sorbet and do something like:

1. First ask Sorbet if it can find the most accurate definition
2. If a definition was found, use that and do nothing else
3. If not found, fallback to the Ruby LSP behaviour (we'd need the inferrer here)

PR #2199, opened 2024-06-17
Fix RBS singleton method owner assignment

vinistock:
### Motivation

In #2177, we prevented the build from failing, but we didn't have a test for singleton methods which made us miss the fact that they're all being associated to the attached class/module.

This PR ensures that we assign the right owner for singleton methods.

### Implementation

If RBS marked the method as `singleton?`, then we need to put it in the singleton class of the current owner.

### Automated Tests

Added a test that prevents regression.

### DIFF
--- a/lib/ruby_indexer/lib/ruby_indexer/rbs_indexer.rb
+++ b/lib/ruby_indexer/lib/ruby_indexer/rbs_indexer.rb
@@ -122,7 +122,26 @@ module RubyIndexer
         Entry::Visibility::PUBLIC
       end
 
-      @index.add(Entry::Method.new(name, file_path, location, comments, [], visibility, owner))
+      real_owner = member.singleton? ? existing_or_new_singleton_klass(owner) : owner
+      @index.add(Entry::Method.new(name, file_path, location, comments, [], visibility, real_owner))
+    end
+
+    sig { params(owner: Entry::Namespace).returns(T.nilable(Entry::Class)) }
+    def existing_or_new_singleton_klass(owner)
+      *_parts, name = owner.name.split("::")
+
+      # Return the existing singleton class if available
+      singleton_entries = T.cast(
+        @index["#{owner.name}::<Class:#{name}>"],
+        T.nilable(T::Array[Entry::SingletonClass]),
+      )
+      return singleton_entries.first if singleton_entries
+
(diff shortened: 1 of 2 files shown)


lib/ruby_indexer/lib/ruby_indexer/rbs_indexer.rb
```diff
@@ -122,7 +122,26 @@ def handle_method(member, owner)
...
+      singleton_entries = T.cast(
+        @index["#{owner.name}::<Class:#{name}>"],
+        T.nilable(T::Array[Entry::SingletonClass]),
+      )
+      return singleton_entries.first if singleton_entries
+
+      # If not available, create the singleton class lazily
+      nesting = owner.nesting + ["<Class:#{name}>"]
+      entry = Entry::SingletonClass.new(nesting, owner.file_path, owner.location, [], nil)
+      @index.add(entry, skip_prefix_tree: true)
```
andyw8, line 143: Can we do the indexing in `handle_method` instead? Otherwise this side-effect is a bit unexpected.

lib/ruby_indexer/lib/ruby_indexer/rbs_indexer.rb
```diff
@@ -122,7 +122,26 @@ def handle_method(member, owner)
...
+      # Return the existing singleton class if available
+      singleton_entries = T.cast(
+        @index["#{owner.name}::<Class:#{name}>"],
+        T.nilable(T::Array[Entry::SingletonClass]),
+      )
+      return singleton_entries.first if singleton_entries
+
+      # If not available, create the singleton class lazily
+      nesting = owner.nesting + ["<Class:#{name}>"]
+      entry = Entry::SingletonClass.new(nesting, owner.file_path, owner.location, [], nil)
```
andyw8, line 142: Perhaps we could introduce a factory method `Entry::Singleton.from_owner(owner)`?

vinistock: Can we re-evaluate after the work for ancestors? I think that'll impact how we're creating those and this might go away.

andyw8: Sure

PR #2200, opened 2024-06-17
WIP: Index method parameters using RBS

andyw8:


andyw8:
Continuing in https://github.com/Shopify/ruby-lsp/pull/2243

PR #2208, opened 2024-06-18
Search aliased namespaces in the top level too

vinistock:
### Motivation

We were treating the top level differently, but in reality that's a mistake. We can have aliased namespaces in the top level too (and we use that in the Ruby LSP itself).

### Implementation

We don't need special treatment for the top level, we just need to invoke the `direct_or_aliased_constant` with the correct name at the end of resolution.

### Automated Tests

Added a test that reproduces the bug.

### DIFF
--- a/lib/ruby_indexer/test/index_test.rb
+++ b/lib/ruby_indexer/test/index_test.rb
@@ -1007,6 +1007,43 @@ module RubyIndexer
       assert_instance_of(Entry::Alias, baz_entry)
     end
 
+    def test_resolving_constants_in_aliased_namespace
+      index(<<~RUBY)
+        module Original
+          module Something
+            CONST = 123
+          end
+        end
+
+        module Other
+          ALIAS = Original::Something
+        end
+
+        module Third
+          Other::ALIAS::CONST
+        end
+      RUBY
+
+      entry = T.must(@index.resolve("Other::ALIAS::CONST", ["Third"])&.first)
+      assert_kind_of(Entry::Constant, entry)
+      assert_equal("Original::Something::CONST", entry.name)
+    end
+
+    def test_resolving_top_level_aliases
+      index(<<~RUBY)
+        class Foo
+          CONST = 123
+        end
+
+        FOO = Foo
+        FOO::CONST
+      RUBY
+
+      entry = T.must(@index.resolve("FOO::CONST", [])&.first)
+      assert_kind_of(Entry::Constant, entry)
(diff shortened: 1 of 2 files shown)


lib/ruby_indexer/test/index_test.rb
```diff
@@ -1472,5 +1472,26 @@ class Foo
...
+          ALIAS = Original::Something
+        end
+
+        module Third
+          Other::ALIAS::CONST
+        end
+      RUBY
+
+      entry = T.must(@index.resolve("Other::ALIAS::CONST", ["Third"])&.first)
+      refute_nil(entry)
```
Morriar, line 1494:
```suggestion
      refute_empty(@index.resolve("Other::ALIAS::CONST", ["Third"]))
```

lib/ruby_indexer/test/index_test.rb
```diff
@@ -1472,5 +1472,26 @@ class Foo
 
       assert_empty(@index.method_completion_candidates("bar", "Foo"))
     end
+
+    def test_resolving_constants_in_aliased_namespace
```
Morriar, line 1476: Let's move this closer from the other tests about constant aliasing?

lib/ruby_indexer/test/index_test.rb
```diff
@@ -1472,5 +1472,26 @@ class Foo
...
+        module Other
+          ALIAS = Original::Something
+        end
+
+        module Third
+          Other::ALIAS::CONST
+        end
+      RUBY
+
+      entry = T.must(@index.resolve("Other::ALIAS::CONST", ["Third"])&.first)
```
Morriar, line 1493: Let's also add a test for top levels constant aliases, I don't think we have one already:

```rb
class Foo
  CONST = 123
end

FOO = Foo
FOO::CONST
```

lib/ruby_indexer/test/index_test.rb
vinistock: When we fail to resolve a constant, we return `nil`, not empty array. This would fail with `undefined method empty? for nil:NilClass`.

Morriar: So I guess you can `assert_nil(@index.resolve("Other::ALIAS::CONST", ["Third"]))`

vinistock: Actually, this test was really weak. I changed it to assert the kind of entry we get back and that it points to the correct name.

PR #2212, opened 2024-06-20
Fix chruby release candidate ordering

### LINKED ISSUE #2169: chruby activation failure for release candidate versions of Ruby
When a repo contains `.ruby-version` file with a Ruby version with a release candidate, e.g. `3.3.0-rc1`, activation fails:

`2024-06-12 08:27:33.756 [info] (once-campfire) Your Ruby version is 3.3.0, but your Gemfile specified 3.3.0.rc1`

This affects chruby but possibly other version managers too.

Shopify folk can see this on the `once-campfire` repo.

vinistock:
### Motivation

Closes #2169

We were actually testing against the wrong behaviour for chruby. Initially, we considered that anything with a `-something` suffix would be considered as older than its stable version. For example: 3.3.0 (stable) would be newer than 3.3.0-rc1 (release candidate).

However, this is not how chruby works. If you have both stable and release candidate, running `chruby 3.3.0` will select the release candidate instead of the stable version. This behaviour is surprising to me, but we should match exactly what chruby does.

### Implementation

This actually simplifies the implementation quite a bit because we initially considered versions including the patch to have a different logic than the ones omitting the patch.

With this approach, we can just always sort the directories and pick whatever is sorted first.

### Automated Tests

Added tests verifying the behaviour.

### DIFF
--- a/vscode/src/ruby/chruby.ts
+++ b/vscode/src/ruby/chruby.ts
@@ -75,16 +75,6 @@ export class Chruby extends VersionManager {
 
   // Returns the full URI to the Ruby executable
   protected async findRubyUri(rubyVersion: RubyVersion): Promise<vscode.Uri> {
-    if (/\d+\.\d+\.\d+/.exec(rubyVersion.version)) {
-      return this.findRubyUriForCompleteVersion(rubyVersion);
-    }
-
-    return this.findRubyUriWithOmittedPatch(rubyVersion);
-  }
-
-  private async findRubyUriWithOmittedPatch(
-    rubyVersion: RubyVersion,
-  ): Promise<vscode.Uri> {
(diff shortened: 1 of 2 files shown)


andyw8:
Should we log this is a bug with chruby?

vinistock:
> Should we log this as a bug with chruby?

We can try, but people could also argue that if you add extra patches to a Ruby version, then it is newer.

For example, if someone builds a custom version of Ruby with 3 extra commits and calls it `3.3.0-custom3`, then in that case it would be newer than 3.3.0.

I suspect they made a compromise here. Since you cannot be sure based on the naming of the Ruby alone, they decided to always just pick whatever gets sorted first.

PR #2213, opened 2024-06-20
Document manager activation for other editors

vinistock:
### Motivation

Document how to configure the executable command for other editors when using version managers. This has been a source of confusion for users of other editors, so we should try to help people configure launching the server the right way.

### DIFF
--- a/EDITORS.md
+++ b/EDITORS.md
@@ -10,6 +10,36 @@ Code, use the official [Ruby LSP extension](https://marketplace.visualstudio.com
 If you need to select particular features to enable or disable, see
 [`vscode/package.json`](vscode/package.json) for the full list.
 
+**IMPORTANT NOTE FOR ALL EDITORS**
+
+The command to launch the language server might depend on which editor and version manager combination you are using.
+In order to work properly, the Ruby LSP must be launched with the Ruby version being used by the project you are working
+on and with the correct Bundler environment set.
+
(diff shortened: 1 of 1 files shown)


EDITORS.md
```diff
@@ -10,6 +10,36 @@ Code, use the official [Ruby LSP extension](https://marketplace.visualstudio.com
...
+If your version manager creates gem executable shims that perform the automatic version switching, then use those:
+
+- `~/.rbenv/shims/ruby-lsp`
+- `~/.asdf/shims/ruby-lsp`
+
+If your version manager doesn't provide either of those, then activate the environment and run the executable:
+
+- `chruby $(cat .ruby-version) && ruby-lsp`
+
+These strategies will ensure that the `ruby-lsp` executable is invoked with the correct Ruby version, GEM_HOME and
```
andyw8, line 40:
```suggestion
These strategies will ensure that the `ruby-lsp` executable is invoked with the correct Ruby version, `GEM_HOME` and
```

EDITORS.md
```diff
@@ -10,6 +10,36 @@ Code, use the official [Ruby LSP extension](https://marketplace.visualstudio.com
...
+GEM_PATH, which are necessary for proper integration with your project.
```
andyw8, line 41:
```suggestion
`GEM_PATH`, which are necessary for proper integration with your project.
```

andyw8 (approved):

PR #2214, opened 2024-06-20
Linearize singleton ancestors

### LINKED ISSUE #1333: Start keeping track of ancestors

### LINKED ISSUE #1938: Support singleton method indexing and resolution

### LINKED ISSUE #1939: Support singleton constants resolution (and/or indexing)

vinistock:
### Motivation

Closes #1939, closes #1938, closes #1333

This PR adds lazy singleton ancestor linearization, which allows us to properly handle constants, methods and instance variables that are inherited by a singleton class.

### Implementation

1. We already had lazy creation of singleton classes spreading across three different places. Moved that to the Index, which is the entity that's actually concerned with those. Started using it in RBSIndexer and the regular declaration listener
2. Removed the concept of `extend`. An extend is nothing but a convenience to include something in the singleton class. If we simply treat all extends as singleton class inclusions, the entire algorithm works as is and we don't have to handle another type of operation
3. Finally, the most important part is how to lazily determine the ancestors based on how many levels of singleton we are checking. Here's how the algorithm works:

For each level of singleton, the ancestors are: the linearization of the same level of singleton + the linearization of the level - 1 of the `Class` class. For example:
```ruby
class Foo; end
class Bar < Foo; end

# First singleton
Bar.singleton_class.ancestors
# => [Bar::<Class:Bar>,
#     Foo::<Class:Foo>,
#     Object::<Class:Object>,
#     BasicObject::<Class:BasicObject>, # <<<< Until here, it's the ancestors of <Class:Bar>
#     Class, Module, Object, Kernel, BasicObject] # <<< Ancestors of Class

# Second singleton
Bar.singleton_class.singleton_class.ancestors
# => [Bar::<Class:Bar>::<Class:<Class:Bar>>,
#     Foo::<Class:Foo>::<Class:<Class:Foo>>,
#     Object::<Class:Object>::<Class:<Class:Object>>,
#     BasicObject::<Class:BasicObject>::<Class:<Class:BasicObject>>,
#     ^^^^^^^ until here, it's the ancestors of <Class:<Class:Bar>>
#     Class::<Class:Class>, Module::<Class:Module>, Object::<Class:Object>,
#     BasicObject::<Class:BasicObject>,
#     ^^^^^^^ then the ancestors of <Class:Class> (the singleton level - 1)
#     Class, Module, Object, Kernel, BasicObject]
#     ^^^^^^^ And finally the ancestors of Class
```

### Automated Tests

Added tests.

### Manual Tests

1. Launch LSP on this branch
5. Type `File.` in any file
6. Verify that you see methods inherited from the `IO` class, such as the method `pipe`

### DIFF
--- a/lib/ruby_indexer/lib/ruby_indexer/declaration_listener.rb
+++ b/lib/ruby_indexer/lib/ruby_indexer/declaration_listener.rb
@@ -314,20 +314,22 @@ module RubyIndexer
           @owner_stack.last,
         ))
       when Prism::SelfNode
-        singleton = singleton_klass
+        owner = @owner_stack.last
 
-        @index.add(Entry::Method.new(
-          method_name,
-          @file_path,
-          node.location,
-          node.name_loc,
-          comments,
-          [Entry::Signature.new(list_params(node.parameters))],
-          current_visibility,
-          singleton,
-        ))
+        if owner
+          singleton = @index.existing_or_new_singleton_class(owner.name)
 
-        if singleton
+          @index.add(Entry::Method.new(
+            method_name,
+            @file_path,
+            node.location,
+            node.name_loc,
+            comments,
+            [Entry::Signature.new(list_params(node.parameters))],
+            current_visibility,
+            singleton,
+          ))
+
           @owner_stack << singleton
           @stack << "<Class:#{@stack.last}>"
         end
@@ -405,7 +407,12 @@ module RubyIndexer
 
       # When instance variables are declared inside the class body, they turn into class instance variables rather than
       # regular instance variables
-      owner = @inside_def ? @owner_stack.last : singleton_klass
+      owner = @owner_stack.last
+
+      if owner && !@inside_def
+        owner = @index.existing_or_new_singleton_class(owner.name)
+      end
+
       @index.add(Entry::InstanceVariable.new(name, @file_path, loc, collect_comments(node), owner))
     end
 
@@ -605,7 +612,8 @@ module RubyIndexer
         when :prepend
           owner.mixin_operations << Entry::Prepend.new(node.full_name)
         when :extend
-          owner.mixin_operations << Entry::Extend.new(node.full_name)
+          singleton = @index.existing_or_new_singleton_class(owner.name)
+          singleton.mixin_operations << Entry::Include.new(node.full_name)
         end
       rescue Prism::ConstantPathNode::DynamicPartsInConstantPathError,
              Prism::ConstantPathNode::MissingNodesInConstantPathError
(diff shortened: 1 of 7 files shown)


lib/ruby_indexer/lib/ruby_indexer/index.rb
```diff
@@ -374,8 +374,25 @@ def linearized_ancestors_of(fully_qualified_name)
       cached_ancestors = @ancestors[fully_qualified_name]
       return cached_ancestors if cached_ancestors
 
+      parts = fully_qualified_name.split("::")
```
andyw8, line 377: `linearized_ancestors_of` is getting really long (149 lines). Should we consider extracting out some methods?

lib/ruby_indexer/test/index_test.rb
```diff
@@ -1573,5 +1568,144 @@ def bar
 
       assert_equal("()", entry.decorated_parameters)
     end
+
+    def test_linearizing_singleton_singleton_ancestors_when_class_has_parent
```
andyw8, line 1572:
```suggestion
    def test_linearizing_singleton_ancestors_when_class_has_parent
```

andyw8 (approved):

lib/ruby_indexer/lib/ruby_indexer/index.rb
vinistock: I did my best in the latest commit. I split `linearize_mixins` and `linearize_superclass` into their own methods.

lib/ruby_indexer/lib/ruby_indexer/declaration_listener.rb
```diff
@@ -314,20 +314,22 @@ def on_def_node_enter(node)
...
-          node.name_loc,
-          comments,
-          list_params(node.parameters),
-          current_visibility,
-          singleton,
-        ))
+        if owner
+          singleton = @index.existing_or_new_singleton_class(owner.name)
+
+          @index.add(Entry::Method.new(
```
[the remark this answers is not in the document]
vinistock: That becomes a bit messy because the case statement helps us ignore dynamic definitions like `def foo.bar`. However, we also allow `owner` to be `nil` at the moment, which means we can't use that as the check for pushing a new entry into the index.

In the end, we clear the case statement, but then get a bunch of duplicate if statements that are already checked inside the case just to avoid pushing incorrect entries.

I think this will become easier once we turn `owner` into non nilable.

lib/ruby_indexer/test/index_test.rb
andyw8: Ahhh, I see. How about:
`test_linearizing_singleton_ancestors_of_singleton_when_class_has_parent`

vinistock: Good idea!

PR #2215, opened 2024-06-20
Vs/avoid showing duplicate completions

vinistock:
### Motivation

While testing #2214, I noticed we were showing duplicate completion entries. The issue is that we were not deduping methods with the same name that appear multiple times in the same hierarchy chain.

We always want to show only the first entry we find in the hierarchy chain, since that's the one that is actually getting invoked.

### Implementation

Since the prefix search returns all possible methods with no ordering guarantees, I started using a hash to remember the entries we found and the index in the ancestor chain where we found it.

Then we only override the entries if we find one earlier in the chain.

The code has a little bit of duplication, but it's hard to avoid because the completion path triggered on the dot character (`@entries.values.flatten`) returns any type of entries included classes. We can either duplicate the entry class checks or the ancestor index checks, but I don't think we can fully remove all of the duplication.

### Automated Tests

Added a test that prevents regression.


lib/ruby_indexer/lib/ruby_indexer/index.rb
```diff
@@ -150,17 +150,33 @@ def method_completion_candidates(name, receiver_name)
...
+      completion_items = candidates.each_with_object({}) do |entry, hash|
         case entry
         when Entry::Member, Entry::MethodAlias
-          entry if ancestors.any?(entry.owner&.name)
+          entry_name = entry.name
+          ancestor_index = ancestors.index(entry.owner&.name)
+          existing_entry = hash[entry_name]
+
+          # If this method belongs to an ancestors and either we haven't found it yet or it appears earlier in the
+          # ancestors chain than the previous candidate, then we need to put it in the hash. Otherwise, continue
```
st0012, line 161: Could you explain this part with an example? I don't quite get it 🤔

lib/ruby_indexer/lib/ruby_indexer/index.rb
```diff
@@ -150,17 +150,33 @@ def method_completion_candidates(name, receiver_name)
...
+          next unless ancestor_index && (!existing_entry || ancestor_index < existing_entry.last)
+
+          hash[entry_name] = [entry, ancestor_index]
         when Entry::UnresolvedMethodAlias
```
st0012, line 165: Since now these conditions have much more in common, maybe we can merge them together and only treat the last part differently based on entry type?

lib/ruby_indexer/lib/ruby_indexer/index.rb
vinistock: I tried expanding the explanation. Let me know if it's clear or confusing.

lib/ruby_indexer/lib/ruby_indexer/index.rb
```diff
@@ -150,17 +150,42 @@ def method_completion_candidates(name, receiver_name)
...
-      candidates.filter_map do |entry|
-        case entry
-        when Entry::Member, Entry::MethodAlias
-          entry if ancestors.any?(entry.owner&.name)
-        when Entry::UnresolvedMethodAlias
-          if ancestors.any?(entry.owner&.name)
-            resolved_alias = resolve_method_alias(entry, receiver_name)
-            resolved_alias if resolved_alias.is_a?(Entry::MethodAlias)
-          end
+      completion_items = candidates.each_with_object({}) do |entry, hash|
```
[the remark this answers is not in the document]
vinistock: We could, but I believe that will be slower because we'll be looping through the rather large array of all methods that matched the prefixed another time.

lib/ruby_indexer/lib/ruby_indexer/index.rb
```diff
@@ -150,17 +150,42 @@ def method_completion_candidates(name, receiver_name)
...
+        unless entry.is_a?(Entry::Member) || entry.is_a?(Entry::MethodAlias) ||
+            entry.is_a?(Entry::UnresolvedMethodAlias)
+          next
+        end
+
+        entry_name = entry.name
+        ancestor_index = ancestors.index(entry.owner&.name)
+        existing_entry = hash[entry_name]
```
st0012, line 161: Let's expand the array here with:

```rb
existing_entry, existing_entry_index = hash[entry_name]
```

And the condition below can be:

```rb
next unless ancestor_index && (!existing_entry || ancestor_index < existing_entry_index)
```

st0012 (approved):

lib/ruby_indexer/lib/ruby_indexer/entry.rb
```diff
@@ -498,12 +501,13 @@ def initialize(target, unresolved_alias)
...
 
         @target = target
+        @owner = T.let(unresolved_alias.owner, T.nilable(Entry::Namespace))
       end
 
-      sig { returns(T.nilable(Entry::Namespace)) }
-      def owner
-        @target.owner
-      end
+      # sig { returns(T.nilable(Entry::Namespace)) }
```
Morriar, line 507: Was this change intentional?

vinistock: Ah, forgot to remove the commented out code. I removed it directly on main https://github.com/Shopify/ruby-lsp/commit/3aa04306728a7cf24a1a1b9432154ea9f1ae3260.

PR #2218, opened 2024-06-21
Make Rails app detection based on Rails::Application superclass

louim:
### Motivation

closes https://github.com/Shopify/ruby-lsp/issues/2096

This change makes the Rails app detection based on the superclass of the main class being `Rails::Application`. This is a more reliable way to detect Rails apps than looking for the presence of the `rails` gem in the Gemfile. Application can require some of the Rails components without requiring the `rails` gem itself.

### Implementation

The implementation is strongly inspired  from #2161, hat tip to @Earlopain. I've added you as a co-author; let me know if that's ok.

### Automated Tests

The test all pass individually. However, when the full file is executed, one of the test doesn't pass. It's most certainly related to something I have changed. I wasn't able to pinpoint if context is leaking between tests or if something else is happening. Since this is also my first time using minitest, an experienced eye would be appreciated.

The problem can be reproduced with:
```
SPEC_REPORTER=1 bundle exec rake TEST="test/setup_bundler_test.rb" TESTOPTS="--seed=55457"
```

Relevant snip in case it doesn't fail:
```
  test_ruby_lsp_rails_is_automatically_included_in_rails_apps     PASS (1.39s)
  test_creates_custom_bundle_for_a_rails_app                      FAIL (0.00s)
        Expected path '.ruby-lsp/Gemfile.lock' to exist.
        /Users/louim/repositories/ruby-lsp/test/setup_bundler_test.rb:97:in `block (2 levels) in test_creates_custom_bundle_for_a_rails_app'
        /Users/louim/repositories/ruby-lsp/test/setup_bundler_test.rb:79:in `chdir'
        /Users/louim/repositories/ruby-lsp/test/setup_bundler_test.rb:79:in `block in test_creates_custom_bundle_for_a_rails_app'
        /Users/louim/.local/share/mise/installs/ruby/3.3.1/lib/ruby/3.3.0/tmpdir.rb:99:in `mktmpdir'
        /Users/louim/repositories/ruby-lsp/test/setup_bundler_test.rb:78:in `test_creates_custom_bundle_for_a_rails_app'
```
### Manual Tests

I haven't yet tested in our app, will do so as soon as possible.

### DIFF
--- a/lib/ruby_lsp/setup_bundler.rb
+++ b/lib/ruby_lsp/setup_bundler.rb
@@ -285,13 +285,11 @@ module RubyLsp
     # Detects if the project is a Rails app by looking if the superclass of the main class is `Rails::Application`
     sig { returns(T::Boolean) }
     def rails_app?
-      /class .* < Rails::Application/.match?(rails_application_content)
-    end
-
-    sig { returns(T.nilable(String)) }
-    def rails_application_content
       config = Pathname.new("config/application.rb").expand_path
-      config.read if config.exist?
+      application_contents = config.read if config.exist?
+      return false unless application_contents
+
+      /class .* < Rails::Application/.match?(application_contents)
     end
   end
 end
--- a/lib/ruby_lsp/setup_bundler.rb
+++ b/lib/ruby_lsp/setup_bundler.rb
@@ -289,7 +289,7 @@ module RubyLsp
       application_contents = config.read if config.exist?
       return false unless application_contents
 
-      /class .* < Rails::Application/.match?(application_contents)
+      /class .* < (::)?Rails::Application/.match?(application_contents)
     end
   end
 end
--- a/lib/ruby_lsp/setup_bundler.rb
+++ b/lib/ruby_lsp/setup_bundler.rb
@@ -50,6 +50,7 @@ module RubyLsp
       @last_updated_path = T.let(@custom_dir + "last_updated", Pathname)
 
       @dependencies = T.let(load_dependencies, T::Hash[String, T.untyped])
+      @rails_app = T.let(rails_app?, T::Boolean)
       @retry = T.let(false, T::Boolean)
     end
 
@@ -62,7 +63,7 @@ module RubyLsp
       # Do not set up a custom bundle if LSP dependencies are already in the Gemfile
       if @dependencies["ruby-lsp"] &&
           @dependencies["debug"] &&
-          (@dependencies["rails"] ? @dependencies["ruby-lsp-rails"] : true)
+          (@rails_app ? @dependencies["ruby-lsp-rails"] : true)
         $stderr.puts(
           "Ruby LSP> Skipping custom bundle setup since LSP dependencies are already in #{@gemfile}",
         )
@@ -148,7 +149,7 @@ module RubyLsp
         parts << 'gem "debug", require: false, group: :development, platforms: :mri'
       end
 
-      if @dependencies["rails"] && !@dependencies["ruby-lsp-rails"]
+      if @rails_app && !@dependencies["ruby-lsp-rails"]
         parts << 'gem "ruby-lsp-rails", require: false, group: :development'
       end
 
@@ -209,7 +210,7 @@ module RubyLsp
         command << " && bundle update "
         command << "ruby-lsp " unless @dependencies["ruby-lsp"]
         command << "debug " unless @dependencies["debug"]
-        command << "ruby-lsp-rails " if @dependencies["rails"] && !@dependencies["ruby-lsp-rails"]
+        command << "ruby-lsp-rails " if @rails_app && !@dependencies["ruby-lsp-rails"]
         command << "--pre" if @experimental
         command.delete_suffix!(" ")
         command << ")"
@@ -244,7 +245,7 @@ module RubyLsp
     def should_bundle_update?
       # If `ruby-lsp`, `ruby-lsp-rails` and `debug` are in the Gemfile, then we shouldn't try to upgrade them or else it
       # will produce version control changes
-      if @dependencies["rails"]
+      if @rails_app
         return false if @dependencies.values_at("ruby-lsp", "ruby-lsp-rails", "debug").all?
 
         # If the custom lockfile doesn't include `ruby-lsp`, `ruby-lsp-rails` or `debug`, we need to run bundle install
@@ -279,6 +280,16 @@ module RubyLsp
       end
 
       @custom_lockfile.write(content)
+    end
+
+    # Detects if the project is a Rails app by looking if the superclass of the main class is `Rails::Application`
+    sig { returns(T::Boolean) }
+    def rails_app?
+      config = Pathname.new("config/application.rb").expand_path
+      application_contents = config.read if config.exist?
+      return false unless application_contents
+
+      /class .* < Rails::Application/.match?(application_contents)
     end
   end
 end
--- a/lib/ruby_lsp/setup_bundler.rb
+++ b/lib/ruby_lsp/setup_bundler.rb
@@ -285,11 +285,13 @@ module RubyLsp
     # Detects if the project is a Rails app by looking if the superclass of the main class is `Rails::Application`
     sig { returns(T::Boolean) }
     def rails_app?
-      config = Pathname.new("config/application.rb").expand_path
-      application_contents = config.read if config.exist?
-      return false unless application_contents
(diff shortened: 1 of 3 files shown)


andyw8:
The test helper has `require "ruby_lsp/internal"` which brings in `prism`.

lib/ruby_lsp/listeners/rails_app.rb
```diff
@@ -0,0 +1,28 @@
+# typed: strict
+# frozen_string_literal: true
+
+module RubyLsp
+  module Listeners
+    class RailsApp
```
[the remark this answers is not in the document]
louim: Just pushed an update with the simplified version. If you have a minute to look at the test to tell me what's causing the problem, the PR will then be ready from my point of view.

lib/ruby_lsp/setup_bundler.rb
```diff
@@ -280,5 +281,15 @@ def correct_relative_remote_paths
...
     end
+
+    # Detects if the project is a Rails app by looking if the superclass of the main class is `Rails::Application`
+    sig { returns(T::Boolean) }
+    def rails_app?
+      config = Pathname.new("config/application.rb").expand_path(Dir.pwd)
+      application_contents = config.read if config.exist?
+      return false unless application_contents
+
+      /class .* < Rails::Application/.match?(application_contents)
```
Morriar, line 292: Should we proof this a bit more? I know the file is generated but I wonder if someone would have some weird cop doing this kind of change?

```suggestion
      /class .* < (::)?Rails::Application/.match?(application_contents)
```

louim:
> The changes look good, but there's a test failing.

@vinistock As I said in the original description, the tests all pass in isolation, but fail when ran together. I was not able to pinpoint the context leak between the tests. I will take another look, but if you have an idea on top of your head, as I'm not very familiar with Minitest so I might have overlooked something obvious.

andyw8:
I'm looking into. It can reproduced consistently with this seed:

```
TESTOPTS="--seed=18973" bundle exec ruby -Itest /Users/andyw8/src/github.com/Shopify/ruby-lsp/test/setup_bundler_test.rb
```

andyw8:
If I check `stderr` in `run_script` I see it contains:

```
Minitest::Assertion: Expected "Ruby LSP> Skipping lockfile copies because there's no top level bundle (, )\n" +
```
🤔

andyw8:
Spoke too soon, there might be something else that needs wrapped in `with_isolated_bundler`.

lib/ruby_lsp/setup_bundler.rb
```diff
@@ -280,5 +281,15 @@ def correct_relative_remote_paths
 
       @custom_lockfile.write(content)
     end
+
+    # Detects if the project is a Rails app by looking if the superclass of the main class is `Rails::Application`
+    sig { returns(T::Boolean) }
+    def rails_app?
+      config = Pathname.new("config/application.rb").expand_path(Dir.pwd)
```
paracycle, line 288:
```suggestion
      config = Pathname.new("config/application.rb").expand_path
```
The default value for `expand_path` is already `Dir.pwd`: https://ruby-doc.org/3.3.0/File.html#method-c-expand_path
> Relative paths are referenced from the current working directory of the process unless `dir_string` is given, in which case it will be used as the starting point

louim: Thanks, updated!

Earlopain:
I tried investigating the test failure again but didn't make any progress. How about just changing the code a bit to not run into that problem? I tested out the following patch locally and it doesn't appear to fail:

```diff
diff --git a/lib/ruby_lsp/setup_bundler.rb b/lib/ruby_lsp/setup_bundler.rb
index 903f43a3..b3ef164b 100644
--- a/lib/ruby_lsp/setup_bundler.rb
+++ b/lib/ruby_lsp/setup_bundler.rb
@@ -285,11 +285,15 @@ module RubyLsp
     # Detects if the project is a Rails app by looking if the superclass of the main class is `Rails::Application`
     sig { returns(T::Boolean) }
     def rails_app?
-      config = Pathname.new("config/application.rb").expand_path
-      application_contents = config.read if config.exist?
-      return false unless application_contents
+      return false unless (application_contents = rails_application_content)
 
       /class .* < Rails::Application/.match?(application_contents)
     end
+
+    sig { returns(T.nilable(String)) }
+    def rails_application_content
+      config = Pathname.new("config/application.rb").expand_path
+      config.read if config.exist?
+    end
   end
 end
diff --git a/test/setup_bundler_test.rb b/test/setup_bundler_test.rb
index 77f3b605..a65d7b3c 100644
--- a/test/setup_bundler_test.rb
+++ b/test/setup_bundler_test.rb
@@ -75,33 +75,33 @@ class SetupBundlerTest < Minitest::Test
   end
 
   def test_creates_custom_bundle_for_a_rails_app
-    Dir.mktmpdir do |dir|
-      Dir.chdir(dir) do
-        FileUtils.mkdir(File.join(dir, "config"))
-        File.write(File.join(dir, "config", "application.rb"), <<~RUBY)
-          module MyApp
-            class Application < Rails::Application
-            end
-          end
-        RUBY
-
-        Object.any_instance.expects(:system).with(
-          bundle_env(".ruby-lsp/Gemfile"),
-          "(bundle check || bundle install) 1>&2",
-        ).returns(true)
-        Bundler::LockfileParser.any_instance.expects(:dependencies).returns({ "rails" => true }).at_least_once
-        run_script
+    Object.any_instance.expects(:system).with(
+      bundle_env(".ruby-lsp/Gemfile"),
+      "(bundle check || bundle install) 1>&2",
+    ).returns(true)
 
-        assert_path_exists(".ruby-lsp")
-        assert_path_exists(".ruby-lsp/Gemfile")
-        assert_path_exists(".ruby-lsp/Gemfile.lock")
-        assert_path_exists(".ruby-lsp/main_lockfile_hash")
-        gemfile_content = File.read(".ruby-lsp/Gemfile")
-        assert_match("ruby-lsp", gemfile_content)
-        assert_match("debug", gemfile_content)
-        assert_match("ruby-lsp-rails", gemfile_content)
+    # There is some unknown state leak with Bundler which prevents us from creating the
+    # folder structure ourselves lest we encounter flaky tests
+    RubyLsp::SetupBundler.any_instance.stubs(:rails_application_content).returns(<<~RUBY)
+      module MyApp
+        class Application < Rails::Application
+        end
       end
-    end
+    RUBY
+
+    Bundler::LockfileParser.any_instance.expects(:dependencies).returns({ "rails" => true }).at_least_once
+
+    run_script
+
+    assert_path_exists(".ruby-lsp")
+    assert_path_exists(".ruby-lsp/Gemfile")
+    assert_path_exists(".ruby-lsp/Gemfile.lock")
+    assert_path_exists(".ruby-lsp/main_lockfile_hash")
+    assert_match("ruby-lsp", File.read(".ruby-lsp/Gemfile"))
+    assert_match("debug", File.read(".ruby-lsp/Gemfile"))
+    assert_match("ruby-lsp-rails", File.read(".ruby-lsp/Gemfile"))
+  ensure
+    FileUtils.rm_r(".ruby-lsp") if Dir.exist?(".ruby-lsp")
   end
 
   def test_changing_lockfile_causes_custom_bundle_to_be_rebuilt
```

andyw8:
@Earlopain I've taken your suggestion but made a slight change so that we use a file fixture instead of needing to stub the private method (which may be brittle).

lib/ruby_lsp/setup_bundler.rb
andyw8: This makes sense, I've committed it.

andyw8 (approved):

louim:
Hey! This looks like it's ready to merge. It would be nice to have this in.

test/setup_bundler_test.rb
```diff
@@ -91,6 +94,7 @@ def test_creates_custom_bundle_for_a_rails_app
     assert_match("ruby-lsp-rails", File.read(".ruby-lsp/Gemfile"))
   ensure
     FileUtils.rm_r(".ruby-lsp") if Dir.exist?(".ruby-lsp")
+    FileUtils.rm_rf("config") if Dir.exist?("config")
```
vinistock, line 97:
```suggestion
    FileUtils.rm_r("config") if Dir.exist?("config")
```

andyw8:
Seems there is an issue when the file contains non-ASCII characters: https://github.com/Shopify/ruby-lsp/issues/2463

PR #2230, opened 2024-06-27
Run multiple Minitest examples via a single ruby command from the VS Code test explorer

thomasmarshall:
### Motivation

Running multiple tests in the VS Code test explorer or via the "Run test" code lens link is currently quite slow. That's because it executes a separate command per example, even if the test runner is capable of running multiple examples at once. That means any start up cost is incurred for each example. This PR attempts to improve that situation for Minitest, by combining the examples into a single command.

### Implementation

We can't use the default Minitest test reporter (or whichever custom reporter the host project is using) for this, because it doesn't have enough structured output in order to know which tests have started or finished, and it's difficult to correlate any failure/error messages to the test in the explorer. This PR introduces a custom reporter that outputs structured JSON, and overrides whatever reporter is normally in place for the host project.

This relies on Minitest conventions to find the plugin, but it also injects some code to disable other Minitest plugins except our custom reporter. This is probably not ideal, but I wasn't sure how to force Minitest to use our custom reporter, other than disabling any other ones. Perhaps there is a way to guarantee our plugin is loaded last, so it overrides any other plugins? Alternatively, maybe our custom reporter could _augment_ the existing reporter and output JSON to a temporary log file instead stdout.

We also need some way to generate a combined test command from a series of separate examples. We're doing this by setting up some "combining" options that include the full name for each example and describe how those names should be combined into a command (in this case via a regex in passed to the `--name` argument). Ideally this should be extensible enough that it could be used by other test frameworks and runners—in particular I'd like to use this for testing Rails apps which often have a multi-second start up cost to running a test.

### Automated Tests

I have updated the expectation tests, I'm not sure how to add automated tests for the VS Code extension part.

### Manual Tests

Open a Minitest test file and click the "Run" code lens link. Try also selecting an arbitrary selection of tests in the explorer and running just those tests.

### Demo

Here's a before/after demo with a fake 1s start up cost.

https://github.com/Shopify/ruby-lsp/assets/770763/0643cf92-40a5-428e-8254-4d062f26c1de

### DIFF
--- a/lib/minitest/load.rb
+++ b/lib/minitest/load.rb
@@ -1,0 +1,6 @@
+# typed: false # rubocop:disable Sorbet/StrictSigil
+# frozen_string_literal: true
+
+require "minitest/manual_plugins"
+
+Minitest.load(:ruby_lsp)
--- a/lib/minitest/ruby_lsp_plugin.rb
+++ b/lib/minitest/ruby_lsp_plugin.rb
@@ -1,0 +1,37 @@
+# typed: false # rubocop:disable Sorbet/StrictSigil
+# frozen_string_literal: true
+
+require "json"
+require "minitest"
+
+module Minitest
+  class RubyLspReporter < Reporter
+    def prerecord(klass, name)
+      @id = "#{klass.name}##{name}"
+      io.puts JSON.generate(event: "start", id: @id)
+      io.flush
+    end
+
+    def record(result)
+      if result.failures.any?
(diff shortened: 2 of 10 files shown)


andyw8:
It might be possible get the reporter selection upstreamed:

> If TextMate, TeamCity, RubyMine or VIM presence is detected, the reporter will be automatically chosen, regardless of any reporters passed to the `use!` method.

https://github.com/minitest-reporters/minitest-reporters

andyw8:
^ Ignore the above, I was forgetting minitest-reporters isn't shipped with minitest.

PR #2233, opened 2024-06-28
Add blank Docusaurus site

andyw8:
### Motivation

This PR adds an empty documentation site powered by Docusaurus, so that we have a starting point for future PRs, based on what I started at Hackdays.

I've also added it as a new workspace in `ruby-lsp.code-workspace`.

We already have a `docs/` directory so I named this `docusaurus/` to distinguish it, but `/docs` would be nicer, so we [could rename the current one](https://github.com/Shopify/ruby-lsp/pull/2234).

### Implementation

Ran `npx create-docusaurus@latest` and chose the classic theme.

I've chosen to use Yarn since we already use that on the extension.

### Automated Tests

Nothing yet.

### Manual Tests

Run `yarn start` and verify the site loads for you.


andyw8:
Closing until we decide on the wider approach.

PR #2234, opened 2024-06-28
Change RDoc docs path

andyw8:
### Motivation

We're considering a new Documentation site (https://github.com/Shopify/ruby-lsp/pull/2233). Let's rename the RDoc output directory so that we can use `/docs` for the new one.

### Implementation

Change output directory, update `.gitignore`.

### Automated Tests

n/a

### Manual Tests

Run `bundle exec rake rdoc` verify the docs are generated in `rdoc_docs`, but git ignored.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,7 +7,7 @@
 /**/spec/reports/
 /**/tmp/
 /**/.dev/
-/docs
+/rdoc_docs
 
 vscode/out
 vscode/dist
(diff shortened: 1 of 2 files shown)


andyw8:
We'll also need to update the GitHub Pages path in https://github.com/Shopify/ruby-lsp/settings/pages

vinistock (commented):
I think GitHub pages only supports the repository root or the `docs` directory. Will it work if you rename it?

andyw8:
Closing until we determine the wider approach for new docs.

PR #2235, opened 2024-06-28
Provide code navigation features to erb files

st0012:
### Motivation

This commit allows Ruby LSP to start handling requests for ERB files, such as definition, completion, hover...etc., which will give users the same level of code navigation features in ERB and Ruby.

However, certain requests are not supported:
- formatting
- on_type_formatting
- diagnostics
- code_actions

### Implementation

1. Update extension so VS Code knows it can send erb requests to Ruby LSP too
2. Introduce `ERBDocument` (written by @vinistock) with a scanner that returns extracted Ruby and remaining html code (not used yet)
3. Make the requests that don't support erb files return early
4. Avoid assuming `Object` as view helper calls receiver type as it'll stop Ruby LSP from suggesting any definition candidates

### Demo

https://github.com/Shopify/ruby-lsp/assets/5079556/39cd8227-721f-40c9-8b9c-66316fc098b3


andyw8:
Worth adding an ERB fixture just for ease of tophatting?

lib/ruby_lsp/erb_document.rb
```diff
@@ -0,0 +1,103 @@
...
+          end
+        when "%"
+          if @inside_ruby && @source[@current_pos + 1] == ">"
+            @inside_ruby = false
+            @current_pos += 1
+            push_char("  ")
+          else
+            push_char(T.must(char))
+          end
+        when "\n"
```
andyw8, line 83: (do we need to consider Windows line endings?)

lib/ruby_lsp/requests/code_actions.rb
```diff
@@ -48,6 +48,8 @@ def initialize(document, range, context)
 
       sig { override.returns(T.nilable(T.all(T::Array[Interface::CodeAction], Object))) }
       def perform
+        return if @document.is_a?(ERBDocument)
```
andyw8, line 51: Perhaps these checks could be something like `@document.supports_code_actions?`, to ease the introduction of other document types in future.

lib/ruby_lsp/erb_document.rb
```diff
@@ -0,0 +1,103 @@
...
+        while @current_pos < @source.length
+          scan_char
+          @current_pos += 1
+        end
+      end
+
+      private
+
+      sig { void }
+      def scan_char
```
Earlopain, line 44: `<%==` is valid syntax, basically does `raw`/`html_safe`. Should that be considered here?

I'm not sure if these are supported via plain `erb` but it certainly works with `erubi`.

andyw8: Thanks, we'll look into that.

Are there any other ERB gems other than those 2? I couldn't see any when searching.

andyw8: Found https://rubygems.org/gems/erubis/ but last release was 2011.

Earlopain: I'm not aware of any other (recent) other implementation myself. `erubi` is the one used by rails and while it technically does allow to switch it out to something else that doesn't seem to be documented/used anymore. Rails 5.2 removed the `erubis` handler

st0012: This is a good catch 👍. Based on [this test case](https://github.com/ruby/erb/blob/5d1c5108b600786a03ac3f38fcbf3f6054294d4c/test/erb/test_erb.rb#L582-L606), I think it's not supported by `erb` but is an expected extension, so I think it makes sense to support it.

lib/ruby_lsp/internal.rb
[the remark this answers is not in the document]
st0012: `erb` is a [default gem](https://github.com/ruby/ruby/blob/master/lib/erb.rb) so it's not required to have it in Gemfile to use it.

lib/ruby_lsp/listeners/definition.rb
```diff
@@ -52,7 +53,15 @@ def on_call_node_enter(node)
         message = node.message
         return unless message
 
-        handle_method_definition(message, @type_inferrer.infer_receiver_type(@node_context))
+        inferrer_receiver_type = @type_inferrer.infer_receiver_type(@node_context)
```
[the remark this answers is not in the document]
st0012: I'm not sure if I understand. Don't we always need to infer the receiver type before making the call?

lib/ruby_lsp/server.rb
```diff
@@ -541,7 +551,11 @@ def text_document_diagnostic(message)
         return
       end
 
-      response = @store.cache_fetch(uri, "textDocument/diagnostic") do |document|
+      document = @store.get(uri)
+
+      return send_empty_response(message[:id]) if document.is_a?(ERBDocument)
+
+      response = document.cache_fetch("textDocument/diagnostic") do |document|
```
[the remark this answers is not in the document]
st0012: Consider `@store.cache_fetch` is literally just `get(uri).cache_fetch(request_name, &block)`, I don't feel we're losing anything by calling `document.cache_fetch`.
In the contrary, I think the fact that we will need a comment for the condition shows that the proposed implementation is potentially confusing?

I also feel the current implementation fits better with our current convention: if we don't need to go further, we return early.

lib/ruby_lsp/store.rb
```diff
@@ -37,13 +37,18 @@ def get(uri)
       return document unless document.nil?
 
       path = T.must(uri.to_standardized_path)
-      set(uri: uri, source: File.binread(path), version: 0)
+      language_id = path.end_with?(".erb") ? "erb" : "ruby"
```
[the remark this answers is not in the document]
st0012: TIL `rhtml`. I think we'll need to add it to the extension's `documentSelector` too?

lib/ruby_lsp/internal.rb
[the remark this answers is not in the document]
st0012: But Ruby LSP isn't run with the client application right? How would we know if the client process has it defined?

lib/ruby_lsp/erb_document.rb
st0012: Updated 👍

lib/ruby_lsp/listeners/definition.rb
```diff
@@ -16,18 +16,19 @@ class Definition
...
         ).void
       end
-      def initialize(response_builder, global_state, uri, node_context, dispatcher, typechecker_enabled) # rubocop:disable Metrics/ParameterLists
+      def initialize(response_builder, global_state, document, node_context, dispatcher, typechecker_enabled) # rubocop:disable Metrics/ParameterLists
         @response_builder = response_builder
         @global_state = global_state
         @index = T.let(global_state.index, RubyIndexer::Index)
         @type_inferrer = T.let(global_state.type_inferrer, TypeInferrer)
-        @uri = uri
+        @document = document
```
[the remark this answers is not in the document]
vinistock: Yes, for the server I believe that's a better idea because Sorbet will narrow types and help us get more out of the type checker.

But in this specific scenario, I really want to avoid passing document down. Maybe we can pass the document's class rather than the instance? Then we can differentiate the behaviour based on that.

lib/ruby_lsp/listeners/definition.rb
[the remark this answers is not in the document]
vinistock: Oh, my bad. Yeah, we need to check if the receiver is `Object`.

lib/ruby_lsp/store.rb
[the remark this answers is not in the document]
vinistock: Not necessary because the grammar already assigns the `erb` language ID to that file type. Similar to how we assign the `ruby` ID to several different files, like `Rakefile`, `Gemfile`, etc.

lib/ruby_lsp/server.rb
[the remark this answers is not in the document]
vinistock: > I also feel the current implementation fits better with our current convention: if we don't need to go further, we return early.

It depends on which convention you are prioritizing. We also don't manipulate the document's cache without going through the store anywhere, which means we'd be breaking a different convention.

In my head, in one case you are changing the implementation of a request, which should be flexible anyway since requests may operate differently and have different requirements. In the other, you are reaching into a lower level API.

It's not a strong opinion and if you feel like the case statement is confusing, feel free to move forward this way.

lib/ruby_lsp/listeners/definition.rb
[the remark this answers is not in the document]
st0012: I ended up making `Document#language_id` returning an enum, which will get us the benefit of Sorbet in all cases and also save us from passing document here.

lib/ruby_lsp/document.rb
```diff
@@ -36,9 +43,12 @@ def initialize(source:, version:, uri:, encoding: Encoding::UTF_8)
 
     sig { params(other: Document).returns(T::Boolean) }
     def ==(other)
-      @source == other.source
+      self.class == other.class && uri == other.uri && @source == other.source
```
vinistock, line 46: I don't believe we need the class part since the URI is already a unique identifier for the document.
```suggestion
      uri == other.uri && @source == other.source
```

lib/ruby_lsp/store.rb
```diff
@@ -37,13 +37,31 @@ def get(uri)
       return document unless document.nil?
 
       path = T.must(uri.to_standardized_path)
-      set(uri: uri, source: File.binread(path), version: 0)
+      language_id = if path.end_with?(".erb") || path.end_with?(".rhtml")
```
vinistock, line 40:
```suggestion
      ext = File.extname(path)
      language_id = if ext == ".erb" || ext == ".rhtml"
```

vinistock (approved):
🚀

lib/ruby_lsp/document.rb
vinistock: Okay, that's fine.

robikovacs:
Hello,

Is there a recommended setup that does format on save for erb files?

PR #2241, opened 2024-07-02
Add better initialization error handling

### LINKED ISSUE #1769: Add a custom error handler for the LSP client
The LSP is highly sensitive to Bundler related issues. If any errors occur, we're currently letting the default LSP error handler do its thing, which is restart 5 times and then fail in a non graceful way.

This is not a great experience. We can certainly do better by defining our own custom error handler.  This can be done by defining an `errorHandler` in the [client options](https://github.com/Shopify/vscode-ruby-lsp/blob/79d53557cd0941d97a93d2a937949850b1f86c39/src/client.ts#L112).

Ideally, we should display the error to the user, give them the chance to fix it and them offer to re-launch.

vinistock:
### Motivation

Closes #1769

Add a custom error handler to the LSP client, so that we can provide a nicer experience if activation fails due to Bundler issues or version manager integrations.

This also gets rid of the incredibly annoying 5 attempt restart loop that we find ourselves in these situations.

### Implementation

In the error handler, we unfortunately don't have access to the output of the server, so we can't easily parse it to understand what type of error it is. So the approach I took is the following:

1. Added an error handler class
2. Added some default handling for when the communication errors, which will just shutdown the LSP
3. Added the special handling to when the server connection is closed, which is what happens when we fail to activate. Note that there are other reasons why the connection might be closed, but they are far less frequent than activation issues. For example, if Ruby, YJIT or Prism were to segfault, then the connection would close and we'd hit that error handler path. Even though we cannot be 100% precise, I still think this is a better experience
4. On the special handling, I'm just showing a message saying we failed to activate, explaining the possible reasons, and then showing 2 different set of actions
    - The links in the markdown are debug actions. Those do not close the dialog and so you can click on them however many times you like
    - The two actual dialog buttons are `Retry` and `Shutdown`, which then close the dialog and do exactly what you'd expect them to

https://github.com/Shopify/ruby-lsp/assets/18742907/9bd8c4b7-a57a-430c-8f80-cbbb2a213dd3

### DIFF
--- a/vscode/src/client.ts
+++ b/vscode/src/client.ts
@@ -13,6 +13,12 @@ import {
   ServerOptions,
   MessageSignature,
   DocumentSelector,
+  ErrorHandler,
+  CloseHandlerResult,
+  ErrorHandlerResult,
+  Message,
+  ErrorAction,
+  CloseAction,
 } from "vscode-languageclient/node";
 
 import { LSP_NAME, ClientInterface, Addon } from "./common";
@@ -152,6 +158,7 @@ function collectClientOptions(
     outputChannel,
     revealOutputChannelOn: RevealOutputChannelOn.Never,
     diagnosticPullOptions,
+    errorHandler: new ClientErrorHandler(workspaceFolder),
     initializationOptions: {
       enabledFeatures,
       experimentalFeaturesEnabled: configuration.get(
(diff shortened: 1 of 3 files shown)


vscode/src/client.ts
```diff
@@ -164,6 +171,41 @@ function collectClientOptions(
...
+  }
+
+  async closed(): Promise<CloseHandlerResult> {
+    const answer = await vscode.window.showErrorMessage(
+      `Launching the Ruby LSP failed. This typically happens due to an error with version manager
+      integration or Bundler issues.
+
+      [Logs](command:workbench.action.output.toggleOutput) |
+      [Troubleshooting](https://github.com/Shopify/ruby-lsp/blob/main/TROUBLESHOOTING.md) |
+      [Bundle install](command:rubyLsp.bundleInstall?"${this.workspaceFolder.uri.toString()}")`,
```
andyw8, line 196: Since this one is an action, and the other two are for information, it is possible to make this a button? (If not, can we say "Run Bundle Install"?)

vscode/src/rubyLsp.ts
```diff
@@ -394,6 +394,25 @@ export class RubyLsp {
...
+          }
+
+          const terminal = vscode.window.createTerminal({
+            name: "Bundle install",
+            cwd: workspace.workspaceFolder.uri.fsPath,
+            env: workspace.ruby.env,
+          });
+
+          terminal.show();
+          terminal.sendText("bundle install");
```
andyw8, line 413: Is it possible to clear any text the user may have typed before sending this? (I sometimes get failures for the terminal test runner because of an incomplete command I had previously typed).

vinistock: We are always creating a brand new terminal in the lines above, so there's no way you could type anything before running the command.

PR #2242, opened 2024-07-02
Simplify class and module grammar for more consistency

vinistock:
### Motivation

One tiny thing that bothered me with the syntax highlighting is that our grammar incorrectly considered the `::` separator as part of a class/module declaration, highlighting the entire thing in the same colour. This PR fixes that by simplifying the grammar regex.

### Implementation

I simplified the regex in the grammar. There are things it was trying to account for that were unnecessary. For example, it tried to handle any arbitrary number of line breaks between the `class` keyword and the constant name (which TIL is valid Ruby). That's not necessary because semantic highlighting will catch that constant for us.

Another example is that it included trying to parse a parent class for modules, which is not valid Ruby `module Foo < Bar`.

We also don't need to account for syntax errors since those are already provided by the server. The summary is: we can make these regexes much simpler and easier to understand.

The way the grammar works is:

1. Write a regex in the `match` attribute
2. Define what highlight token should be assigned to each regex capture group

### DIFF
--- a/vscode/grammars/ruby.cson.json
+++ b/vscode/grammars/ruby.cson.json
@@ -51,20 +51,21 @@
         "2": {
           "name": "entity.name.type.class.ruby"
         },
-        "4": {
-          "name": "entity.other.inherited-class.ruby"
-        },
         "5": {
-          "name": "punctuation.separator.inheritance.ruby"
-        },
-        "6": {
-          "name": "variable.other.object.ruby"
+          "name": "punctuation.separator.namespace.ruby"
         },
         "7": {
-          "name": "punctuation.definition.variable.ruby"
+          "name": "punctuation.separator.inheritance.ruby"
+        },
+        "8": {
+          "name": "entity.other.inherited-class.ruby"
+        },
+        "11": {
+          "name": "punctuation.separator.namespace.ruby"
         }
       },
-      "match": "(?x)\n^\\s*(class)\\s+\n(\n  (\n    [.a-zA-Z0-9_:]+\n    (\\s*(<)\\s*[.a-zA-Z0-9_:]+)?   # class A < B\n  )\n  |\n  ((<<)\\s*[.a-zA-Z0-9_:]+)         # class << C\n)",
+      "comment": "class Namespace::ClassName < OtherNamespace::OtherClassName",
+      "match": "(class)\\s*(([a-zA-Z0-9_]+)((::)[a-zA-Z0-9_]+)*)\\s*((<)\\s*(([a-zA-Z0-9_]+)((::)[a-zA-Z0-9_]+)*))?",
       "name": "meta.class.ruby"
     },
     {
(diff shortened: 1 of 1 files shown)


vinistock:
@andyw8 I think that's a great idea and necessary in the long run to maintain these grammar files in a healthy way. The Ruby one is huge and understanding all regexes is no easy task.

However, I'd prefer leaving that to a separate task. VS Code doesn't provide an easy way to test grammar files as far as I know. We will need to explore building a test framework that applies the grammar rules to a string of Ruby code and then allows us to assert the token types for every position. It's bound to take a little bit of effort to get right. I created https://github.com/Shopify/ruby-lsp/issues/2249 so that we don't forget.

vscode/grammars/ruby.cson.json
```diff
@@ -51,20 +51,21 @@
...
         },
-        "7": {
-          "name": "punctuation.definition.variable.ruby"
+        "11": {
+          "name": "punctuation.separator.namespace.ruby"
         }
       },
-      "match": "(?x)\n^\\s*(class)\\s+\n(\n  (\n    [.a-zA-Z0-9_:]+\n    (\\s*(<)\\s*[.a-zA-Z0-9_:]+)?   # class A < B\n  )\n  |\n  ((<<)\\s*[.a-zA-Z0-9_:]+)         # class << C\n)",
+      "comment": "class Namespace::ClassName < OtherNamespace::OtherClassName",
+      "match": "(class)\\s*(([a-zA-Z0-9_]+)((::)[a-zA-Z0-9_]+)*)\\s*((<)\\s*(([a-zA-Z0-9_]+)((::)[a-zA-Z0-9_]+)*))?",
```
andyw8, line 68: For illustration:

```ruby
s = "class Namespace::ClassName < OtherNamespace::OtherClassName"
re = "(?x)\n^\\s*(class)\\s+\n(\n  (\n    [.a-zA-Z0-9_:]+\n    (\\s*(<)\\s*[.a-zA-Z0-9_:]+)?   # class A < B\n 
 )\n  |\n  ((<<)\\s*[.a-zA-Z0-9_:]+)         # class << C\n)"
s.match(re)
=> 
#<MatchData
 "class Namespace::ClassName < OtherNamespace::OtherClassName"
 1:"class"
 2:"Namespace::ClassName < OtherNamespace::OtherClassName"
 3:"Namespace::ClassName < OtherNamespace::OtherClassName"
 4:" < OtherNamespace::OtherClassName"
 5:"<"
 6:nil
 7:nil>
```

vscode/grammars/ruby.cson.json
```diff
@@ -75,27 +76,24 @@
...
+          "name": "keyword.control.class.ruby"
         },
-        "8": {
+        "2": {
           "name": "punctuation.separator.inheritance.ruby"
         }
       },
-      "match": "(?x)\n^\\s*(module)\\s+\n(\n  ([A-Z]\\w*(::))?\n  ([A-Z]\\w*(::))?\n  ([A-Z]\\w*(::))*\n  [A-Z]\\w*\n)",
-      "name": "meta.module.ruby"
+      "match": "(class)\\s*(<<)\\s",
```
andyw8, line 95:
```suggestion
      "match": "(class)\\s*(<<)\\s*",
```
Since `class<<self` is valid.

andyw8 (approved):

PR #2243, opened 2024-07-02
WIP: Parse RBS overloads into signatures

andyw8:


andyw8:
(have moved the Ruby LSP changes out to `andyw8/support-multiple-signatures`)

andyw8:
@soutaro this is still WIP, but thought you may be interested to see our approach.

lib/ruby_indexer/test/rbs_indexer_test.rb
```diff
@@ -74,5 +74,229 @@ def test_location_and_name_location_are_the_same
...
+      assert_equal([:by, :to], parameters.map(&:name))
+      assert_kind_of(Entry::OptionalKeywordParameter, parameters[0])
+      assert_kind_of(Entry::OptionalKeywordParameter, parameters[1])
+    end
+
+    def test_rbs_method_with_required_keywords
+      # There are no methods in Core that have required keyword arguments,
+      # so we test against RBS directly
+
+      # TODO: why does removing the block break this?
```
andyw8, line 253: Need to figure out why
```
def foo: (::Numeric a, ::Numeric b) -> self
```
is failing with:
```
Minitest::Assertion: Expected #<RubyIndexer::Entry::RequiredParameter:0x0000000129a54158 @name=:a>
 to be a kind of RubyIndexer::Entry::KeywordParameter, not RubyIndexer::Entry::RequiredParameter.
```

lib/ruby_indexer/test/method_test.rb
```diff
@@ -123,8 +123,9 @@ def bar(a)
 
       assert_entry("bar", Entry::Method, "/fake/path/foo.rb:1-2:2-5")
       entry = T.must(@index["bar"].first)
-      assert_equal(1, entry.parameters.length)
-      parameter = entry.parameters.first
+      parameters = entry.signatures.first.parameters
+      assert_equal(1, parameters.length)
```
andyw8, line 127: These kind of length assertions were useful during development, but we remove them to make the tests a little shorter.

lib/ruby_indexer/test/rbs_indexer_test.rb
```diff
@@ -74,5 +74,229 @@ def test_location_and_name_location_are_the_same
...
+      entries = @index["crypt"] # https://www.rubydoc.info/stdlib/core/String#crypt-instance_method
+      assert_equal(1, entries.length)
+
+      entry = entries.first
+      signatures = entry.signatures
+      assert_equal(1, signatures.length)
+
+      first_signature = signatures.first
+
+      # (::string salt_str) -> ::String
```
andyw8, line 88: I copied in the RBS definitions for easy reference, but we could remove them.

lib/ruby_indexer/test/rbs_indexer_test.rb
```diff
@@ -74,5 +74,229 @@ def test_location_and_name_location_are_the_same
...
+      parameters = signatures[1].parameters
+      assert_equal([:read_array, :write_array, :error_array, :timeout], parameters.map(&:name))
+      assert_kind_of(Entry::RequiredParameter, parameters[0])
+      assert_kind_of(Entry::OptionalParameter, parameters[1])
+      assert_kind_of(Entry::OptionalParameter, parameters[2])
+      assert_kind_of(Entry::OptionalParameter, parameters[3])
+    end
+
+    def test_rbs_method_with_optional_keywords
+      entries = @index["step"] # https://www.rubydoc.info/stdlib/core/Numeric#step-instance_method
```
andyw8, line 215: `step` is an interesting one since it can called with either positional or keyword args.

andyw8:
There's one thing I need to fix but I'd like to start getting reviews on this.

lib/ruby_indexer/lib/ruby_indexer/rbs_indexer.rb
```diff
@@ -20,15 +20,21 @@ def index_ruby_core
       end
     end
 
-    private
-
-    sig { params(source: T.untyped, pathname: Pathname, declarations: T::Array[RBS::AST::Declarations::Base]).void }
+    sig do
+      params(
+        source: T.any(Symbol, String), # If source is from Ruby Core, it will be the symbol `:core`
```
vinistock, line 25: It doesn't look like we're using `source`. Should we remove it?

lib/ruby_indexer/lib/ruby_indexer/rbs_indexer.rb
```diff
@@ -122,7 +128,91 @@ def handle_method(member, owner)
...
+
+    sig { params(function: RBS::Types::Function).returns(T::Array[Entry::Parameter]) }
+    def parse_arguments(function)
+      parameters = []
+      parameters += process_required_positionals(function) if function.required_positionals
+      parameters += process_optional_positionals(function) if function.optional_positionals
+      parameters += process_trailing_positionals(function) if function.trailing_positionals
+      parameters += process_rest_positionals(function) if function.rest_positionals
+      parameters += process_required_keywords(function) if function.required_keywords
+      parameters += process_optional_keywords(function) if function.optional_keywords
```
vinistock, line 162: Looking at our implementation for Ruby, are we missing keyword rest? (e.g.: `**kwargs`)?

lib/ruby_indexer/lib/ruby_indexer/rbs_indexer.rb
```diff
@@ -122,7 +128,91 @@ def handle_method(member, owner)
...
+        Entry::RequiredParameter.new(name: name)
+      end
+    end
+
+    sig { params(function: RBS::Types::Function).returns(T::Array[Entry::OptionalParameter]) }
+    def process_optional_positionals(function)
+      function.optional_positionals.map do |param|
+        # Optional positionals may be unnamed, e.g.
+        #  def self.polar: (Numeric, ?Numeric) -> Complex
+        name = param.name || :anon # TODO: what should we call this?
```
vinistock, line 182: I wouldn't use `anon` since you cannot actually have an anonymous optional parameter, they must always have some name.

My vote goes for simplicity with `arg0`, `arg1`, `arg2` and so on using a map with index.

lib/ruby_indexer/lib/ruby_indexer/rbs_indexer.rb
```diff
@@ -122,7 +128,91 @@ def handle_method(member, owner)
...
+
+        Entry::OptionalParameter.new(name: name)
+      end
+    end
+
+    sig { params(function: RBS::Types::Function).returns(T::Array[Entry::OptionalParameter]) }
+    def process_trailing_positionals(function)
```
vinistock, line 189: I'll defer to you if you prefer this style, but I'm not a fan of this tiny methods that just map through one parameter type. I would just inline them all.

lib/ruby_indexer/test/rbs_indexer_test.rb
vinistock: I like those, it's easier to understand 👍

lib/ruby_indexer/test/rbs_indexer_test.rb
```diff
@@ -74,5 +74,229 @@ def test_location_and_name_location_are_the_same
...
+        end
+      RBS
+      signatures = parse_rbs_methods(rbs, "foo")
+      parameters = signatures[0].parameters
+      assert_equal([:a, :b], parameters.map(&:name))
+      assert_kind_of(Entry::KeywordParameter, parameters[0])
+      assert_kind_of(Entry::KeywordParameter, parameters[1])
+    end
+
+    def test_parse_simple_rbs
```
vinistock, line 266: Let's add a test for keyword rest `**kwargs`.

vinistock (commented):
This is great

lib/ruby_indexer/lib/ruby_indexer/rbs_indexer.rb
```diff
@@ -122,7 +127,102 @@ def handle_method(member, owner)
...
+      parameters += process_rest_keywords(function) if function.rest_keywords
+      parameters
+    end
+
+    sig { params(function: RBS::Types::Function).returns(T::Array[Entry::RequiredParameter]) }
+    def process_required_positionals(function)
+      function.required_positionals.map do |param|
+        # Some parameters don't have names, e.g.
+        #   def self.try_convert: [U] (untyped) -> ::Array[U]?
+        name = param.name || :object
```
vinistock, line 171: Let's also use `arg{index}` here instead of just `object`.

lib/ruby_indexer/lib/ruby_indexer/rbs_indexer.rb
```diff
@@ -122,7 +127,102 @@ def handle_method(member, owner)
...
+    end
+
+    # TODO: add test
+    sig { params(function: RBS::Types::Function).returns(T::Array[Entry::KeywordRestParameter]) }
+    def process_rest_keywords(function)
+      param = function.rest_keywords
+
+      name = param.name || Entry::KeywordRestParameter::DEFAULT_NAME
+
+      [Entry::KeywordRestParameter.new(name: name)]
```
vinistock, line 225: I'm disliking this pattern of tiny methods more and more. Can we at least redesign this so that we are not forced to create arrays just to concatenate in the call site?

lib/ruby_indexer/lib/ruby_indexer/rbs_indexer.rb
```diff
@@ -122,7 +127,102 @@ def handle_method(member, owner)
...
+      block = overload.method_type.block
+      parameters << Entry::BlockParameter.anonymous if block&.required
+
+      parameters
+    end
+
+    sig { params(function: RBS::Types::Function).returns(T::Array[Entry::Parameter]) }
+    def parse_arguments(function)
+      parameters = []
+      parameters += process_required_positionals(function) if function.required_positionals
```
vinistock, line 156: All of these `+=` will generate intermediate arrays. Let's switch them all for `concat`.

lib/ruby_indexer/lib/ruby_indexer/rbs_indexer.rb
andyw8: I do generally prefer this way, I picked it from the guidance in [Clean Code](https://www.amazon.ca/Clean-Code-Handbook-Software-Craftsmanship/dp/0132350882) that that methods should have only one level of abstraction.

lib/ruby_indexer/lib/ruby_indexer/rbs_indexer.rb
```diff
@@ -122,7 +127,106 @@ def handle_method(member, owner)
...
+      parameters.concat(process_trailing_positionals(function)) if function.trailing_positionals
+      parameters << process_rest_positionals(function) if function.rest_positionals
+      parameters.concat(process_required_keywords(function)) if function.required_keywords
+      parameters.concat(process_optional_keywords(function)) if function.optional_keywords
+      parameters << process_rest_keywords(function) if function.rest_keywords
+      parameters
+    end
+
+    sig { params(function: RBS::Types::Function).returns(T::Array[Entry::RequiredParameter]) }
+    def process_required_and_optional_positionals(function)
```
andyw8, line 166: This was prevously two methods, I combined them because we need a way to keep track of an index to use for `arg0` etc, and this seemed the simplest way.

vinistock (approved):

PR #2246, opened 2024-07-02
Upgrade Node to v20.15.0

vinistock:
### Motivation

Since #2220, we're getting flaky NodeJS crashes in CI which are very annoying. I suggest we try to upgrade Node to the latest minor version.

It's a bit ahead of what VS Code uses, but there shouldn't be any breaking changes in minor versions, so I think it should be fine.

### Implementation

Upgraded NodeJS.


vinistock:
Well, it didn't fix it, so I guess we can close.

PR #2250, opened 2024-07-03
Make `RubyIndexer::Index#entries` public

andyw8:
### Motivation

We commonly access the index's entries in tests, but we need to awkwardly read the internal state of the index. Let's make this public instead.

I've added a note to highlight that this is only intended for tests.

This is a slight risk that addon authors could mis-use this API, but they could have circumvented the access control anyway if they wanted.

### Implementation

Add an `attr_reader`.

### Automated Tests

n/a

### Manual Tests

n/a

### DIFF
--- a/exe/ruby-lsp
+++ b/exe/ruby-lsp
@@ -97,8 +97,7 @@ if options[:time_index]
   index = RubyIndexer::Index.new
 
   result = Benchmark.realtime { index.index_all }
-  entries = index.instance_variable_get(:@entries)
-  entries_by_entry_type = entries.values.flatten.group_by(&:class)
+  entries_by_entry_type = index.entries.values.flatten.group_by(&:class)
 
   puts <<~MSG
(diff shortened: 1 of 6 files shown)


vinistock (changes requested):
The decision to not make `@entries` public was very intentional. We don't want users of the API to interact with the internal state of the index without going through one of the public APIs, as someone could accidentally corrupt the internal state.

If we need a new API for some specific need, we can add it to the index, but we should not expose any of its internal state.

Because it's Ruby, one can always do something like `instance_variable_get`, but at least it communicates very clearly that they are reaching into non public internal state and could get into trouble.

vinistock:
I wouldn't do that. The method will show up as available and people will only discover they are not supposed to use it when things break. I would honestly just keep the `instance_variable_get` occurrences we have in tests.

Morriar:
You could return `@entries.dup` or just `@entries.values`

andyw8:
I've pushed an update with @Morriar `dup` suggestion for consideration. (GH won't let me re-open since I force pushed).

PR #2253, opened 2024-07-03
Fix display of label details for completion requests

andyw8:
### Motivation

While working on https://github.com/Shopify/ruby-lsp/pull/2243 we noticed that the details for completions weren't appearing. The label details cannot be added in the resolve, they have to be pre-computed.

This PR also introduces a label detail showing where an instance variable is defined.

### Implementation

Move the behaviour from `completion_resolve` to `completion`.

As can be seen below, we're not yet showing the available arguments for Core methods, that will come next.

Screenshot 2024-07-03 at 1 05 32 PM

### Automated Tests

Updated

### Manual Tests

Type a class name, hit `.`, and verify you see the corresponding file names.

### DIFF
--- a/lib/ruby_lsp/listeners/completion.rb
+++ b/lib/ruby_lsp/listeners/completion.rb
@@ -209,8 +209,13 @@ module RubyLsp
         @index.instance_variable_completion_candidates(name, type).each do |entry|
           variable_name = entry.name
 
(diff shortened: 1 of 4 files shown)


vinistock (approved):
Let's also add an assertion for the label details for constant completion in `completion_test`

PR #2255, opened 2024-07-04
Set `DISABLE_V8_COMPILE_CACHE=1` for Yarn in CI

andyw8:
We're seeing a common failure recently for Node in CI:

> Could not get yarn cache folder path for /home/runner/work/ruby-lsp/ruby-lsp/vscode/.vscode"

https://github.com/Shopify/ruby-lsp/actions/runs/9783340748/job/27011714149

The post at https://github.com/nodejs/node/issues/51555 suggests this may help to avoid it.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -55,6 +55,8 @@ jobs:
       - uses: actions/checkout@v4
 
       - name: Set up Node
+        env:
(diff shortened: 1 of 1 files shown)


andyw8:
@st0012 this is only for the JIT, not the package caching.

It does seem to add about 10s to the build, but we can live with that, since the other parts are much slower.

st0012 (approved):

PR #2259, opened 2024-07-05
Use new telemetry API

### LINKED ISSUE #1778: Refactor our telemetry to use VS Code's `TelemetryLogger`
VS Code already provides a telemetry logger with a pluggable `TelemetrySender`, which would be the equivalent of our private API. We can get rid of our implementation and just use the one provided by the API.

Docs: https://code.visualstudio.com/api/references/vscode-api#TelemetryLogger

vinistock:
### Motivation

Closes #1778

Refactor our telemetry to follow VS Code API's conventions and use the `vscode.TelemtrySender` interface. This not only simplifies our code, but also maintains consistency with how other extensions would be collecting metrics.

### Implementation

1. Started using the VS Code API to create a new telemetry logger instance. I did this in `extension.ts` because creating the logger is async (which means we could not do it inside a constructor) and I didn't want the telemetry object to be possibly undefined. This code is a little awkward, but essentially this is how it breaks down:
    - For some reason, when we first invoke the command to get the sender object, it sometimes returns `undefined` during the private extension's activation. And then in subsequent calls, it finally returns the actual object. I'm not 100% sure if this is a VS Code bug or something we're not doing right and I frankly couldn't figure it out. I'm just invoking the command a few times until we get something back. If calling the command fails or if the extension is in development or test mode, we use a no-op sender
2. Got rid of our original telemetry
3. Started using the new one everywhere. This allowed us to simplify the benchmark middleware considerably

### Automated Tests

Adapted the existing tests.


vscode/README.md
```diff
@@ -389,39 +389,47 @@ Please note that only Docker is officially supported as a backend by the Dev Con
...
-service if desired.
+The Ruby LSP does not collect any telemetry by default, but it supports hooking up to a private metrics service if
+desired. This can be useful if you'd like to understand adoption, performance or catch errors of the Ruby LSP within
+your team or company.
 
-In order to receive metrics requests, a private plugin must export the `ruby-lsp.getPrivateTelemetryApi` command, which
-should return an object that implements the `TelemetryApi` interface defined
-[here](https://github.com/Shopify/ruby-lsp/blob/main/vscode/src/telemetry.ts).
+To collect metrics, another VS Code extension (typically a private one) should define the command
+`getTelemetrySenderObject`. This command should return an object that implements the
```
andyw8, line 397: (nit) we control this name, right? Could it just be `getTelemetrySender`?

vinistock: That would be a breaking change in our own private extension which already uses `getTelemetrySenderObject`. Unless you feel really strongly about this, I'd prefer to avoid changing it.

andyw8: Sure.

PR #2270, opened 2024-07-08
Require space for class keyword match

### LINKED ISSUE #2269: Ruby Syntax Highlighting "class" where not actually a class i.e. "classification"
### Description
When using the latest version of VSCode and the RubyLSP, words that contain `class` are highlighted like an actual `class`.

### Reproduction steps

Ruby LSP extension: v0.7.8
VSCode version: 
- 1.91.0 (Universal)
  Commit: ea1445cc7016315d0f5728f8e8b12a45dc0a7286
  Date: 2024-07-01T18:54:18.541Z (1 wk ago)
  Electron: 29.4.0
  ElectronBuildId: 9728852
  Chromium: 122.0.6261.156
  Node.js: 20.9.0
  V8: 12.2.281.27-electron.0
  OS: Darwin arm64 23.5.0

1. Open VSCode
2. Install the Ruby extension pack containing sorbet and the LSP
3. Open/create a new Ruby file
4. Paste the following code
5. See that words that contain `class` are highlighted weird
6. If I disable the LSP extension, the issue disappears (because nothing is highlighting the syntax).

```ruby
# frozen_string_literal: true

# A test class
class MyClass
  class_attribute :banana

  def my_function
    classification = 1

    classification + 1
  end
end
```

Screenshot 2024-07-08 at 1 27 09 PM

vinistock:
### Motivation

Closes #2269

In order for `class` to be a keyword, there needs to be a space between it and the constant name.

### Implementation

Switched to `\s+` rather than `\s*`.

### DIFF
--- a/vscode/grammars/ruby.cson.json
+++ b/vscode/grammars/ruby.cson.json
@@ -65,7 +65,7 @@
         }
       },
       "comment": "class Namespace::ClassName < OtherNamespace::OtherClassName",
-      "match": "(class)\\s*(([a-zA-Z0-9_]+)((::)[a-zA-Z0-9_]+)*)\\s*((<)\\s*(([a-zA-Z0-9_]+)((::)[a-zA-Z0-9_]+)*))?",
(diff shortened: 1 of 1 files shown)


st0012 (approved):
This was fast 👍

andyw8:
(although less likely to be encountered)

vinistock:
@andyw8 good point, we should fix that too https://github.com/Shopify/ruby-lsp/pull/2271.

PR #2272, opened 2024-07-09
Remove ? and ! from editor.wordSeparators

andyw8:
Closes https://github.com/Shopify/vscode-shopify-ruby/issues/422


andyw8:
Note that there is slight trade-off here: If you have a negation such as "!foo", selecting the token will include the leading `!`, which may not always be the intention.

andyw8:
Yeah it's unfortunate that VS Code doesn't allow for more fine-grained control over this.

PR #2275, opened 2024-07-09
Ensure test group exists before trying to add examples

vinistock:
### Motivation

This fixes another error found in our telemetry. We are sometimes getting test example code lens that are associated to a `group_id` for a group that doesn't exist.

It's difficult to figure out exactly in what scenario this happens, but given that addons can contribute test code lens, we should protect the extension from invalid data anyway.

### Implementation

We still had code in the test controller to account for the previous behaviour of code lens, which didn't include group IDs. It's been a long time since we released that, so I think it's safe to remove at this point. If anyone reports issues related to very old server versions, we can just ask them to update.

Other than simplifying the code after removing that old code lens handling, the fix is simply the check that the group exists before trying to add children. Otherwise, we print an error to the output channel.

### Automated Tests

Created a test for the test controller, which was missing.

### DIFF
--- a/vscode/src/test/suite/testController.test.ts
+++ b/vscode/src/test/suite/testController.test.ts
@@ -1,0 +1,61 @@
+import * as assert from "assert";
+
+import * as vscode from "vscode";
+import { CodeLens } from "vscode-languageclient/node";
+
+import { TestController } from "../../testController";
+import { Command } from "../../common";
+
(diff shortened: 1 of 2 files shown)


st0012 (approved):
Tested against `ruby-lsp-rspec`-using projects and the tree displays the same AFAICT 👍

PR #2283, opened 2024-07-10
Eagerly activate workspaces for opened files

vinistock:
### Motivation

Fixes the issue reported in https://github.com/Shopify/ruby-lsp/issues/1897#issuecomment-2219056807

In a multi-root workspace configuration, if you have a Ruby file opened for a secondary workspace, we don't eagerly activate that workspace, despite the file already being opened. You need to switch to another Ruby file for that workspace to have the language server activate, which is a bit odd.

### Implementation

Started eagerly activating the workspace for the currently opened file if it's a Ruby file and if the workspace associated with it is not the same as the first one (to avoid any duplicate activations).

### DIFF
--- a/vscode/src/rubyLsp.ts
+++ b/vscode/src/rubyLsp.ts
@@ -3,7 +3,12 @@ import { Range } from "vscode-languageclient/node";
 
 import DocumentProvider from "./documentProvider";
 import { Workspace } from "./workspace";
-import { Command, LOG_CHANNEL, STATUS_EMITTER } from "./common";
+import {
+  Command,
+  LOG_CHANNEL,
+  STATUS_EMITTER,
+  SUPPORTED_LANGUAGE_IDS,
+} from "./common";
 import { ManagerIdentifier, ManagerConfiguration } from "./ruby";
 import { StatusItems } from "./status";
 import { TestController } from "./testController";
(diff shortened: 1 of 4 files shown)


andyw8 (approved):

vscode/src/rubyLsp.ts
```diff
@@ -98,6 +98,20 @@ export class RubyLsp {
       await this.activateWorkspace(firstWorkspace, true);
     }
 
+    // If the user has the editor already opened on a Ruby file and that file does not belong to the first workspace,
+    // eagerly activate the workspace for that file too
+    const activeDocument = vscode.window.activeTextEditor?.document;
+
+    if (activeDocument && activeDocument.languageId === "ruby") {
```
st0012, line 105: Does it make sense to include `erb` as well?

st0012 (approved):
Just a question 👍

vscode/src/rubyLsp.ts
vinistock: It does make sense and now that you mentioned it, I realized we needed a better way of sharing which languages are supported for multiple places in the extension.

The latest commit creates a `SUPPORT_LANGUAGE_IDS` constant, that we then use to register for lazy activation, for the LSP client and for the language status item.

vscode/src/status.ts
```diff
@@ -17,10 +22,11 @@ export abstract class StatusItem {
   public item: vscode.LanguageStatusItem;
 
   constructor(id: string) {
-    this.item = vscode.languages.createLanguageStatusItem(id, {
-      scheme: "file",
-      language: "ruby",
-    });
+    const documentSelector = SUPPORTED_LANGUAGE_IDS.map((language) => ({
```
st0012, line 25: Should this be plural? `documentSelectors`?

vinistock: I matched the type name, which is `DocumentSelector` and is an array of entries.

PR #2289, opened 2024-07-11
Add extension and server version to telemetry

vinistock:
### Motivation

A couple more adjustments to our telemetry. It's useful to have the extension and server versions as part of errors, so that we can more easily tell if something was already fixed.

### Implementation

Add extension version to the common properties of the logger (included for all metrics). Also moved `environment` there.

Started including server version for the errors.

### DIFF
--- a/vscode/src/client.ts
+++ b/vscode/src/client.ts
@@ -360,7 +360,7 @@ export default class Client extends LanguageClient implements ClientInterface {
             name: error.data.errorClass,
             stack,
           },
-          { ...error.data },
+          { ...error.data, serverVersion: this.serverVersion },
         );
       }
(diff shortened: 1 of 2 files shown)


vscode/src/client.ts
```diff
@@ -360,7 +360,7 @@ export default class Client extends LanguageClient implements ClientInterface {
             name: error.data.errorClass,
             stack,
           },
-          { ...error.data },
+          { ...error.data, serverVersion: this.serverVersion },
```
andyw8, line 363: I can't remember if we report telemetry events while working on ruby-lsp itself? If so, should we have some way to distinguish these?

PR #2310, opened 2024-07-15
Unify ruby-lsp-doctor with ruby-lsp command

andyw8:
### Motivation

Improve discoverability of the troubleshooting tool.

### Implementation

I chose not to deprecate rather than fully remove since we have a number of issues mentioning the tool. We can revisit in future.

### Automated Tests

None

### Manual Tests

* Run `./exe/ruby-lsp -d` and verify the checks are performed.
* Run `./exe/ruby-lsp-doctor` and verify it prints a deprecation notice.


exe/ruby-lsp
```diff
@@ -36,6 +36,10 @@ parser = OptionParser.new do |opts|
     options[:experimental] = true
   end
 
+  opts.on("--doctor", "Run troubleshooting steps") do
```
andyw8, line 39: I didn't add a `-d` short prefix since it's not something that would be run often, and also to prevent unwanted clashes with other options we may want to add in future.

exe/ruby-lsp-doctor
```diff
@@ -1,23 +1,5 @@
...
-
-index = RubyIndexer::Index.new
-
-puts "Globbing for indexable files"
-
-RubyIndexer.configuration.indexables.each do |indexable|
-  puts "indexing: #{indexable.full_path}"
-  index.index_single(indexable)
-end
+puts "The ruby-lsp-doctor command is deprecated. Please run `ruby-lsp --doctor` instead."
```
vinistock, line 4: I vote for getting rid of this file. It's just used for troubleshooting anyway, it's not going to be super disruptive for users.

Let's just ensure it's not referenced in any docs or instructions.

vinistock (approved):

exe/ruby-lsp-doctor
andyw8: Any strong opinion @st0012 ?

st0012: Yeah let's remove it 👍 Remember to update `gemspec` too.

PR #2311, opened 2024-07-15
Allow turning off ERB support through setting

### LINKED ISSUE #2282: VSCode: ERB file false positive problems
### Description
The latest Ruby LSP VSCode extension (v0.7.11) generates numerous code problems for ERB files which should not be indicated as problems. The issue does not occur with v0.7.10. I've attached two screenshots - one of the code:
code

and another of the 'Problems' output:
problems

I think I narrowed it down to it only occurring when the `solargraph` gem is installed from my `Gemfile`, but the odd behavior occurs whether the VSCode "castwide.solargraph" extension is enabled or not (even occurs with the `solargraph` gem present when the "castwide.solargraph" extension has been fully uninstalled).

### To Reproduce
1. Create a new empty directory with the following files:
   * `.ruby-version` containing an rbenv ruby version (3.2.2 in my case)
   * `Gemfile` containing only `gem "solargraph", require: false`
   * A dummy `test.html.erb` file with any valid HTML
2. run `bundle install` and then open VSCode in that directory
3. Notice the problems in the `test.html.erb` file once the LSP extension starts the language server

vinistock:
### Motivation

Closes #2282

If the user is working on a project with a dependency that prevents the Ruby LSP server from upgrading, the server will not have ERB support.

Unfortunately, mutating the document selector after the client has already launched doesn't stop it from receiving requests for the removed language IDs. And we have no way of knowing if the server version supports ERB before actually launching. It's a chicken and egg problem.

### Implementation

I believe the best we can do is the following (implemented in this PR):

1. After launching the server, we check if there's ERB support. If there is, then the server is up to date and we don't have to do anything
2. Otherwise, we print a warning to output tab (should it be a dialog?) and then we restart the server, passing a new option to turn off ERB support before launch (which then has the desired effect of preventing requests for ERB files)

### Concerns

Should we even do this? This means that anyone on old server versions will always need to boot the server twice until they manage to upgrade to newer versions.

### DIFF
--- a/vscode/src/client.ts
+++ b/vscode/src/client.ts
@@ -21,6 +21,7 @@ import {
   ErrorAction,
   CloseAction,
   State,
+  DocumentFilter,
 } from "vscode-languageclient/node";
 
 import {
@@ -119,7 +120,7 @@ function collectClientOptions(
   const enabledFeatures = Object.keys(features).filter((key) => features[key]);
 
   const fsPath = workspaceFolder.uri.fsPath.replace(/\/$/, "");
-  const documentSelector: DocumentSelector = SUPPORTED_LANGUAGE_IDS.map(
+  let documentSelector: DocumentSelector = SUPPORTED_LANGUAGE_IDS.map(
     (language) => {
       return { language, pattern: `${fsPath}/**/*` };
     },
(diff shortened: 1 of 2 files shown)


st0012:
> should it be a dialog?

Yeah I think it should be dialog

> Should we even do this?

If the user still can not upgrade the server after seeing the warning, to revert the erb file behaviour the only alternative would be to downgrade the extension version? If that's the case, then I think restarting the server twice seems more acceptable.

vinistock:
I really dislike the double-start on launch, so I want to propose that we expose a setting to turn off ERB support as an escape hatch for users that cannot upgrade the server. After a few months, we can get rid of the setting and enable ERB support generally.

vscode/src/client.ts
```diff
@@ -158,6 +159,12 @@ function collectClientOptions(
     });
   }
 
+  if (!configuration.get<boolean>("erbSupport")) {
```
st0012, line 162: Can we add a comment here to document it's a temporary solution?

st0012 (approved):

PR #2317, opened 2024-07-16
Add keywords to completion candidates

### LINKED ISSUE #2304: Completion should avoid fuzzy searching methods when the typed characters match keywords
### Reproduction steps

1. In a project that does NOT use Sorbet, create a `test.rb` file
2. Type `foo do`
3. The Ruby LSP would try to suggest `define_singleton_method` when `do` is typed, like:

    Screenshot 2024-07-14 at 22 39 55

I consider this a bug because it means users will not be able to type `do` and then immediately type newline, which is very interruptive.

st0012:
### Motivation

Closes #2304

When typing `foo d`, VS Code would send a completion request for `d`. And then when `foo do` is typed, it would do fuzzy search on the previous completion candidates. So currently, the completion candidates for `foo do` would be `define_singleton_method` as demonstrated in #2304.

Sorbet LSP solved this issue by adding keywords to the completion candidates. And this PR does the same.

### Implementation

1. Bump Sorbet so array constants don't require `T.let` anymore
2. Add `KEYWORDS` constant (from [Prism's Ripper translation layer](https://github.com/ruby/prism/blob/main/lib/prism/translation/ripper.rb)) 
3. Add `add_keyword_completions`, which adds keywords to candidates list based on prefix match

### Automated Tests

Added a new test case for keywords completion and updated a few existing completion tests too.


lib/ruby_lsp/listeners/completion.rb
```diff
@@ -7,6 +7,50 @@ class Completion
...
+        "undef",
+        "unless",
+        "until",
+        "when",
+        "while",
+        "yield",
+        "__ENCODING__",
+        "__FILE__",
+        "__LINE__",
+      ]
```
vinistock, line 52:
```suggestion
      ].freeze
```

vinistock (commented):
Let's also remove the keywords from our [snippets](https://github.com/Shopify/ruby-lsp/blob/main/vscode/snippets.json), so that they don't appear duplicated.

Completions can also return snippets and the advantage is that it will work on any editor, so we can start with simple completion for all keywords and gradually improve with more specific context about each one of them.

vinistock (approved):

PR #2319, opened 2024-07-17
Document new indexing configuration approach

andyw8:
### Motivation

Document the changes in https://github.com/Shopify/ruby-lsp/pull/2300

### DIFF
--- a/README.md
+++ b/README.md
@@ -78,7 +78,12 @@ default gems, except for
 - Gems that only appear under the `:development` group
 - All Ruby files under `test/**/*.rb`
 
-By creating a `.index.yml` file, these configurations can be overridden and tuned. Note that indexing dependent behavior, such as definition, hover, completion or workspace symbol will be impacted by the configurations placed here.
+This behaviour can be overridden and tuned. Learn how to configure it [for VS Code](vscode/README.md#Indexing-Configuration) or [for other editors](EDITORS.md#Indexing-Configuration).
+
+Note that indexing-dependent behavior, such as definition, hover, completion or workspace symbol will be impacted by
+the configuration changes.
+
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -78,7 +78,22 @@ default gems, except for
 - Gems that only appear under the `:development` group
 - All Ruby files under `test/**/*.rb`
 
-By creating a `.index.yml` file, these configurations can be overridden and tuned. Note that indexing dependent behavior, such as definition, hover, completion or workspace symbol will be impacted by the configurations placed here.
+To change this behaviour, you can pass configuration as initalization options, for example:
+
+```jsonc
+// PROJECT/.vscode/settings.json
+{
+  "rubyLsp.excludedPatterns": ["**/test/**.rb"],
```
andyw8, line 86: @vinistock is this syntax correct? I tried to test it by excluding `lib/ruby_lsp/parameter_scope.rb` then using Go To Definition on `ParameterScope`, but it was still able to locate it.

andyw8: Hmm, no, same issue.

I'm unclear on the intended syntax, since we call into `configuration.get("indexing"),` but in `lsp.code-workspace` there's no `indexing` key.

vinistock: It should be like this (and we need to fix the [workspace file](https://github.com/Shopify/ruby-lsp/blob/085e6772162a3714e6b01847178ca181f2c1e065/lsp.code-workspace#L21)):

```json
{
  "rubyLsp.indexing": {
    "excludedPatterns": ["**/test/**.rb"]
  }
}
```

Notice that in the workspace configuration, you need to apply these in `lsp.code-workspace`. I'm not sure `.vscode/settings.json` will work.

vinistock (approved):

Issue #2294, opened 2024-07-11
Conditionally provide hints/definitions on typed files.


Jesse-Cameron:
### I have checked that this feature is not already implemented

- [X] This feature does not exist

### Use case

Our ruby codebase uses sorbet types in some of it'sfiles (lots of `typed: false` still hanging around). So devs like to have both ruby-lsp and the sorbet lsp running, one for the typed files and one for untyped files.

This works well for the untyped files as sorbet tends not display signature helpers/definitions etc. But once we are in the realm of `typed: true`, both ruby-lsp and sorbet will try do similar things. So sometimes both of the LSP's will provide similar information/details.

### Description

I think that it would be great if the ruby lsp were to integrate a little tighter with the sorbet LSP. We could give users control of which typed files are worth displaying hints on. I'm thinking a setting that allows users to toggle whether we want ruby-lsp to be suggesting on `typed: {true|strict}` files.

vinistock:
Thanks for opening the issue!

Before we can have deeper integration with Sorbet, there are already some things we can do:

1. I believe Sorbet is always able to provide completion, go to definition and hover for methods that are invoked on a class or on `self`. We should double-check if this is the case and then there's no need to offer these for declarations made inside the workspace on `typed: true` files or higher
2. We should not show any instance variable functionality on `typed: strict` files, since Sorbet enforces that every variable must be annotated

PR #2322, opened 2024-07-18
Handle Sorbet levels to prevent some feature duplication

vinistock:
### Motivation

This PR is a step towards #2294

We added a bunch of new features recently, but we weren't super careful about checking which parts overlapped with Sorbet's behaviour and under which scenarios. This resulted in a bit of duplicate behaviour and we can improve that.

Note that full removal of any kind of duplication is going to require deep integration with the type checker and is a much more involved effort.

### Implementation

Essentially, requests have to know the level of Sorbet type checking available for a file to make decisions about what to provide.

Instead of trying to use a blanket statement like `typechecker_enabled?`, I switched to passing the Sorbet level of type checking to the relevant requests.

I left some comments on the PR to help understand the decisions.

### Automated Tests

Added tests.

### DIFF
--- a/lib/ruby_lsp/document.rb
+++ b/lib/ruby_lsp/document.rb
@@ -10,6 +10,16 @@ module RubyLsp
       end
     end
 
+    class SorbetLevel < T::Enum
+      enums do
+        None = new("none")
+        Ignore = new("ignore")
+        False = new("false")
+        True = new("true")
+        Strict = new("strict")
+      end
+    end
+
     extend T::Sig
     extend T::Helpers
 
@@ -213,10 +223,23 @@ module RubyLsp
       NodeContext.new(closest, parent, nesting_nodes, call_node)
     end
 
-    sig { returns(T::Boolean) }
-    def sorbet_sigil_is_true_or_higher
-      parse_result.magic_comments.any? do |comment|
-        comment.key == "typed" && ["true", "strict", "strong"].include?(comment.value)
+    sig { returns(SorbetLevel) }
+    def sorbet_level
+      sigil = parse_result.magic_comments.find do |comment|
+        comment.key == "typed"
+      end&.value
+
+      case sigil
+      when "ignore"
+        SorbetLevel::Ignore
+      when "false"
+        SorbetLevel::False
+      when "true"
+        SorbetLevel::True
+      when "strict", "strong"
+        SorbetLevel::Strict
+      else
+        SorbetLevel::None
       end
     end
 
--- a/lib/ruby_lsp/listeners/completion.rb
+++ b/lib/ruby_lsp/listeners/completion.rb
@@ -56,7 +56,7 @@ module RubyLsp
           response_builder: ResponseBuilders::CollectionResponseBuilder[Interface::CompletionItem],
           global_state: GlobalState,
           node_context: NodeContext,
-          typechecker_enabled: T::Boolean,
+          sorbet_level: Document::SorbetLevel,
           dispatcher: Prism::Dispatcher,
           uri: URI::Generic,
           trigger_character: T.nilable(String),
@@ -66,7 +66,7 @@ module RubyLsp
         response_builder,
         global_state,
         node_context,
-        typechecker_enabled,
+        sorbet_level,
         dispatcher,
         uri,
         trigger_character
@@ -76,7 +76,7 @@ module RubyLsp
         @index = T.let(global_state.index, RubyIndexer::Index)
         @type_inferrer = T.let(global_state.type_inferrer, TypeInferrer)
         @node_context = node_context
-        @typechecker_enabled = typechecker_enabled
+        @sorbet_level = sorbet_level
         @uri = uri
         @trigger_character = trigger_character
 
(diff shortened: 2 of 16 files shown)


lib/ruby_lsp/listeners/completion.rb
```diff
@@ -322,11 +324,13 @@ def complete_require_relative(node)
       sig { params(node: Prism::CallNode, name: String).void }
       def complete_methods(node, name)
         # If the node has a receiver, then we don't need to provide local nor keyword completions
-        if !@global_state.has_type_checker && !node.receiver
+        if @sorbet_level == Document::SorbetLevel::Disabled && !node.receiver
           add_local_completions(node, name)
           add_keyword_completions(node, name)
         end
 
+        return if @sorbet_level != Document::SorbetLevel::Disabled && self_receiver?(node)
```
vinistock, line 332: In `typed: true` files or above, Sorbet can always provide completion for methods invoked directly on `self`, regardless of whether they are annotated or not.

lib/ruby_lsp/listeners/definition.rb
```diff
@@ -53,6 +53,8 @@ def initialize(response_builder, global_state, language_id, uri, node_context, d
 
       sig { params(node: Prism::CallNode).void }
       def on_call_node_enter(node)
+        return if @sorbet_level != Document::SorbetLevel::Disabled && self_receiver?(node)
```
vinistock, line 56: On `typed: true` and above, Sorbet properly handles methods invoked on `self`.

lib/ruby_lsp/listeners/definition.rb
```diff
@@ -149,6 +151,8 @@ def on_forwarding_super_node_enter(node)
 
       sig { void }
       def handle_super_node_definition
+        return unless @sorbet_level == Document::SorbetLevel::Disabled
```
vinistock, line 154: On `typed: true` and above, Sorbet properly handles `super`.

lib/ruby_lsp/listeners/hover.rb
```diff
@@ -180,6 +182,8 @@ def handle_method_hover(message, inherited_only: false)
 
       sig { params(name: String).void }
       def handle_instance_variable_hover(name)
+        return if @sorbet_level == Document::SorbetLevel::Strict
```
vinistock, line 185: On `typed: strict` and above, Sorbet enforces that all instance variables be declared and can therefore provide all features on them.

lib/ruby_lsp/document.rb
```diff
@@ -10,6 +10,14 @@ class LanguageId < T::Enum
       end
     end
 
+    class SorbetLevel < T::Enum
```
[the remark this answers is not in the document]
andyw8: What is we include all the levels but implement comparable? That would allow us to cleanly check the levels without having to consider any mapping.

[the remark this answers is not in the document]
andyw8: But wasn't that originally intended for Core? There could be case for disabling it for Ruby LSP.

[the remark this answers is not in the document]
vinistock: I would argue it's even more important for the LSP than for a web app. Our performance constraints are much more strict. Whatever few milliseconds we manage to shave off, the more responsive the editor becomes.

lib/ruby_lsp/document.rb
```diff
@@ -213,10 +221,19 @@ def locate(node, char_position, node_types: [])
...
+    sig { returns(SorbetLevel) }
+    def sorbet_level
+      sigil = parse_result.magic_comments.find do |comment|
+        comment.key == "typed"
+      end&.value
+
+      case sigil
+      when "true"
+        SorbetLevel::Enabled
+      when "strict", "strong"
```
st0012, line 233: Oh I didn't know there's even a `strong` level 😮

lib/ruby_lsp/listeners/completion.rb
```diff
@@ -128,28 +128,30 @@ def on_constant_path_node_enter(node)
...
-
-            constant_path_completion(
-              "#{name}::",
-              Interface::Range.new(
-                start: Interface::Position.new(line: start_loc.start_line - 1, character: start_loc.start_column),
-                end: Interface::Position.new(line: end_loc.end_line - 1, character: end_loc.end_column),
-              ),
-            )
-            return
+        unless @global_state.has_type_checker
```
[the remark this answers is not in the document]
st0012: I think we can check if `sorbet_level` is `Disabled` for this case?

Basically, `@global_state.has_type_checker` won't be needed anymore.

lib/ruby_lsp/listeners/completion.rb
st0012: Can the context be capture in a comment above this line?

lib/ruby_lsp/listeners/definition.rb
```diff
@@ -161,6 +165,8 @@ def handle_super_node_definition
 
       sig { params(name: String).void }
       def handle_instance_variable_definition(name)
+        return if @sorbet_level == Document::SorbetLevel::Strict
```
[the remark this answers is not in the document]
st0012: Let's add this as a comment.

lib/ruby_lsp/listeners/completion.rb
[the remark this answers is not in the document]
vinistock: We still need `@global_state.has_type_checker` to prevent registering for workspace symbols (since that's workspace wide and not document specific), but indeed we don't need `has_type_checker` for any text document requests anymore.

lib/ruby_lsp/document.rb
vinistock: It's quite challenging to use. It will throw error if there are any `T.untyped` bindings for the entire file. It's essentially a full typed mode, where escape hatches are not available.

vinistock:
I dug a little deeper and there were a few other scenarios I hadn't realized. For example, Sorbet can provide constant completion even if a file has no `typed` sigil. The only time it cannot provide it is on `typed: ignore`.

The PR should be good for another round of reviews.

st0012 (approved):

PR #2330, opened 2024-07-18
Add task to verify indexing against Top 100 RubyGems

andyw8:
### Motivation

To help catch issues related to indexing, we can expand our indexing testing to cover a set of RubyGems

As per @kddnewton's suggestion, let's begin with just the top 100. We can later consider a periodic job which runs against _every_ gem on RubyGems.org.

### Implementation

Shamelessly copied from https://github.com/ruby/prism/blob/main/rakelib/lex.rake

Questions:

* Do we want to run this as part of CI? Locally: `17.04s user 3.65s system 60% cpu 34.393 total`
* If so, do we care about caching the downloads between runs? (~120MB).

### Automated Tests

None, since this is a test itself.

### Manual Tests

Run with `bundle exec rake index:topgems`.


lsp.code-workspace
```diff
@@ -18,9 +18,11 @@
             "**/out": true,
         },
         "rubyLsp.bypassTypechecker": true,
-        "rubyLsp.excludedPatterns": [
```
st0012, line 21: I thought this has been updated in #2319?

st0012 (approved):
I say let's run it on CI so it'll be maintained (or deleted when not used). I'm fine with downloading things on build if it's not too flaky.

lsp.code-workspace
andyw8: Add, bad rebase, will fix.

sorbet/config
```diff
@@ -1,6 +1,7 @@
 --dir
 .
 --ignore=vendor/
+--ignore=tmp/
```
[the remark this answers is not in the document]
andyw8: I'd rather keep the ignore in case the indexing fails part way through, and leaves some files in there. The machine grinds to a halt if tries to start typechecking those gems directories.

lib/tasks/index.rake
vinistock, line 1: I think we need to remove this file.

.github/workflows/ci.yml
```diff
@@ -117,3 +117,15 @@ jobs:
...
+    steps:
+      - uses: actions/checkout@v4
+
+      - name: Set up Ruby
+        uses: ruby/setup-ruby@v1
+        with:
+          bundler-cache: true
+
+      - name: Index Top 100 Ruby gems
+        run: bundle exec rake index:topgems
```
vinistock, line 131: This should only run if the lockfile or the `ruby_indexer` dir have been modified. There's no need to run in other PRs.

lib/tasks/top_100_gems.yml
```diff
@@ -0,0 +1,101 @@
+---
+- actioncable-7.0.4.3
```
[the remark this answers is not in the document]
andyw8: Yeah, I think we can consider them as fixtures.

Also, if we follow Prism's approach, then we can later expand this to checking all gems, for the latest version.

.github/workflows/ci.yml
```diff
@@ -117,3 +117,15 @@ jobs:
           yarn --frozen-lockfile
           /usr/bin/Xvfb :99 -screen 0 1024x768x24 > /dev/null 2>&1 &
           yarn run test
+  verify_indexing:
```
st0012, line 120: nit: I don't think we should use `verify` when we actually don't check the result's correctness. Maybe something like `indexing_sanity_check` will be closer to what it does?

vinistock (commented):
If indexing succeeds, this is fine. However, if it fails, we will keep the copies of every gem still installed, even if only a single file out of the 100 gems have failed - which is why we need to ignore the `tmp` directory for Sorbet.

You also get prints for every single file path, even the ones that succeed, which is not useful information for someone running the script.

Finally, if there is more than one error, you need to re-run the script because it will stop at the first failure.

How about we do something like this instead?

```ruby
errors = Dir[File.join(directory, "**", "*.rb")].filter_map do |filepath|
  print "."
  code = File.read(filepath)
  index.index_single(RubyIndexer::IndexablePath.new(nil, filepath), code)
  nil
rescue => e
  { message: e.message, file: filepath }
end

# => Copy only the files with errors to some directory 
# => group errors by filepath and then print them
# => include the error message as part of what is printed
```

.github/workflows/ci.yml
vinistock: Yes! Let's do that an run the action on any commit or PR that modifies the lockfile or ruby_indexer.

.github/workflows/indexing.yml
```diff
@@ -0,0 +1,25 @@
...
+
+on:
+  push:
+    paths:
+      - 'Gemfile.lock'
+      - 'lib/ruby_indexer/**'
+  pull_request:
+    paths:
+      - 'Gemfile.lock'
+      - 'lib/ruby_indexer/**'
```
andyw8, line 11: (it does not actually run on this PR since neither of these are changed).

PR #2331, opened 2024-07-18
Rescue `NonExistingDocumentError` in base server

vinistock:
### Motivation

In #2313, I forgot that we need to rescue the new `NonExistingDocumentError` in base server, otherwise the TypeScript client tests will fail when VS Code opens some fake files automatically.

### Implementation

Added the missing rescue.


st0012 (approved):

st0012:
Should there be a test case for this though? Otherwise it'll be hard to tell if/when we'll need this in the future.

andyw8 (approved):

PR #2336, opened 2024-07-19
Unify class and module handling for RBS

andyw8:
### Motivation

@vinistock pointed out that a lot of this behaviour is shared between the class and module handling.

### Implementation

Unify

### Automated Tests

No need to update.

### Manual Tests

n/a


lib/ruby_indexer/lib/ruby_indexer/rbs_indexer.rb
```diff
@@ -37,45 +37,33 @@ def process_signature(pathname, declarations)
     sig { params(declaration: RBS::AST::Declarations::Base, pathname: Pathname).void }
     def process_declaration(declaration, pathname)
       case declaration
-      when RBS::AST::Declarations::Class
-        handle_class_declaration(declaration, pathname)
-      when RBS::AST::Declarations::Module
-        handle_module_declaration(declaration, pathname)
+      when RBS::AST::Declarations::Class, RBS::AST::Declarations::Module
```
andyw8, line 40: The parent of these is `RBS::AST::Declarations::Base` but it includes things we don't want (or not yet), such as `Interface` and `TypeAlias`.

vinistock (approved):
Perfect! Thank you

st0012 (approved):

Issue #2340, opened 2024-07-22
Fetch index comments lazily


vinistock:
Currently, we are indexing all comments for every entry eagerly. This is wasteful both from a CPU and memory standpoint since we spend more time indexing and then store a bunch of documentation that the user may actually never interact with. I believe we can get some pretty interesting gains if we lazily collect comments for entries.

For this to work, we first need to ensure that we can index files on changes (rather than on save), so that we can show the correct comments as the user is typing.

Then we can stop eagerly indexing comments and start gathering them lazily, memoizing the results.

PR #2341, opened 2024-07-22
Switch to storing URIs in entries instead of file paths

### LINKED ISSUE #1908: Completions are unavailable until file is saved
Currently we only index when a file is saved, meaning that completion will be unavailable for newly-added symbols within the same file. This could be unexpected, especially if people are coming from other editors which behave differently.

A couple of options that come to mind:

- Index the current file on changes
- Index the current file on-the-fly when the autocompletion is triggered

We should also consider the case of multiple unsaved files, with one referencing the other.

vinistock:
### Motivation

First step towards #2340 and #1908

Currently, we save the file paths for entries in the index. This decision is incompatible with how language servers work, because you can have files that exist in different URI schemes (such as the one for unsaved files like `untitled:Untitled-1`).

In general, we should always strive to use URIs everywhere and only get the file paths when we want to return something or when we're reading from disk.

This PR changes our entries to store the URIs where they were declared, rather than tying ourselves with file paths. This change is necessary in order to fix #1908, because if we index files before they are saved, there's no file path associated to the declarations being made in the unsaved file.

### Implementation

The change is essentially switching from `file_path` to URI pretty much everywhere.

As an added benefit, we now no longer need to build the URI when serving features, because that's already available directly in the entries - which is another indication that standardizing on URIs is likely the best choice.

### Automated Tests

Adapted existing tests.

### DIFF
--- a/lib/ruby_indexer/lib/ruby_indexer/index.rb
+++ b/lib/ruby_indexer/lib/ruby_indexer/index.rb
@@ -25,10 +25,11 @@ module RubyIndexer
 
       # Holds references to where entries where discovered so that we can easily delete them
       # {
-      #  "/my/project/foo.rb" => [#<Entry::Class>, #<Entry::Class>],
-      #  "/my/project/bar.rb" => [#<Entry::Class>],
+      #  "file:///my/project/foo.rb" => [#<Entry::Class>, #<Entry::Class>],
+      #  "file:///my/project/bar.rb" => [#<Entry::Class>],
+      #  "untitled:Untitled-1" => [#<Entry::Class>],
       # }
-      @files_to_entries = T.let({}, T::Hash[String, T::Array[Entry]])
+      @uris_to_entries = T.let({}, T::Hash[String, T::Array[Entry]])
 
       # Holds all require paths for every indexed item so that we can provide autocomplete for requires
       @require_paths_tree = T.let(PrefixTree[IndexablePath].new, PrefixTree[IndexablePath])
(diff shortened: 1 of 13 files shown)


st0012 (approved):

lib/ruby_indexer/lib/ruby_indexer/indexable_path.rb
```diff
@@ -2,6 +2,8 @@
 # frozen_string_literal: true
 
 module RubyIndexer
+  # Represents a file system path that can be indexed. This class should only be used for files in the file system and
+  # not for other URIs that may be indexed, such as unsaved file URIs using the `untitled` scheme
```
andyw8, line 6: Can we not enforce this by using `URI::File` instead of `URI::Generic`?

lib/ruby_indexer/lib/ruby_indexer/index.rb
```diff
@@ -25,10 +25,11 @@ def initialize
...
       # Holds references to where entries where discovered so that we can easily delete them
       # {
-      #  "/my/project/foo.rb" => [#<Entry::Class>, #<Entry::Class>],
-      #  "/my/project/bar.rb" => [#<Entry::Class>],
+      #  "file:///my/project/foo.rb" => [#<Entry::Class>, #<Entry::Class>],
+      #  "file:///my/project/bar.rb" => [#<Entry::Class>],
+      #  "untitled:Untitled-1" => [#<Entry::Class>],
       # }
-      @files_to_entries = T.let({}, T::Hash[String, T::Array[Entry]])
+      @uris_to_entries = T.let({}, T::Hash[String, T::Array[Entry]])
```
andyw8, line 32: Should we change this to use the URI as the key?

andyw8 (approved):

vinistock:
Closing this for now. We need to re-think the approach.

PR #2342, opened 2024-07-22
Index files as they are modified

### LINKED ISSUE #1908: Completions are unavailable until file is saved

vinistock:
### Motivation

Closes #1908

Index files as they are modified rather than waiting until they are saved. This better reflects the state of the codebase as the user is coding.

### Implementation

The idea is the following:

1. If a document is changed, without being managed by the client (that is, the file was never opened in the editor and it's not present in our Store), nothing changes. We still index the changes based on their file path
2. If the document is already managed by the client, then we simply add the declaration listener as part of all of the others in the combined requests and re-index the file as we are providing other features

The only extra thing is that we need to be able to handle ancestor changes even if the user hasn't saved the file, so I created a way to track those.

### Automated Tests

Added some tests to ensure that we index declarations in unsaved files as they are being modified and also to ensure that we don't accidentally create duplicate entries, which is a risk if we accidentally forget to clean things up.

### Manual Tests

1. Create a new file in the UI without saving it
2. Change the language mode to Ruby
3. Add a class declaration
4. Verify you get completion for the newly created class

### DIFF
--- a/lib/ruby_indexer/lib/ruby_indexer/index.rb
+++ b/lib/ruby_indexer/lib/ruby_indexer/index.rb
@@ -469,16 +469,23 @@ module RubyIndexer
 
     # Synchronizes a change made to the given indexable path. This method will ensure that new declarations are indexed,
     # removed declarations removed and that the ancestor linearization cache is cleared if necessary
-    sig { params(indexable: IndexablePath).void }
-    def handle_change(indexable)
-      uri = indexable.to_uri.to_s
+    sig { params(uri: URI::Generic, indexable: IndexablePath).void }
+    def handle_change(uri, indexable)
+      handling_ancestor_change(uri) do
+        delete(indexable)
+        index_single(indexable)
+      end
+    end
+
+    # Handle ancestor changes that may occur when invoking the given block. The goal of this method is to automatically
+    # synchronize ancestors when handling changes in the index
+    sig { params(indexable_or_uri: T.any(IndexablePath, URI::Generic), block: T.proc.void).void }
+    def handling_ancestor_change(indexable_or_uri, &block)
+      uri = (indexable_or_uri.is_a?(IndexablePath) ? indexable_or_uri.to_uri : indexable_or_uri).to_s
+
       original_entries = @uris_to_entries[uri]
-
-      delete(indexable)
-      index_single(indexable)
-
+      block.call
       updated_entries = @uris_to_entries[uri]
-
       return unless original_entries && updated_entries
 
       # A change in one ancestor may impact several different others, which could be including that ancestor through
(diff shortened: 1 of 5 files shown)


lib/ruby_indexer/lib/ruby_indexer/index.rb
```diff
@@ -469,16 +469,23 @@ def instance_variable_completion_candidates(name, owner_name)
 
     # Synchronizes a change made to the given indexable path. This method will ensure that new declarations are indexed,
     # removed declarations removed and that the ancestor linearization cache is cleared if necessary
-    sig { params(indexable: IndexablePath).void }
-    def handle_change(indexable)
-      uri = indexable.to_uri.to_s
-      original_entries = @uris_to_entries[uri]
+    sig { params(uri: URI::Generic, indexable: IndexablePath).void }
+    def handle_change(uri, indexable)
```
st0012, line 473: Are there cases where the `uri` would be different than `indexable.to_uri`? If there are, can we capture it in test cases?

andyw8:
Do we need to consider the case where someone saves a file outside of the current workspace?

lib/ruby_indexer/lib/ruby_indexer/index.rb
andyw8: Also perhaps we can swap the arguments order, and have `uri` default to `indexable.uri`?

vinistock: Yeah, I think we need to change `IndexablePath` to be simply an `Indexable` and then include the URI there. I was just trying to avoid parsing the same URI twice, but the code is ugly.

vinistock:
> Do we need to consider the case where someone saves a file outside of the current workspace?

I don't believe so. If the document is not opened in the editor and it's outside of the current workspace, how would we know that it is relevant for the current project? Or in other words, how could we know which files to watch (we wouldn't want to watch all Ruby files in the computer)?

**Note**

I'm afraid that this PR is going to degrade performance very significantly. I'm going to re-think this and #2341 to ensure that we're not making the experience worse.

andyw8:
Here's a reproduction of what I'm thinking of:

* Create new unsaved file (VS Code will recognize it as Ruby):
```
# typed: strict
# frozen_string_literal: true

module RubyLsp
  class Foo
  end
end
```
* Save it outside of workspace, e.g. to the Desktop as `foo.rb`
* Observe that completion is offered for `RubyLsp::Foo` from `foo.rb`

andyw8 (changes requested):
(I know changes are planned, just marking as reviewed so GitHub stops ping me).

PR #2344, opened 2024-07-23
Update LazyVim docs

andyw8:


ianschmitz (approved):

EDITORS.md
```diff
@@ -142,29 +142,7 @@ require("lspconfig").ruby_lsp.setup({
...
-          autostart = false
-        },
-        -- ruby_lsp will be automatically installed with mason and loaded with lspconfig
-        ruby_lsp = {},
-      },
-    },
-  },
-}
-```
+[As of v12.33.0](https://github.com/LazyVim/LazyVim/pull/3652), Ruby LSP is the default LSP for Ruby and no configuration should be needed.
```
vinistock, line 145: Does it make sense to keep the previous instructions for people on older versions? Or is it common practice to be on the latest one for LazyVim?

vinistock (approved):

EDITORS.md
andyw8: It seems like [an easy update](https://github.com/LazyVim/LazyVim/issues/1563#issuecomment-1744228821):

> `LazyVim` itself is the same as a regular plugin, so there is no need for a special update command. You update LAzyVim like you update any other plugin using `:Lazy`, the plugin mamager.

andyw8: Let's see if people object to upgrading, if so we can re-add the old information.

PR #2358, opened 2024-07-24
Allow for indexing enhancements through included hooks

vinistock:
### Motivation

This PR is the first step for allowing addons to enhance the indexing mechanism and a proposal of how we could handle included hooks.

Example usages https://github.com/Shopify/ruby-lsp-rails/pull/422, https://github.com/Shopify/ruby-lsp-rails/pull/423

**What does a concern do?**

For context, in order to make a module into a concern one has to `extend ActiveSupport::Concern`. Upon extending, the concern module will then dynamically define an `included` hook into the module that has extended it. Finally, the included hook will dynamically extend any `ClassMethods` module that exists inside of the original module.

```ruby
module Concern
  def self.extended(mod)
    mod.define_singleton_method(:included) do |other_module|
      other_module.extend(mod.const_get(:ClassMethods))
    end
  end
end
```

So we need a way to
- Handle the extend event, so that we can remember that a module has been made a concern
- Provide a way to run included callbacks during ancestor linearization, so that the ancestor chain can be modified dynamically

**Other use cases**

I have also considered other use cases, like creating a new instance method during an `extend` or handling something like `belongs_to` (which creates new methods on the receiver).

```ruby
module Concern
  def self.extended(base)
    base.define_method(:new_method) {}
  end
end
```

### Naming

I ended up going with "indexing augmentations" for the name of concept. The idea is to avoid using extension and plugin (too associated with editors) or addon (now associated with the language server).

Do we agree with the naming? Is indexing enhancement better? Indexing contribution?

### Implementation

The implementation has a few parts:

1. We need a way to register for augmentations and those have to be passed down to the declaration listener, so that we can invoke the events as needed
2. As with most things, we want to expose the minimum amount of methods to addons to help guide authors in their implementations. I created an interface that any object can implement, containing the only events we want to allow
3. Finally, the bulk of the logic and decisions are part of the included hooks. The augmentations will need to be able to register lazy included hooks, because during indexing we have no guarantees that we won't find an include for a concern that hasn't been found yet. We also have no guarantees that we will know which module an include refers to before we finished indexing. The idea is to allow registering blocks that get invoked once per linearization. These hooks should mutate the index with the extra information they want to include. The only time we run these blocks again is if the ancestor cache gets busted, which is exactly what we want otherwise we would lose the information forever

### Notes

I believe we do not have to support lazy extended, prepended or inherited hooks. The reason is because the concern pattern is broadly used to mix instance and class methods in a single operation. You include one module and then you get everything you need.

The opposite pattern of extending a module and then getting a dynamic include is not really common and would essentially just be the exact same thing as a concern.

And inheriting from a class already carries both instance and class methods, so there's also reduced need for meta-programming on that side.

We may discover other scenarios later, but I think it should be fine to handle included hooks only for now.

### Automated Tests

Added tests.

### Manual Tests

There's no way to test this without the changes being made to the Rails addon too. I did it locally and recorded it in action in a toy app.

https://github.com/user-attachments/assets/08255211-6bcf-4d21-b6f9-4dd264791ee8

### DIFF
--- a/lib/ruby_indexer/lib/ruby_indexer/declaration_listener.rb
+++ b/lib/ruby_indexer/lib/ruby_indexer/declaration_listener.rb
@@ -9,11 +9,18 @@ module RubyIndexer
     BASIC_OBJECT_NESTING = T.let(["BasicObject"].freeze, T::Array[String])
 
     sig do
-      params(index: Index, dispatcher: Prism::Dispatcher, parse_result: Prism::ParseResult, file_path: String).void
+      params(
+        index: Index,
+        dispatcher: Prism::Dispatcher,
+        parse_result: Prism::ParseResult,
+        file_path: String,
+        enhancements: T::Array[Enhancement],
+      ).void
     end
-    def initialize(index, dispatcher, parse_result, file_path)
+    def initialize(index, dispatcher, parse_result, file_path, enhancements: [])
       @index = index
       @file_path = file_path
+      @enhancements = enhancements
       @visibility_stack = T.let([Entry::Visibility::PUBLIC], T::Array[Entry::Visibility])
       @comments_by_line = T.let(
         parse_result.comments.to_h do |c|
@@ -279,6 +286,8 @@ module RubyIndexer
       when :private
         @visibility_stack.push(Entry::Visibility::PRIVATE)
       end
+
+      @enhancements.each { |aug| aug.on_call_node(@index, @owner_stack.last, node, @file_path) }
     end
 
     sig { params(node: Prism::CallNode).void }
--- a/lib/ruby_indexer/lib/ruby_indexer/declaration_listener.rb
+++ b/lib/ruby_indexer/lib/ruby_indexer/declaration_listener.rb
@@ -293,8 +293,8 @@ module RubyIndexer
 
       @enhancements.each do |enhancement|
         enhancement.on_call_node(@index, @owner_stack.last, node, @file_path)
-      rescue StandardError
-        @indexing_errors << "Error occurred when indexing #{@file_path} with '#{enhancement.class.name}' enhancement"
+      rescue StandardError => e
+        @indexing_errors << "Indexing error in #{@file_path} with '#{enhancement.class.name}' enhancement: #{e.message}"
       end
     end
 
--- a/lib/ruby_indexer/lib/ruby_indexer/declaration_listener.rb
+++ b/lib/ruby_indexer/lib/ruby_indexer/declaration_listener.rb
@@ -8,6 +8,9 @@ module RubyIndexer
     OBJECT_NESTING = T.let(["Object"].freeze, T::Array[String])
     BASIC_OBJECT_NESTING = T.let(["BasicObject"].freeze, T::Array[String])
 
+    sig { returns(T::Array[String]) }
+    attr_reader :indexing_errors
+
     sig do
       params(
         index: Index,
@@ -36,6 +39,7 @@ module RubyIndexer
 
       # A stack of namespace entries that represent where we currently are. Used to properly assign methods to an owner
       @owner_stack = T.let([], T::Array[Entry::Namespace])
+      @indexing_errors = T.let([], T::Array[String])
 
       dispatcher.register(
         self,
@@ -287,7 +291,11 @@ module RubyIndexer
         @visibility_stack.push(Entry::Visibility::PRIVATE)
       end
 
-      @enhancements.each { |aug| aug.on_call_node(@index, @owner_stack.last, node, @file_path) }
+      @enhancements.each do |enhancement|
+        enhancement.on_call_node(@index, @owner_stack.last, node, @file_path)
+      rescue StandardError
+        @indexing_errors << "Error occurred when indexing #{@file_path} with '#{enhancement.class.name}' enhancement"
+      end
     end
 
     sig { params(node: Prism::CallNode).void }
--- a/lib/ruby_indexer/lib/ruby_indexer/index.rb
+++ b/lib/ruby_indexer/lib/ruby_indexer/index.rb
@@ -35,6 +35,27 @@ module RubyIndexer
 
       # Holds the linearized ancestors list for every namespace
       @ancestors = T.let({}, T::Hash[String, T::Array[String]])
+
+      # List of classes that are enhancing the index
+      @enhancements = T.let([], T::Array[Enhancement])
+
+      # Map of module name to included hooks that have to be executed when we include the given module
+      @included_hooks = T.let(
+        {},
+        T::Hash[String, T::Array[T.proc.params(index: Index, base: Entry::Namespace).void]],
+      )
+    end
+
+    # Register an enhancement to the index. Enhancements must conform to the `Enhancement` interface
+    sig { params(enhancement: Enhancement).void }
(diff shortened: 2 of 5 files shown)


jacobdaddario:
Fantastic bit of technical writing as always. Excited to see this go.

> Do we agree with the naming? Is indexing enhancement better? Indexing contribution?

I think that one of the latter two choices may be better depending on the technique used. If customizations are adding modifying the indexing techniques themselves, then I think `enhancement` feels more appropriate. If the customizations are adding directly on to the `Index` instance, then `contribution` feels more appropriate. Based on your PR description, it sounds like the first option is closer to the case.

andyw8:
Another possible name - supplement. Here's what GPT says:

---

The terms "supplement" and "augmentation" both refer to the process of adding something, but they are used in slightly different contexts and have nuanced meanings.

### Supplement
- **Definition**: To supplement something means to add to it in order to complete it, enhance it, or make up for a deficiency.
- **Context**: Often used when referring to adding something that is lacking or to improve the overall quality or completeness of something.
- **Example**: Taking vitamin supplements to ensure you get all the necessary nutrients that might be missing from your diet.

### Augmentation
- **Definition**: To augment something means to increase its size, value, or effectiveness by adding something to it.
- **Context**: Typically used when the goal is to make something larger or more effective, rather than just completing it.
- **Example**: Augmenting a software application with new features to improve its functionality.

### Key Differences
- **Purpose**: Supplementing is often about filling gaps or enhancing completeness, while augmenting is about increasing or amplifying.
- **Usage**: Supplement is more commonly used in contexts like nutrition, education, and information. Augmentation is often used in contexts like technology, medical procedures, and performance enhancement.

In summary, while both terms involve adding something, "supplement" focuses on completion and enhancement, whereas "augmentation" focuses on increasing and amplifying.

lib/ruby_indexer/lib/ruby_indexer/index.rb
```diff
@@ -457,6 +478,12 @@ def linearized_ancestors_of(fully_qualified_name)
         end
       end
 
+      # We only need to run included hooks when linearizing singleton classes. Included hooks are typically used to add
+      # new singleton methods or to extend a module through an include. There's no need to support instance methods, the
+      # inclusion of another module or the prepending of another module, because those features are already a part of
+      # Ruby and can be used directly without any meta-programming
```
andyw8, line 484:
```suggestion
      # Ruby and can be used directly without any metaprogramming
```

lib/ruby_indexer/test/index_test.rb
```diff
@@ -1822,5 +1822,161 @@ class Child < Namespace::Parent
         @index.linearized_ancestors_of("Foo::Child::<Class:Child>"),
       )
     end
+
+    def test_augmenting_indexing_included_hook
```
andyw8, line 1826: How about we introduce a seperate tes file for things related to the augmentations?

lib/ruby_indexer/test/index_test.rb
```diff
@@ -1822,5 +1822,161 @@ class Child < Namespace::Parent
...
+      aug_class = Class.new do
+        include Augmentation
+
+        def on_call_node(index, owner, node, file_path)
+          return unless owner
+
+          if node.name == :extend
```
andyw8, line 1833: Could this be an early return?

st0012:
I'll test the API with the RSpec addon too.

lib/ruby_indexer/lib/ruby_indexer/augmentation.rb
```diff
@@ -0,0 +1,24 @@
+# typed: strict
+# frozen_string_literal: true
+
+module RubyIndexer
+  module Augmentation
```
st0012, line 5: Why do we use mixin here instead of inheritance, like `Addon`?

lib/ruby_indexer/lib/ruby_indexer/declaration_listener.rb
```diff
@@ -9,11 +9,18 @@ class DeclarationListener
...
+      params(
+        index: Index,
+        dispatcher: Prism::Dispatcher,
+        parse_result: Prism::ParseResult,
+        file_path: String,
+        augmentations: T::Array[Augmentation],
+      ).void
     end
-    def initialize(index, dispatcher, parse_result, file_path)
+    def initialize(index, dispatcher, parse_result, file_path, augmentations = [])
```
st0012, line 20: I think `augmentations` can be a keyword argument as I consider it a additional option.

lib/ruby_indexer/test/index_test.rb
vinistock: Good idea.

st0012 (commented):
For naming, it would be great to have a more approachable name. I think `supplement` is better than `augmentation`, but it's a bit rare to see it in software context too (from my limited experience). So `enhancement` will be my favorite.

lib/ruby_indexer/lib/ruby_indexer/augmentation.rb
vinistock: Because we're not providing any functionality to enhancements at the moment. We just need them to comply to a specific interface.

vinistock:
Switched the name to enhancements.

st0012:
I added basic error reporting around enhancement so when it encounters an error, the file's indexing can still be completed and we can get some messages for debugging, like:

```
2024-08-05 15:22:46.219 [info] (ruby-lsp-rspec) Error occurred when indexing /Users/hung-wulo/src/github.com/st0012/ruby-lsp-rspec/spec/ruby_lsp_rspec_spec.rb with 'RubyLsp::RSpec::IndexingEnhancement' enhancement: 'undefined method `foo' for an instance of Prism::ArgumentsNode'

Backtrace: /Users/hung-wulo/src/github.com/st0012/ruby-lsp-rspec/lib/ruby_lsp/ruby_lsp_rspec/indexing_enhancement.rb:29:in `on_call_node'  
/Users/hung-wulo/src/github.com/Shopify/ruby-lsp/lib/ruby_indexer/lib/ruby_indexer/declaration_listener.rb:295:in `block in on_call_node_enter'  
/Users/hung-wulo/src/github.com/Shopify/ruby-lsp/lib/ruby_indexer/lib/ruby_indexer/declaration_listener.rb:294:in `each'  
/Users/hung-wulo/src/github.com/Shopify/ruby-lsp/lib/ruby_indexer/lib/ruby_indexer/declaration_listener.rb:294:in `on_call_node_enter'  
/Users/hung-wulo/.gem/ruby/3.3.3/gems/prism-0.30.0/lib/prism/dispatcher.rb:216:in `block in visit_call_node'  
/Users/hung-wulo/.gem/ruby/3.3.3/gems/prism-0.30.0/lib/prism/dispatcher.rb:216:in `each'  
/Users/hung-wulo/.gem/ruby/3.3.3/gems/prism-0.30.0/lib/prism/dispatcher.rb:216:in `visit_call_node'  
/Users/hung-wulo/.gem/ruby/3.3.3/gems/prism-0.30.0/lib/prism/node.rb:2478:in `accept'  
/Users/hung-wulo/.gem/ruby/3.3.3/gems/prism-0.30.0/lib/prism/visitor.rb:31:in `block in visit_child_nodes'  
```

st0012 (approved):

Issue #2329, opened 2024-07-18
Index Core aliases using RBS (if supported)


st0012:

andyw8:
(reminder for self) Branch: `andyw8/support-method-aliases`

PR #2359, opened 2024-07-24
Support RBS signature aliases

andyw8:
### Motivation

Closes https://github.com/Shopify/ruby-lsp/issues/2329

Presumably to reduce duplication, RBS uses aliases to indicate that two methods have the same signature.

### Implementation

Handle using a `Entry::UnresolvedMethodAlias`

### Automated Tests

Added.

### Manual Tests

Not sure we can currently...

### DIFF
--- a/lib/ruby_indexer/test/rbs_indexer_test.rb
+++ b/lib/ruby_indexer/test/rbs_indexer_test.rb
@@ -332,8 +332,8 @@ module RubyIndexer
     end
 
     def test_signature_alias
-      # In RBS, an alias means that two methods have the same signature. It does not mean the same thing as a Ruby
-      # alias.
+      # In RBS, an alias means that two methods have the same signature.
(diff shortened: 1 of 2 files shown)


lib/ruby_indexer/test/rbs_indexer_test.rb
```diff
@@ -331,6 +331,23 @@ def self?.open: (String name, ?String mode, ?Integer perm) -> IO?
       assert_kind_of(Entry::BlockParameter, parameters[3])
     end
 
+    def test_signature_alias
+      # In RBS, an alias means that two methods have the same signature. It does not mean the same thing as a Ruby
+      # alias.
+      any_entries = @index["any?"]
+
+      assert_equal(["Array", "Enumerable", "Hash"], any_entries.map { _1.owner.name })
```
andyw8, line 339: This is more for illustration than testing.

Issue #2302, opened 2024-07-13
Another "classification" highlighting issue


borama:
### Description

(VSCode 1.91.0 , Ruby LSP 0.7.12)

Related to #2269, I recently noticed another issue with words containing `class` being highlighted as keywords. This time, it is words _ending_ with `class`. 

I [noticed](https://github.com/borama/vscode-ruby-slim/issues/2) it in a Slim template similar to this:

```slim
div class=wrapper_classes data-something="value"

div class=wrapper_class data-something="value"
```

Supposing `wrapper_classes` and `wrapper_class` are some ruby methods, the first example is highlighted correctly while the second is messed up:

![obrazek](https://github.com/user-attachments/assets/e711d31b-6e34-41fa-93d1-257c09b30ffe)

Adding a word boundary check to the beginning of the [class regexp](https://github.com/Shopify/ruby-lsp/blob/main/vscode/grammars/ruby.cson.json#L68) in the grammar file seems to fix it:

```diff
- "match": "(class)\\s+(([a-zA-Z0-9_]+)((::)[a-zA-Z0-9_]+)*)\\s*((<)\\s*(([a-zA-Z0-9_]+)((::)[a-zA-Z0-9_]+)*))?",
+ "match": "\b(class)\\s+(([a-zA-Z0-9_]+)((::)[a-zA-Z0-9_]+)*)\\s*((<)\\s*(([a-zA-Z0-9_]+)((::)[a-zA-Z0-9_]+)*))?",
```

I can add a PR if you like. Thanks!

andyw8:
@borama when you are editing this file, does VS Code show `{ } Ruby` in the footer?

borama:
@andyw8 No, it says "Slim" as the file is a [Slim template](https://github.com/slim-template/slim) but, in Slim notation, everything after the equal sign in an attribute is [embedded ruby](https://github.com/slim-template/slim?tab=readme-ov-file#ruby-attributes) and that's where I think the ruby LSP comes into play and that's also where the `class` keyword gets matched via the regexp. See the Editor tokens inspection:

![image](https://github.com/user-attachments/assets/93b5537b-451b-4e9d-a352-805c45e18e9e)

BTW I also managed to replicate the same problem in a [HAML template](https://haml.info/docs/yardoc/file.REFERENCE.html): suppose I have a `some_class` helper method available, then the following HAML template renders correctly:

```ruby
def some_class(item)
  "some_class #{item}"
end

hi = "hi!"
[...]

PR #2364, opened 2024-07-25
Prevent non keywords from accidentally matching on type formatting

### LINKED ISSUE #2301: On type formatting incorrectly adds `end` after hitting Enter key at the end of method's arguments when it has a trailing comma
### Environment

* Ruby LSP VS Code extension version: v0.7.12
* VS Code version: Version: 1.91.1 (Universal)
* Ruby version: ruby 3.1.6p260 (2024-05-29 revision a777087be6) [arm64-darwin24]
* Ruby version manager version: asdf v0.12.0
* Project setup: workspace

### Reproduction steps

1. Open VSCode with the Ruby LSP extension enabled.
2. Open a blank Ruby file.
3. Write the following:

```rb
link_to "Something",
  some_path,
  class: "some_class",
```

4. Hit enter after the last comma:

Expected:

```rb
link_to "Something",
  some_path,
  class: "some_class",
  # cursor here
```

Actual:

```rb
link_to "Something",
  some_path,
  class: "some_class",
    # cursor here
  end
```

This also happens when doing it on a single line:

1. Write the following:

```rb
link_to "Something", some_path, class: "some_class",
```

2. Hit enter after the last comma.

Expected:

```rb
link_to "Something", some_path, class: "some_class",
# cursor here
``` 

Actual:

```rb
link_to "Something", some_path, class: "some_class",
  # cursor here
end
```

vinistock:
### Motivation

Closes #2301

Our regexes for matching when to add the `end` token were a little too permissive. We were accidentally matching something like `class: "bg-white"` as a `class` keyword.

### Implementation

The reality is that there are only two possibilities for all of the keywords we provide on type formatting for: either there needs to be a space right after the keyword or they need to be at the end of the line.

### Automated Tests

Added a test to ensure we don't regress.


lib/ruby_lsp/requests/on_type_formatting.rb
```diff
@@ -32,8 +32,8 @@ def provider
 
       END_REGEXES = T.let(
         [
-          /\b(if|unless|for|while|class|module|until|def|case)\b.*/,
-          /.*\s\bdo\b/,
+          /\b(if|unless|for|while|class|module|until|def|case)\b($|\s)/,
```
andyw8, line 35: 🤔 What if someone uses parents directly after the keyword, e.g. `if(true)`

andyw8 (approved):

andyw8:
@vinistock does this also solve https://github.com/Shopify/ruby-lsp/issues/2302 ?

PR #2366, opened 2024-07-25
Add Ruby Copilot chat agent with domain driven design command

vinistock:
### Motivation

Add a Ruby Copilot chat agent with a domain driven design command (`/design` for short). The idea of the command is to assist the user in designing a concept into the model and database implementation.

For example, they can ask the following: `I'm working on a school web application. How can I model courses? And how do they relate to students?`. The chat will then reply with a suggested schema and buttons to run the Rails generate command for the proposed solution. Users can continue chatting back and forth until they arrive at the desired state.

### Implementation

The first commit only bumps the VS Code engine and `@types` package to v1.90, which is the first to include the chat related types we need.

The second commit includes the chat agent implementation. I will add comments in the PR explaining each part.

### Manual Tests

https://github.com/user-attachments/assets/ff1c41e4-7c6a-43d1-a66e-d48ea634a46b


vscode/src/chatAgent.ts
```diff
@@ -0,0 +1,214 @@
...
+    request: vscode.ChatRequest,
+    context: vscode.ChatContext,
+    stream: vscode.ChatResponseStream,
+    token: vscode.CancellationToken,
+  ) {
+    if (this.withinConversation("design", request, context)) {
+      return this.runDesignCommand(request, context, stream, token);
+    }
+
+    stream.markdown(
```
vinistock, line 49: If no command is used for the agent, there's really not much we can provide that regular Copilot chat won't.

Here, we only provide specialized behaviour if the user has already initiated a conversation with a previous command.

vscode/src/chatAgent.ts
```diff
@@ -0,0 +1,214 @@
...
+          title: "Generate with Rails",
+          arguments: [commandList, workspace],
+        });
+
+        stream.button({
+          command: Command.RailsDestroy,
+          title: "Revert previous generation",
+          arguments: [commandList, workspace],
+        });
+      }
```
vinistock, line 126: We ask in the prompt to return Rails generate commands inside a shell Markdown code block. The reason is so that we can extract which commands were proposed right here and display the buttons for Rails generate and destroy.

vscode/src/chatAgent.ts
```diff
@@ -0,0 +1,214 @@
...
+    } catch (err) {
+      this.handleError(err, stream);
+    }
+
+    return { metadata: { command: "design" } };
+  }
+
+  private async schema(workspace: Workspace) {
```
vinistock, line 134: This reading of the schema is extremely simple, but I propose we start with this before investing more effort and complexity into trying to detect only relevant models or anything of that nature.

vscode/package.json
```diff
@@ -26,6 +28,21 @@
   ],
   "main": "./out/extension.js",
   "contributes": {
+    "chatParticipants": [
+      {
+        "id": "rubyLsp.chatAgent",
+        "fullName": "Ruby",
+        "name": "ruby",
+        "description": "How can I help with your Ruby application?",
```
alexcameron89, line 36: Hello! 👋 Out of curiosity, would it be better to specify "Ruby on Rails application"? Scanning the PR, the prompt seems to be more Rails specific applications rather than plain Ruby applications.

vscode/src/chatAgent.ts
```diff
@@ -0,0 +1,214 @@
...
+      const match = /(?<=```shell)[^.$]*(?=```)/.exec(response);
+
+      if (workspace && match && match[0]) {
+        // The shell code block includes all of the `rails generate` commands. We need to strip out the `rails generate`
+        // from all of them since our commands only accept from the generator forward
+        const commandList = match[0]
+          .trim()
+          .split("\n")
+          .map((command) => {
+            return command.replace(/\s*(bin\/rails|rails) generate\s*/, "");
```
andyw8, line 112: Need to handle Windows here?

vinistock: Here we're sanitizing the commands generated by the LLM. It's the `Rails.Generate` and the `Rails.Destroy` commands that actually execute it and they already handle Windows.

isidorn:
Thanks for sharing the thinking process. I love some of the ideas you are outlining.

> How are the Python and Java teams organizing their chat participants? Do you have suggestions on how to organize ours with these possible use cases?

They do **not** contribute a chat participant. They are all experimenting with different approaches. I want to brainstorm more with the team next week on the recommended approach for language extensions on extending GH Copilot. But it might be something like:

* Improve the Inline Completions experience (similar to what C++ did in this [blog](https://devblogs.microsoft.com/cppblog/improving-github-copilot-completions-in-vs-code-for-c-developers/)). For this we need to have nicer pluggable story for GH Copilot.
* (still in discussion) Contribute #ruby variable, that we would automatically append to user queries based on the current file / intent detection.  Then you would resolve that variable. This would cover your `ask_rails`, `ask_sorbet`, `ask_workspace` scenarios. Basically you could dynamically resolve it based on the question and provide proper context.
* Help us improve the core Copilot experience for your specific language. Example (we need a cross platform docker-based Ruby diagnostics provider, and tree-sitter queries for ruby that we need for summarisastion of files). The downside here is that Copilot extension is closed source, so I do not know if you want to contribute.
* Only contribute a participant if there is still some functionality that is not improved by the above 

Having said this, things are still in the air, and from our side you can go ahead with your current approach to learn and get user feedback.

fyi @roblourens @aeschli

vscode/README.md
```diff
@@ -29,6 +29,29 @@ See complete information about features [here](https://shopify.github.io/ruby-ls
 If you experience issues, please see the [troubleshooting
 guide](https://github.com/Shopify/ruby-lsp/blob/main/TROUBLESHOOTING.md).
 
+### Copilot chat agent
```
dirceu, line 32: "Copilot" is a heavily used term nowadays, let's clarify this is about "GitHub Copilot"

vscode/README.md
```diff
@@ -29,6 +29,29 @@ See complete information about features [here](https://shopify.github.io/ruby-ls
...
+
+For users of Copilot, the Ruby LSP contributes a Ruby agent for AI assisted development of Ruby applications. Below you
+can find the documentation of each command for the Ruby agent. For information about how to interact with Copilot Chat,
+check [VS Code's official documentation](https://code.visualstudio.com/docs/copilot/copilot-chat).
+
+#### Design command
+
+The `@ruby /design` command is intended to be a domain driven design expert to help users model concepts for their
+applications. Users should describe what type of application they are building and which concept they are trying to
```
dirceu, line 41:
```suggestion
Rails applications. Users should describe what type of application they are building and which concept they are trying to
```

vscode/README.md
dirceu: Also, would it be worth explicitly saying this is an experimental feature? 

```suggestion
### [Experimental] GitHub Copilot chat agent
```

vscode/README.md
```diff
@@ -29,6 +29,29 @@ See complete information about features [here](https://shopify.github.io/ruby-ls
...
+```
+@ruby /design I'm working on a web application for schools. How do I model courses? And how do they relate to students?
+```
+
+The output is a suggested schema for courses including relationships with users. In the chat window, two buttons will appear: `Generate with Rails`, which invokes the Rails generators to create the models suggested, and `Revert previous generation`, which will delete files generated by a previous click in the generate button.
+
+As with most LLM chat functionality, suggestions may not be fully accurate, especially in the first iteration. Users can
+continue chatting with the `@ruby` agent to fine tune the suggestions given, before deciding to move forward with
+generation.
+
```
dirceu, line 54: Is it worth pointing people to the DX Slack for feedback or questions, or is the general link is enough?

dirceu (approved):
Thanks for pushing this forward! This is a worthwhile experiment to run 👍

vscode/README.md
vinistock: Yeah, good idea!

Issue #2296, opened 2024-07-12
Help users diagnose indexing issues


andyw8:
We've had couple of recent issues relating to indexing:

https://github.com/Shopify/ruby-lsp/issues/2287
https://github.com/Shopify/ruby-lsp/issues/2295

To aid with troubleshooting, we could add a section in the docs and link to it from that error. The docs can explain how to use `ruby-lsp-doctor`, and how to temporarily ignore problematic gems/files.

andyw8:
Note: The `ruby-lsp-doctor` command is now `ruby-lsp --doctor` (https://github.com/Shopify/ruby-lsp/pull/2310)

PR #2367, opened 2024-07-25
Add steps for indexing troubleshooting

andyw8:
Closes https://github.com/Shopify/ruby-lsp/issues/2296

### DIFF
--- a/TROUBLESHOOTING.md
+++ b/TROUBLESHOOTING.md
@@ -210,7 +210,7 @@ In rare cases, Ruby LSP will encounter an error which prevents parsing from comp
 
 Firstly, ensure that you are using the latest release of the `ruby-lsp` gem, as the problem may have been already fixed.
 
-To diagnose the particular file(s) causing a problem, run `ruby-lsp --doctor`. Please log an issue so that we we can address it. If the code is not open source then you can try write a minimal reproduction.
+To diagnose the particular file(s) causing a problem, run `ruby-lsp --doctor`. Please log an issue so that we can address it. If the code is not open source then please provide a minimal reproduction.
 
 In the meantime, you can [configure Ruby LSP to ignore a particular gem or file for indexing]([README.md#configuring-code-indexing]).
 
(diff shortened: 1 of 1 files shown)


TROUBLESHOOTING.md
```diff
@@ -202,6 +202,19 @@ ruby-lsp
 Is there any extra information given from booting the server manually? Or does it only fail when booting through the
 extension?
 
+## Indexing ##
```
vinistock, line 205:
```suggestion
## Indexing
```

TROUBLESHOOTING.md
```diff
@@ -202,6 +202,19 @@ ruby-lsp
...
+
+When Ruby LSP starts, it attempts to index your code as well as your dependencies as described in [Configuring code indexing](README.md#configuring-code-indexing).
+
+In rare cases, Ruby LSP will encounter an error which prevents parsing from completing, meaning Ruby LSP cannot start.
```
vinistock, line 209: We rescue all indexing errors, so that the language server doesn't crash. The only consequence of an indexing failure is incomplete indexing data.
```suggestion
In rare cases, Ruby LSP will encounter an error which prevents indexing from completing, which will result in incomplete information in the editor.
```

vinistock (commented):
Let's link to this new section from the indexing error dialog https://github.com/Shopify/ruby-lsp/blob/a5e17758c74c6559b13a715f20d81ee212fb21f6/lib/ruby_lsp/server.rb#L778.

TROUBLESHOOTING.md
```diff
@@ -202,6 +202,19 @@ ruby-lsp
...
+
+Firstly, ensure that you are using the latest release of the `ruby-lsp` gem, as the problem may have been already fixed.
+
+To diagnose the particular file(s) causing a problem, run `ruby-lsp --doctor`. Please log an issue so that we we can address it. If the code is not open source then you can try write a minimal reproduction.
```
st0012, line 213: I think we should require a minimal repro, not just ask them to try.

```suggestion
To diagnose the particular file(s) causing a problem, run `ruby-lsp --doctor`. Please log an issue so that we can address it. If the code is not open source then please provide a minimal reproduction.
```

andyw8:
Since the new docs have been moved around, I'll close this and open a seperate PR.

PR #2371, opened 2024-07-26
Fix grammar for class and module keyword matching

andyw8:
### Motivation

See https://github.com/Shopify/ruby-lsp/issues/2302

### Implementation

Only match if there is a preceding word boundary character.

Before:
Screenshot 2024-07-26 at 11 43 20 AM

After:
Screenshot 2024-07-26 at 11 43 11 AM

### Automated Tests

None

### Manual Tests

Try enter the same code as the screenshots above.

You will need to launch via the extension host to verify.

### DIFF
--- a/vscode/grammars/ruby.cson.json
+++ b/vscode/grammars/ruby.cson.json
@@ -65,7 +65,7 @@
         }
       },
       "comment": "class Namespace::ClassName < OtherNamespace::OtherClassName",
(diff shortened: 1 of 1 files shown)


vinistock (approved):

RedCMD:
@andyw8
`\b` is the JSON character `backspace`
`\\b` is the regex Word Boundary Assertion

PR #2372, opened 2024-07-26
Add switch block refactor

vinistock:
### Motivation

This PR adds a new refactor code action to switch the style of a block. If the block uses `do..end`, it becomes a one liner brace block and vice-versa.

### Implementation

The first commit adds a new method to locate the first node matching a given list of types within the selected range. This is important for refactors, because users will select the code they want to apply some refactor on and we need to ignore the surrounding extra selection if it doesn't match the expected node.

The second commit adds the code action.

### Automated Tests

Added tests.

### Manual Tests

Select a block, click the light bulb (CMD + .) and select switch block style.

https://github.com/user-attachments/assets/96ee9615-2975-486d-9579-81360cd862fd

### DIFF
--- a/lib/ruby_lsp/document.rb
+++ b/lib/ruby_lsp/document.rb
@@ -223,6 +223,40 @@ module RubyLsp
       NodeContext.new(closest, parent, nesting_nodes, call_node)
     end
 
+    sig do
+      params(
+        range: T::Hash[Symbol, T.untyped],
+        node_types: T::Array[T.class_of(Prism::Node)],
+      ).returns(T.nilable(Prism::Node))
+    end
+    def locate_first_within_range(range, node_types: [])
+      scanner = create_scanner
+      start_position = scanner.find_char_position(range[:start])
+      end_position = scanner.find_char_position(range[:end])
+      desired_range = (start_position...end_position)
+      queue = T.let(@parse_result.value.child_nodes.compact, T::Array[T.nilable(Prism::Node)])
+
+      until queue.empty?
+        candidate = queue.shift
+
+        # Skip nil child nodes
+        next if candidate.nil?
+
+        # Add the next child_nodes to the queue to be processed. The order here is important! We want to move in the
+        # same order as the visiting mechanism, which means searching the child nodes before moving on to the next
+        # sibling
+        T.unsafe(queue).unshift(*candidate.child_nodes)
+
+        # Skip if the current node doesn't cover the desired position
+        loc = candidate.location
+
+        if desired_range.cover?(loc.start_offset...loc.end_offset) &&
+            (node_types.empty? || node_types.any? { |type| candidate.class == type })
+          return candidate
+        end
+      end
+    end
+
     sig { returns(SorbetLevel) }
     def sorbet_level
       sigil = parse_result.magic_comments.find do |comment|
--- a/lib/ruby_lsp/requests/code_action_resolve.rb
+++ b/lib/ruby_lsp/requests/code_action_resolve.rb
@@ -47,6 +47,8 @@ module RubyLsp
 
       sig { override.returns(T.any(Interface::CodeAction, Error)) }
       def perform
+        return Error::EmptySelection if @document.source.empty?
+
         case @code_action[:title]
         when CodeActions::EXTRACT_TO_VARIABLE_TITLE
           refactor_variable
@@ -63,8 +65,6 @@ module RubyLsp
 
       sig { returns(T.any(Interface::CodeAction, Error)) }
       def switch_block_style
-        return Error::EmptySelection if @document.source.empty?
-
         source_range = @code_action.dig(:data, :range)
         return Error::EmptySelection if source_range[:start] == source_range[:end]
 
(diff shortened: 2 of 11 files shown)


lib/ruby_lsp/document.rb
```diff
@@ -223,6 +223,48 @@ def locate(node, char_position, node_types: [])
...
+      params(
+        range: T::Hash[Symbol, T.untyped],
+        node_types: T::Array[T.class_of(Prism::Node)],
+      ).returns(T.nilable(Prism::Node))
+    end
+    def locate_first_within_range(range, node_types: [])
+      scanner = create_scanner
+      start_position = scanner.find_char_position(range[:start])
+      end_position = scanner.find_char_position(range[:end])
+      desired_range = (start_position...end_position)
```
andyw8, line 236: Might be good to extract this to `Scanner`, then you could write:

`desired_range = scanner.range(range[:start], range[:end])`

lib/ruby_lsp/document.rb
```diff
@@ -223,6 +223,48 @@ def locate(node, char_position, node_types: [])
...
+        next unless desired_range.cover?(loc.start_offset...loc.end_offset)
+
+        # If the node's start character is already past the position, then we should've found the closest node
+        # already
+        break if end_position < loc.start_offset
+
+        # If there are node types to filter by, and the current node is not one of those types, then skip it
+        next if node_types.any? && node_types.none? { |type| candidate.class == type }
+
+        closest = candidate
```
amomchilov, line 262: Can we just:

```suggestion
        return candidate
```

Then we can get rid of the nilable `closest` local variable

lib/ruby_lsp/requests/code_action_resolve.rb
```diff
@@ -206,8 +267,6 @@ def refactor_method
         )
       end
 
-      private
```
amomchilov, line 209: intentional?

amomchilov (approved):
Great feature!

lib/ruby_lsp/document.rb
```diff
@@ -223,6 +223,48 @@ def locate(node, char_position, node_types: [])
...
+        range: T::Hash[Symbol, T.untyped],
+        node_types: T::Array[T.class_of(Prism::Node)],
+      ).returns(T.nilable(Prism::Node))
+    end
+    def locate_first_within_range(range, node_types: [])
+      scanner = create_scanner
+      start_position = scanner.find_char_position(range[:start])
+      end_position = scanner.find_char_position(range[:end])
+      desired_range = (start_position...end_position)
+      queue = T.let(@parse_result.value.child_nodes.compact, T::Array[T.nilable(Prism::Node)])
```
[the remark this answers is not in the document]
vinistock: Not at this exact line, but further down in the code because we keep pushing the child nodes into the queue recursively and eventually we may find a node that is `nil`.

test/expectations/code_action_resolve/nested_block_calls.exp.json
```diff
@@ -0,0 +1,47 @@
...
+                                "start": {
+                                    "line": 0,
+                                    "character": 29
+                                },
+                                "end": {
+                                    "line": 3,
+                                    "character": 3
+                                }
+                            },
+                            "newText": "{ |a| nested_call(fourth_call).each do |b|;  end }"
```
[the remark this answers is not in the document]
vinistock: That's a good point. I think we should be opinionated here, but I wasn't sure which way to go.

Essentially, we have two options:

- Find the first block within the selection and switch the style of that one. The implementation is straightforward and (at least on my opinion) the behaviour matches what I expect
- Find all blocks inside the selection and switch the style of all of them. This feels a bit surprising to me, but maybe others feel differently. Also, the implementation would need to be recursive since we don't know how many blocks are inside

Thoughts? cc @andyw8

lib/ruby_lsp/requests/code_action_resolve.rb
vinistock: Yeah, we had added a new method that should be private as public, so I moved it.

lib/ruby_lsp/document.rb
vinistock: Good point. I simplified the implementation.

test/expectations/code_action_resolve/nested_block_calls.exp.json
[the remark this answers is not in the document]
st0012: > Find the first block within the selection and switch the style of that one. The implementation is straightforward and (at least on my opinion) the behaviour matches what I expect

Do you mean we'd get:

```ruby 
method_call(other_call).each { |a| 
  nested_call(fourth_call).each do |b|
  end 
}
```

To me this is ok too if we want to be a bit more conservative. I simply do not expect to see `do; end` block being compressed to one-line.

But if I wanted to make the `do; end` block one line, I probably want to apply it to the second one too? In those cases, not having to it twice should be a lot more convenient.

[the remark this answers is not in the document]
st0012: > do we base it on the first block we find and then turn all of them to that style?

Personally, this makes the most sense.

st0012 (approved):

PR #2374, opened 2024-07-26
Add BBEdit to EDITORS page

andyw8:
It seems there are still some issues to be resolved, but in the mean time let's direct people to this thread.

cc [therbob](https://github.com/etherbob)

### DIFF
--- a/EDITORS.md
+++ b/EDITORS.md
@@ -50,6 +50,7 @@ new H2 header in this file containing the instructions. -->
 - [Sublime Text LSP](#sublime-text-lsp)
 - [Zed](#zed)
 - [RubyMine](#RubyMine)
+- [BBEdit](#BBEdit)
 
 ## Emacs Eglot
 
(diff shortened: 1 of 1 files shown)


etherbob:
@andyw8  https://github.com/Shopify/ruby-lsp/issues/2382 might also be relevant

FWIW this can be worked around on a per-project basis by wrapping lsp-ruby in shell script that sets the current working directory to a ruby project, but it does not appear to trigger indexing for multiple workspaces, nor does it work by setting CWD to an empty folder. It might be premature to say ruby-lsp works with BBEdit as of now. I'm still trying to figure out what feature BBEdit is leaning on from other language servers to make this work and/or if there's a configuration change I need to pass in.

andyw8:
We are going to move BBEdit support to the wiki instead (see https://github.com/Shopify/ruby-lsp/pull/2414).

EDITORS.md
```diff
@@ -185,6 +186,10 @@ Note that there might be overlapping functionality when using it with RubyMine,
 
 [Ruby LSP plugin](https://plugins.jetbrains.com/plugin/24413-ruby-lsp)
 
+## BBEdit
+
+See https://github.com/Shopify/ruby-lsp/issues/2365 for the latest information.
```
vinistock, line 191: I have the PR open to address the issue with BBEdit partially. Can we put configuration instructions here rather than pointing to the issue?

PR #2392, opened 2024-07-30
Fix definition listener's sorbet early return condition

st0012:
The condition missed the return so it's not working. The test still passed because the definition candidate was rejected by the check in `handle_method_definition` method instead.

This commit fixes the condition and tweaks the test to make sure the listener is working as expected.


test/requests/definition_expectations_test.rb
```diff
@@ -819,6 +819,10 @@ def bar
   end
 
   def test_definition_on_self_is_disabled_for_typed_true
+    # We need this stub to make sure the test is testing the early return inside on_call_node_enter
+    # not the one inside handle_method_definition
+    RubyLsp::Listeners::Definition.any_instance.stubs(:not_in_dependencies?).returns(false)
```
andyw8, line 824: `not_in_dependencies?` doesn't seem to be getting called here (you can see by changing the `stubs` to `expects`)

andyw8: Ok, then would it be better to write `RubyLsp::Listeners::Definition.any_instance.expects(:not_in_dependencies?).never`?

st0012: That's great, thanks!

PR #2395, opened 2024-07-31
Add go-to-definition support to autoloaded constants

### LINKED ISSUE #1893: Add go to definition support for autoloaded constants

st0012:
### Motivation

This PR is based on #1995 

Closes #1893

### Implementation

Most of the feature is already implemented in #1995 thanks to @Super-Xray.

My changes are:

- Use the symbol node instead of the `autoload` call node to trigger the request.
- Update the test case to account for the latest Sorbet check changes.

### Manual Tests

1. Pull this branch
2. Restart the server
3. Go to `lib/ruby_lsp/requests.rb`
4. Click around the autoload statements in the file and see if it lands at the right definition

### DIFF
--- a/test/requests/definition_expectations_test.rb
+++ b/test/requests/definition_expectations_test.rb
@@ -358,12 +358,20 @@ class DefinitionExpectationsTest < ExpectationsTestRunner
 
       class Foo
         autoload :Bar, "bar"
-
-        class Bar; end
       end
     RUBY
 
     with_server(source) do |server, uri|
+      server.global_state.index.index_single(
+        RubyIndexer::IndexablePath.new(nil, "/fake/path/bar.rb"), <<~RUBY
+          class Foo::Bar; end
+        RUBY
+      )
+      server.global_state.index.index_single(
+        RubyIndexer::IndexablePath.new(nil, "/fake/path/baz.rb"), <<~RUBY
+          class Foo::Bar; end
+        RUBY
+      )
       server.process_message(
         id: 1,
         method: "textDocument/definition",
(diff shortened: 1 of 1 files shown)


lib/ruby_lsp/listeners/definition.rb
```diff
@@ -251,6 +263,17 @@ def handle_require_definition(node, message)
         end
       end
 
+      sig { params(node: Prism::CallNode).void }
+      def handle_autoload_definition(node)
+        argument = node.arguments&.arguments&.first
+        return unless argument.is_a?(Prism::SymbolNode)
+
+        value = argument.value
```
andyw8, line 271: Would it be clearer to call this something like `constant_name`

lib/ruby_lsp/requests/definition.rb
```diff
@@ -79,7 +79,7 @@ def initialize(document, global_state, position, dispatcher, sorbet_level)
             position,
           )
         elsif target.is_a?(Prism::CallNode) && target.name != :require && target.name != :require_relative &&
-            !covers_position?(target.message_loc, position)
+            !covers_position?(target.message_loc, position) && target.name != :autoload
```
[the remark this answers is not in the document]
st0012: Updated 👍

lib/ruby_lsp/requests/definition.rb
```diff
@@ -78,8 +84,8 @@ def initialize(document, global_state, position, dispatcher, sorbet_level)
             parent,
             position,
           )
-        elsif target.is_a?(Prism::CallNode) && target.name != :require && target.name != :require_relative &&
-            !covers_position?(target.message_loc, position)
+        elsif target.is_a?(Prism::CallNode) && !covers_position?(target.message_loc, position) &&
+            !SPECIAL_METHOD_CALLS.include?(target.message)
```
paracycle, line 88: Wouldn't it be more performant to check first that the call node is not one of the special method calls?
```suggestion
        elsif target.is_a?(Prism::CallNode) && !SPECIAL_METHOD_CALLS.include?(target.message) && !covers_position?(target.message_loc, position)
```
which was what the old code was doing.

PR #2396, opened 2024-07-31
List install VS Code extensions as part of `--doctor`

andyw8:
Just a little idea I had: What if we list the extensions as part of `--doctor` to aid with troubleshooting? The intention would be that people can paste this into an issue.

### DIFF
--- a/exe/ruby-lsp
+++ b/exe/ruby-lsp
@@ -129,6 +129,10 @@ if options[:doctor]
     puts "indexing: #{indexable.full_path}"
     index.index_single(indexable)
   end
+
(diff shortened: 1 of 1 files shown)


Earlopain:
Stock VSCode is surely by far the most popular flavor but there are other like VSCodium https://github.com/VSCodium/vscodium which enjoys some usage. Another plus when delegating this to the extenion since it uses a different binary name and `code` doesn't exist, or worse, prints entirely different extensions.

st0012:
@andyw8 do you still want to add extensions to `--doctor` flag after #2456?

PR #2398, opened 2024-07-31
Disable certain LSP features for .erb files through registration options

### LINKED ISSUE #2394: ERB support seems to have disabled HTML folding
Before erb support was added, it was possible to fold HTML tags in erb files

Since erb support, only ruby expressions are foldable.
Disabling erb support reverts the behaviour as expected.

It would be amazing to be able to fold both ruby expressions and HTML tags with Ruby LSP

st0012:
### Motivation

Closes #2394

Also improves the existing approach of disabling requests on erb files.

### Implementation

For editors supporting LSP's [register capability](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#client_registerCapability), the LSP server can tell them which document type the particular request supports.

Currently, VS Code, Emacs (Eglot), and Neovim have all supported it. So I think it's fair to assume most editors should work with this feature.

#### Folding Range

VS Code currently doesn't support multiple folding range providers for the same document. So if Ruby LSP provides folding range for `erb` files, users would lose their original folding range options. Since in most cases folding HTML structure is more useful than folding erb code, I decided to disable folding range for erb files.

**Before**

Screenshot 2024-07-31 at 22 51 31

**After**

Screenshot 2024-07-31 at 23 41 48

#### Formatting, OnTypeFormatting, Diagnostics, and Code Action

Instead of using early return to disable these requests, I also migrated them to use document selector to inform the client instead.

### DIFF
--- a/lib/ruby_lsp/requests/code_actions.rb
+++ b/lib/ruby_lsp/requests/code_actions.rb
@@ -25,9 +25,12 @@ module RubyLsp
       class << self
         extend T::Sig
 
-        sig { returns(Interface::CodeActionOptions) }
+        sig { returns(Interface::CodeActionRegistrationOptions) }
         def provider
-          Interface::CodeActionOptions.new(resolve_provider: true)
+          Interface::CodeActionRegistrationOptions.new(
+            document_selector: [Interface::DocumentFilter.new(language: "ruby")],
+            resolve_provider: true,
+          )
         end
       end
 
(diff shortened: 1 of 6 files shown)


andyw8:
> Since in most cases folding HTML structure is more useful than folding erb code, I decided to disable folding range for erb files.

I agree. If you're writing long sections of Ruby in an ERB template that you're probably don't something wrong.

Intrepidd:
Thanks a lot for this !

PR #2403, opened 2024-08-01
Fix custom bundle reference to Gemfile above top level of working directory

mckeed:
### Motivation

When running ruby-lsp from a directory where the gemfile is in a parent directory, it works fine if the bundle includes the required dependencies (ruby-lsp and debug). If it doesn't, the custom bundle built at `.ruby-lsp/Gemfile` has `eval_gemfile('../Gemfile')` which causes the `bundle install` to fail.

It also fails if the project has a custom gemfile path set via the bundle config. This PR doesn't address that but will help with fixing it.

### Implementation

Updated `write_custom_gemfile` to use the correct path for the project gemfile.

### Automated Tests

Added a test for the change and another that was missing.

### Manual Tests

With a directory tree like:

```
test/
|- Gemfile
|- Gemfile.lock
|- proj/
|  |- test.rb
```

Run `ruby-lsp` from `proj/` or open `proj` in VSCode and open `test.rb` to trigger the extension to activate in the workspace.

### DIFF
--- a/lib/ruby_lsp/setup_bundler.rb
+++ b/lib/ruby_lsp/setup_bundler.rb
@@ -42,6 +42,8 @@ module RubyLsp
 
       @gemfile_name = T.let(@gemfile&.basename&.to_s || "Gemfile", String)
 
+      $stderr.puts("Ruby LSP> Gemfile: #{@gemfile}") if @gemfile && @gemfile != Pathname.pwd.join("Gemfile")
+
       # Custom bundle paths
       @custom_dir = T.let(Pathname.new(".ruby-lsp").expand_path(Dir.pwd), Pathname)
       @custom_gemfile = T.let(@custom_dir + @gemfile_name, Pathname)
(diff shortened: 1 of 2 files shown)


vinistock:
Thank you for the pull request, but this solution is not general enough. If the server in being launched in a workspace and the Gemfile is outside of it, then we essentially don't have a way of finding it that would work for all setups.

Nothing stops someone from making the same request, but wanting two levels above instead of a single one. Or maybe looking in a sibling directory. Or any other combination that cannot be accurately guessed.

The recommendation is to always launch the language server from the root of the workspace (where the Gemfile is) and to use [VS Code's multi-root workspace support](https://github.com/Shopify/ruby-lsp/tree/main/vscode#multi-root-workspaces) if you wish to slice a directory into multiple conceptual workspaces.

vinistock:
Sorry, I may have misunderstood the issue. Are you saying that `Bundler.default_gemfile` will keep checking any directories above the current one until it finds a Gemfile?

If that's what you mean, I indeed had not understood and then indeed we have to fix it. In that case, please re-open the PR.

PR #2414, opened 2024-08-07
Direct to wiki for community-supported editors

andyw8:
As discussed with @vinistock, we're going to try using the public Wiki for community-supported editors. Anyone with a GitHub account can edit those pages.

I've added a page for BBEdit, so we won't need https://github.com/Shopify/ruby-lsp/pull/2374

I also moved Sublime to there since it's not so popular.


EDITORS.md
```diff
@@ -1,8 +1,10 @@
 # Editors
 
-This file contains community driven instructions on how to set up the Ruby LSP in editors other than VS Code. For VS
+This file contains community driven instructions on how to set up the Ruby LSP in some popular editors. For VS
```
st0012, line 3:
```suggestion
This file contains community driven instructions on how to set up the Ruby LSP in some of the more popular editors. Due to reviewing capacity, we aren't able to keep all the available editors here. For VS
```

EDITORS.md
```diff
@@ -1,8 +1,10 @@
...
 Code, use the official [Ruby LSP extension](https://marketplace.visualstudio.com/items?itemName=Shopify.ruby-lsp).
 
+For other editors, we encourage the community to the use [public wiki](https://github.com/Shopify/ruby-lsp/wiki).
```
st0012, line 6:
```suggestion
For editors that aren't covered in this document, we encourage the community to the use [public wiki](https://github.com/Shopify/ruby-lsp/wiki).
```

st0012 (commented):
Let's also explain why we need to use the wiki in addition to this file.

st0012 (approved):

andyw8:
Have restored Sublime for now.

Issue #2363, opened 2024-07-25
Show all available overloads for signature help


st0012:
Currrently we only show the first.

st0012:
Hey @andyw8 I think in the last call we mentioned that there's a WIP PR to actually display the multiple overloads in the signature help? If that's correct, can you link the PR here please?

andyw8:
https://github.com/Shopify/ruby-lsp/pull/2495

PR #2415, opened 2024-08-07
Indicate multiple signatures for Hover and Completion Resolve

andyw8:
### Motivation

Partially addresses https://github.com/Shopify/ruby-lsp/issues/2363

Co-authored with @vinistock 

### Implementation

We are following TypeScript's lead on how to indicate that there are overloads.

In a follow-up we will start trying to intelligently match the correct signature as the arguments are typed.

### Automated Tests

Included

### Manual Tests

* Enter a method with overloads, e.g. `String.try_convert`
* Verify that hovering indicates there are multiple overloads
* Verify that completion indicates there are multiple overloads

### DIFF
--- a/lib/ruby_indexer/lib/ruby_indexer/entry.rb
+++ b/lib/ruby_indexer/lib/ruby_indexer/entry.rb
@@ -542,6 +542,11 @@ module RubyIndexer
       def decorated_parameters
         @target.decorated_parameters
       end
+
+      sig { returns(T::Array[Signature]) }
+      def signatures
+        @target.signatures
+      end
     end
 
     # Ruby doesn't support method overloading, so a method will have only one signature.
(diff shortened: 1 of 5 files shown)


lib/ruby_lsp/requests/support/helpers.rb
```diff
@@ -0,0 +1,25 @@
...
+# frozen_string_literal: true
+
+module RubyLsp
+  module Requests
+    module Support
+      module Helpers
+        extend T::Sig
+
+        sig { params(signatures: T::Array[RubyIndexer::Entry::Signature]).returns(String) }
+        def format_overloads_count(signatures)
```
andyw8, line 11: Any better place to put this?

vinistock: Agreed, this should be in the `Entry::Member` class.

PR #2430, opened 2024-08-12
Avoid requiring json for chruby activation

vinistock:
### Motivation

There's an annoying bug in our chruby activation because of the dependency on the json default gem.

These are the steps to reproduce it:

1. Ensure you are using Ruby 3.3.4 (or any Ruby with a patch higher than zero)
2. Chruby overrides the `GEM_HOME` to use `~/.gem/ruby/3.3.4` instead of the default `~/.gem/ruby/3.3.0`. We perform this correction inside our activation script (in Ruby)
3. However, the `-rjson` to require json will also require `strscan`. This happens before we had a chance to activate the chruby environment, which means it will try to require `strscan` with a `GEM_HOME` still pointing to `~/.gem/ruby/3.3.0`, even though the Ruby version is 3.3.4 and chruby would've pointed us to `~/.gem/ruby/3.3.4`
4. This leads to the `linked to incompatible Ruby version` error and breaks the activation

### Implementation

I can't think of another way for fixing this that isn't removing the json dependency. The approach that I took is quite simple: I started printing a separator string, so that we can easily break the pieces that we need from the activation script's result.

### Automated Tests

It's not very trivial to write tests for this specific scenario because it would involve installing more than one Ruby version.

### DIFF
--- a/vscode/src/ruby/chruby.ts
+++ b/vscode/src/ruby/chruby.ts
@@ -13,6 +13,8 @@ interface RubyVersion {
   version: string;
 }
 
+const ACTIVATION_SEPARATOR = "ACTIVATION_SEPARATOR";
+
 // A tool to change the current Ruby version
 // Learn more: https://github.com/postmodern/chruby
 export class Chruby extends VersionManager {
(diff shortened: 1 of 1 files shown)


vscode/src/ruby/chruby.ts
```diff
@@ -179,14 +179,17 @@ export class Chruby extends VersionManager {
...
     ].join(";");
 
     const result = await this.runScript(
-      `${rubyExecutableUri.fsPath} -W0 -rjson -e '${script}'`,
+      `${rubyExecutableUri.fsPath} -W0 -e '${script}'`,
     );
 
-    return this.parseWithErrorHandling(result.stderr);
+    const [defaultGems, gemHome, yjit, version] = result.stderr.split(
+      "ACTIVATION_SEPARATOR",
```
st0012, line 190: nit: can this value be stored in a variable/constant so it always stay in sync? I know it's unlikely to be changed but I'm just paranoid about it somehow got out of sync and creates mysterious bugs 😅

st0012 (approved):

PR #2432, opened 2024-08-13
Bugfix: Recognise ERB files correctly in Neovim

tsvallender:
### Motivation

Though ruby-lsp supports ERB files, in Neovim these were either not being parsed at all, or, following [this update to lspconfig](https://github.com/neovim/nvim-lspconfig/pull/3266), parsed as Ruby (with all the associated errors that brings).

### Implementation

Checking for "eruby" as the filetype, as well as "erb" gets things working.

There is one caveat: I still have a single error in each file being thrown by Rubocop: `Lint/Syntax: unexpected token tLT This offense is not auto-correctable`. I presume this does not occur in VS Code but can’t see any code preventing Rubocop in ERB files, so would appreciate any thoughts on that. Still worth merging as-is, I believe, as it reduces a screen full of errors and no functionality to one error plus functionality.

### Automated Tests

Not applicable

### Manual Tests

In VS Code there should be no change, in Neovim ERB HTML files should now be parsed correctly.

### DIFF
--- a/lib/ruby_lsp/server.rb
+++ b/lib/ruby_lsp/server.rb
@@ -296,7 +296,7 @@ module RubyLsp
       @mutex.synchronize do
         text_document = message.dig(:params, :textDocument)
         language_id = case text_document[:languageId]
-        when "erb"
+        when "erb", "eruby"
           Document::LanguageId::ERB
         else
           Document::LanguageId::Ruby


tsvallender:
I have signed the CLA!

lib/ruby_lsp/server.rb
```diff
@@ -296,7 +296,7 @@ def text_document_did_open(message)
       @mutex.synchronize do
         text_document = message.dig(:params, :textDocument)
         language_id = case text_document[:languageId]
-        when "erb"
+        when "erb", "eruby"
```
st0012, line 299: Why doesn't neovim have the `erb` as a filetype tho?

tsvallender: Looks like [this](https://github.com/neovim/neovim/blob/9768e88f3891b5594eb2d556bd3bdf40c61d46e1/runtime/lua/vim/filetype.lua#L434) is where this is defined, though I couldn’t say if there’s a reason for "eruby" over "erb".

vinistock: Normally, editors allow you to tell which language ID a file extension should be associated to.

That said, I have two comments about this:

1. I think it's fine to add `eruby` if NeoVim standardizes the `erb` extension as `eruby`
2. I think it's worth asking @adam12 if it's working properly for him since he added the configs. Adam, is the ERB support working properly for you? Do you need any additional changes or is shipping this PR the way to go?

adam12: > I think it's fine to add eruby if NeoVim standardizes the erb extension as eruby
Vim/Neovim have both standardized on the `eruby` filetype for `.erb` files.

>  Adam, is the ERB support working properly for you?
This _was_ working for me. When I tested it and submitted the PR to `nvim-lspconfig`, it looked like it worked fine. No warnings, no errors, autocomplete/etc seemed fine.

Today I see the same issue tho.

After making the change, I went ahead and did a lot of anxillary tasks, which may have affected things. Updated Neovim, updated Ruby-LSP, etc.

tsvallender:
I’ll also note the Rubocop warning I mentioned can be removed simply by telling Rubocop to ignore ERB files. Should Rubocop be able to successfully run in ERB files normally though? (i.e. am I missing out on some other functionality by doing that?)

vinistock:
> I’ll also note the Rubocop warning I mentioned can be removed simply by telling Rubocop to ignore ERB files. Should Rubocop be able to successfully run in ERB files normally though? (i.e. am I missing out on some other functionality by doing that?)

No, RuboCop should only run in Ruby files. We turn off diagnostics for ERB documents, but eventually we'd like to offer diagnostics + formatting through erb-lint.

> Confirming this patch corrects the errors seen in Neovim.

Checking other Ruby language servers for NeoVim, it seems like it standardizes on `eruby` as the language ID. Not sure why that was picked over `erb`, but I think it makes sense to move forward with this fix.

vinistock (approved):

mbriggs:
Would it be possible to cut a release for this? If you are an erb enjoyer, this is a pretty disruptive bug :)

adam12:
> Would it be possible to cut a release for this? If you are an erb enjoyer, this is a pretty disruptive bug :)

Might be worth adjusting your nvim config temporarily to remove eruby from the filetype.

Here's two examples, tho I wasn't able to test Mason directly because I only use it for managing non-Ruby LSPs.

```lua
local lspconfig = require('lspconfig')
lspconfig.ruby_lsp.setup({
  filetypes = { 'ruby' }
})
```

```lua
require('mason-lspconfig').setup_handlers {
  function (server_name) -- default handler
    require('lspconfig')[server_name].setup({})
  end,

  ['ruby_lsp'] = function ()
    require('lspconfig').ruby_lsp.setup({
      filetypes = { 'ruby' },
    })
  end,
}
```

rodrigolj:
Is this a regex issue, perhaps? If we need to narrow the diagnostics only between <% %> tags and such, I could spend some time on this. I just need someone to point me where to look, because I haven't worked on an LSP before.

adam12:
> Is this a regex issue, perhaps?

The parser is already aware of ERB formatted files. The issue corrected here was the filetype sent by vim/neovim wasn't the same filetype sent by the VSCode extension.

vinistock:
Just cut a release with the fix https://github.com/Shopify/ruby-lsp/releases/tag/v0.17.13.

To force the upgrade, you can either

- `rm -r .ruby-lsp` delete the custom bundle directory altogether and then re-launch the editor to create it from scratch
- `BUNDLE_GEMFILE=.ruby-lsp/Gemfile bundle update ruby-lsp` update the LSP manually

> Is this a regex issue, perhaps? 

No, this was a mismatch of the language ID for ERB documents expected by the server vs the one provided by NeoVim - which does not match the one used by VS Code.

adam12:
Just to round out this PR, I checked Zed's support for the Ruby LSP and it appears that it sends `erb` as well, so no special handling required in that case.

https://github.com/zed-industries/zed/blob/88a12b60a98f5d7980054cfef007fb79f8fc7be3/extensions/ruby/extension.toml#L15

AvidRambo:
Does this fix this issue?

CleanShot 2024-08-15 at 07 51 04@2x

vinistock:
When the Ruby LSP broadcasts its diagnostic capabilities, it scopes it to only Ruby files https://github.com/Shopify/ruby-lsp/blob/e18d4565c0733f7b8fb909c3f47592b1143d6053/lib/ruby_lsp/requests/diagnostics.rb#L28

Maybe NeoVim is not respecting the server's document selector for diagnostics? If that's the case, it's a NeoVim bug since the spec prescribes the possibility for the server to dictate which files it wants to handle for diagnostics.

See [Diagnostic registration options](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#diagnosticRegistrationOptions), which extends [Text document registration options](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#textDocumentRegistrationOptions) and allows for a document selector.

In addition to correctness, this is important for performance reasons too. We don't want the client to keep sending requests that the server doesn't want to handle. VS Code respects the server selection and never sends diagnostic requests for ERB files.

In terms of tactics on how to move forward with this, our 3 person team does not have NeoVim experience. My suggestion is to revert https://github.com/neovim/nvim-lspconfig/pull/3266 and then someone with more NeoVim experience would need to do a more in depth investigation to understand what's not going right before shipping that configuration PR again (to avoid disruption for NeoVim users).

PR #2435, opened 2024-08-13
Move Sorbet level and `locate_first_within_range` to RubyDocument

vinistock:
### Motivation

Continuing the work to properly separate the document concerns and eventually support RBS files, both Sorbet level and `locate_first_within_range` are specific to Ruby files only.

### Implementation

Moved both `sorbet_level` and `locate_first_within_range` to `RubyDocument` and updated the references accordingly.

### Automated Tests

Existing tests should cover this refactor.

### DIFF
--- a/lib/ruby_lsp/document.rb
+++ b/lib/ruby_lsp/document.rb
@@ -10,16 +10,6 @@ module RubyLsp
       end
     end
 
-    class SorbetLevel < T::Enum
-      enums do
-        None = new("none")
-        Ignore = new("ignore")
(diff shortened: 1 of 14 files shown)


st0012 (approved):
I think `Document#locate` can also be moved to `RubyDocument`?

vinistock:
> I think Document#locate can also be moved to RubyDocument?

Indeed, but I wanted to try to reduce the size of the PRs as much as I could. It's implemented in #2436.

PR #2444, opened 2024-08-14
Fix parsing erb with bare `yield`

### LINKED ISSUE #2442: "Invalid yield" diagnostic in erb
### Description

On Neovim v0.10.1 going through lspconfig/mason with ruby-lsp v0.17.13

### Reproduction steps

given

```html
<!DOCTYPE>
<html>
  <head>
    <%= yield :head %>
  </head>
  <body>
    <%= yield %>
  </body>
</html>
```
I am seeing "Invalid yield" on both variants

image

Love the work you are doing, and appreciate how well things work at this point in erb :)

Earlopain:
### Motivation

Some things that are valid in erb would not be valid in a standalone ruby file. This particularly impacts `yield`.
Instead treat the provided code as being evaled

Closes #2442

### Implementation

Switch prism into eval mode. Prior work for the prism parser translator at https://github.com/ruby/prism/pull/2741

### Automated Tests

Yup

### Manual Tests

Create a new erb file with the content from the test and observe that there are no syntax errors


andyw8 (approved):

lib/ruby_lsp/erb_document.rb
```diff
@@ -12,7 +12,7 @@ def parse
       @needs_parsing = false
       scanner = ERBScanner.new(@source)
       scanner.scan
-      @parse_result = Prism.parse(scanner.ruby)
+      @parse_result = Prism.parse(scanner.ruby, scopes: [[]])
```
st0012, line 15:
```suggestion
      # assigning empty scopes to turn Prism into eval mode
      @parse_result = Prism.parse(scanner.ruby, scopes: [[]])
```

st0012 (approved):
Thanks for the fix. Let's add a comment to explain what it does as well.

kddnewton:
> Are there any possible downsides to using eval mode here? (cc @kddnewton)

It will also turn off unused local variable warnings and void statement warnings, which is probably fine.

vinistock:
That wouldn't be the case for Kevin's comment though. We strip out the Ruby code from the ERB template and parse it with Prism as if it were a regular Ruby file, so in that particular example it would be able to tell you if the local variable was indeed unused inside the entire template.

We don't yet run RuboCop or ERB Lint on ERB files, so all of the existing diagnostics were coming from Prism itself.

PR #2445, opened 2024-08-14
Document use of `init_options` to configure LSP inside Neovim

adam12:
### Motivation

Documents the use of `init_options` to configure LSP options inside Neovim.

Provides an example of using these `init_options` to enable the Standard add-on for pull-style linting and formatting.

### DIFF
--- a/EDITORS.md
+++ b/EDITORS.md
@@ -72,6 +72,22 @@ When you run `eglot` command it will run `ruby-lsp` process for you.
 The [nvim-lspconfig](https://github.com/neovim/nvim-lspconfig/blob/master/lua/lspconfig/server_configurations/ruby_lsp.lua)
 plugin has support for Ruby LSP.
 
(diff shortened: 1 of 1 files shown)


EDITORS.md
```diff
@@ -72,6 +72,22 @@ When you run `eglot` command it will run `ruby-lsp` process for you.
...
+
+```lua
+local lspconfig = require('lspconfig')
+lspconfig.ruby_lsp.setup({
+  init_options = {
+    formatter = 'standard',
+    linters = { 'standard' },
+  },
+})
+```
```
andyw8, line 89: Not a blocker, but I wonder if we should generalize this so that each editor config shows how to pass initialization options (we could use `standard` as the example for consistency).

jacobdaddario:
👏🏻 This is great. Took me a while to figure that one out. Thank you for documenting!

vinistock (approved):
Nice! Thank you for the contribution

PR #2460, opened 2024-08-19
Run signature blocks ahead of time and compact after indexing

vinistock:
### Motivation

This PR adds two minor changes that may provide some performance gains.

### Running all signatures ahead of time

This eliminates the need to run them on the first invocation of each method and eagerly unwraps the original method back in place. This will spend a bit more time during boot, but it should ensure that we see fewer of Sorbet's method wrapping logic in our stacks.

### Compacting the GC after indexing

Indexing produces a lot of short lived objects for all of the ASTs of every file we interact with. This expands the heap pages a lot and may lead to some fragmentation. Compacting the GC helps us ensure that we're using the minimum number of pages needed and should make future GC runs a bit faster.

### DIFF
--- a/exe/ruby-lsp
+++ b/exe/ruby-lsp
@@ -81,6 +81,8 @@ require "ruby_lsp/load_sorbet"
 $LOAD_PATH.unshift(File.expand_path("../lib", __dir__))
 require "ruby_lsp/internal"
 
+T::Utils.run_all_sig_blocks
+
 if options[:debug]
   if ["x64-mingw-ucrt", "x64-mingw32"].include?(RUBY_PLATFORM)
     $stderr.puts "Debugging is not supported on Windows"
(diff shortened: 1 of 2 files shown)


exe/ruby-lsp
```diff
@@ -81,6 +81,8 @@ require "ruby_lsp/load_sorbet"
 $LOAD_PATH.unshift(File.expand_path("../lib", __dir__))
 require "ruby_lsp/internal"
 
+T::Utils.run_all_sig_blocks
```
KaanOzkan, line 84: I don't see a check for Sorbet, what happens if a project doesn't use Sorbet?

st0012 (approved):
On my machine I didn't notice any index/boot slowdown when running `--time-index` with this branch compared to `main` 👍

PR #2464, opened 2024-08-20
Document initialization options

neanias:
### Motivation

In [a discussion on a recent PR](https://github.com/Shopify/ruby-lsp/pull/2445/files#r1722069148), it was mentioned that it would be helpful to document all of the initialization options. I thought I'd go ahead and do that since I've fiddled around with it a few times in my own Neovim configuration.

### Implementation

I've added a new section at the bottom of the EDITORS.md file. I've also added a small example to the Neovim configuration section. I can add examples to the other editors if it helps?

### DIFF
--- a/EDITORS.md
+++ b/EDITORS.md
@@ -40,6 +40,54 @@ If your version manager doesn't provide either of those, then activate the envir
 These strategies will ensure that the `ruby-lsp` executable is invoked with the correct Ruby version, `GEM_HOME` and
 `GEM_PATH`, which are necessary for proper integration with your project.
 
+## All initialization options
+
+Each LSP client can control various abilities of the LSP at startup. The following JSON dictionary contains all of the
+available initialization options. Generally, editor LSP clients will configure LSP servers using a dictionary in their
+configuration languages (JSON, Lua, ELisp, etc.).
+
+```json
+{
+  "initializationOptions": {
+    "enabledFeatures": {
+      "codeActions": true,
+      "codeLens": true,
+      "completion": true,
+      "definition": true,
+      "diagnostics": true,
+      "documentHighlights": true,
+      "documentLink": true,
+      "documentSymbols": true,
+      "foldingRanges": true,
+      "formatting": true,
+      "hover": true,
+      "inlayHint": true,
+      "onTypeFormatting": true,
+      "selectionRanges": true,
+      "semanticHighlighting": true,
+      "signatureHelp": true,
+      "typeHierarchy": true,
+      "workspaceSymbol": true
+    },
+    "featuresConfiguration": {
+      "inlayHint": {
+        "implicitHashValue": true,
+        "implicitRescue": true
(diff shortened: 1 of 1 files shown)


neanias:
I have signed the CLA!

EDITORS.md
```diff
@@ -254,3 +280,52 @@ To configure indexing, pass a JSON hash as part of the initialization options fo
   }
 }
 ```
+
+# All initialization options
+
+Each LSP client can control various abilities of the LSP at startup. The following JSON dictionary contains all of the
+available initialization options. Generally, editor LSP clients will configure LSP servers using a dictionary in their
+configuration languages (JSON, Lua, ELisp, etc.). _NOTE: `enabledFeatures` may also be an array with feature names as
+strings rather than a string->bool dictionary as in this documentation.
```
vinistock, line 289: The array style is deprecated and we're only keeping it for backwards compatibility. Let's not advertise it.

EDITORS.md
```diff
@@ -254,3 +280,52 @@ To configure indexing, pass a JSON hash as part of the initialization options fo
   }
 }
 ```
+
+# All initialization options
```
vinistock, line 284: Since this applies to all editors, let's move it up, right below the `IMPORTANT NOTE` section.

EDITORS.md
```diff
@@ -88,6 +90,24 @@ lspconfig.ruby_lsp.setup({
 })
 ```
 
+To enable or disable features, you can add an `enabledFeatures` table to the `init_options` options, like so:
```
vinistock, line 93: Do we need this section considering we have the list of initialization options?

EDITORS.md
```diff
@@ -116,6 +136,12 @@ mason_lspconfig.setup_handlers {
 }
 ```
 
+> [!NOTE]
+> As documented under [LazyVim LSP](#lazyvim-lsp), Mason and Mason LspConfig may interfere with choosing the correct
+> version of Ruby LSP. If you're having problems with Ruby LSP starting, try adding `mason = false, cmd = {
+> vim.fn.expand("~/.rbenv/shims/ruby-lsp") }` to your `ruby_lsp.setup` table. If you're not using rbenv, this path may
+> be different.
```
vinistock, line 143: This note feels a bit short on the details and unrelated to the overall PR.

It doesn't explain why Mason may interfere with the LSP versions, why setting it to false will fix it or the relationship with version managers.

If this problem applies to NeoVim and LazyVim, then maybe we can create a separate section to discuss this and link to it. But let's do it in a separate PR.

EDITORS.md
neanias: It makes more sense being in the actual Neovim lsp-config docs anyway

neanias:
Hold on, removing indexing section

vinistock (approved):
Thank you very much for the contribution!

Issue #2463, opened 2024-08-20
setup_bundler's `rails_app?` incompatible with non-ascii characters


D-system:
### Reproduction steps
Having the config/application.rb with Japanese characters (eg: あ)

In https://github.com/Shopify/ruby-lsp/blob/88b85b810061aafd7d8d7078df6b5aa713979e54/test/fixtures/rails_application.rb add `あ`
```ruby
module MyApp
  class Application < Rails::Application
    # あ
  end
end
```

### Details

From this change https://github.com/Shopify/ruby-lsp/pull/2218 ruby-lsp error out on 
https://github.com/Shopify/ruby-lsp/blob/88b85b810061aafd7d8d7078df6b5aa713979e54/lib/ruby_lsp/setup_bundler.rb#L292

Ruby-lsp: 0.17.14
Ruby: 3.3.4
Rails: 6.1.7.8

Error: `setup_bundler.rb:295:in 'match?': invalid byte sequence in US-ASCII (ArgumentError)`
Full output:
```
2024-08-20 17:32:33.321 [info] (app) Checking if chruby is available on the path with command: /bin/zsh -i -c 'chruby --version'
2024-08-20 17:32:35.982 [info] (app) Checking if rbenv is available on the path with command: /bin/zsh -i -c 'rbenv --version'
2024-08-20 17:32:36.935 [info] (app) Discovered version manager rbenv
2024-08-20 17:32:36.935 [info] (app) Running command: `rbenv exec ruby -W0 -rjson -e 'STDERR.print({env: ENV.to_h,yjit:!!defined?(RubyVM::YJIT),version:RUBY_VERSION}.to_json)'` in /Users/thomas/projects/app using shell: /bin/zsh
2024-08-20 17:32:38.382 [info] (app) /Users/thomas/.rbenv/versions/3.3.4/lib/ruby/gems/3.3.0/gems/ruby-lsp-0.17.14/lib/ruby_lsp/setup_bundler.rb:295:in `match?': invalid byte sequence in US-ASCII (ArgumentError)

      /class .* < (::)?Rails::Application/.match?(application_contents)
                                                  ^^^^^^^^^^^^^^^^^^^^
	from /Users/thomas/.rbenv/versions/3.3.4/lib/ruby/gems/3.3.0/gems/ruby-lsp-0.17.14/lib/ruby_lsp/setup_bundler.rb:295:in `rails_app?'
[...]

PR #2466, opened 2024-08-20
Revert "Make Rails app detection based on Rails::Application superclass"

vinistock:
Reverting Shopify/ruby-lsp#2218 while https://github.com/Shopify/ruby-lsp/issues/2463 is better understood.

There are issues with encoding and multibyte characters and it's not yet clear why. This bug blocks people from using the LSP, so we should revert first and then work to figure out what happened.


Earlopain:
I openend #2467 which I'm fairly sure will fix the issue. But if you want to get something known good out first or think about a different solution that's fine

PR #2467, opened 2024-08-20
Handle `US_ASCII` encoding when parsing rails application config

Earlopain:
### Motivation

Connects to #2463

### Implementation

Added ascii encoding tests to the CI pipeline. Surprisingly to me, this was the only test that failed (and some multibyte which make less sense for `US_ASCII` encoding).

Not that knowledgable on this encoding stuff, but I do know that if ruby gets run with proper `LANG` environment variables, `Encoding.default_external` will reflect that which then causes `File.read` and the like to also read files in that encoding.

Could in theory force the encoding by overwriting `Encoding.default_external` but however the user has set this up, there is probably a reason for it.

### Automated Tests

Run the test suite with a encoding different to utf8.

### Manual Tests

None, I'm not really sure how ):. I did at least confirm that this doesn't break when the encoding is utf8.


lib/ruby_lsp/setup_bundler.rb
```diff
@@ -289,6 +289,7 @@ def rails_app?
       application_contents = config.read if config.exist?
       return false unless application_contents
 
+      application_contents = application_contents.encode(Encoding::UTF_8, Encoding::UTF_8, invalid: :replace)
```
Earlopain, line 292: `force_encoding` would do the job as well but this is more robust against whatever else may be present in that file

PR #2468, opened 2024-08-20
Add literal types to inferrer

vinistock:
### Motivation

This PR starts returning the expected type for literals, so that we can provide completion, go to definition, hover and signature help when something is invoked directly on one.

### Implementation

I also fixed a minor bug that we had not caught. We were showing any type in the completion documentation as if it were a guessed type, but that should only appear if it was guessed.

Then I added a bunch of branches to cover literals.

### Automated Tests

Added tests.

### DIFF
--- a/lib/ruby_lsp/type_inferrer.rb
+++ b/lib/ruby_lsp/type_inferrer.rb
@@ -36,9 +36,47 @@ module RubyLsp
     def infer_receiver_for_call_node(node, node_context)
       receiver = node.receiver
 
+      # For receivers inside parenthesis, such as ranges like (0...2), we need to unwrap the parenthesis to get the
+      # actual node
+      if receiver.is_a?(Prism::ParenthesesNode)
+        statements = receiver.body
+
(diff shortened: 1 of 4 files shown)


andyw8:
Do we have a list of others literals not yet covered here? `Range` is the one I noticed, but there are probably others.

lib/ruby_lsp/type_inferrer.rb
```diff
@@ -39,6 +39,26 @@ def infer_receiver_for_call_node(node, node_context)
       case receiver
       when Prism::SelfNode, nil
         self_receiver_handling(node_context)
+      when Prism::StringNode
```
Morriar, line 42: I think you can also handle ranges and maybe procs

andyw8: Also percent-literals (`%w{}` etc)?

vinistock: Added ranges and literal lambdas. `%{}` are string nodes, so they are already handled.

PR #2469, opened 2024-08-20
WIP: Assume UTF-8 when detecting Rails

andyw8:


test/setup_bundler_test.rb
```diff
@@ -483,6 +483,9 @@ def test_ensures_lockfile_remotes_are_relative_to_default_gemfile
   end
 
   def test_ruby_lsp_rails_is_automatically_included_in_rails_apps
+    original_encoding = Encoding.default_external
```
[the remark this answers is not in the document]
vinistock: Yeah, I think it's worth doing a separate test since it needs the specific encoding change.

vinistock (approved):
Looks good

test/setup_bundler_test.rb
```diff
@@ -511,6 +514,8 @@ def test_ruby_lsp_rails_is_automatically_included_in_rails_apps
         assert_match('gem "ruby-lsp-rails"', File.read(".ruby-lsp/Gemfile"))
       end
     end
+  ensure
+    Encoding.default_external = original_encoding
```
vinistock, line 518:
```suggestion
    Encoding.default_external = T.must(original_encoding)
```

paracycle: Can I suggest to extract a helper method for this?
```ruby
def with_default_external_encoding(encoding, &block)
  original_encoding = Encoding.default_external
  begin
    Encoding.default_external = encoding
    block.call
  ensure
    Encoding.default_external = original
  end
end
```
which should need the must and makes the test code read nicer.

vinistock: Yeah, I like that 👍

lib/ruby_lsp/setup_bundler.rb
```diff
@@ -286,7 +286,7 @@ def correct_relative_remote_paths
     sig { returns(T::Boolean) }
     def rails_app?
       config = Pathname.new("config/application.rb").expand_path
-      application_contents = config.read if config.exist?
+      application_contents = config.read(external_encoding: Encoding::UTF_8) if config.exist?
```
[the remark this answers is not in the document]
paracycle: Honestly, I am not sure if it is worth using `Pathname` here for this.
```ruby
config = File.expand_path("config/application.rb")
application_contents = File.read(config, external_encoding: Encoding::UTF_8) if File.exist?(config)
```
is equally readable IMO

andyw8:
Updated based on feedback. Still trying to understand the remaining type error, seems the shim isn't being applied?

sorbet/rbi/shims/encoding.rbi
```diff
@@ -0,0 +1,7 @@
+# typed: true
+
+class Encoding < Object
+  sig {returns(Encoding)}
+  sig {returns(String)}
+  def self.default_external; end
```
st0012, line 6: I think this is not needed? Typechecking passed without it.

st0012 (approved):
I think a shim is not needed, but overall looks good.

Issue #2287, opened 2024-07-11
Indexing error in `existing_or_new_singleton_class`


andyw8:
As reported in https://github.com/Shopify/ruby-lsp/issues/1897#issuecomment-2222048806, indexing the httplog gem is failing.

andyw8:
@zachahn-gusto are you on the latest server gem? This does not error for me.

zachahn-gusto:
Yes I am! 0.17.14. To be honest, I wasn't really able to reproduce this issue consistently. I'll see if I can find a better test case.

zachahn-gusto:
Quick update, I wasn't able to make it crash, but here's a test case that demonstrates what I'm seeing

```ruby
    def test_indexing_singleton
      index(<<~RUBY)
        module ::Foo
          class Bar
            class << self
            end
          end
        end
      RUBY

      # We want to explicitly verify that we didn't introduce the leading `::` by accident, but `Index#[]` deletes the
      # prefix when we use `refute_entry`
      entries = @index.instance_variable_get(:@entries)
      assert_empty(entries.keys.select { _1.start_with?("::") })
    end
```

That fails with this error:

```
Failure:
RubyIndexer::ClassesAndModulesTest#test_indexing_singleton [~/ruby-lsp/lib/ruby_indexer/test/classes_and_modules_test.rb:601]
Minitest::Assertion: Expected ["::Foo::Bar::<Class:Bar>"]
 to be empty.
```

vinistock:
Ah, that's another part of the bug we missed indeed. I'm cutting a PR for that.

vinistock:
https://github.com/Shopify/ruby-lsp/pull/2471

PR #2471, opened 2024-08-20
Fix singleton class nesting when inside top level reference

vinistock:
### Motivation

Fixes the bug reported in https://github.com/Shopify/ruby-lsp/issues/2287#issuecomment-2299656280

We need to use our proper nesting resolution `actual_nesting` to discover the correct place where we discovered a singleton class, just like we're doing for regular classes. Otherwise, we incorrectly insert a name prefixed with `::` in the index state, which we do not want.

### Implementation

Started using our `actual_nesting` method to figure out the right nesting.

### Automated Tests

Added a test.


zachahn-gusto:
Thank you so much for the quick turnaround!!!

st0012 (approved):

Issue #2461, opened 2024-08-19
Add support for semantic tokens delta


vinistock:
While investigating performance issues in very large files (3k+ lines), I discovered something which had been off our radar. We always wondered why the Ruby LSP was not performing well in super large files despite our benchmarks showing that the server would respond really fast.

I believe the issue we've been facing is that, for very large files, the number of semantic tokens returned from the server is insane and then editors like VS Code lag when trying to apply all of them in such long files.

By simply not returning any tokens from the server, the lag is gone and the Ruby LSP becomes responsive in large files just like it is in smaller ones.

If this hypothesis is correct, then the resolution is to add support for [semantic tokens delta](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#semanticTokens_deltaRequest). Essentially, instead of returning all tokens on every request, we return all tokens only once and then the editor will ask for the delta (the difference) in tokens between the previous request and the current one.

This is potentially related to #1756 because since the editor falls behind in processing incoming responses, it may be the culprit for dropping diagnostics coming from the server. By reducing the pressure in the editor, it will not fall behind and might resolve that problem too.

### Notes

- Not every editor supports delta requests. We have to check for the client capability during initialization, save it in the global state and pass that along to semantic highlighting
- Currently, we cache the encoded tokens in the document they belong to. I think we may be able to implement this in a somewhat straight forward way if we instead cache the tokens before they are encoded. That will make it significantly easier to compute the delta

PR #2478, opened 2024-08-23
Support semantic token deltas

vinistock:
### Motivation

First step in #2461

This PR adds support for [semantic token deltas](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#semanticTokens_deltaRequest).

Here's how semantic highlighting works before and after this change.

**Before**

On every keypress, the editor sends a `textDocument/semanticTokens/full` request and the server returns all tokens for the entire file. In large files, the editor will sometimes send `textDocument/semanticTokens/range` to compute tokens for a limited range instead of doing it for the entire file.

**After**

When the file is first opened, the editor will send a `textDocument/semanticTokens/full` to compute all tokens for that file. From that point on, the editor will only send `textDocument/semanticTokens/full/delta`, to receive only the difference in tokens between the previous response and the current state of the file.

#### Why does this matter

Computing semantic tokens on the server is cheap. In some of my benchmarks, the Ruby LSP can compute ~20k tokens in roughly ~10ms. However, applying these tokens in the editor is very expensive. The fewer tokens we return, the faster the editor becomes.

In fact, if we return ~20k tokens on every keypress, the editor becomes backlogged with work and the lag prevents users from working normally.

#### Does this PR completely fix the problem

No. It improves the problem, but I will follow this PR up with an audit of our semantic tokens implementations. We start out returning tokens for pretty much everything, but we need to be a lot more conservative and only return tokens for the things that are really ambiguous to avoid performance issues.

### Implementation

There are many important parts in this implementation. First, for semantic token requests to work properly with deltas, we need to keep track of the semantic token result IDs and we need a separate cache for those tokens that cannot be cleared upon typing. The delta request needs to remember the previous result, so clearing upon typing would make it impossible to implement.

Secondly, running the full semantic tokens with the other listeners makes it incredibly difficult to ensure the correct result IDs. Accidentally bumping the ID will make the editor miss the delta and then we lose the benefits. It is clearly more impactful from a performance standpoint to support delta than it is to run semantic highlighting with the other listeners.

Having it separate will also play well with the next step to audit our semantic tokens, because we can convert the implementation to a visitor and have much better control over which tokens are returned.

Finally, the last part is computing the delta. The way the algorithm works is:

1. Find the first token from the start that differs between the previous and the current result. Strip the token arrays up to that point
2. From the end, find the first token that differs and strip the tail of the token arrays

Whatever is left will be the single edit necessary to turn the previous tokens into the current ones.

### Automated Tests

Added a bunch of tests.

### DIFF
--- a/lib/ruby_lsp/document.rb
+++ b/lib/ruby_lsp/document.rb
@@ -16,6 +16,7 @@ module RubyLsp
     extend T::Generic
 
     ParseResultType = type_member
+    EMPTY_CACHE = T.let(Object.new.freeze, Object)
 
     abstract!
 
@@ -34,9 +35,13 @@ module RubyLsp
     sig { returns(Encoding) }
     attr_reader :encoding
 
+    sig { returns(T.any(Interface::SemanticTokens, Object)) }
+    attr_accessor :semantic_tokens
+
     sig { params(source: String, version: Integer, uri: URI::Generic, encoding: Encoding).void }
     def initialize(source:, version:, uri:, encoding: Encoding::UTF_8)
-      @cache = T.let({}, T::Hash[String, T.untyped])
+      @cache = T.let(Hash.new(EMPTY_CACHE), T::Hash[String, T.untyped])
+      @semantic_tokens = T.let(EMPTY_CACHE, T.any(Interface::SemanticTokens, Object))
       @encoding = T.let(encoding, Encoding)
       @source = T.let(source, String)
       @version = T.let(version, Integer)
@@ -63,7 +68,7 @@ module RubyLsp
     end
     def cache_fetch(request_name, &block)
       cached = @cache[request_name]
-      return cached if cached
+      return cached if cached != EMPTY_CACHE
 
       result = block.call(self)
       @cache[request_name] = result
--- a/lib/ruby_lsp/requests/semantic_highlighting.rb
+++ b/lib/ruby_lsp/requests/semantic_highlighting.rb
@@ -35,28 +35,125 @@ module RubyLsp
               token_modifiers: ResponseBuilders::SemanticHighlighting::TOKEN_MODIFIERS.keys,
             ),
             range: true,
-            full: { delta: false },
+            full: { delta: true },
           )
+        end
+
+        # The compute_delta method receives the current semantic tokens and the previous semantic tokens and then tries
+        # to compute the smallest possible semantic token edit that will turn previous into current
+        sig do
+          params(
+            current_tokens: T::Array[Integer],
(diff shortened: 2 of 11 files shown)


lib/ruby_lsp/requests/semantic_highlighting.rb
```diff
@@ -35,28 +35,129 @@ def provider
...
+        @result_id = T.let(SemanticHighlighting.next_result_id.to_s, String)
         @response_builder = T.let(
           ResponseBuilders::SemanticHighlighting.new(global_state.encoding),
           ResponseBuilders::SemanticHighlighting,
         )
-        Listeners::SemanticHighlighting.new(dispatcher, @response_builder, range: range)
+        Listeners::SemanticHighlighting.new(dispatcher, @response_builder)
 
         Addon.addons.each do |addon|
           addon.create_semantic_highlighting_listener(@response_builder, dispatcher)
```
st0012, line 126: Since we still need to produce full tokens as usual, I assume this won't affect addons and they also don't need to know anything about result id?

st0012 (approved):

lib/ruby_lsp/requests/semantic_highlighting.rb
```diff
@@ -35,28 +35,129 @@ def provider
...
+          # Filter the tokens based on the first different position. This must happen at this stage, before we try to
+          # find the next position from the end or else we risk confusing sets of token that may have different lengths,
+          # but end with the exact same token
+          old_tokens = T.must(old_tokens[first_different_position...])
+          new_tokens = T.must(new_tokens[first_different_position...])
+
+          # Then search from the end to find the first token that doesn't match. Since the user is normally editing the
+          # middle of the file, this will minimize the number of edits since the end of the token array has not changed
+          reverse_old_tokens = old_tokens.reverse
+          first_different_token_from_end = new_tokens.reverse.index.with_index do |token, index|
```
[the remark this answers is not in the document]
vinistock: See my comment for the other one.

lib/ruby_lsp/document.rb
```diff
@@ -16,6 +16,7 @@ class LanguageId < T::Enum
     extend T::Generic
 
     ParseResultType = type_member
+    EmptyCache = T.let(Object.new.freeze, Object)
```
st0012, line 19: Nit: `EmptyCache` feels like a class/module name. Maybe it could be `EMPTY_CACHE` instead?

vinistock: Ah, yeah, good point. It's more in the style of constants.

PR #2482, opened 2024-08-23
Minimize semantic tokens returned for local variables

vinistock:
### Motivation

Similar to #2479, this PR minimizes the number of semantic tokens returned to the editor for local variables.

We don't need to return semantic tokens for most local variable assignments, because they can be detected with Text Mate grammars.

### Implementation

In the grammar, I added rules for local variable assignments. In the server, I removed handling of local variable assignments. We still need to handle local variable read, to differentiate between parameters and local variables.

### Automated Tests

Updated the expectations.


vscode/grammars/ruby.cson.json
```diff
@@ -100,6 +100,51 @@
...
+      "comment": "A local variable operation assignment (+=, -=, *=, /=)"
+    },
+    {
+      "captures": {
+        "1": {
+          "name": "variable.ruby"
+        }
+      },
+      "match": "^\\s*([a-z]([A-Za-z0-9_])*)\\s*=[^=]",
+      "comment": "A local variable assignment"
```
st0012, line 146: Comparing against the list in https://github.com/ruby/irb/pull/961, I think there also are `%=`, `**=`, `&=`, `|=`, `^=`, `<<=`, `>>=`

vinistock: Ah, nice! Thanks for sharing the others. I added a new commit to improve the regex.

st0012 (approved):

PR #2483, opened 2024-08-23
Disable diagnostics on large files

vinistock:
### Motivation

Similar to #2481, this PR disables "external diagnostics" for files with over 100k characters. Diagnostics such as syntax errors (collected by Prism) are still displayed, but linters like RuboCop are not shown.

Diagnostics runs on every keypress and trying to compute it for very large files ends up causing too much lag in the editor. After this PR, working on a ~4k line file is fully responsive, with response times around ~10-50ms.

### Implementation

Just changed the message on did open and started returning before running RuboCop in large files.

### Automated Tests

We don't have such a large file inside our codebase and we intentionally don't run diagnostics on files that are outside of the workspace, so I couldn't use Prism's `node.rb` file to write a test. I think this is simple enough that it's okay to skip it.

### DIFF
--- a/lib/ruby_lsp/requests/diagnostics.rb
+++ b/lib/ruby_lsp/requests/diagnostics.rb
@@ -46,7 +46,9 @@ module RubyLsp
         diagnostics.concat(syntax_error_diagnostics, syntax_warning_diagnostics)
 
         # Running RuboCop is slow, so to avoid excessive runs we only do so if the file is syntactically valid
-        return diagnostics if @document.syntax_error? || @active_linters.empty?
+        if @document.syntax_error? || @active_linters.empty? || @document.past_expensive_limit?
+          return diagnostics
+        end
 
         @active_linters.each do |linter|
           linter_diagnostics = linter.run_diagnostic(@uri, @document)
--- a/lib/ruby_lsp/server.rb
+++ b/lib/ruby_lsp/server.rb
@@ -315,13 +315,14 @@ module RubyLsp
           language_id: language_id,
         )
 
-        if document.source.length > Requests::SemanticHighlighting::MAXIMUM_CHARACTERS_FOR_HIGHLIGHT
+        if document.past_expensive_limit?
           send_message(
             Notification.new(
               method: "window/showMessage",
               params: Interface::ShowMessageParams.new(
                 type: Constant::MessageType::WARNING,
-                message: "This file is too long. For performance reasons, semantic highlighting will be disabled",
+                message: "This file is too long. For performance reasons, semantic highlighting and " \
+                  "diagnostics will be disabled",
               ),
             ),
           )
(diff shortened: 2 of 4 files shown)


lib/ruby_lsp/requests/diagnostics.rb
```diff
@@ -19,6 +19,10 @@ module Requests
     class Diagnostics < Request
       extend T::Sig
 
+      # Computing "external diagnostics", like RuboCop violations, can be very expensive in super large files. Similar
+      # to semantic highlighting, we limit the feature in favor of performance and responsiveness
+      MAXIMUM_CHARACTERS_FOR_EXTERNAL_DIAGNOSTICS = 100_000
```
st0012, line 24: Should we just use `MAXIMUM_CHARACTERS_FOR_HIGHLIGHT` and maybe rename it to `MAXIMUM_CHARACTERS_FOR_HIGHLIGHT_AND_DIAGNOSTICS` (or something more generic)? Otherwise, there could be a mismatch between features and messages' condition over time. Like if `MAXIMUM_CHARACTERS_FOR_HIGHLIGHT` is reduced, the message in did open request will disappear in some cases while the diagnostics is still disabled.

lib/ruby_lsp/server.rb
```diff
@@ -315,13 +315,14 @@ def text_document_did_open(message)
...
 
-        if document.source.length > Requests::SemanticHighlighting::MAXIMUM_CHARACTERS_FOR_HIGHLIGHT
+        if document.source.length > MAXIMUM_CHARACTERS_FOR_EXPENSIVE_FEATURES
           send_message(
             Notification.new(
               method: "window/showMessage",
               params: Interface::ShowMessageParams.new(
                 type: Constant::MessageType::WARNING,
-                message: "This file is too long. For performance reasons, semantic highlighting will be disabled",
+                message: "This file is too long. For performance reasons, semantic highlighting and " \
```
andyw8, line 324: What if we just say "For performance reasons, some Ruby LSP features will be disabled". That can then cover any aspect that we might want to disable in future.

st0012 (approved):

lib/ruby_lsp/server.rb
vinistock: It's not a strong opinion, but I would rather be specific about what we're turning off so that users know what exactly what they can expect to be working. If we're too generic, people won't know if something is a bug or if it's intentionally disabled.

lib/ruby_lsp/server.rb
```diff
@@ -315,13 +315,14 @@ def text_document_did_open(message)
           language_id: language_id,
         )
 
-        if document.source.length > Requests::SemanticHighlighting::MAXIMUM_CHARACTERS_FOR_HIGHLIGHT
+        if document.source.length > MAXIMUM_CHARACTERS_FOR_EXPENSIVE_FEATURES
```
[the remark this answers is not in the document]
vinistock: Nice suggestion. Extracted it.

PR #2484, opened 2024-08-23
Improve Ruby LSP extension's status item names

### LINKED ISSUE #2391: Consolidate all the status items into one

### Use case

It takes up way too much space, and crowds out the room for other status bar items.

We should combine the 2 status items, remove excess information, make it narrower.

### Description

Here's what a typical status bar looks like when using the Ruby LSP:

Screenshot 2024-07-30 at 12 27 31 PM

Some criticisms:

1. The word "Ruby" appears 4 times
2. The RubyLSP version takes up a lot of space, and isn't particularly relevant information most of the time.
    * Perhaps that's useful when working on the RubyLSP, so perhaps it could only be shown in dev miode?
3. "Running" is also wording, and is the most common status (hopefully!). Perhaps we should only show a status while it's stopped/starting/etc., and then hide it once started and running.
4. The distinction between the [`RubyVersionStatus`](https://github.com/Shopify/ruby-lsp/blob/935c7e63aba77ff283ca95e3c28e9891b070b794/vscode/src/status.ts#L39) and the [`ServerStatus`](https://github.com/Shopify/ruby-lsp/blob/935c7e63aba77ff283ca95e3c28e9891b070b794/vscode/src/status.ts#L64) feels contrived. Can't these be combined?
5. Both of these status items have `.name = "Ruby LSP Status"` ([`RubyVersionStatus`](https://github.com/Shopify/ruby-lsp/blob/935c7e63aba77ff283ca95e3c28e9891b070b794/vscode/src/status.ts#L43), [`ServerStatus`](https://github.com/Shopify/ruby-lsp/blob/935c7e63aba77ff283ca95e3c28e9891b070b794/vscode/src/status.ts#L68)), so you get this funky menu:

    Screenshot 2024-07-30 at 12 33 56 PM

Desired outcome:

A single, short status item, perhaps even with a Ruby icon. (e.g. an `item.text = "$(ruby) RubyLSP"`, [doc ref](https://code.visualstudio.com/api/references/vscode-api#StatusBarItem))

### Implementation

This looks like an easy change, I'd love to take a stab at it, if we decide we want it!

st0012:
### Motivation

Closes #2391 

Screenshot 2024-08-23 at 22 04 47

### DIFF
--- a/vscode/src/status.ts
+++ b/vscode/src/status.ts
@@ -40,7 +40,7 @@ export class RubyVersionStatus extends StatusItem {
   constructor() {
     super("rubyVersion");
 
-    this.item.name = "Ruby LSP Status";
+    this.item.name = "Ruby Version";
     this.item.command = {
       title: "Configure",
       command: Command.SelectVersionManager,
@@ -65,7 +65,7 @@ export class ServerStatus extends StatusItem {
   constructor() {
     super("server");
 
-    this.item.name = "Ruby LSP Status";
+    this.item.name = "Ruby LSP Server Status";
     this.item.text = "Ruby LSP: Starting";
     this.item.severity = vscode.LanguageStatusSeverity.Information;
     this.item.command = {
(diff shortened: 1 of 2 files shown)


amomchilov (approved):
Ah perfect, thanks!

PR #2491, opened 2024-08-26
WIP: Make rails runner stdin available through global state

andyw8:
See also:

- https://github.com/Shopify/tapioca/pull/2001
- https://github.com/Shopify/ruby-lsp-rails/pull/438


st0012:
Do we need to have an API for it? I thought `instance_variable_set|get` with `T.cast` would already work?

andyw8:
No longer needed with new approach.

PR #2493, opened 2024-08-27
fix(grammar): Fix syntax highlighting for Kernel.` method in Ruby grammar

### LINKED ISSUE #1470: Syntax highlight breaks on Kernel.` method
### Description

The following code breaks syntax highlighting feature:

```ruby
# frozen_string_literal: true

def do_test
  _h = { a: 1, b: 2 }

  Kernel.`"cat /bin/false"

  { a: 1, b: 2 }
end
```

Screenshot from 2024-03-07 04-15-48

Prism's AST seems to be OK:

```
$ bin/parse -e 'Kernel.`"cat /bin/false"'
@ ProgramNode (location: (1,0)-(1,24))
├── locals: []
└── statements:
    @ StatementsNode (location: (1,0)-(1,24))
    └── body: (length: 1)
        └── @ CallNode (location: (1,0)-(1,24))
            ├── flags: ∅
            ├── receiver:
            │   @ ConstantReadNode (location: (1,0)-(1,6))
            │   └── name: :Kernel
            ├── call_operator_loc: (1,6)-(1,7) = "."
            ├── name: :`
            ├── message_loc: (1,7)-(1,8) = "`"
            ├── opening_loc: ∅
            ├── arguments:
            │   @ ArgumentsNode (location: (1,8)-(1,24))
            │   ├── flags: ∅
            │   └── arguments: (length: 1)
            │       └── @ StringNode (location: (1,8)-(1,24))
            │           ├── flags: ∅
            │           ├── opening_loc: (1,8)-(1,9) = "\""
            │           ├── content_loc: (1,9)-(1,23) = "cat /bin/false"
            │           ├── closing_loc: (1,23)-(1,24) = "\""
            │           └── unescaped: "cat /bin/false"
            ├── closing_loc: ∅
            └── block: ∅

```

latest VScode
VScode ruby-lsp plugin v0.5.12
Gem ruby-lsp 0.14.4

VSCode `show-syntax-tree` result:

[show-syntrax-tree.txt](https://github.com/Shopify/ruby-lsp/files/14517441/show-syntrax-tree.txt)

Sean0628:
### Motivation

Closes #1470

The issue was that syntax highlighting breaks when using the ``Kernel.\` `` method in Ruby code. The grammar was incorrectly treating the backtick in this context as the start of an interpolated string literal, leading to incorrect tokenization and highlighting.

### Implementation

To address this, the Ruby grammar was updated to properly distinguish between backtick string literals and the ``Kernel.` `` method calls. Specifically, the `begin` pattern for backticks was modified to include a negative lookbehind `(?<!\\.)`, ensuring that only standalone backticks are treated as string delimiters.

> (?<!…) — Negative Lookbehind. Matches if the provided pattern would not match the preceding characters, but does not advance the current position. The pattern must have a fixed length (unbounded [quantifiers](https://rbuckton.github.io/regexp-features/engines/oniguruma.html#feature-quantifiers) are not permitted).

ref. https://rbuckton.github.io/regexp-features/engines/oniguruma.html

Additionally, I added new test cases to cover both standard backtick strings and ``Kernel.\` `` method calls. These tests verify that the grammar correctly tokenizes each scenario, ensuring that backtick strings are recognized as interpolated strings, while ``Kernel.\` `` methods are handled without string interpretation.

### Automated Tests

Unit tests were added as part of this change to ensure that both scenarios (standard backtick strings and ``Kernel.\` ``method calls) are correctly handled by the grammar.

#### Before
Screenshot 2024-08-27 at 00 37 02

#### After
Screenshot 2024-08-27 at 00 36 27

### DIFF
--- a/vscode/grammars/ruby.cson.json
+++ b/vscode/grammars/ruby.cson.json
@@ -486,7 +486,7 @@
       ]
     },
     {
-      "begin": "`",
+      "begin": "(?<!\\.)`",
       "beginCaptures": {
         "0": {
           "name": "punctuation.definition.string.begin.ruby"
--- a/vscode/src/test/suite/grammars.test.ts
+++ b/vscode/src/test/suite/grammars.test.ts
@@ -270,6 +270,61 @@ suite("Grammars", () => {
       });
     });
 
+    suite("Backtick String Literals", () => {
+      test("Standard backtick string", () => {
+        const ruby = "`ruby`";
+        const expectedTokens = [
+          [
+            "`",
+            [
+              "source.ruby",
+              "string.interpolated.ruby",
+              "punctuation.definition.string.begin.ruby",
+            ],
+          ],
+          ["ruby", ["source.ruby", "string.interpolated.ruby"]],
+          [
+            "`",
+            [
+              "source.ruby",
+              "string.interpolated.ruby",
+              "punctuation.definition.string.end.ruby",
+            ],
+          ],
+        ];
+        const actualTokens = tokenizeRuby(ruby);
+        assert.deepStrictEqual(actualTokens, expectedTokens);
+      });
+
+      test("Kernel backtick method", () => {
+        const ruby = 'Kernel.`"ls"';
+        const expectedTokens = [
+          ["Kernel", ["source.ruby", "variable.other.constant.ruby"]],
+          [".", ["source.ruby", "punctuation.separator.method.ruby"]],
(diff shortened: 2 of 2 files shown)


Sean0628:
I have signed the CLA!

andyw8 (approved):

vinistock (approved):
Nice. Thank you for the contribution!

PR #2498, opened 2024-08-27
Warn users if shadownev is not installed

vinistock:
### Motivation

We received an interesting report where activating the Ruby environment with shadowenv was failing. It turned out that it was not failing because the workspace was untrusted, but because the extension couldn't find the shadowenv executable because sourcing the user's `~/.zshrc` was taking too long and exceeded VS Code's timeout for launching the shell.

The problem ended up being more confusing than it had to be because we always assume that any Shadowenv failures are because the workspace is untrusted.

This PR just adds a custom error for when Shadowenv is not found, so that users are better informed of why it failed.

### Implementation

Started checking the error message. Each shell shows a different message when an executable is not available, so I tried to go for a catch all implementation.

### Automated Tests

The Shadowenv tests were both brittle and a bit redundant, so I did a little bit of restructuring and added a new example.


vscode/src/test/suite/ruby/shadowenv.test.ts
```diff
@@ -77,8 +110,7 @@ suite("Shadowenv", () => {
   });
 
   afterEach(() => {
-    fs.rmSync(rootPath, { recursive: true, force: true });
-    bundleGemfileStub.restore();
+    // fs.rmSync(rootPath, { recursive: true, force: true });
```
egiurleo, line 113: Any reason this is commented?

egiurleo (approved):
shadowenv strikes again!

vscode/src/test/suite/ruby/shadowenv.test.ts
vinistock: Because I forgot 😅. Removed it!

vscode/src/ruby/shadowenv.ts
```diff
@@ -37,18 +37,41 @@ export class Shadowenv extends VersionManager {
...
-      // activate the Ruby environment).
-      const answer = await vscode.window.showErrorMessage(
-        `Failed to run shadowenv exec. Is ${this.bundleUri.fsPath} trusted? Run 'shadowenv trust --help' to know more`,
-        "Trust workspace",
-        "Cancel",
-      );
+      const message = error.message;
+
+      // Depending on the user's shell, the message for not finding the executable is slightly different. For example,
+      // in zsh, it's "zsh: command not found: shadowenv", but in bash, it's "bash: shadowenv: command not found". Here
```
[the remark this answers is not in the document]
vinistock: Switched to using `which shadowenv`, which I think will work for any shell on Linux and MacOS (Shadowenv does not support Windows).

vscode/src/ruby/shadowenv.ts
```diff
@@ -37,18 +37,34 @@ export class Shadowenv extends VersionManager {
...
     } catch (error: any) {
-      // If running `shadowev exec` fails, it's typically because the workspace has not been trusted yet. Here we offer
-      // to trust it and fail it the user decides to not the trust the workspace (since in that case, we are unable to
-      // activate the Ruby environment).
-      const answer = await vscode.window.showErrorMessage(
-        `Failed to run shadowenv exec. Is ${this.bundleUri.fsPath} trusted? Run 'shadowenv trust --help' to know more`,
-        "Trust workspace",
-        "Cancel",
-      );
+      const { stdout } = await this.runScript("which shadowenv");
```
sambostock, line 40: Apparently `command -v shadowenv` is more portable. Should we use that?

PR #2501, opened 2024-08-28
Add addon client and server base classes

st0012:
### Motivation

The client/server architecture of `ruby-lsp-rails` to a large degree can be adopted by many other addons too. For example, other web frameworks, like Hanami, can also use it to provide runtime information. And if we extract the base client/server implementation into `ruby-lsp`, it'll make such applications easier to implement.

Also, through such extraction we can standardize things like:

- Logging interface (to VS Code's Output tab)
- Communications between the client and server
- Server's lifecycle management

And that will make `ruby-lsp-rails` easier to maintain and understand as well.

Draft PR from `ruby-lsp-rails`: https://github.com/Shopify/ruby-lsp-rails/pull/439

### Implementation

- Add `RubyLsp::Addon::ProcessClient|ProcessServer` classes. To use them, addons should require `ruby_lsp/addon/process_client` in the client file and `ruby_lsp/addon/process_server` in the server file. 
- I added `process` prefix with the hope to indicate that they're only needed when the addon needs to interact with another process.

### Automated Tests

I added a new `test/addon/process_client_server_test.rb` with a fake addon to test the client and server end-to-end.

### DIFF
--- a/lib/ruby_lsp/addon/process_server.rb
+++ b/lib/ruby_lsp/addon/process_server.rb
@@ -1,0 +1,50 @@
+# typed: strict
+# frozen_string_literal: true
+
+require "sorbet-runtime"
+
+module RubyLsp
+  class Addon
+    class ProcessServer
+      extend T::Sig
+      extend T::Generic
+
+      abstract!
+
+      VOID = Object.new
+
+      sig { void }
+      def initialize
+        $stdin.sync = true
+        $stdout.sync = true
+        $stdin.binmode
+        $stdout.binmode
+        @running = T.let(true, T.nilable(T::Boolean))
+      end
+
+      sig { void }
+      def start
+        initialize_result = generate_initialize_response
+        $stdout.write("Content-Length: #{initialize_result.length}\r\n\r\n#{initialize_result}")
+
+        while @running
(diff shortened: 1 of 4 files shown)


lib/ruby_lsp/addon/process_server.rb
```diff
@@ -0,0 +1,50 @@
+# typed: strict
+# frozen_string_literal: true
+
+require "sorbet-runtime"
```
andyw8, line 4: We had some issues in the past when requiring `sorbet-runtime` within the Rails side. I can't recall the exact cause, but we might need to be wary of that or do some extra checking.

test/addon/fake_process_server.rb
```diff
@@ -0,0 +1,29 @@
...
+
+      def execute(request, params)
+        case request
+        when "echo"
+          { result: { echo_result: params[:message] } }
+        when "shutdown"
+          @running = false
+          { result: {} }
+        else
+          VOID
```
andyw8, line 22: We shouldn't need the `VOID` stuff any more now that https://github.com/Shopify/ruby-lsp-rails/pull/453 is merged.

st0012:
FYI, I'll update this after Tapioca LSP is merged.

PR #2507, opened 2024-08-28
Raise AddonNotFoundError exception when the addon is not found

andyw8:
### Motivation

Fail in more graceful way for addons such as Tapioca LSP.

### Implementation

Raise and error that be e rescued downstream.

### Automated Tests

Added

### Manual Tests

n/a


st0012 (approved):
LGTM and I also updated the PR title to make the change clearer

PR #2510, opened 2024-08-29
feat: Migrate rubocop watching to server

Sean0628:
### Motivation

The Ruby LSP currently watches for changes in `.rubocop.yml` from the VS Code extension and restarts the server when the configuration is modified. However, this approach is limited to the VS Code environment and unnecessarily restarts the entire server. To enhance this functionality and make it available across other editors, the file watching should be initiated from the server itself. Instead of restarting the server, the RuboCop runner object can be re-initialized to use the updated configurations, providing a more efficient and editor-agnostic solution.

Closes [#1457](https://github.com/Shopify/ruby-lsp/issues/1457).

### Implementation

- Introduced a new request from the Ruby LSP server to watch `.rubocop.yml` files, enabling this functionality across different editors, not just VS Code.
- Added a specific check for `.rubocop.yml` in the `workspace/didChangeWatchedFiles` handler to detect changes to the RuboCop configuration file.
- Updated the RuboCop runner configurations by re-initializing the RuboCop runner object when `.rubocop.yml` is modified, allowing the server to use the new configurations without needing a full restart.

### Automated Tests

Unit tests have been added to:
- Verify that the `.rubocop.yml` watcher is correctly registered.
- Ensure that changes to `.rubocop.yml` trigger the re-initialization of the RuboCop runner with updated configurations.
- Confirm that only changes to `.rubocop.yml` trigger configuration reloads, while other files do not affect the RuboCop runner.

### Manual Tests

To manually test these changes:
1. Modify the `.rubocop.yml` file within a watched workspace.
2. Observe that the RuboCop runner reloads its configuration without restarting the server.
3. Confirm that editing unrelated files does not affect the RuboCop runner's configuration.

https://github.com/user-attachments/assets/24feedca-895e-4496-9590-befa7121f2e4

### DIFF
--- a/test/server_test.rb
+++ b/test/server_test.rb
@@ -428,6 +428,65 @@ class ServerTest < Minitest::Test
     })
   end
 
+  def test_registers_rubocop_yml_watcher
+    @server.process_message({
+      id: 1,
+      method: "initialize",
+      params: {
+        capabilities: {
+          workspace: {
+            didChangeWatchedFiles: {
+              dynamicRegistration: true,
+              relativePatternSupport: true,
+            },
+          },
+        },
+      },
+    })
+
+    registration = find_message(RubyLsp::Request, "client/registerCapability")
+    rubocop_registration =
+      registration.params.registrations.any? { |reg| reg.id == "workspace/didChangeWatchedFiles-rubocop" }
+
+    assert(rubocop_registration, "Expected RuboCop .rubocop.yml watcher to be registered")
+  end
+
+  def test_handles_rubocop_yml_changes
+    @server.global_state.expects(:register_formatter).with(
+      "rubocop",
+      instance_of(RubyLsp::Requests::Support::RuboCopFormatter),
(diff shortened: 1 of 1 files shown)


lib/ruby_lsp/server.rb
```diff
@@ -265,6 +265,19 @@ def run_initialize(message)
...
                   ),
                 ),
+                # Register watching .rubocop.yml
+                Interface::Registration.new(
+                  id: "workspace/didChangeWatchedFiles-rubocop",
+                  method: "workspace/didChangeWatchedFiles",
+                  register_options: Interface::DidChangeWatchedFilesRegistrationOptions.new(
+                    watchers: [
+                      Interface::FileSystemWatcher.new(
+                        glob_pattern: "**/.rubocop.yml",
```
andyw8, line 275: Since this is now server-side, I wonder if we could hook into RuboCop to know which files to watch 🤔 .

PR #2513, opened 2024-08-29
Forward initialization settings to addons

### LINKED ISSUE #1204: Forward initialization options to addons
Certain addons might need to allow users to configure functionality. For example, Steep allows you to define how many typechecking workers to fork.

We should have a generic configuration object in the extension that gets forwarded to addons based on their name. Something like

```json
{
  "rubyLsp.addonConfigurations": {
    "nameOfAddon": { configs }
  }
}
```

vinistock:
### Motivation

Closes #1204
First step to allow https://github.com/Shopify/ruby-lsp-rails/issues/389

While we work super hard to minimize the amount of configuration for the Ruby LSP, it's not realistic to assume that addons are not going to need any settings to be exposed.

We need a way to allow addons to receive relevant editor settings, so that their behaviour can be customized.

### Implementation

The idea is to allow for configuration objects like

```jsonc
"rubyLsp.addonSettings": {
  "Ruby LSP Rails": {
    "runtimeServerCommand": "bin/dev server",
    "somethingElse": false,
  }
}
```

And then we remember those settings in the global state. Addons will then have access to these settings and can decide what to do accordingly.

### Automated Tests

Added tests.

### DIFF
--- a/lib/ruby_lsp/global_state.rb
+++ b/lib/ruby_lsp/global_state.rb
@@ -40,6 +40,12 @@ module RubyLsp
       @supports_watching_files = T.let(false, T::Boolean)
       @experimental_features = T.let(false, T::Boolean)
       @type_inferrer = T.let(TypeInferrer.new(@index, @experimental_features), TypeInferrer)
+      @addon_settings = T.let({}, T::Hash[String, T.untyped])
+    end
+
+    sig { params(addon_name: String).returns(T.nilable(T::Hash[Symbol, T.untyped])) }
+    def settings_for_addon(addon_name)
+      @addon_settings[addon_name]
     end
 
     sig { params(identifier: String, instance: Requests::Support::Formatter).void }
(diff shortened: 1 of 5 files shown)


st0012 (approved):

lib/ruby_lsp/global_state.rb
```diff
@@ -40,6 +40,12 @@ def initialize
       @supports_watching_files = T.let(false, T::Boolean)
       @experimental_features = T.let(false, T::Boolean)
       @type_inferrer = T.let(TypeInferrer.new(@index, @experimental_features), TypeInferrer)
+      @addon_settings = T.let({}, T::Hash[String, T.untyped])
+    end
+
+    sig { params(addon_name: String).returns(T.nilable(T::Hash[Symbol, T.untyped])) }
+    def settings_for_addon(addon_name)
+      @addon_settings[addon_name]
```
andyw8, line 48: Worth logging to `stderr` if the addon doesn't exist, to help with troubleshooting?

vinistock: By design, if the user didn't define settings for an addon this will return `nil`, so addons should always handle that scenario.

And given that addons may be different depending on which project you are working on, I feel like trying to validate the user's editor settings could be a bit tricky.

PR #2524, opened 2024-09-04
Respect `-c .rubocop.yml` in `.rubocop`

### LINKED ISSUE #2521: LSP not honoring `.rubocop` config file
### Description

I think the lsp is not honoring the `.rubocop` file

I have this in `.rubocop`

```
-c .rubocop-local.yml
```

But the lsp is using `.rubopcop.yml` instead which is not what `bundle exec rubocop` would do.

From the docs:

> Default command-line options are loaded from .rubocop and RUBOCOP_OPTS and are combined with command-line options that are explicitly passed to rubocop. Thus, the options have the following order of precedence (from highest to lowest):
> 
> Explicit command-line options
> 
> Options from RUBOCOP_OPTS environment variable
> 
> Options from .rubocop file.

Earlopain:
### Motivation

Closes #2521

### Implementation

This does actually already contain all the options from `.rubocop` and `RUBOCOPOPT`. The logic for extracting these resides in [`RuboCop::Options`](https://github.com/rubocop/rubocop/blob/f1108d1c4b8b9db8bab794203fb6b4f2f111e000/lib/rubocop/options.rb#L28-L30), it just needs a bit of special attention.

I also added a restart watcher to the vscode extension for this.

[--force-default-config](https://github.com/rubocop/rubocop/blob/f1108d1c4b8b9db8bab794203fb6b4f2f111e000/lib/rubocop/cli.rb#L160-L161) is special as well but I say lets wait until someone proves that this actually gets used.

### Automated Tests

Check that providing a non-existant config raises. I don't think there is much more to do here, what this actually does is up to RuboCop.

### Manual Tests
In ruby-lsp:
* Copy `.rubocop.yml` to `.rubocop2.yml`
* Append the following:
  ```yml
  Style/StringLiterals:
    EnforcedStyle: single_quotes
  ```
* Create `.rubocop` with `-c .rubocop2.yml`
* Reload and observe a bunch of offenses


st0012 (approved):

vinistock (approved):

PR #2527, opened 2024-09-05
Downgrade engine to v1.91

vinistock:
### Motivation

I just realized that we don't need v1.92 for the chat functionality. v1.91 already has what we need.


st0012 (approved):
🎉

Issue #2387, opened 2024-07-30
Possible Bundler lock problem


andyw8:
A couple of times recently, running `bundle` on the command line has been hanging for me.

I spotted notices [this](https://github.com/rubygems/rubygems/issues/7732) issue, and found that killing off all the active Ruby processes on my machine solved the problem, so it seems something was indeed locking the update.

When it happens again, I can investigate more to try find the culprit.

cc @bdewater

andyw8:
Interesting. I wonder if we should detect if there's a bisect in progress, and avoid running Bundler if so.

vinistock:
If the Ruby LSP is already running bundle install, it indeed makes it impossible to run another bundle install process in the command line due to the lock. Honestly, I think that's the right behaviour since we wouldn't want two processes trying to compile and install the same gems.

That said, we should improve the Ruby LSP's bundle install reporting so that instead of just printing to STDERR, we show a progress notification (similar to indexing) with the current state of bundle install.

Earlopain:
A progress notification like that would be pretty nice to show whats going on a bit more prominently. It makes sense for bundler to not trip over itself but having to wait a minute for gems to finish compiling (gemfile specific of course) before you can continue bisecting is :neutral_face:

PR #2530, opened 2024-09-06
Inhibit server restarts during bisect as well

Earlopain:
### Motivation

This is a revival of what I wrote in https://github.com/Shopify/ruby-lsp/issues/2387#issuecomment-2288146864

I consistently run into the issue of the lockfile changing during a bisect, restarting the server which then runs bundler, which blocks `bundler/inline` from processing until the gems are installed which may or may not finish in a reasonable amount of time.

This stops the server from restarting when a bisect is in process, where the lockfile is likely to constantly change.

### Implementation

Piggy-back of the already existing code for checking against a rebase. This does still cause additional restarts if you bisect during a rebase or the other way around but it seems very rare and this implementation is very simple.

### Automated Tests

No. There are no tests for rebase

### Manual Tests

Start a bisect, and observe that the server doesn't restart itself. Similarly, with a rebase


vscode/src/workspace.ts
```diff
@@ -22,7 +22,7 @@ export class Workspace implements WorkspaceInterface {
   private readonly isMainWorkspace: boolean;
   private readonly telemetry: vscode.TelemetryLogger;
   private needsRestart = false;
-  #rebaseInProgress = false;
+  #inhibitRestart = false;
```
andyw8, line 25: 👍 good name

vscode/src/workspace.ts
```diff
@@ -44,7 +44,11 @@ export class Workspace implements WorkspaceInterface {
     this.isMainWorkspace = isMainWorkspace;
 
     this.registerRestarts(context);
-    this.registerRebaseWatcher(context);
+    this.registerCreateDeleteWatcher(
+      context,
+      ".git/{rebase-merge,rebase-apply}",
```
andyw8, line 49: I tried find a list of all the various temporary files that git creates, for example during cherry-picking there's `CHERRY_PICK_HEAD`, but couldn't find a comprehensive list.

vscode/src/workspace.ts
```diff
@@ -335,35 +339,32 @@ export class Workspace implements WorkspaceInterface {
...
     };
 
     context.subscriptions.push(
       workspaceWatcher,
       parentWatcher,
-      // When one of the rebase files are created, we set this flag to prevent restarting during the rebase
-      workspaceWatcher.onDidCreate(startRebase),
-      // Once they are deleted and the rebase is complete, then we restart
-      workspaceWatcher.onDidDelete(stopRebase),
+      // When one of the rebase/bisect files are created, we set this flag to prevent restarting during that action
```
andyw8, line 364:
```suggestion
      // When one of the 'inhibit restart' files are created, we set this flag to prevent restarting during that action
```
(just so we're not maintaining the list in two places)

Earlopain:
Thanks for merging! Should make bisecting large codebases much more bearable

PR #2536, opened 2024-09-10
Uses Jekyll to generate the documentation

st0012:
### Motivation

Compared to RDoc, Jekyll is a much better platform for non-API-focused documentation. Some useful features for us are:

- Rich themes to choose from
- Easily-customizable templates
- Easy configuration with `_config.yml`
- Feature-rich markdown syntax support (e.g. supporting html tags too)

### Implementation

- Uses Jekyll to generate the documentation, which's source and configs live in the `/jekyll` folder.
- The documentation will still be built to the `/docs` folder and be ignored by git.
- Currently, the documentation is one big page (similar to rust-analyzer's).
- All the current feature demos and introduction are copied from the Code Navigation post on the Rails At Scale blog. We will gradually fill in the rest of the documentation.
- To avoid duplication, most of README's content has been moved to the documentation too.
- Because Jekyll server isn't able to link to assets outside of the source directory, we need to copy the Ruby LSP icon to `/jekyll` folder.

### Demo

https://github.com/user-attachments/assets/173dc595-74d7-4467-a562-97aa79253ec4


.github/workflows/publish_docs.yml
```diff
@@ -23,13 +23,10 @@ jobs:
...
       - name: Generate documentation
-        run: bundle exec rake rdoc
-
-      - name: Copy README images to the docs folder
+        working-directory: ./jekyll
         run: |
-          mkdir docs/vscode
-          cp vscode/icon.png docs/vscode/icon.png
-          cp -R vscode/extras docs/vscode/
+          bundle install
```
vinistock, line 28: This is never going to cache dependencies. Can we use [working-directory](https://github.com/ruby/setup-ruby?tab=readme-ov-file#working-directory) from the setup-ruby action?

README.md
```diff
@@ -15,119 +15,9 @@ experience to Ruby developers using modern standards for cross-editor features,
...
-searching for the `ruby-lsp` prefix.
-
-For instructions on how to create addons, see the [addons documentation](ADDONS.md).
-
-## Learn More
-
-* [RubyConf 2022: Improving the development experience with language servers](https://www.youtube.com/watch?v=kEfXPTm1aCI) ([Vinicius Stock](https://github.com/vinistock))
-* [Remote Ruby: Ruby Language Server with Vinicius Stock](https://remoteruby.com/221)
-* [RubyKaigi 2023: Code indexing - How language servers understand our code](https://www.youtube.com/watch?v=ks3tQojSJLU) ([Vinicius Stock](https://github.com/vinistock))
+Please refer to the official [documentation](https://shopify.github.io/ruby-lsp) for more information about [installation and usage](https://shopify.github.io/ruby-lsp#usage) and [supported features](https://shopify.github.io/ruby-lsp#features).
```
vinistock, line 20: I wonder if we should still keep at least the basics of usage in the readme, pointing people to the VS Code extension and the more extensive documentation.

jekyll/Gemfile
```diff
@@ -0,0 +1,37 @@
...
+  gem "tzinfo", ">= 1", "< 3"
+  gem "tzinfo-data"
+end
+
+# Performance-booster for watching directories on Windows
+gem "wdm", "~> 0.1.1", platforms: [:mingw, :x64_mingw, :mswin]
+
+# Lock `http_parser.rb` gem to `v0.6.x` on JRuby builds since newer versions of the gem
+# do not have a Java counterpart.
+gem "http_parser.rb", "~> 0.6.0", platforms: [:jruby]
```
vinistock, line 37: Do we need to keep these dependencies given that they are exclusive for Windows and JRuby?

jekyll/Gemfile
vinistock, line 1: Could we avoid the separate Gemfile by having an optional group in the main top level one? Something like

```ruby
group :jekyll do
  gem "jekyll"
end
```

Otherwise, we need to setup dependabot for these dependencies too.

jekyll/_config.yml
```diff
@@ -0,0 +1,63 @@
...
+# https://learnxinyminutes.com/docs/yaml/
+#
+# Site settings
+# These are used to personalize your new site. If you look in the HTML files,
+# you will see them accessed via {{ site.title }}, {{ site.email }}, and so on.
+# You can create any custom variable you would like, and they will be accessible
+# in the templates via {{ site.myvariable }}.
+
+title: Ruby LSP
+description: >- # this means to ignore newlines until "baseurl:"
```
andyw8, line 22: Will we do update this (and `url`) in a separate PR?

andyw8:
Is there somewhere else we could host the videos to avoid bloating the repo? They don't take up much space currently, but each time we update them the repo will grow.

andyw8:
Once shipped, we should link to this from https://shopify.github.io/ to help with SEO.

st0012:
> Is there somewhere else we could host the videos to avoid bloating the repo? They don't take up much space currently, but each time we update them the repo will grow.

Vini and I talked about this before, specifically with [Git LFS](https://git-lfs.com/) as the potential solution. But IMO we should focus on improving the content until RailsWorld and then worry about this later.

jekyll/Gemfile
```diff
@@ -0,0 +1,37 @@
...
+# Happy Jekylling!
+gem "jekyll", "~> 4.3.3"
+
+# Theme
+gem "just-the-docs", "~> 0.10.0"
+
+# If you want to use GitHub Pages, remove the "gem "jekyll"" above and
+# uncomment the line below. To upgrade, run `bundle update github-pages`.
+# gem "github-pages", group: :jekyll_plugins
+# If you have any plugins, put them here!
```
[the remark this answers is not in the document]
st0012: No I don't think it's for our use cases as we build the assets ourselves and publish it manually. We can later migrate to it if it's proven easier tho.

README.md
st0012: I lean towards to leading users to the official website from the very beginning, especially with the one-page design.
This way, users could immediately learn and explore other content like feature demos, or at least know there presence, when they start using Ruby LSP.

st0012:
@vinistock because `sass-embedded`, a dependency of `jekyll`, fails to build with Ruby 3.1 and takes significantly longer to build with 3.2 (due to its **build script** relying on psych 5.1+, which only uses the version bundled with Ruby). Since Jekyll is only needed for documentation, I don't want to complicate the CI setup for it.
Given that `ruby-lsp` doesn't have dependabot working atm anyway (due to `rbi` requires Ruby 3.1 constantly failing it), can we maintain a separate Gemfile for Jekyll for now?

vinistock:
@st0012 would this work?

- Put jekyll related gems in an optional Gemfile group, like `group: :documentation`
- Use The setup-ruby's action [bundle config](https://github.com/ruby/setup-ruby?tab=readme-ov-file#bundle-config) through environment variables to only include the group when generating the docs. It should just be a matter of using `"BUNDLE_WITH": "documentation"` in the config

README.md
andyw8: We can certainly cut down a lot of the content in the README but I do think we should keep a minimal explanation of what Ruby LSP does and how to install it (for VS Code at least).

st0012: I added basic usage instructions back to readme.

PR #2538, opened 2024-09-11
Add rake task to run indexer tests

vinistock:
### Motivation

I find myself wanting to run only the indexer tests first when making changes, so that I can get a quicker signal if I broke something.

I think it's worth splitting the indexer tests into its separate rake task and making the default run both. That way, `bundle exec rake` still does the same thing, but we have the opportunity to run `bundle exec rake indexer_test` to run only indexing related ones.


Rakefile
```diff
@@ -8,7 +8,13 @@ require "ruby_lsp/check_docs"
 Rake::TestTask.new(:test) do |t|
   t.libs << "test"
   t.libs << "lib"
-  t.test_files = FileList["test/**/*_test.rb", "lib/ruby_indexer/test/**/*_test.rb"].exclude("test/fixtures/prism/**/*")
+  t.test_files = FileList["test/**/*_test.rb"].exclude("test/fixtures/prism/**/*")
+end
+
+Rake::TestTask.new(:indexer_test) do |t|
```
andyw8, line 14: Could we put this inside the `test` namespace so it's `rake test:indexer`? That would be closer to conventions elsewhere, e.g. Rails.

andyw8 (approved):

Rakefile
vinistock: Great idea!

PR #2541, opened 2024-09-11
Add signature matching for positional parameters

vinistock:
### Motivation

Step towards https://github.com/Shopify/ruby-lsp/issues/2363

To determine which signature overload to show, we are going to need to match a declaration signature to the arguments of a method call. Doing it all in a single PR felt a bit too complex, so this PR only adds the method `matches?` to `Signature` and only implements matching for positional arguments. I will follow up with the implementation for keyword arguments.

### Implementation

For matching positional arguments, we essentially need to check if the number of arguments in the method call are within the range of possible arguments that the method accepts. For example:

```ruby
def bar(a, b = 2)
end

# The range for valid calls here is 1..2 because you can call `bar` with 1 or 2 arguments
bar(1) # ok!
bar(1, 2) # ok!
```

There are also special rules for splats and forwarding arguments. In general, these are not statically analyzable and we just bail out saying that the signature does match, but there are a few scenarios where we can do at least partial matching. For example:

```ruby
def bar(a, *b)
end

# We can't statically analyze anything related to the splat parameter, but we know
# that `bar` requires at least one positional parameter. In this case, the range is
# 1..INFINITY

bar # not ok!
bar(1) # ok!
bar(1, 2) # ok!
bar(*a) # ok!
```

See the tests for more examples.

### Automated Tests

Added tests.


lib/ruby_indexer/test/method_test.rb
```diff
@@ -488,5 +488,169 @@ def bar
...
+      assert_signature_matches(entry, "bar(1, ...)")
+      assert_signature_matches(entry, "bar(*a)")
+      assert_signature_matches(entry, "bar(1, *a)")
+      assert_signature_matches(entry, "bar(1, *a, 2)")
+      assert_signature_matches(entry, "bar(*a, 2)")
+      assert_signature_matches(entry, "bar(1, **a)")
+      assert_signature_matches(entry, "bar(1) {}")
+
+      # Non matching calls
+
```
andyw8, line 515: ✂️

lib/ruby_indexer/lib/ruby_indexer/entry.rb
```diff
@@ -597,6 +597,36 @@ def initialize(parameters)
...
+            max_pos += 1
+          when RestParameter
+            max_pos = Float::INFINITY
+          when ForwardingParameter
+            max_pos = Float::INFINITY
+          end
+        end
+
+        _keyword_hash_node, positional_args = arguments.partition { |arg| arg.is_a?(Prism::KeywordHashNode) }
+        unknown_argument_length = positional_args.any? do |arg|
```
andyw8, line 624:
```suggestion
        argument_length_is_unknown = positional_args.any? do |arg|
```
(since `unknown_argument_length` suggests it's a count).

lib/ruby_indexer/test/method_test.rb
```diff
@@ -488,5 +488,169 @@ def bar
       assert_equal(6, name_location.start_column)
       assert_equal(9, name_location.end_column)
     end
+
+    def test_signature_matches_the_for_a_method_with_positional_params
```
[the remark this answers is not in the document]
andyw8: (and in the test below)

andyw8 (approved):

PR #2545, opened 2024-09-12
Allow env variables in bundleGemfile vscode plugin config

jlahtinen:
### Motivation

```
{
  "rubyLsp.bundleGemfile": "../../path/to/the/directory/Gemfile",
}
```
Didn't accept env variables like

```
{
  "rubyLsp.bundleGemfile": "${HOME}/$CUSTOM/Gemfile",
}
```

### Implementation

Didn't find from node core function for this. Created a new one in vscode/src/bundleGemfileHelper.ts
Also extracted same code logic from versionManager.ts and ruby.ts to use this shared helper.

### Automated Tests

I didn't. I have too restricted environment at the moment to be able to run tests that uses window managers (I think this is the reason why yarn run test fails)

### Manual Tests

I updated my remote extension folder with extesion.js file after running yarm run package and tested env variables are expanded as expected.

### DIFF
--- a/vscode/src/bundleGemfileHelper.ts
+++ b/vscode/src/bundleGemfileHelper.ts
@@ -1,0 +1,25 @@
+import path from "path";
+import * as vscode from "vscode";
+
+function expandEnvVariables(filepath: string): string {
+  return filepath.replace(/\$\{(\w+)\}|\$(\w+)/g, (match: string, p1: string, p2: string): string => {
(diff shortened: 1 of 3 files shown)


jlahtinen:
I have multiple projects in one remote ssh server. I would like to set them all to use $HOME/ruby-lsp-custom/Gemfile without changing all of the settings to use relative paths.

One way is to use absolute paths without envs but then sharing those settings with others is a little harder. Not too big deal but I think this was easy enough to implement.

Issue #2529, opened 2024-09-06
Implement request delegation for ERB


vinistock:
Currently, we provide features only for the embedded Ruby part of ERB, but not for the host language. It's possible, although not super trivial, to add request delegation support so that features for the host languages will work in VS Code if there's a language service registered for them.

There's an example on how to do this for `<style>` tags embedded inside HTML in the samples repository https://github.com/microsoft/vscode-extension-samples/tree/main/lsp-embedded-request-forwarding.

The main idea behind this is that the extension has to implement a [document provider](https://github.com/microsoft/vscode-extension-samples/blob/7749ad8b24005e611047832b2e55bc2661055080/lsp-embedded-request-forwarding/client/src/extension.ts#L32) for virtual documents, where the virtual document includes only the content of the host language.

Using a middleware, the client then checks if a request is [inside or outside the embedded portion](https://github.com/microsoft/vscode-extension-samples/blob/7749ad8b24005e611047832b2e55bc2661055080/lsp-embedded-request-forwarding/client/src/extension.ts#L45) and then either does nothing (invoking `next` to move on to the next layer of middleware) or performs the delegation to VS Code using the virtual document's URI, which will makes VS Code look for the right language service and fulfill the features.

If you check their [embedded support](https://github.com/microsoft/vscode-extension-samples/blob/main/lsp-embedded-request-forwarding/client/src/embeddedSupport.ts) in the sample, they use a pretty similar approach as the one we're using in the server right now. A scanner separates the host language parts from the embedded language parts.

Since our scanner lives in the server, our middleware may have to be a little bit different. We may need to use a special response to tell VS Code that the request needs to be fulfilled by another language service, but we also need a way to get the host language's into the virtual document storage - this is likely the most challenging part.

Finally, one last detail that we shouldn't overlook is that ERB can be used with pretty much any host language, so we can't limit this to HTML, but instead should get the URI for the virtual document based on the original URI.

ehutzelman:
I'm assuming this would allow VS Code to correctly handle .html.erb files as HTML outside of the ERB blocks and give us back those standard HTML editing features like tag completion, formatting, etc. I've had to remap my ERB files as HTML in VS Code go get those features back since I would rather have the HTML features over the ERB features. Would be great to get them both.

PR #2551, opened 2024-09-13
Add support for delegating completion requests for ERB files

vinistock:
### Motivation

Implements the majority of what's needed for #2529. This PR only adds completion support, but it implements all of the building blocks necessary and adding the other features should be relatively straight forward after this.

This PR makes the Ruby LSP capable of delegating requests that occur in the host language of an ERB template to the appropriate language service. Before this PR, we only serve Ruby related features for the ERB files.

After this PR, we continue to serve all features for the embedded Ruby portion of the documents, but if a request is triggered for the host language (like inside the HTML part of the code), then the Ruby LSP server will signal to the client that it needs to delegate the request to the appropriate language service.

The result is that users get features for both the embedded Ruby part and the host language (provided they have a language service registered to handle the features for the host language). Since most of the time ERB is used with HTML, this should work by default in VS Code.

I will follow up adding the remaining features, which are `hover`, `definition`, `signature help` and `document highlight`.

**Note**: the HTML language server also supports references and rename, but it's not possible to delegate a request that the Ruby LSP itself doesn't implement yet, so those will have to wait.

### Important note

This implementation is heavily based on the [sample for request delegation](https://github.com/microsoft/vscode-extension-samples/tree/main/lsp-embedded-request-forwarding). However, request delegation is not yet standardized into the LSP specification, which means this needs custom implementation in the client and will only work for VS Code.

Other editors will need to implement the custom delegation too if they wish to support this feature - at least until this becomes a part of the spec.

### Implementation

Here's the flow of delegation:

1. Someone edits an ERB file
2. The server receive requests for it
3. If the document had anything new to be parsed, if it's an ERB document and the client advertised that it supports request delegation, then we notify the client with the state of the virtual host language source, so that it can be used during delegation
4. Then if we receive a completion request and the position is inside the host language, we raise a custom error to signal that the request is being delegated
5. The client then catches the error and delegates using VS Code's completion provider

It's essentially the same implementation as the sample, but instead of detecting the virtual host language source directly in the client, we use the server since that's where the ERB scanner is.

I will add comments to explain the specific changes inline.

### Automated Tests

Added tests.

### Manual Tests

https://github.com/user-attachments/assets/7e073aa8-446c-4f0f-8b10-187b9a71a438

### DIFF
--- a/lib/ruby_lsp/base_server.rb
+++ b/lib/ruby_lsp/base_server.rb
@@ -59,7 +59,7 @@ module RubyLsp
                 # If the client supports request delegation and we're working with an ERB document and there was
                 # something to parse, then we have to maintain the client updated about the virtual state of the host
                 # language source
-                if document.parse && @global_state.supports_request_delegation && document.is_a?(ERBDocument)
+                if document.parse! && @global_state.supports_request_delegation && document.is_a?(ERBDocument)
                   send_message(
                     Notification.new(
                       method: "delegate/textDocument/virtualState",
--- a/lib/ruby_lsp/base_server.rb
+++ b/lib/ruby_lsp/base_server.rb
@@ -31,6 +31,7 @@ module RubyLsp
         Thread,
       )
 
+      @global_state = T.let(GlobalState.new, GlobalState)
       Thread.main.priority = 1
     end
 
@@ -52,7 +53,26 @@ module RubyLsp
               message[:params][:textDocument][:uri] = parsed_uri
 
               # We don't want to try to parse documents on text synchronization notifications
-              @store.get(parsed_uri).parse unless method.start_with?("textDocument/did")
+              unless method.start_with?("textDocument/did")
+                document = @store.get(parsed_uri)
+
+                # If the client supports request delegation and we're working with an ERB document and there was
+                # something to parse, then we have to maintain the client updated about the virtual state of the host
+                # language source
+                if document.parse && @global_state.supports_request_delegation && document.is_a?(ERBDocument)
+                  send_message(
+                    Notification.new(
+                      method: "delegate/textDocument/virtualState",
+                      params: {
+                        textDocument: {
+                          uri: uri,
+                          text: document.host_language_source,
+                        },
+                      },
+                    ),
+                  )
+                end
+              end
             rescue Store::NonExistingDocumentError
               # If we receive a request for a file that no longer exists, we don't want to fail
             end
--- a/lib/ruby_lsp/document.rb
+++ b/lib/ruby_lsp/document.rb
@@ -52,7 +52,7 @@ module RubyLsp
       @uri = T.let(uri, URI::Generic)
       @needs_parsing = T.let(true, T::Boolean)
       @parse_result = T.let(T.unsafe(nil), ParseResultType)
-      parse
+      parse!
     end
 
     sig { params(other: Document[T.untyped]).returns(T::Boolean) }
@@ -109,7 +109,7 @@ module RubyLsp
 
     # Returns `true` if the document was parsed and `false` if nothing needed parsing
     sig { abstract.returns(T::Boolean) }
-    def parse; end
+    def parse!; end
 
     sig { abstract.returns(T::Boolean) }
     def syntax_error?; end
--- a/lib/ruby_lsp/document.rb
+++ b/lib/ruby_lsp/document.rb
@@ -51,7 +51,8 @@ module RubyLsp
       @version = T.let(version, Integer)
       @uri = T.let(uri, URI::Generic)
       @needs_parsing = T.let(true, T::Boolean)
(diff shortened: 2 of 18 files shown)


lib/ruby_lsp/base_server.rb
```diff
@@ -31,6 +31,7 @@ def initialize(test_mode: false)
         Thread,
       )
 
+      @global_state = T.let(GlobalState.new, GlobalState)
```
vinistock, line 34: I had to move global state to the base server, so that we can notify the client about the virtual source content.

lib/ruby_lsp/erb_document.rb
```diff
@@ -6,17 +6,30 @@ class ERBDocument < Document
...
+    attr_reader :host_language_source
+
     ParseResultType = type_member { { fixed: Prism::ParseResult } }
 
-    sig { override.returns(ParseResultType) }
+    sig { params(source: String, version: Integer, uri: URI::Generic, encoding: Encoding).void }
+    def initialize(source:, version:, uri:, encoding: Encoding::UTF_8)
+      # This has to be initialized before calling super because we call `parse` in the parent constructor, which
+      # overrides this with the proper virtual host language source
+      @host_language_source = T.let("", String)
```
vinistock, line 18: I started calling this host language source, since ERB is not only used with HTML.

lib/ruby_lsp/global_state.rb
```diff
@@ -21,7 +21,7 @@ class GlobalState
     attr_reader :encoding
 
     sig { returns(T::Boolean) }
-    attr_reader :supports_watching_files, :experimental_features
+    attr_reader :supports_watching_files, :experimental_features, :supports_request_delegation
```
vinistock, line 24: Since this is custom and not a part of the specification, we need to know if the client has support for it before raising the custom signals or sending custom notifications.

vscode/src/client.ts
```diff
@@ -224,6 +229,25 @@ class ClientErrorHandler implements ErrorHandler {
   }
 }
 
+class ExperimentalCapabilities implements StaticFeature {
```
vinistock, line 232: This might seem complicated, but all we need is adding the extra experimental capabilities in `fillClientCapabilities`. We don't use anything else.

vscode/src/client.ts
```diff
@@ -273,6 +301,13 @@ export default class Client extends LanguageClient implements ClientInterface {
     this.#context = context;
     this.ruby = ruby;
     this.#formatter = "";
+
+    this.onNotification("delegate/textDocument/virtualState", (params) => {
```
vinistock, line 305: When we receive the custom notification, we remember the virtual source for the host language in a hash.

vscode/src/client.ts
```diff
@@ -379,6 +420,58 @@ export default class Client extends LanguageClient implements ClientInterface {
...
+    }
+
+    const hostLanguage = /\.([^.]+)\.erb$/.exec(originalUri)?.[1] || "html";
+    const vdocUriString = `embedded-content://${hostLanguage}/${encodeURIComponent(
+      originalUri,
+    )}.${hostLanguage}`;
+
+    if (request === "textDocument/completion") {
+      return vscode.commands
+        .executeCommand<CompletionList>(
```
vinistock, line 442: Call the completion provider for the delegated URI, which will invoke whatever language service is registered to handle it.

lib/ruby_lsp/document.rb
```diff
@@ -51,7 +51,8 @@ def initialize(source:, version:, uri:, encoding: Encoding::UTF_8)
       @version = T.let(version, Integer)
       @uri = T.let(uri, URI::Generic)
       @needs_parsing = T.let(true, T::Boolean)
-      @parse_result = T.let(parse, ParseResultType)
+      @parse_result = T.let(T.unsafe(nil), ParseResultType)
+      parse
```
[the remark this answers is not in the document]
st0012: Nit: if `parse` doesn't return the result anymore but only assign it to another attribute, IMO it should be renamed to `parse!`

lib/ruby_lsp/requests/request.rb
```diff
@@ -17,6 +17,23 @@ def perform; end
...
+        params(
+          global_state: GlobalState,
+          document: Document[T.untyped],
+          char_position: Integer,
+        ).void
+      end
+      def delegate_request_if_needed!(global_state, document, char_position)
+        if global_state.supports_request_delegation &&
+            document.is_a?(ERBDocument) &&
+            document.inside_host_language?(char_position)
```
st0012, line 32: I like this chained condition as they make the intention very clear 👍

st0012 (approved):
Left a few comments but it looks great 👍 
I think all the PR comments in the extension changes could also be code comments tho.

st0012 (approved):

PR #2560, opened 2024-09-16
Remove unused documentation utilities

st0012:
### Motivation

Since we now use Jekyll to generate documentation, some previous documentation utilities can be removed.

### Implementation

- We don't need the RDoc declaration and tasks anymore
- We don't need the check_docs.rb file anymore
- We also need to update CI configs to account for the above changes


test/setup_bundler_test.rb
```diff
@@ -639,8 +639,15 @@ def run_script(path = Dir.pwd, expected_path: nil, **options)
   end
 
   def stub_bundle_with_env(env, command = "(bundle check || bundle install) 1>&2")
+    puts "????????????"
     Object.any_instance.expects(:system).with do |actual_env, actual_command|
+      $stderr.puts "!!!!!!!!!!!!!!!!"
       actual_env.delete_if { |k, _v| k.start_with?("BUNDLE_PKGS") }
+      puts "================"
```
vinistock, line 646: ✂️

vinistock (approved):

test/setup_bundler_test.rb
st0012: 🤦‍♂️

PR #2561, opened 2024-09-17
Consolidate docs to Jekyll site

st0012:

PR #2563, opened 2024-09-17
Delegate all possible features for ERB

### LINKED ISSUE #2529: Implement request delegation for ERB

vinistock:
### Motivation

Closes #2529

This PR starts delegating the remaining features that occur in the host language part of an ERB file to the appropriate language services.

The features that will be available for the host languages are now: definition, hover, signature help, document highlight, folding range and document link.

#### JavaScript embedded in HTML in embedded Ruby

When using ERB in an HTML template, there's double delegation happening if the user is writing CSS or JavaScript. For example, consider this ERB template:

```erb
<% @users.each do |user| %>
  <button onclick="document."></button>
<% end %>
```

When typing the dot inside the `onclick` attribute, the HTML language server is in JavaScript mode, delegating all requests to the JS service.

This is currently working, but I noticed that when the double-delegation is in place the completion results for JavaScript are not exactly the same as when you use a regular HTML file. I created a [discussion](https://github.com/microsoft/vscode-discussions/discussions/1628) for this because I'm not sure if this is an issue with our implementation, with VS Code or with the `html-language-features` extension.

After #2561 goes in, I can add some documentation highlighting that this is a current limitation.

#### Semantic highlighting

The only missing feature that the Ruby LSP does implement is semantic highlighting. It is possible to support delegation for it by firing two requests and then concatenating the responses, but that requires that the Ruby LSP server stops encoding the tokens for ERB documents, so that they can be concatenated, ordered and only then encoded in the client.

It's not very trivial and I'm not sure how semantic tokens delta would work in this case. We would probably need to re-implement delta on the client so that we could compute deltas between fully merged semantic token responses. For now, I'm going to skip semantic highlighting due to the complexity.

#### Other features

Request delegation can only work for features that are supported by the Ruby LSP, so references and rename support are currently missing.

### Implementation

For position based requests, we just followed the exactly same approach as we did for completion. For full document requests, we need to delegate the request, run the real request on the server and merge the responses.

This allows you to get features for the mixed document, like being able to fold both the Ruby and the HTML code in an ERB template at the right places.

### Automated Tests

Added tests.

### DIFF
--- a/dev.yml
+++ b/dev.yml
@@ -18,8 +18,7 @@ commands:
     aliases: ["tc"]
     run: "bundle exec srb tc"
   ci:
-    run: "bundle exec rake ruby_lsp:check_docs &&
-      bundle exec srb tc &&
+    run: "bundle exec srb tc &&
       bin/rubocop &&
       bin/test &&
       bundle exec exe/ruby-lsp-check"
--- a/lib/ruby_lsp/requests/completion.rb
+++ b/lib/ruby_lsp/requests/completion.rb
@@ -17,7 +17,7 @@ module RubyLsp
         def provider
           Interface::CompletionOptions.new(
             resolve_provider: true,
-            trigger_characters: ["/", "\"", "'", ":", "@", "."],
+            trigger_characters: ["/", "\"", "'", ":", "@", ".", "=", "<"],
             completion_item: {
               labelDetailsSupport: true,
             },
--- a/vscode/src/client.ts
+++ b/vscode/src/client.ts
@@ -424,6 +424,9 @@ export default class Client extends LanguageClient implements ClientInterface {
     }
   }
 
+  // Delegate a request to the appropriate language service. Note that only position based requests are delegated here.
+  // Full file requests, such as folding range, have their own separate middleware to merge the embedded Ruby + host
+  // language responses
   private async executeDelegateRequest(
     type: string | MessageSignature,
     params: any,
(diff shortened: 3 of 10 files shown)


dev.yml
```diff
@@ -18,8 +18,7 @@ commands:
     aliases: ["tc"]
     run: "bundle exec srb tc"
   ci:
-    run: "bundle exec rake ruby_lsp:check_docs &&
-      bundle exec srb tc &&
+    run: "bundle exec srb tc &&
```
vinistock, line 21: The `check_docs` task was removed and I'm a big user of this little helper.

lib/ruby_lsp/requests/completion.rb
```diff
@@ -36,7 +36,7 @@ class << self
         def provider
           Interface::CompletionOptions.new(
             resolve_provider: true,
-            trigger_characters: ["/", "\"", "'", ":", "@", "."],
+            trigger_characters: ["/", "\"", "'", ":", "@", ".", "="],
```
vinistock, line 39: To delegate requests properly, the server needs to register for the expected characters for delegation. For HTML attributes, we want to show completion after the equals.

Earlopain: If I understand this correctly, should it include `<` for tag name completion? Seems to work if I add it and feels better.

Earlopain: `*` as well for emmet abbreviations like `div*5`, though that doesn't work for some reason. Adding `"*"` does get one character further (the `*`) but when typing the number it disappears. Doesn't seem to work just like that.

They can start with a few characters (`.` for class, `#` for id), and maybe more.

https://code.visualstudio.com/docs/editor/emmet
https://docs.emmet.io/abbreviations/

andyw8 (approved):

vscode/src/client.ts
```diff
@@ -471,6 +471,31 @@ export default class Client extends LanguageClient implements ClientInterface {
 
           return response;
         });
+    } else if (request === "textDocument/hover") {
```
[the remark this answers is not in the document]
vinistock: I don't mind either style to be honest. I changed to `switch`.

Issue #2227, opened 2024-06-26
Hover Definitions for Ruby Keywords


jamesmengo:
## Set up the tooltips

Add tooltips to these keywords. The markdown files should have the description below and a newbie-friendly description.

Try to avoid having tooltips which are over-zealous (e.g. `DefNode` will work on any keyword within a method definition).

```[tasklist]
### Tooltips
- [ ] `__ENCODING__` - The script encoding of the current file.
- [ ] `__LINE__` - The line number of this keyword in the current file.
- [ ] `__FILE__` - The path to the current file.
- [ ] `BEGIN` - Runs before any other code in the current file.
- [ ] `END` - Runs after any other code in the current file.
- [x] `alias` - Creates an alias between two methods (and other things).
- [x] `and` - Short-circuit Boolean and with lower precedence than `&&`.
- [x] `begin` - Starts an exception handling block.
- [x] `break` - Leaves a block early.
- [x] `case` - Starts a case expression.
- [x] `class` - Creates or opens a class.
- [x] `def` - Defines a method.
- [x] `defined?` - Returns a string describing its argument.
- [x] ~`do` - Starts a block.~
- [x] `else` - The unhandled condition in `case`, `if` and `unless` expressions.
- [ ] `elsif` - An alternate condition for an `if` expression.
- [ ] `end` - The end of a syntax block. Used by classes, modules, methods, exception handling and control expressions.
- [x] `ensure` - Starts a section of code that is always run when an exception is raised.
- [x] `false` - Boolean false.
- [x] `for` - A loop that is similar to using the `each` method.
- [x] `if` - Used for `if` and modifier `if` expressions.
- [x] `in` - Used to separate the iterable object and iterator variable in a `for` loop.
- [x] `module` - Creates or opens a module.
- [x] `next` - Skips the rest of the block.
- [x] `nil` - A false value usually indicating “no value” or “unknown”.
- [ ] `not` - Inverts the following boolean expression. Has a lower precedence than `!`.
- [x] `or` - Boolean or with lower precedence than `||`.
- [x] `redo` - Restarts execution in the current block.
- [x] `rescue` - Starts an exception section of code in a `begin` block.
- [x] `retry` - Retries an exception block.
- [x] `return` - Exits a method.
- [x] `self` - The object the current method is attached to.
- [x] `super` - Calls the current method in a superclass.
- [ ] `then` - Indicates the end of conditional blocks in control structures.
- [x] `true` - Boolean true.
[...]

PR #2581, opened 2024-09-19
Add infrastructure for keyword documentation

vinistock:
### Motivation

This PR adds the basic infrastructure for #2227, so that we can slowly document keywords and even receive help from the community in doing so.

### Implementation

The idea is to show a short sentence on hover/completion with a link to longer form documentation.

The way I went about this is to add a separate directory with all of the static documentation in Markdown files. Those files are never read by the LSP, we just link to them from the hover markdown. Clicking on the link will open the documentation on the user's editor.

After this infrastructure is merged, the work to document more keywords is simply:

1. Add the related Prism class to the list of `ALLOWED_TARGETS`
2. Register for the event on the dispatcher (e.g.: `:on_next_node_enter`)
3. Define the exact same handler as yield
```ruby
sig { params(node: Prism::SomeKeywordNode).void }
def on_some_keyword_node_enter(node)
  show_keyword_documentation(node.keyword)
end
```
4. Write both the short and long documentation, in `static_docs.rb` and `static_docs/some_keyword.md`, respectively

### Automated Tests

Added tests.

### Manual Tests

1. On this branch, hover over yield
2. Verify you see the documentation
3. Click the read more link
4. Verify you land on the `yield.md` file


lib/ruby_lsp/listeners/hover.rb
```diff
@@ -193,6 +200,18 @@ def generate_heredoc_hover(node)
         end
       end
 
+      sig { params(keyword: String).void }
+      def show_keyword_documentation(keyword)
```
andyw8, line 204:
```suggestion
      def handle_keyword_documentation(keyword)
```
(for consistency with other methods in `Hover`).

lib/ruby_lsp/static_docs.rb
```diff
@@ -0,0 +1,15 @@
+# typed: strict
+# frozen_string_literal: true
+
+module RubyLsp
+  # The path to the `static_docs` directory, where we keep long form static documentation
```
andyw8, line 5:
```suggestion
  # The path to the `static_docs` directory, where we keep long-form static documentation
```

static_docs/yield.md
```diff
@@ -0,0 +1,81 @@
...
+
+## Block parameter
+
+In addition to implicit blocks, Ruby also allows developers to use explicit block parameters as part of the method's
+signature. In this scenario, we can use the reference to the block directly instead of relying on the `yield` keyword.
+
+```ruby
+# Block parameters are prefixed with & and a name
+def foo(&my_block_param)
+  # If a block was passed to `foo`, my_block_param will be a `Proc` object. Otherwise, it will be `nil`. We can use
```
andyw8, line 71:
```suggestion
  # If a block was passed to `foo`, `my_block_param` will be a `Proc` object. Otherwise, it will be `nil`. We can use
```

andyw8 (approved):

PR #2589, opened 2024-09-20
Remove all autoloads

vinistock:
### Motivation

Sorbet can only unwrap methods back in place if they are already loaded by the time we run `T::Utils.run_all_sig_blocks`. However, reflecting about it, that's never going to be the case because we autoload everything.

There's really not much of a point in doing so, since the LSP needs its entire code loaded anyway. By eagerly requiring everything, we ensure that Sorbet is unwrapping the methods back and not calling them through `bind_call`.

### Implementation

Just replaced all autoloads with requires. I also found an outdated `autoload` to `rails_documentation_client`, which has been moved to the Rails addon a long time ago, but we missed this autoload. Since the class is never referred to in code, the autoload never triggers and we never noticed the path doesn't exist.

### Automated Tests

Existing tests should cover it.


andyw8 (approved):
Would Zeitwerk be an option here?

lib/ruby_lsp/internal.rb
```diff
@@ -30,15 +30,54 @@
...
+require "ruby_lsp/requests/support/sorbet"
+require "ruby_lsp/requests/support/common"
+require "ruby_lsp/requests/support/formatter"
 require "ruby_lsp/requests/support/rubocop_runner"
 require "ruby_lsp/requests/support/rubocop_formatter"
 require "ruby_lsp/requests/support/syntax_tree_formatter"
+
+# Requests
+require "ruby_lsp/requests/request"
+require "ruby_lsp/requests/document_symbol"
```
andyw8, line 62: Let's order alphabetically?

vinistock: I ordered only the requests, but since there are dependencies between these files, we need to maintain the rest of the require order.

For example, `Request` needs to exist before we can inherit from that class, so it needs to be required first.

PR #2595, opened 2024-09-23
Fix bundler env

epoberezhny:
### Motivation

Bundler has [array configs](https://github.com/rubygems/rubygems/blob/master/bundler/lib/bundler/settings.rb#L73) like `BUNDLE_WITHOUT` and `settings[e]` [returns array](https://github.com/rubygems/rubygems/blob/master/bundler/lib/bundler/settings.rb#L438) that later converted to string incorrectly (e.g. `'[:development, :test]'`, but must be `'development:test'`). Because of that bundler tries to install all gems in Gemfile, when `ruby-lsp` is not included in your application's Gemfile

### Automated Tests

Couldn't find any related tests, so tests are missing.

### Manual Tests

You need to have a group of gems in your Gemfile. Then you should run `bundle config set without 'YOUR_GROUP'`. And you should not have gems in this group installed after the LSP server starts

### DIFF
--- a/lib/ruby_lsp/setup_bundler.rb
+++ b/lib/ruby_lsp/setup_bundler.rb
@@ -255,7 +255,9 @@ module RubyLsp
       # "vendor/bundle"`
       settings.all.to_h do |e|
         key = Bundler::Settings.key_for(e)
-        [key, settings[e].to_s]
+        value = Array(settings[e]).join(":").tr(" ", ":")
+
+        [key, value]
       end
     end
 
--- a/test/setup_bundler_test.rb
+++ b/test/setup_bundler_test.rb
@@ -600,6 +600,31 @@ class SetupBundlerTest < Minitest::Test
     end
   end
 
+  def test_uses_correct_bundler_env_when_there_is_bundle_config
+    Dir.mktmpdir do |dir|
+      Dir.chdir(dir) do
+        File.write(File.join(dir, "Gemfile"), <<~GEMFILE)
+          source "https://rubygems.org"
+          gem "irb"
+        GEMFILE
+
+        Bundler.with_unbundled_env do
+          system("bundle config set --local with production staging")
+
+          assert_path_exists(File.join(dir, ".bundle", "config"))
+
+          capture_subprocess_io do
+            system("bundle install")
+
+            env = run_script(dir)
+
+            assert_equal("production:staging", env["BUNDLE_WITH"])
+          end
+        end
+      end
+    end
+  end
+
   private
 
   def with_default_external_encoding(encoding, &block)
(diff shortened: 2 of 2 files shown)


epoberezhny:
I've signed the CLA!

lib/ruby_lsp/setup_bundler.rb
```diff
@@ -255,7 +255,9 @@ def bundler_settings_as_env
       # "vendor/bundle"`
       settings.all.to_h do |e|
         key = Bundler::Settings.key_for(e)
-        [key, settings[e].to_s]
+        value = Array(settings[e]).join(":")
```
vinistock, line 258: Does Bundler always use `:` as the separator? Or would it use `;` on Windows?

vinistock (commented):
Thank you for the PR! We definitely missed this.

We should add a test for this to prevent regression. [This test](https://github.com/Shopify/ruby-lsp/blob/95c35f6e013e323eaeb818b0adaee7c5a7b5104d/test/setup_bundler_test.rb#L576) isn't exactly the same, but it also mutates project specific Bundler settings.

Can we please add a test case that configures any Bundler setting as an array so that we can confirm that the environment is mutated properly?

lib/ruby_lsp/setup_bundler.rb
vinistock: Oh, since they have helper methods for those, we should probably just use them directly instead of re-implementing it on our side.

Can we change it to use their API?

```ruby
value = settings.array_to_s(settings[e]).to_s
```

epoberezhny: Unfortunately, `array_to_s` is private 😞

test/setup_bundler_test.rb
```diff
@@ -658,7 +692,9 @@ def bundle_env(base_path = Dir.pwd, bundle_gemfile = "Gemfile")
 
     env = settings.all.to_h do |e|
       key = Bundler::Settings.key_for(e)
-      [key, settings[e].to_s]
+      value = Array(settings[e]).join(":")
```
vinistock, line 695:
```suggestion
      value = Array(settings[e]).join(":").tr(" ", ":")
```

vinistock (approved):
Looks great! Let's just ensure we invoke `tr` too and we can ship

test/setup_bundler_test.rb
```diff
@@ -600,6 +600,37 @@ def test_respects_overridden_bundle_path_when_there_is_bundle_config
...
+        File.write(File.join(dir, "Gemfile"), <<~GEMFILE)
+          source "https://rubygems.org"
+          gem "irb"
+        GEMFILE
+
+        Bundler.with_unbundled_env do
+          system("bundle config set --local without development test")
+          system("bundle config set --local with production")
+          system("bundle config set --local jobs 1")
+          system("bundle config set --local clean true")
```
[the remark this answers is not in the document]
epoberezhny: Sure, let's try your suggestion

epoberezhny:
> Thank you for the contribution!

Thank you for creating Ruby LSP and quick responses!

PR #2597, opened 2024-09-23
Never set shell on Windows for any version manager

vinistock:


andyw8 (approved):
Can we ask a Windows user to verify?

vinistock:
Given that it's currently broken, we can probably ship a release and ask them to report back. FWIW, we had the same issue with `RubyInstaller` and it got fixed properly after a similar change.

PR #2603, opened 2024-09-23
Generalize scope so that it can be used with any locals

vinistock:
### Motivation

There will be other cases where we want to track proper scope of locals. We can currently only track parameters, which is only used for semantic highlighting, but we can use this to provide renaming and references for locals too.

This PR turns our `ParameterScope` into a more general `Scope` implementation, so that we can use it in other places.

### Implementation

I recommend reviewing by commit:

1. Renamed the class
2. Renamed the `<<` method to `add`, so that we can pass more parameters
3. Generalized the scope

### Automated Tests

Adapted existing tests.

### DIFF
--- a/lib/ruby_lsp/listeners/semantic_highlighting.rb
+++ b/lib/ruby_lsp/listeners/semantic_highlighting.rb
@@ -128,39 +128,39 @@ module RubyLsp
       sig { params(node: Prism::BlockParameterNode).void }
       def on_block_parameter_node_enter(node)
         name = node.name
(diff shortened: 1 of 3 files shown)


alexcrocha (approved):
Looks good to me. 👍 Are there any potential pitfalls or edge cases we should be aware of with this change?

vinistock:
Not for now. It's mostly just renaming and refactoring things. When we add the ability to track location of locals then it might become more complicated.

PR #2610, opened 2024-09-24
Only require RuboCop diagnostic inside the formatter

vinistock:
### Motivation

The RuboCop diagnostic class should only be required if RuboCop is present. In #2589, I mistakenly moved the require to `internal.rb`, but trying to require it when RuboCop is not present fails.

### Implementation

Moved the require inside the class that needs the diagnostic class. That require will only occur if RuboCop is present.

### Manual tests

1. Remove all rubocop related gems and syntax tree from the Gemfile
2. Reload VS Code
3. Verify it doesn't crash

### DIFF
--- a/lib/ruby_lsp/internal.rb
+++ b/lib/ruby_lsp/internal.rb
@@ -47,7 +47,6 @@ require "ruby_lsp/response_builders/semantic_highlighting"
 require "ruby_lsp/response_builders/signature_help"
 
 # Request support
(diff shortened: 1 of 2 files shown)


andyw8 (approved):

andyw8:
Let's ship this but I like @Earlopain's suggestion in the other PR of having a CI job that doesn't have RuboCop available.

vinistock:
Fair. We should totally add that, but I do want to cut the release with the fix asap.

PR #2616, opened 2024-09-24
Avoid 'puts' in addons example

andyw8:
Using `puts` can break the LSP client/server connection.


andyw8:
> Should be fine since https://github.com/Shopify/ruby-lsp/pull/2514?

It should be but we want this to match with an upcoming blog post about the add-on system.

PR #2618, opened 2024-09-25
Avoid a deprecation warning on Ruby 3.4 with the uri gem

Earlopain:
### Motivation

Calling `URI::DEFAULT_PARSER.escape` emits a deprecation warning with the uri gem shipped with Ruby 3.4:
> lib/core_ext/uri.rb:23: warning: URI::RFC3986_PARSER.escape is obsoleted. Use URI::RFC2396_PARSER.escape explicitly.

### Implementation

Check for the precence of the new `RFC2396_PARSER` constant and use that if available. Otherwise fall back to `DEFAULT_PARSER`, which will be an `RFC2396_Parser` instance in that case.

Tapioca changes: https://github.com/Shopify/tapioca/pull/2026

### Automated Tests

n/A

### Manual Tests

Run tests with ruby-head and warnings enabled: `RUBYOPT=-w VERBOSE=1 bundle exec rake`

### DIFF
--- a/lib/core_ext/uri.rb
+++ b/lib/core_ext/uri.rb
@@ -3,6 +3,11 @@
 
 module URI
   class Generic
+    # Avoid a deprecation warning with Ruby 3.4 where the default parser was changed to RFC3986.
+    # This condition must remain even after support for 3.4 has been dropped for users that have
+    # `uri` in their lockfile, decoupling it from the ruby version.
+    PARSER = T.let(const_defined?(:RFC2396_PARSER) ? RFC2396_PARSER : DEFAULT_PARSER, RFC2396_Parser)
+
     class << self
       extend T::Sig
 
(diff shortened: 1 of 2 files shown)


lib/core_ext/uri.rb
```diff
@@ -3,19 +3,22 @@
 
 module URI
   class Generic
+    # Avoid a deprecation warning with Ruby 3.4 where the default parser was changed to RFC3986
+    PARSER = T.let(const_defined?(:RFC2396_PARSER) ? RFC2396_PARSER : DEFAULT_PARSER, RFC2396_Parser)
```
vinistock, line 7: If the default changed, can we use the new default? Or is there some sort of incompatibility?

Earlopain: You can't, there is no equivalent in the RFC3986 parser for this. https://github.com/ruby/uri/blob/be858ed90548425eb278cee2116947c7b749cb4e/lib/uri/rfc3986_parser.rb#L155-L160 I have no idea why it is this way.

I'll get this PR ready tomorrow.

vinistock (approved):

PR #2619, opened 2024-09-25
Handle multibyte characters in indexing

NotFounds:
### Motivation

issue: https://github.com/Shopify/ruby-lsp/issues/1251
prev PR: https://github.com/Shopify/ruby-lsp/pull/2051#issuecomment-2371386060

When a Ruby file contains multibyte characters (such as Japanese, Chinese, emojis, etc.), the "go to definition" and hover features do not work correctly. The definition location or hover documentation will be incorrect.

This issue arises because the current implementation assumes single-byte characters when calculating offsets during index building and document referencing. We need to properly handle multibyte characters to ensure these features work reliably for all users.

### Implementation

I implemented this based on the following comment to resolve the handling problem of multibyte characters in indexing.

> Let's take the approach of saving the code units with the new Prism API in the index. So essentially, we need to:
> 
> 1. Configure the index with the encoding that was negotiated between editor and server. After setting the global state encoding [here](https://github.com/Shopify/ruby-lsp/blob/95c35f6e013e323eaeb818b0adaee7c5a7b5104d/lib/ruby_lsp/global_state.rb#L111), you want to grab the `@index.configuration` object and set the encoding there, so that we can check what is the encoding being used during indexing
> 2. You then need to pass the encoding (or maybe the entire config object?) to the [declaration listener](https://github.com/Shopify/ruby-lsp/blob/main/lib/ruby_indexer/lib/ruby_indexer/declaration_listener.rb), where we will use the encoding to invoke the Prism location API `location.start_code_units_column(encoding)` to get the proper locations for multibyte characters
> 3. Finally, we should add a few tests to ensure that we don't accidentally regress. One test per entity type should be okay. These would be:
>    
>    * [A class and a module](https://github.com/Shopify/ruby-lsp/blob/main/lib/ruby_indexer/test/classes_and_modules_test.rb)
>    * [A constant](https://github.com/Shopify/ruby-lsp/blob/main/lib/ruby_indexer/test/constant_test.rb)
>    * [An instance variable](https://github.com/Shopify/ruby-lsp/blob/main/lib/ruby_indexer/test/instance_variables_test.rb)
>    * [A method](https://github.com/Shopify/ruby-lsp/blob/main/lib/ruby_indexer/test/method_test.rb)

ref. https://github.com/Shopify/ruby-lsp/pull/2051#issuecomment-2368350689

### Automated Tests

Add a test case for each entity types.

### DIFF
--- a/lib/ruby_indexer/lib/ruby_indexer/entry.rb
+++ b/lib/ruby_indexer/lib/ruby_indexer/entry.rb
@@ -512,7 +512,13 @@ module RubyIndexer
 
       sig { params(target: String, unresolved_alias: UnresolvedConstantAlias, encoding: Encoding).void }
       def initialize(target, unresolved_alias, encoding)
-        super(unresolved_alias.name, unresolved_alias.file_path, unresolved_alias.location, unresolved_alias.comments, encoding)
+        super(
+          unresolved_alias.name,
+          unresolved_alias.file_path,
+          unresolved_alias.location,
+          unresolved_alias.comments,
+          encoding
+        )
 
         @visibility = unresolved_alias.visibility
         @target = target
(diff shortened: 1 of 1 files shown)


NotFounds:
Oops! I fixed the type error and formatting problem.

PR #435, opened 2023-01-06
Set up cspell (via Code Spell Checker extension)

andyw8:
### Motivation

Catch mistakes automatically. I have it installed locally, and I've been adding entries to my local dictionary, but I thought it would be beneficial to have as a shared dictionary.

(Partly addresses https://github.com/Shopify/ruby-dev-exp-issues/issues/590)

### Implementation

I used the command line `cspell` app to generate a list of the unknown words.

We can add this on other repos if folk like it.

PR #2622, opened 2024-09-25
Run `cspell` in CI

andyw8:
I [added](https://github.com/Shopify/ruby-lsp/pull/435) a VS Code `cspell` configuration a while ago but it relies on people having the extension enabled, and noticing the things it detects. Instead, let's enforce this in CI. I've verified that the extension is still able to read the config.

### DIFF
--- a/lib/ruby_indexer/test/instance_variables_test.rb
+++ b/lib/ruby_indexer/test/instance_variables_test.rb
@@ -209,7 +209,7 @@ module RubyIndexer
         end
       RUBY
 
-      # If the surrounding method is behind defined on any dynamic value that isn't `self`, then we attribute the
+      # If the surrounding method is being defined on any dynamic value that isn't `self`, then we attribute the
       # instance variable to the wrong owner since there's no way to understand that statically
       entry = T.must(@index["@a"]&.first)
       owner = T.must(entry.owner)


.github/workflows/ci.yml
```diff
@@ -3,6 +3,21 @@ name: CI
...
+      - name: Set up Node
+        env:
+          DISABLE_V8_COMPILE_CACHE: "1"
+        uses: actions/setup-node@v4
+        with:
+          node-version: "20.9"
+          cache: "yarn"
+
+      - name: Run cspell
+        run: yarn global add cspell && cspell
```
andyw8, line 20: TODO: how should we set this up locally in `dev.yml`? I'm hesitant about introducing a `package.json` for just this one thing.

project-words
```diff
@@ -0,0 +1,85 @@
+activestorage
```
andyw8, line 1: I generated the list by running `cspell` then used GPT to turn the output into a list, but there may be a better way.

dirceu:
Oh nice, that caught a bunch of stuff already!

Do we _have_ to check in that new `yarn.lock` file?

.github/workflows/ci.yml
```diff
@@ -3,6 +3,22 @@ name: CI
...
+      - name: Set up Node
+        env:
+          DISABLE_V8_COMPILE_CACHE: "1"
+        uses: actions/setup-node@v4
+        with:
+          node-version: "20.9"
+
+      - name: Run cspell
+        run: |
+          yarn
```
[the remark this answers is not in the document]
andyw8: Isn't that flag redundant if we're not committing `yarn.lock`?

test/fixtures/prism
vinistock, line 1: This shouldn't be a part of this PR.

vinistock (approved):
I'm slightly hesitant of making this a required part of CI. I agree that we should avoid typos and misspellings whenever possible, but I wonder if many PRs will end up being blocked on unknown terms and require regular updates to the cspell configuration.

I think there's no harm in trying this out, but if it becomes really annoying, then I will suggest removing the check.

PR #2626, opened 2024-09-25
Add rename support for constants

### LINKED ISSUE #57: Add rename support to the `ruby-lsp`
Full support for renaming things, like method names and constants is pretty difficult and should be left for the Sorbet LSP. However, our LSP can support renaming some simple things that don't require complete type checking, like:
- local variables
- method/block/lambda arguments

Remember that our LSP doesn't parse every file in the code base, so our renames are limited to the current file.

Documentation: https://microsoft.github.io/language-server-protocol/specification#textDocument_rename

vinistock:
### Motivation

Closes #57

This PR adds rename support for constants, modules and classes.

### Implementation

1. Created a `ReferenceFinder`. This is essentially a simplified copy of our declaration listener. I believe that either the declaration listener can be modified to support collecting references or we can extract some abstraction for both to use. However, I would like to start with a copy, so that we can see the patterns emerge as we add support for more renames
2. Then I added the rename request

### Automated Tests

Added tests.


vinistock:
There's one thing I missed: for documents currently opened in the UI, we should be reading their contents from the state, since you can rename content that was not saved yet. Only documents that are not being handled by the client should be read from disk.

I'll update the PR.

lib/ruby_indexer/test/reference_finder_test.rb
```diff
@@ -0,0 +1,43 @@
+# typed: true
+# frozen_string_literal: true
+
+require "test_helper"
+
+module RubyIndexer
+  class ReferenceFinderTest < Minitest::Test
+    def test_finds_constant_references
```
st0012, line 8: Can we have tests to cover more scenarios? Especially around singletons and top-level namespaces, like:

```rb
class Foo
  class << self
    Bar # should be a reference
  end
end
```

And maybe 

```rb
class Foo
  class ::Baz
    Bar # shouldn't be a reference
  end
end
```

vinistock: Added more tests 👍

lib/ruby_lsp/requests/rename.rb
```diff
@@ -0,0 +1,184 @@
...
+          # Do not rename files that are not part of the workspace
+          next unless entry.file_path.start_with?(@global_state.workspace_path)
+
+          case entry
+          when RubyIndexer::Entry::Class, RubyIndexer::Entry::Module, RubyIndexer::Entry::Constant,
+               RubyIndexer::Entry::ConstantAlias, RubyIndexer::Entry::UnresolvedConstantAlias
+
+            file_from_constant_name = short_name
+              .gsub(/([a-z])([A-Z])|([A-Z])([A-Z][a-z])/, '\1\3_\2\4')
+              .downcase
```
st0012, line 109: I feel this gsub and downcase can be extracted to a helper method with a clear name.

lib/ruby_lsp/requests/rename.rb
```diff
@@ -0,0 +1,184 @@
...
+          document: T.any(RubyDocument, ERBDocument),
+          params: T::Hash[Symbol, T.untyped],
+        ).void
+      end
+      def initialize(global_state, store, document, params)
+        super()
+        @global_state = global_state
+        @store = store
+        @document = document
+        @params = params
```
st0012, line 26: nit: what do you think about storing `position` and `newName` in ivars instead? I feel it improves readability by stating what params we need directly in the constructor.

lib/ruby_lsp/requests/rename.rb
```diff
@@ -0,0 +1,184 @@
...
+            end
+          end
+        end
+      end
+
+      sig { params(fully_qualified_name: String, name: String).returns(T::Hash[String, T::Array[Interface::TextEdit]]) }
+      def collect_text_edits(fully_qualified_name, name)
+        changes = {}
+
+        Dir.glob(File.join(@global_state.workspace_path, "**/*.rb")).each do |path|
```
st0012, line 131: Is it possible to apply renames to erb files too?

lib/ruby_indexer/lib/ruby_indexer/reference_finder.rb
```diff
@@ -0,0 +1,264 @@
...
+        fully_qualified_name: String,
+        index: RubyIndexer::Index,
+        dispatcher: Prism::Dispatcher,
+        parse_result: Prism::ParseResult,
+      ).void
+    end
+    def initialize(fully_qualified_name, index, dispatcher, parse_result)
+      @fully_qualified_name = fully_qualified_name
+      @index = index
+      @inside_def = T.let(false, T::Boolean)
```
andyw8, line 38: Unused?

lib/ruby_lsp/requests/rename.rb
andyw8: Since `position` is the only thing we use from `@params`, should we pass it directly?

andyw8: Oh, `newName` too, so maybe not.

jekyll/index.markdown
```diff
@@ -45,6 +45,7 @@ Want to discuss Ruby developer experience? Consider joining the public
     - [Show syntax tree](#show-syntax-tree)
     - [ERB support](#erb-support)
     - [Guessed types](#guessed-types)
+    - [Rename](#rename)
```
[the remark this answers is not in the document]
vinistock: I'm not opposed to this, but do you feel like we need to mention symbol? The name of the feature is just rename.

lib/ruby_lsp/requests/rename.rb
vinistock: Moved it to a private method.

test/requests/rename_test.rb
```diff
@@ -0,0 +1,47 @@
+# typed: true
+# frozen_string_literal: true
+
+require "test_helper"
+
+class RenameTest < Minitest::Test
```
andyw8, line 6: Looks like we're not testing the file renaming here?

vinistock: Added behaviour to test the renaming.

jekyll/index.markdown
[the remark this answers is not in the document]
vinistock: Switched.

lib/ruby_lsp/global_state.rb
```diff
@@ -133,6 +137,9 @@ def apply_options(options)
       end
 
       @supports_request_delegation = options.dig(:capabilities, :experimental, :requestDelegation) || false
+      supported_resource_operations = options.dig(:capabilities, :workspace, :workspaceEdit, :resourceOperations)
+      @supported_resource_operations = supported_resource_operations if supported_resource_operations
```
andyw8, line 141: Isn't this the same as:

```
@supported_resource_operations ||= options.dig(:capabilities, :workspace, :workspaceEdit, :resourceOperations)
```

vinistock: No because `@supported_resource_operations` defaults to an empty array to avoid nilability with Sorbet.

lib/ruby_lsp/requests/rename.rb
st0012: No I mean still taking `params` in the constructor, but just do:

```rb
@position = params[:position]
@new_name = params[:newName]
```

vinistock: Oh, sorry I misunderstood what you said. Yeah, I agree that it's nicer like that.

andyw8 (approved):

PR #2638, opened 2024-09-30
Add version constraint for add-ons

vinistock:
### Motivation

This PR proposes an API for add-ons to declare version dependencies for the Ruby LSP and for other add-ons.

If the proposed API is accepted, I will add it to our add-ons documentation page.

### Implementation

There are two API changes here.

The new `depend_on!` method declares a dependency on the Ruby LSP. This should be invoked at the top level, so that we skip requiring the add-on completely if it's depends on a version that cannot be satisfied by the current API. This method is only intended to be invoked by add-ons that are defined in gems that do not depend on the `ruby-lsp` explicitly.

The second change is requiring the version constraint as part of `Addon.get`. This ensures that add-ons depending on other add-ons have to "declare" which version constraint they are comfortable depending on, so that if a breaking change happens they can handle accordingly and run in partial support mode.

### Example

```ruby

RubyLsp::Addon.depend_on!(">= 0.18.0", "< 0.19.0")

# ... requires


module RubyLsp
  module MyGem
    class Addon < ::RubyLsp::Addon
      def activate(global_state, outgoing_queue)
        rails_addon = ::RubyLsp::Addon.get("Ruby LSP Rails", "> 0.3.0", "< 1.0.0")
      rescue RubyLsp::Addon::IncompatibleApiError
        # decide how to provide partial support
      end
    end
  end
end
```

### Automated Tests

Added test.

### DIFF
--- a/lib/ruby_lsp/addon.rb
+++ b/lib/ruby_lsp/addon.rb
@@ -32,6 +32,8 @@ module RubyLsp
 
     AddonNotFoundError = Class.new(StandardError)
 
+    class IncompatibleApiError < StandardError; end
+
     class << self
       extend T::Sig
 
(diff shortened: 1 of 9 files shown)


lib/ruby_lsp/addon.rb
```diff
@@ -80,13 +82,49 @@ def load_addons(global_state, outgoing_queue)
...
+      # RubyLsp::Addon.depend_on!(">= 0.18.0")
+      #
+      # module MyGem
+      #   class MyAddon < RubyLsp::Addon
+      #     # ...
+      #   end
+      # end
+      # ```
+      sig { params(version_constraints: String).void }
+      def depend_on!(*version_constraints)
```
andyw8, line 120:
```suggestion
      def depend_on_ruby_lsp!(*version_constraints)
```
to make it self-descriptive

lib/ruby_lsp/addon.rb
```diff
@@ -80,13 +82,49 @@ def load_addons(global_state, outgoing_queue)
...
+        unless version_constraints.all? { |constraint| Gem::Requirement.new(constraint).satisfied_by?(version_object) }
+          raise IncompatibleApiError,
+            "Constraints #{version_constraints.inspect} is incompatible with #{addon_name} version #{addon.version}"
+        end
+
         addon
       end
+
+      # Depend on a specific version of the Ruby LSP. This method should only be used if the add-on is distributed in a
+      # gem that does not have a runtime dependency on the ruby-lsp gem. This method should be invoked at the top of the
```
andyw8, line 107: > that does not have a runtime dependency on the ruby-lsp gem

We still expect the addon to have `test` dependency on ruby-lsp, yes?

lib/ruby_lsp/addon.rb
```diff
@@ -132,6 +170,11 @@ def deactivate; end
     sig { abstract.returns(String) }
     def name; end
 
+    # Add-ons should override the `version` method to return a semantic version string representing the add-on's
+    # version. This is used for compatibility checks
+    sig { abstract.returns(String) }
```
andyw8, line 175: Are we anticipating the add-on would be versioned separately from the gem its contained within? Or would we return the gem's own `VERSION` is most cases?

st0012: What'd happen when the `version` is called on a add-on that doesn't use Sorbet itself? Would it raise an error or return `nil`?

PR #2646, opened 2024-10-01
Use Netlify to build documentation preview for PRs

st0012:
### Motivation

Currently, to review a documentation change, one needs to:

1. Pull the change and checkout to it with `gh` command
2. Run `cd jekyll`
3. Run `bundle exec jekyll serve`

And it's assuming that Jekyll is already installed to the latest version locally.

If we can have a way to preview the doc changes directly, it'd be much easier for maintainers and anyone who's interested in the changes to review. And hopefully lowering the reviewing costs can also make doc contribution (feel) easier.

### Implementation

1. Use Ruby 3.3.3 for `jekyll` folder. I explained more in the comments below.
2. Add `netlify.toml` to store all build configurations.

With the build ignore config in place, Netlify's free plan should cover our current usages.


jekyll/.ruby-version
```diff
@@ -1 +0,0 @@
-../.ruby-version
```
st0012, line 1: We need to stay 3.3.3 for now because Netlify uses rvm, which Ruby 3.3.4+ currently fails to build on its
containers due to:

- https://github.com/ruby/bigdecimal/issues/297
- https://github.com/rubygems/rubygems/issues/8015

And because Netlify doesn't allow installing apt packages, we can't address the issue to make the install succeed. So instead, we need to use a version of Ruby that can be built on Netlify's containers.

netlify.toml
```diff
@@ -0,0 +1,10 @@
+[build]
+  base = "jekyll"
+  publish = "../docs"
+
+  # Set the baseurl to "/" and url to "$DEPLOY_PRIME_URL" so that the built site is served from the root
+  command = """
+  printf "url: \"$DEPLOY_PRIME_URL\"\nbaseurl: \"/\"" > _config_netlify.yml; \
```
st0012, line 7: Since we want to build new sites for PR preview, the url will vary PR by PR. So this line is needed to dynamically generate suitable Jekyll configs for Netlify.
And because it needs to be dynamic, it can't be a committed `_config_netlify.yml` file.

andyw8 (approved):

st0012:
Since Netlify's usage at Shopify is still not approved, I'm closing this now.

Issue #2640, opened 2024-10-01
Add references support for methods


vinistock:
Build on top of #2632 to add support for finding method references.

For methods, since we can't always be sure they refer to the exact declaration requested due to missing type information, it would be nice to brainstorm some way of communicating that to the user. We can still show all possible references we find, but it's good to make it clear that some of them (when types are unknown) may not actually refer to the same method.

PR #2650, opened 2024-10-02
Add method support to References request

andyw8:
This allows the use of Find All References for methods.

As discussed in https://github.com/Shopify/ruby-lsp/issues/2640, we cannot have _full_ confidence in the matches due the lack of a type system.

Note: In VS Code, Rename Symbol will show as available for methods, but isn't yet implemented.

Co-authored with @vinistock 

Closes https://github.com/Shopify/ruby-lsp/issues/2640

### DIFF
--- a/lib/ruby_indexer/lib/ruby_indexer/reference_finder.rb
+++ b/lib/ruby_indexer/lib/ruby_indexer/reference_finder.rb
@@ -5,6 +5,37 @@ module RubyIndexer
   class ReferenceFinder
     extend T::Sig
 
+    class Target
+      # def initialize
+      # end
+    end
+
+    class ConstTarget < Target
+      extend T::Sig
+
+      sig { params(fully_qualified_name: String).void }
+      def initialize(fully_qualified_name)
+        super()
+        @fully_qualified_name = fully_qualified_name
+      end
+
+      sig { returns(String) }
+      attr_reader :fully_qualified_name
+    end
+
+    class MethodTarget < Target
+      extend T::Sig
+
+      sig { params(method_name: String).void }
+      def initialize(method_name)
+        super()
+        @method_name = method_name
+      end
+
+      sig { returns(String) }
+      attr_reader :method_name
+    end
+
     class Reference
       extend T::Sig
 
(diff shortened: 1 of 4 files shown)


andyw8:
@vinistock should we also try cover `attr_*` references, or does that add too much complexity?

lib/ruby_indexer/lib/ruby_indexer/reference_finder.rb
```diff
@@ -5,6 +5,35 @@ module RubyIndexer
   class ReferenceFinder
     extend T::Sig
 
+    class Target
+    end
+
+    class ConstTarget < Target
```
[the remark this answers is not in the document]
vinistock: I don't think so. Usually, interfaces are used to ensure that an object **must** implement the interface. If there's no common interface between all of the target types, then it's a bit odd to use an interface.

lib/ruby_indexer/lib/ruby_indexer/reference_finder.rb
```diff
@@ -5,6 +5,35 @@ module RubyIndexer
...
+      extend T::Sig
+
+      sig { params(fully_qualified_name: String).void }
+      def initialize(fully_qualified_name)
+        super()
+        @fully_qualified_name = fully_qualified_name
+      end
+
+      sig { returns(String) }
+      attr_reader :fully_qualified_name
```
vinistock, line 21: NIT: Move the attributes above the regular method declarations.

lib/ruby_indexer/lib/ruby_indexer/reference_finder.rb
```diff
@@ -5,6 +5,35 @@ module RubyIndexer
...
+      extend T::Sig
+
+      sig { params(method_name: String).void }
+      def initialize(method_name)
+        super()
+        @method_name = method_name
+      end
+
+      sig { returns(String) }
+      attr_reader :method_name
```
vinistock, line 34: NIT: Move the attributes above the regular method declarations.

vinistock (commented):
Nice. We can get rename working after this

vinistock:
> @vinistock should we also try cover attr_* references, or does that add too much complexity?

We should definitely do it, but you can do that in a subsequent PR.

lib/ruby_indexer/lib/ruby_indexer/reference_finder.rb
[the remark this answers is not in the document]
vinistock: Yes, that make sense.

lib/ruby_indexer/lib/ruby_indexer/reference_finder.rb
vinistock: Do we really need a cop for this though?

andyw8: If it's a style decision we care about, and something that trivially handled by an exising cop, then I don't see why not.

PR #2651, opened 2024-10-02
Reduce noise from "This file is too long" warning

### LINKED ISSUE #2550: Reduce noise from "This file is too long" warning.

### Use case

We have several large files that I end up opening frequently. I've been closing the notification popup probably a dozen~ times a day, which can get pretty frustrating when it keeps happening.

Image

I know I could refactor these large files into smaller ones, but they're pretty long for a reason, and I'm not sure how I'd easily break them apart.

### Description

I'm not sure what the best fix would be. Here are some ideas, but my guess is that you all will probably be able to think of something better.

1. A way to disable the warning.
2. Only warn for a file once, not every time you open the file.
3. A file ignore list.
4. Send the warning to a less intrusive place (I think there's a difference between notification panels and popups, but I don't know how that works).
5. Remove the warning entirely.

### Implementation

_No response_

alexcrocha:
### Motivation

Resolves #2550 

Frequent notifications for large files are disruptive. Refactoring the files isn't always feasible, so a better solution is needed to reduce the noise.

### Implementation

- **Log Only for File Scheme**: Notifications will now only log if the scheme is a file, preventing unnecessary alerts for other URI types.
- **Output Tab Logging**: Notifications will be printed to the output tab instead of showing a popup, reducing interruptions.

### DIFF
--- a/lib/ruby_lsp/server.rb
+++ b/lib/ruby_lsp/server.rb
@@ -320,7 +320,7 @@ module RubyLsp
           language_id: language_id,
         )
 
-        if document.past_expensive_limit?
+        if document.past_expensive_limit? && text_document[:uri].scheme == "file"
           log_message = <<~MESSAGE
             The file #{text_document[:uri].path} is too long. For performance reasons, semantic highlighting and
             diagnostics will be disabled.


lib/ruby_lsp/server.rb
```diff
@@ -320,11 +320,11 @@ def text_document_did_open(message)
...
-        if document.past_expensive_limit?
+        if document.past_expensive_limit? && text_document[:uri].scheme == "file"
           send_message(
             Notification.new(
-              method: "window/showMessage",
-              params: Interface::ShowMessageParams.new(
+              method: "window/logMessage",
+              params: Interface::LogMessageParams.new(
                 type: Constant::MessageType::WARNING,
                 message: "This file is too long. For performance reasons, semantic highlighting and " \
```
andyw8, line 329: Since this is now 'separated' from the active editor tab, I think we should include the path in the message.

vinistock (approved):

PR #2652, opened 2024-10-02
Add indexing enhancements and version to add-on docs

vinistock:
### Motivation

This PR documents indexing enhancements and adds the `version` requirement to our add-on docs.

### DIFF
--- a/jekyll/add-ons.markdown
+++ b/jekyll/add-ons.markdown
@@ -82,6 +82,12 @@ module RubyLsp
       def name
         "Ruby LSP My Gem"
       end
+
+      # Defining a version for the add-on is mandatory. This version doesn't necessarily need to match the version of
+      # the gem it belongs to
+      def version
+        "0.1.0"
+      end
     end
   end
 end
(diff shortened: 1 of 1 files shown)


jekyll/add-ons.markdown
```diff
@@ -126,6 +132,45 @@ This approach enables all add-on responses to be captured in a single round of A
 
 ### Enhancing features
 
+There are two ways to enhance Ruby LSP features. One is handling DSLs that occur at a call site and that do not change
+which declarations exist in the project. A great example of this the Rails validation method, which accepts a symbol
```
andyw8, line 136:
```suggestion
which declarations exist in the project. A great example of this the Rails `validate` method, which accepts a symbol
```

jekyll/add-ons.markdown
```diff
@@ -126,6 +132,45 @@ This approach enables all add-on responses to be captured in a single round of A
...
+that represents a method that gets dynamically invoked. That style of DSL is what we refer to as a [call site
+DSL](#dealing-with-call-site-dsls).
```
andyw8, line 138: Would it be more correct to call this a 'DSL method' or something like that?

jekyll/add-ons.markdown
```diff
@@ -195,6 +244,103 @@ module RubyLsp
 end
 ```
 
+#### Dealing with declaration DSLs
+
+Add-ons can inform the Ruby LSP about declarations that are made via meta-programming. By ensuring that the index is
+populated with all declarations, features like go to definition, hover, completion, signature help and workspace symbol
+will all automatically work.
+
+To achieve this the add-on must create an indexing enhancement class and then it needs to register that. Here's an
```
andyw8, line 253:
```suggestion
To achieve this the add-on must create an indexing enhancement class and register it. Here's an
```

jekyll/add-ons.markdown
```diff
@@ -195,6 +244,103 @@ module RubyLsp
...
+    # Create the array of signatures that this method will accept. Every signatures is composed of a list of
+    # parameters. The parameter classes represent each type of parameter
+    signatures = [
+      Entry::Signature.new([Entry::RequiredParameter.new(name: :a)])
+    ]
+
+    new_entry = Entry::Method.new(
+      "new_method", # The name of the method that gets created via meta-programming
+      file_path,    # The file_path where the DSL call was found. This should always just be the file_path received
+      location,     # The location where the DSL call was found
```
andyw8, line 300:
```suggestion
      location,     # The Prism node location where the DSL call was found
```

jekyll/add-ons.markdown
```diff
@@ -346,14 +504,56 @@ end
...
+For add-ons that have a runtime dependency on the `ruby-lsp` gem, you can simply use regular gemspec constraints to
+define which version is supported.
 
 ```ruby
 spec.add_dependency("ruby-lsp", "~> 0.6.0")
 ```
 
+#### Add-ons that do not have a runtime dependency on the ruby-lsp
+
+For add-ons that are defined inside other gems that do not wish to take a runtime dependency on `ruby-lsp`, please use
```
andyw8, line 522:
```suggestion
For add-ons that are defined inside other gems that do not wish to have a runtime dependency on `ruby-lsp`, please use
```

andyw8:
@alexcrocha it would be good to get your input here as a 'newcomer' to developing add-ons.

jekyll/add-ons.markdown
vinistock: I'm not sure what the best term is. A method that mutates the index like `belongs_to` is also a "DSL method".

I was looking for terms that can differentiate DSL on the call site vs DSLs that create declarations.

andyw8: I was meaning just the 'DSL' part, to me the 'DSL' is the _collection_ of all the available 'constructs'. Not a big deal though.

Issue #2644, opened 2024-10-01
Add definition, hover and completion support for global variables


vinistock:
We should provide features for global variables.

**Note: this issue includes instructions on how to add indexing capabilities for globals, definition support, hover support and completion support. If you wish to contribute this, please create 4 separate pull requests for each step.**

## Implementation

I think the most natural implementation is to add global variables as a new entry type in the index. Hopefully, RBS should have global declarations with documentation that we can simply index like we do core declarations.

If RBS does not have global declarations, please raise it in this issue and we can reach out to Soutaro and the team to ask for guidance.

### Changes to the indexer

1. Create a new [entry](https://github.com/Shopify/ruby-lsp/blob/main/lib/ruby_indexer/lib/ruby_indexer/entry.rb) for global variables
2. Start handling `RBS::AST::Declarations::Global` in [process_declaration](https://github.com/Shopify/ruby-lsp/blob/c00f2bce3aaebd9c0d7427e6acadccdac581cd13/lib/ruby_indexer/lib/ruby_indexer/rbs_indexer.rb#L38), so that we can insert the global entries into the index

### Changes to definition

1. We need to allow global variables (`Prism::GlobalVariableReadNode`) as a possible target [here](https://github.com/Shopify/ruby-lsp/blob/c00f2bce3aaebd9c0d7427e6acadccdac581cd13/lib/ruby_lsp/requests/definition.rb#L58)
2. Then the listener has to [register](https://github.com/Shopify/ruby-lsp/blob/c00f2bce3aaebd9c0d7427e6acadccdac581cd13/lib/ruby_lsp/listeners/definition.rb#L51) for the `on_global_variable_read_enter` event and
3. define a [handler](https://github.com/Shopify/ruby-lsp/blob/c00f2bce3aaebd9c0d7427e6acadccdac581cd13/lib/ruby_lsp/listeners/definition.rb#L159) for `on_global_variable_read_enter`
4. Since we're talking about global variables, there's no need to perform any type checking. We just need to access the index and get the declarations based on the name

```ruby
@index[name_of_global_variable]
```
5. Then loop through the entries and push to the response as we do [here](https://github.com/Shopify/ruby-lsp/blob/c00f2bce3aaebd9c0d7427e6acadccdac581cd13/lib/ruby_lsp/listeners/definition.rb#L192)

### Changes to hover

Essentially, the same exact steps as definition. Simply look for the `hover` request and listener and apply analogous changes.

### Changes to completion

[...]

PR #2656, opened 2024-10-02
Index global variables

snutij:
### Motivation

Before providing features related to global variables, we must index them via [RBS declarations](https://github.com/ruby/rbs/blob/master/core/global_variables.rbs), as we do for core constants for example.

### Implementation

Handle `RBS::AST::Declarations::Global` in `RBSIndexer`, and create `Entry::GlobalVariable` for each declaration.

### Automated Tests

I added a basic test, following `index_methods` example, but if specific assertions are required, I can add them. 

### Manual Tests

After indexation process, a search in the index with `.prefix_search("$")`, it should return all entries from RBS declaration.

### DIFF
--- a/lib/ruby_indexer/lib/ruby_indexer/entry.rb
+++ b/lib/ruby_indexer/lib/ruby_indexer/entry.rb
@@ -526,20 +526,7 @@ module RubyIndexer
     end
 
     # Represents a global variable e.g.: $DEBUG
-    class GlobalVariable < Entry
-      sig do
-        params(
-          name: String,
-          file_path: String,
-          location: RubyIndexer::Location,
-          comments: T.nilable(String),
-          encoding: Encoding,
-        ).void
-      end
-      def initialize(name, file_path, location, comments, encoding)
-        super
-      end
-    end
+    class GlobalVariable < Entry; end
 
     # Represents an instance variable e.g.: @a = 1
     class InstanceVariable < Entry


lib/ruby_indexer/lib/ruby_indexer/entry.rb
```diff
@@ -525,6 +525,22 @@ def initialize(target, unresolved_alias, encoding)
...
+        params(
+          name: String,
+          file_path: String,
+          location: RubyIndexer::Location,
+          comments: T.nilable(String),
+          encoding: Encoding,
+        ).void
+      end
+      def initialize(name, file_path, location, comments, encoding)
+        super(name, file_path, location, comments, encoding)
```
andyw8, line 540:
```suggestion
        super
```
(since the arguments are identical)

lib/ruby_indexer/lib/ruby_indexer/rbs_indexer.rb
```diff
@@ -271,6 +273,25 @@ def handle_constant(declaration, nesting, file_path)
       ))
     end
 
+    sig do
```
andyw8, line 276: (nit) let's put on one line

andyw8 (approved):
Great!

lib/ruby_indexer/lib/ruby_indexer/entry.rb
```diff
@@ -525,6 +525,22 @@ def initialize(target, unresolved_alias, encoding)
...
+          name: String,
+          file_path: String,
+          location: RubyIndexer::Location,
+          comments: T.nilable(String),
+          encoding: Encoding,
+        ).void
+      end
+      def initialize(name, file_path, location, comments, encoding)
+        super
+      end
```
vinistock, line 541: If all we're doing is invoking super, can we not get rid of this definition altogether and invoke the parent class' definition directly?

vinistock (commented):
Just a minor comment, but this looks great!

PR #2657, opened 2024-10-02
Add support for range formatting with Syntax Tree

### LINKED ISSUE #203: Add range formatting to the `ruby-lsp`
Range formatting is a request very similar to formatting, but it applies only to the selected range and not the entire document.

Currently, trying to format ranges with either RuboCop or SyntaxTree fails, because both believe they are always looking at whole files instead of ranges.

For RuboCop, this will be difficult to overcome, given that trying to format a range in the middle of the document will even add `# frozen_string_literal` comments.

However, we may be able to use SyntaxTree for this if we add the ability of informing the formatter about the current indentation (or nesting) level. In the current behaviour, the code inside the range will be considered as a whole file and everything is pushed to indentation level zero.

Documentation: https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#textDocument_rangeFormatting

vinistock:
### Motivation

Closes #203

A long time ago, I added the [bits missing in Syntax Tree for partial formating](https://github.com/ruby-syntax-tree/syntax_tree/pull/199). I always wanted to finish the work and offer range formatting, which is what this PR does.

As far as I know, there's no way to support this on RuboCop without some new API changes, but at least Syntax Tree users can benefit from it.

### Implementation

The request is extremely similar to formatting. The only difference is that instead of formatting the entire document, we need to first locate the first node within the requested range and then calculate the indentation level for formatting (which is just the start column / 2).

Then we format only that node's slice and return a text edit to replace it.

### Automated Tests

Added a test.


andyw8 (approved):
(did not 🎩 )

PR #2661, opened 2024-10-03
Prevent lazy comment fetching from failing if file gets deleted

vinistock:
### Motivation

Due to concurrency, there are situations where we may try to lazily fetch comments for an entry right after it was deleted from disk, but before we fully removed the entry from the index. In those cases, we don't want to fail. 

### Implementation

Started rescuing from missing files.

### Automated Tests

Added a test that fails without the fix.


lib/ruby_indexer/lib/ruby_indexer/entry.rb
```diff
@@ -110,6 +110,8 @@ def comments
         else
           ""
         end
+      rescue Errno::ENOENT
```
andyw8, line 113: Could we worth a brief explanatory comment.

andyw8 (approved):

PR #2662, opened 2024-10-03
Display add-on version in the displayAddons command

st0012:
### Motivation

Since add-ons now can specify their versions, we can also display them in VS Code.

### Implementation

- Add `version` to `rubyLsp/workspace/addons`'s response
- Add a nullable `version` attribute to the `Addon` extension class as older versions of `ruby-lsp` don't return it
- Add `version` information to `displayAddon` command's output:

Screenshot 2024-10-03 at 14 51 59

### Automated Tests

Updated an existing case for it.

### Manual Tests

1. Run Ruby LSP extension with this branch
2. Use the latest `ruby-lsp-rails` as the development host
3. Update `ruby-lsp-rails`'s Gemfile to use this branch of `ruby-lsp`
    - `gem "ruby-lsp", github: "Shopify/ruby-lsp", branch: "display-add-on-version"`
    - Run `bundle update`
4. After Ruby LSP is initialized, run `Ruby LSP: Display addons` command in VS Code's command palette

### DIFF
--- a/lib/ruby_lsp/server.rb
+++ b/lib/ruby_lsp/server.rb
@@ -89,7 +89,7 @@ module RubyLsp
             id: message[:id],
             response:
               Addon.addons.map do |addon|
-                { name: addon.name, errored: addon.error? }
+                { name: addon.name, version: addon.version, errored: addon.error? }
               end,
           ),
         )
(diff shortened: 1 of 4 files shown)


vinistock:
This may be a bit premature. If add-ons haven't upgrade to the latest API yet, the moment the extension is released it's going to break. Why do we need to add this immediately? Can it wait a bit until people had the chance to upgrade?

PR #2666, opened 2024-10-04
Enable RuboCop class structure

andyw8:
Seeing if we can handle https://github.com/Shopify/ruby-lsp/pull/2650/files#r1787743844 automatically.


andyw8:
https://github.com/rubocop/rubocop/issues/13295

.rubocop.yml
```diff
@@ -58,3 +58,28 @@ Sorbet/StrictSigil:
...
     - "lib/ruby_indexer/lib/ruby_indexer/prefix_tree.rb"
     - "lib/ruby_lsp/load_sorbet.rb"
+
+Layout/ClassStructure:
+  Enabled: true
+  ExpectedOrder:
+    - module_inclusion
+    - constants
+    - public_attribute_macros
+    - public_delegate
```
Earlopain, line 68: `delegate` comes from rails, not necessary here

andyw8: Ah, was thinking this was Ruby's `def_delegator(s)`. Can remove.

PR #2669, opened 2024-10-04
Handle multibyte characters in RubyDocument

NotFounds:
### Motivation

Close https://github.com/Shopify/ruby-lsp/issues/1251

In [Shopify/ruby-lsp#2619](https://github.com/Shopify/ruby-lsp/pull/2619), we modified the Index creation to handle multibyte characters. However, when processing requests from the LSP to calculate the target node, the character positions are calculated based on the PrismResult rather than the Index, so adjustments are needed in this part of the code as well.

### Implementation

This Pull Request updates the `locate` function in RubyDocument to handle multibyte characters .

### Automated Tests

I added simple test cases for it and the definition request.

### Manual Tests

I have tested this on multiple projects (including those with Japanese) and confirmed that the definition jump and hover functions work as intended.

### DIFF
--- a/lib/ruby_lsp/erb_document.rb
+++ b/lib/ruby_lsp/erb_document.rb
@@ -50,7 +50,12 @@ module RubyLsp
       ).returns(NodeContext)
     end
     def locate_node(position, node_types: [])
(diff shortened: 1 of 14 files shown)


vinistock (approved):

PR #2691, opened 2024-10-08
Add hover for global variables

snutij:
### Motivation
Handle global variables in hover requests.

### Implementation
Handle each kind of global variable node in the listener pattern, then search for entries in the index with the variable name, and finally, push the documentation for entries to the `response_builder`.

### Automated Tests
- Added a test with assertions for each kind of global variable
- Added a test with assertions to ensure the target is unset if the hover is not exactly on the variable name

### Manual Tests
- Write `$DEBUG` within a file, then hover over it. You should see documentation from the [RBS declaration](https://github.com/ruby/rbs/blob/master/core/global_variables.rbs). 
- Write a custom `$foo` variable with a comment, hover over it, you should see the hover.
- You should not see documentation when hovering not exactly on variable name

https://github.com/user-attachments/assets/149bc72d-56b9-41d8-b3fc-531c0adbd2e2

### DIFF
--- a/lib/ruby_lsp/requests/hover.rb
+++ b/lib/ruby_lsp/requests/hover.rb
@@ -98,9 +98,6 @@ module RubyLsp
       sig { params(position: T::Hash[Symbol, T.untyped], target: T.nilable(Prism::Node)).returns(T::Boolean) }
       def position_outside_target?(position, target)
         case target
-        when Prism::CallNode
-          target.name != :require && target.name != :require_relative &&
-            !covers_position?(target.message_loc, position)
         when Prism::GlobalVariableAndWriteNode,
           Prism::GlobalVariableOperatorWriteNode,
           Prism::GlobalVariableOrWriteNode,


lib/ruby_lsp/listeners/hover.rb
```diff
@@ -13,6 +13,7 @@ class Hover
           Prism::ConstantReadNode,
           Prism::ConstantWriteNode,
           Prism::ConstantPathNode,
+          Prism::GlobalVariableReadNode,
```
andyw8, line 16: We should also handle `GlobalVariableWriteNode`.

(there's also `GlobalVariableAndWriteNode` and a few others, but they are very rarely used).

snutij: I'm wondering, after reading your feedback, if I missed this point with the indexer part as well.

```ruby
# a new comment, that will not be displayed in the hover
$DEBUG = "foo"
```

Here, on hover, if I'm not mistaken, we want to display both comments: the one from the RBS declarations and the one from our own implementation. If yes, then I need to handle all these kinds of nodes within `RubyIndexer::DeclarationListener`, not only on RBS side. 

WDYT?

vinistock: Also, I don't see any issues with shipping this PR and following it up with the rest of the nodes, if you prefer that approach.

andyw8: Yes, we probably should. We index comments for constants defined in the user's own code, so I would expect globals to behave the same way.

andyw8: (wrote that before I saw Vini had already replied)

snutij: I would prefer to hold-off this one, create a separate PR focusing solely on indexing all kinds of global variables, and then adjust this one, taking advantage of all the additions. This way, we can keep concerns clearly separated, WDYT?

vinistock: It's very possible that the condition is no longer necessary. In the case of hovering over global assignments, we would need something along these lines:

```ruby
if ...
elsif target.is_a?(Prism::GlobalVariableWriteNode) && !covers_position?(target.name_loc, position)
 target = nil
end
```

PR #2698, opened 2024-10-09
Build locations before creating entries

vinistock:
### Motivation

Currently, our index entries accept either a `Prism::Location` or a `RubyIndexer::Location`. This is causing some weird patterns, like the need to pass the encoding everywhere even after the initial indexing has been performed.

Prism is about to ship a new [code units cache concept](https://github.com/ruby/prism/pull/3173), which will hopefully solve the performance degradation we've been seeing, but it will also make our current design even more difficult to deal with.

This PR forces us to build locations outside of entries, which prevents the coupling with the encoding and makes the changes required much more straight forward.

### Implementation

Every entry only accepts a `RubyIndexer::Location` now. I added a helper method to build them from a `Prism::Location`, so that the transition to use the code units cache is easier.

### Automated Tests

Existing tests should cover it.

### DIFF
--- a/lib/ruby_indexer/lib/ruby_indexer/declaration_listener.rb
+++ b/lib/ruby_indexer/lib/ruby_indexer/declaration_listener.rb
@@ -33,6 +33,7 @@ module RubyIndexer
         T::Hash[Integer, Prism::Comment],
       )
       @inside_def = T.let(false, T::Boolean)
+      @encoding = T.let(@index.configuration.encoding, Encoding)
 
       # The nesting stack we're currently inside. Used to determine the fully qualified name of constants, but only
       # stored by unresolved aliases which need the original nesting to be lazily resolved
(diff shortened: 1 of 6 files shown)


lib/ruby_indexer/lib/ruby_indexer/declaration_listener.rb
```diff
@@ -106,10 +107,9 @@ def on_class_node_enter(node)
       entry = Entry::Class.new(
         nesting,
         @file_path,
-        node.location,
-        constant_path.location,
+        Location.from_prism_location(node.location, @encoding),
```
andyw8, line 110: What do you think about adding factory methods to the `Entry` classes instead? e.g.:

```ruby
      entry = Entry::Class.from_prism_location(
        nesting,
        @file_path,
        node,
        constant_path,
        @encoding,
        comments,
        parent_class,
      )
```
(as well as `from_location` for when we're dealing with `Location`).

vinistock: That's going to preserve the coupling that I'm trying to get rid of. We need to be able to build entries without depending on the encoding, so that in certain situations we instantiate `RubyIndexer::Location` outside without passing the encoding around.

andyw8: Ok, let's go with this for now, but I think there's probably a way to clean up the API later.

andyw8 (approved):

PR #2699, opened 2024-10-09
Index all missing global variable nodes

snutij:
### Motivation
The first PR to index global variables (https://github.com/Shopify/ruby-lsp/pull/2656) focused only on RBS declarations. This covers only partial use cases, as it is possible for a user to define their own global variables.

### Implementation
From [prism/config.yml](https://github.com/ruby/prism/blob/d6e9b8de36b4d18debfe36e4545116539964ceeb/config.yml#L2110), listen to all kinds of global variable writing nodes, and push `GlobalVariable` entries into the index.

### Automated Tests
Added a new file with all kinds of global variable writing.

### Manual Tests
Because I have already merged go-to-definition changes for global variables, we can already check that it's working.

https://github.com/user-attachments/assets/70245e1d-7778-4c2e-b733-d565e60a69e1


andyw8 (approved):
Thank you!

Issue #2688, opened 2024-10-08
Possible regression with Code Lens for running spec-style tests


andyw8:
There seems to have been some regression with the Code Lens for running spec-style tests. On Tapioca, Kaan says it was working previously, but now it is failing, e.g.

```
tapioca main % bundle exec ruby -Itest /Users/andyw8/src/github.com/Shopify/tapioca/spec/tapioca/cli/annotations_spec.rb --name "/^T
apioca::AnnotationsTest(#|::)/"
/opt/rubies/3.3.4/lib/ruby/3.3.0/bundled_gems.rb:74:in `require': cannot load such file -- spec_helper (LoadError)
        from /opt/rubies/3.3.4/lib/ruby/3.3.0/bundled_gems.rb:74:in `block (2 levels) in replace_require'
        from /Users/andyw8/src/github.com/Shopify/tapioca/spec/tapioca/cli/annotations_spec.rb:4:in `<main>'
```

(this could relate to a change within Tapioca itself)

paracycle:
The problem is with the `-Itest` part. For that project, it should be `-Ispec`.

vinistock:
If that's the only problem, then this is a duplicate of #2308 and we can close this one.

andyw8:
I think https://github.com/Shopify/ruby-lsp/pull/2522 is partly the cause. If I try the ruby-lsp version prior to that (v0.17.17), and just change `-Itest` to `-Ispec` then it works, e.g.

```
bundle exec ruby -Ispec /Users/andyw8/src/github.com/Shopify/tapioca/spec/tapioca/cli/gem_spec.rb --name /must\ remove\ outdated\ RBIs/
Started with run options --name "/must remove outdated RBIs/" --seed 41260
```

andyw8:
This is likely related to the specs in Tapioca being within a class definition, which is a bit different from normal specs.

andyw8:
I'll closet his since it wasn't a regression, it had never worked previously. But it should all be working once https://github.com/Shopify/ruby-lsp/pull/2706 is released.

PR #2706, opened 2024-10-10
Fix minitest spec code lens load path

### LINKED ISSUE #2308: VSCode test "Run in Terminal" uses hardcoded `-Itest`
### Reproduction steps

1. Open a Ruby project in VSCode
2. Have a `spec_helper.rb` at toplevel of `spec` directory
3. Open a Ruby spec file in a `spec` directory (I used `minitest/spec`) that has a `require 'spec_helper'`
4. Wait for the LSP to add the `Run | Run in Terminal | Debug` underneath each `describe` entry)
5. Click `Run in Terminal` on one
6. Observe that running the test fails because of `-Itest` instead of `-Ispec`

Minimal project layout:

```
spec\
  spec_helper.rb
  my_spec.rb
```

```ruby
# my_spec.rb
require 'minitest/spec'
require 'minitest/autorun'
require 'spec_helper'

# Run | Run in Terminal | Debug
describe 'foo' do
  it { _(something).must_equal true }
end
```

### Analysis

Ruby LSP appears to generate the command from here: https://github.com/Shopify/ruby-lsp/blob/db5002a1c2fd085954745e1e9b1bbf9ef1a0a11b/lib/ruby_lsp/listeners/code_lens.rb#L18

```ruby
      BASE_COMMAND = T.let(
        begin
          Bundler.with_original_env { Bundler.default_lockfile }
          "bundle exec ruby"
        rescue Bundler::GemfileNotFound
          "ruby"
        end + " -Itest ",
        String,
      )
```

And then use it here: https://github.com/Shopify/ruby-lsp/blob/db5002a1c2fd085954745e1e9b1bbf9ef1a0a11b/lib/ruby_lsp/listeners/code_lens.rb#L227

```ruby
      def generate_test_command(group_stack: [], spec_name: nil, method_name: nil)
        command = BASE_COMMAND + T.must(@path)

        # a bunch of conditional `command +=`

        command
      end
```

This results in this:

```console
$ bundle exec ruby -Itest /path/to/project/spec/my_spec.rb --name /foo/
/nix/store/rzpdrazbld7rjppb8433md41s385g1fg-ruby-3.3.1/lib/ruby/3.3.0/bundled_gems.rb:74:in `require': cannot load such file -- spec_helper (LoadError)
        from /path/to/ruby-3.3.1/lib/ruby/3.3.0/bundled_gems.rb:74:in `block (2 levels) in replace_require'
        from /path/to/project/spec/my_spec.rb:3:in `<main>'
```

Hacking that `BASE_COMMAND` by adding `-Ispec` results in:

```console
$ bundle exec ruby -Itest -Ispec /Users/loic.nageleisen/Source/github.com/DataDog/graft-rb/spec/graft/hook_point_spec.rb --name /when\ giv
en\ invalid\ string/

# Running tests with run options --name "/when given invalid string/" --seed 25736:

.SS.

Finished tests in 0.000867s, 4613.6109 tests/s, 2306.8055 assertions/s.
```

So that `-Itest` definitely comes from there.

### Root cause

1. `BASE_COMMAND` hardcodes `-Itest`
2. Nothing of either `BASE_COMMAND` or `generate_test_command` allow for any sort of configurability of the actual command

But:

- in my case/the example: there is no `test` dir as I have only specs
- in general: the `test` dir may be irrelevant or even contain conflicting files that would be mistakenly picked up by some `require`s in `spec` if a filename match (e.g. say you have both a `spec/support/helper.rb` and a `test/support/helper.rb`)

### Proposed solution

`BASE_COMMAND` should not hardcode `-Itest`.

Instead:

- `generate_test_command` can typically infer a sensible `-I#{directory}` from the test file or spec file relative path.
- `generate_test_command` should have some configurability: for a given project there may be more `-I` needed, and/or the inferred ones may actually need to **not** be present at all.

Notably, when using VSCode, `generate_test_command`  could probably make use of things from `.vscode/launch.json` in some way.

thomasmarshall:
### Motivation

Fixes #2308 and #2688.

This PR changes the code lens base test command to add `spec` to the load path instead of `test` if the current path is inside a `spec` directory (for example [tapioca](https://github.com/Shopify/tapioca/tree/main/spec)).

### Implementation

It checks to see whether the current path is inside a `spec` directory or a `test` directory and adds an appropriate `-I` argument to the base command.

This changes the default behavior—if the current path is in neither a `spec` or `test` directory it no longer adds `-Itest`. Is this reasonable behavior?

There are of course weird edge cases here, for example if the path is inside both a `spec` and `test` directory (like `test/spec/foo_test.rb`). In that case both `test` and `spec` would be added to the load path, but I don't currently see a better way of guessing which directory should be in the load path.

### Automated Tests

I added a new test case to capture the behavior for minitest spec code lens. I also regenerated the expectation JSON files, which have mostly changed just to remove `-Itest`. We could change the logic so `-Itest` is added by default, even if it's not in a `test` directory, and only switch to `-Ispec` if necessary. That would minimize the expectation JSON changes here.

### Manual Tests

Use code lens to run specs in [the tapioca codebase](https://github.com/Shopify/tapioca/tree/main/spec) (upgrade to latest minitest first):

https://github.com/user-attachments/assets/ba99a00f-fc6c-49bc-a6ec-f81339845c10

Use code lens to make sure non-spec tests still work fine:

https://github.com/user-attachments/assets/5c1bd148-b9f0-40c7-a640-18ea75dcac99

### DIFF
--- a/lib/ruby_lsp/listeners/code_lens.rb
+++ b/lib/ruby_lsp/listeners/code_lens.rb
@@ -229,10 +229,11 @@ module RubyLsp
         ).returns(String)
       end
       def generate_test_command(group_stack: [], spec_name: nil, method_name: nil)
+        path = T.must(@path)
         command = BASE_COMMAND
-        command += " -Itest" if T.must(@path).include?("#{File::SEPARATOR}test#{File::SEPARATOR}")
-        command += " -Ispec" if T.must(@path).include?("#{File::SEPARATOR}spec#{File::SEPARATOR}")
-        command += " #{T.must(@path)}"
+        command += " -Itest" if File.fnmatch?("**/test/**/*", path, File::FNM_PATHNAME)
+        command += " -Ispec" if File.fnmatch?("**/spec/**/*", path, File::FNM_PATHNAME)
+        command += " #{path}"
 
         case @global_state.test_library
         when "minitest"


lib/ruby_lsp/listeners/code_lens.rb
```diff
@@ -229,7 +229,10 @@ def add_test_code_lens(node, name:, command:, kind:, id: name)
         ).returns(String)
       end
       def generate_test_command(group_stack: [], spec_name: nil, method_name: nil)
-        command = BASE_COMMAND + T.must(@path)
+        command = BASE_COMMAND
+        command += " -Itest" if T.must(@path).include?("#{File::SEPARATOR}test#{File::SEPARATOR}")
+        command += " -Ispec" if T.must(@path).include?("#{File::SEPARATOR}spec#{File::SEPARATOR}")
+        command += " #{T.must(@path)}"
```
vinistock, line 235: I think we can use `File.fnmatch?` to check if the path matches a specific glob. I'm afraid it could cause some weirdness on Windows since paths are sometimes represented with backslash or forwardslash and I'm not 100% sure which string we would get here.

```suggestion
        path = T.must(@path)
        command = BASE_COMMAND
        command += " -Itest" if File.fnmatch?("**/test/**/*", path, File::FNM_PATHNAME)
        command += " -Ispec" if File.fnmatch?("**/spec/**/*", path, File::FNM_PATHNAME)
        command += " #{path}"
```

vinistock (approved):
This is great!

vinistock:
Thanks, Thomas!

Issue #1767, opened 2024-01-23
"Debugger exited with status 1" for request=launch even though I'm able to execute the same command myself and then debug with request=attach


Nowaker:
### Operating System

Linux

### Ruby version

3.3.0 / 3.1.4

### Project has a bundle

- [X] Has bundle

### Ruby version manager being used

rvm

### Description

```json
    {
      "name": "Rails",
      "type": "ruby_lsp",
      "request": "launch",
      "program": "puma",
      "env": {
        "RAILS_DEBUG": "true"
      },
    }
```

> Debugger exited with status 1. Check the output channel for more information.

![image](https://github.com/Shopify/vscode-ruby-lsp/assets/611746/0074da70-4af2-4028-a05a-99b8055de41c)

```
2024-01-22 20:32:40.983 [info] (worldmodern) Checking if asdf is available on the path with command: /usr/bin/zsh -i -c 'asdf --version'
2024-01-22 20:32:41.353 [info] (worldmodern) Checking if chruby is available on the path with command: /usr/bin/zsh -i -c 'chruby --version'
2024-01-22 20:32:41.513 [info] (worldmodern) Checking if rbenv is available on the path with command: /usr/bin/zsh -i -c 'rbenv --version'
2024-01-22 20:32:41.673 [info] (worldmodern) Checking if rvm is available on the path with command: /usr/bin/zsh -i -c 'rvm --version'
2024-01-22 20:32:41.885 [info] (worldmodern) Discovered version manager rvm
2024-01-22 20:32:41.885 [info] (worldmodern) Trying to activate Ruby environment with command: /usr/bin/zsh -i -c 'rvm-auto-ruby -rjson -e "STDERR.printf(%{RUBY_ENV_ACTIVATE%sRUBY_ENV_ACTIVATE}, JSON.dump({ env: ENV.to_h, ruby_version: RUBY_VERSION, yjit: defined?(RubyVM::YJIT) }))"' inside directory: /home/nowaker/projekty/modern/worldmodern
2024-01-22 20:32:42.978 [info] (worldmodern) Starting Ruby LSP v0.13.1...

2024-01-22 20:32:43.150 [info] (worldmodern) Rails server is not running. To get Rails features in the editor, boot the Rails server
Ruby LSP is ready

2024-01-22 20:33:18.537 [info] Spawning debugger in directory /home/nowaker/projekty/modern/worldmodern
[...]

PR #2707, opened 2024-10-10
Allow custom gemfile for debug

malcolmohare:
### Motivation

Potentially closes:
https://github.com/Shopify/ruby-lsp/issues/1767
https://github.com/st0012/ruby-lsp-rspec/issues/33

When triggering debug (run and debug) from the editor, this extension was setting the env BUNDLE_GEMFILE to be the RubyLsp generated one, not the workspaces one.

### Implementation

I added a new setting, `rubyLsp.debugGemfile`, allowing users to specify the Gemfile to use for debug runs.  This way, no users are affected by this change unless they opt in.

### Automated Tests

When I try and run the tests, I get this error.  So I haven't bothered to write one for this change. 

`AssertionError [ERR_ASSERTION]: Ruby bin path does not exist /opt/rubies/3.3.4/bin`

### Manual Tests

You can see this test in action by triggering debug runs from the editor.  If you set the setting, you should see the value of the BUNDLE_GEMFILE change in the output of RubyLsp when it dumps all the envs into the output.

### DIFF
--- a/vscode/package.json
+++ b/vscode/package.json
@@ -390,6 +390,11 @@
           "type": "string",
           "default": ""
         },
+        "rubyLsp.debugGemfile": {
+          "description": "Relative or absolute path to the Gemfile to use when running debug.  If unset, the Gemfile used for the Ruby LSP server will be used",
+          "type": "string",
+          "default": ""
+        },
         "rubyLsp.testTimeout": {
           "description": "The amount of time in seconds to wait for a test to finish before timing out. Only used when running tests from the test explorer",
           "type": "integer",
--- a/vscode/src/debugger.ts
+++ b/vscode/src/debugger.ts
@@ -1,12 +1,12 @@
 import net from "net";
 import os from "os";
 import { ChildProcessWithoutNullStreams, spawn } from "child_process";
+import path from "path";
 
 import * as vscode from "vscode";
 
 import { LOG_CHANNEL, asyncExec } from "./common";
 import { Workspace } from "./workspace";
-import path from "path";
 
 class TerminalLogger {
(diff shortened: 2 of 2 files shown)


malcolmohare:
investigating CLA status

malcolmohare:
Perhaps there are multiple issues then?  My specific issue was that the lockfile you are generating in the lsp directory is getting different dependency versions that my application lock file.

vinistock:
> Perhaps there are multiple issues then? My specific issue was that the lockfile you are generating in the lsp directory is getting different dependency versions that my application lock file.

If that happened, then the issue is rooted elsewhere and has consequences that extend beyond debugging. The language server itself would index incorrect dependencies if that happens, which is not solved by this PR.

Our [composed bundle](https://shopify.github.io/ruby-lsp/composed-bundle.html) should always be based on your project's lockfile and we only request updates for the `ruby-lsp`, `ruby-lsp-rails` and `debug` gems **if** they are not in the project's Gemfile.

Whenever the lockfile is modified or the LSP is booted, we check if the main lockfile matches exactly the same versions as when we last booted and if not, we re-generate the composed bundle to avoid mismatches.

Now, it could have gotten into a bad state due to some bug either in the LSP or in Bundler. For example, there was a recent Bundler bug where running `bundle check` would result in undesired modifications to the lockfile.

For ensuring that they are sync again, I'd suggest removing the entire directory and then reloading the editor/rebooting the LSP `rm -r .ruby-lsp`. If you still notice that you're getting different dependencies in the composed bundle vs the main one, then please report back and we can try to dig into why that is happening. We need to fix the issue at the root cause.

PR #2716, opened 2024-10-12
Ignore, if given path is directory while referencing and renaming

### LINKED ISSUE #2687: Rename and references fail if there's a directory using the `rb` extension
This is a bit of a silly bug, but one that has bitten us before in indexing. If you name a directory with the `.rb` extension, like `dir.rb`, then it will match the glob pattern we use to find Ruby files. Both rename and references will fail if a directory exists with that extension.

To fix this, we should do the same thing we did in indexing. Instead of making everyone pay the performance penalty of checking every single path if it's a directory or not, we just [rescue](https://github.com/Shopify/ruby-lsp/blob/844c0c357710dbb2eeeedb80ce1fa489f12a7e43/lib/ruby_indexer/lib/ruby_indexer/index.rb#L403) if the path is a directory and continue.

ro-gun:
### Motivation

Fixes #2687

### DIFF
--- a/lib/ruby_lsp/requests/references.rb
+++ b/lib/ruby_lsp/requests/references.rb
@@ -78,6 +78,8 @@ module RubyLsp
 
           parse_result = Prism.parse_file(path)
           collect_references(reference_target, parse_result, uri)
+        rescue Errno::EISDIR, Errno::ENOENT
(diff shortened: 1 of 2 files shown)


rogancodes:
I have signed the CLA

vinistock (approved):

PR #2615, opened 2024-09-24
Document custom bundle process

vinistock:
### Motivation

This PR documents the custom bundle process, which was missing from our docs.

PR #2732, opened 2024-10-17
Update docs and code for remaining references to 'custom bundle'

andyw8:
Following up from https://github.com/Shopify/ruby-lsp/pull/2615#issuecomment-2383485967

### DIFF
--- a/jekyll/design-and-roadmap.markdown
+++ b/jekyll/design-and-roadmap.markdown
@@ -201,7 +201,6 @@ Interested in contributing? Check out the issues tagged with [help-wanted] or [g
 - [Add rename support]
 - [Add show type hierarchy support]
 - [Show index view on the VS Code extension allowing users to browse indexed gems]
-- Remove custom bundle in favor of using bundler-compose
 - [Add more refactoring code actions such as extract to method, extract to class/module, etc]
 - [Explore speeding up indexing by caching the index for gems]
 - Explore speeding up indexing by making Prism AST allocations lazy
--- a/jekyll/troubleshooting.markdown
+++ b/jekyll/troubleshooting.markdown
@@ -29,16 +29,8 @@ As an example, the activation script for `zsh` using `rbenv` as a version manage
 ```
 
 After activating the Ruby version, we then proceed to boot the server gem (`ruby-lsp`). To avoid having users include
-the `ruby-lsp` in their `Gemfile`, we currently create a custom bundle under the `.ruby-lsp` directory inside your
-project. That directory contains another `Gemfile`, that includes the `ruby-lsp` gem in addition to your project's
-dependencies. This approach allows us to automatically detect which formatter your project uses and which gems we need
-to index for features such as go to definition.
-
-{: .note }
-We are working with the RubyGems/Bundler team to have this type of mechanism properly supported from within
-Bundler itself, which is currently being experimented with in a plugin called `bundler-compose`. Once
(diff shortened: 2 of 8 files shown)


jekyll/design-and-roadmap.markdown
```diff
@@ -201,7 +201,7 @@ Interested in contributing? Check out the issues tagged with [help-wanted] or [g
 - [Add rename support]
 - [Add show type hierarchy support]
 - [Show index view on the VS Code extension allowing users to browse indexed gems]
-- Remove custom bundle in favor of using bundler-compose
+- Remove composed bundle in favor of using bundler-compose
```
vinistock, line 204: Let's remove this item completely. It's not really clear that we'll be able to use bundler compose as the current state won't allow us to perform auto updates.

jekyll/troubleshooting.markdown
```diff
@@ -29,15 +29,15 @@ As an example, the activation script for `zsh` using `rbenv` as a version manage
 ```
 
 After activating the Ruby version, we then proceed to boot the server gem (`ruby-lsp`). To avoid having users include
-the `ruby-lsp` in their `Gemfile`, we currently create a custom bundle under the `.ruby-lsp` directory inside your
+the `ruby-lsp` in their `Gemfile`, we currently create a composed bundle under the `.ruby-lsp` directory inside your
```
vinistock, line 32: We should probably link to the composed bundle page from here instead of explaining everything again.

jekyll/troubleshooting.markdown
```diff
@@ -29,15 +29,15 @@ As an example, the activation script for `zsh` using `rbenv` as a version manage
...
 project. That directory contains another `Gemfile`, that includes the `ruby-lsp` gem in addition to your project's
 dependencies. This approach allows us to automatically detect which formatter your project uses and which gems we need
 to index for features such as go to definition.
 
 {: .note }
 We are working with the RubyGems/Bundler team to have this type of mechanism properly supported from within
 Bundler itself, which is currently being experimented with in a plugin called `bundler-compose`. Once
-> `bundler-compose`is production ready, the entire custom bundle created under the `.ruby-lsp` directory will go away
+> `bundler-compose`is production ready, the entire composed bundle created under the `.ruby-lsp` directory will go away
```
vinistock, line 40: Let's remove this node completely too.

lib/ruby_lsp/setup_bundler.rb
```diff
@@ -202,7 +202,7 @@ def run_bundle_install(bundle_gemfile = @gemfile)
       end
 
       # If `ruby-lsp` and `debug` (and potentially `ruby-lsp-rails`) are already in the Gemfile, then we shouldn't try
-      # to upgrade them or else we'll produce undesired source control changes. If the custom bundle was just created
+      # to upgrade them or else we'll produce undesired source control changes. If the composed bundle was just created
       # and any of `ruby-lsp`, `ruby-lsp-rails` or `debug` weren't a part of the Gemfile, then we need to run `bundle
       # install` for the first time to generate the Gemfile.lock with them included or else Bundler will complain that
       # they're missing. We can only update if the custom `.ruby-lsp/Gemfile.lock` already exists and includes all gems
```
alexcrocha, line 208:
```suggestion
      # they're missing. We can only update if the composed `.ruby-lsp/Gemfile.lock` already exists and includes all gems
```

We have references to a composed lockfile. I am assuming this is it. Should we also refer to it here as `composed`?

alexcrocha (approved):

Issue #2653, opened 2024-10-02
Support `module_function` in the indexer


vinistock:
When `module_function` is invoked, the given method becomes available in the singleton class and it is turned into a private method on including namespaces. For example:

```ruby
module A
  def foo; end
  module_function :foo
end

A.foo # okay!

class Something
  include A
end

Something.new.foo # not okay, instance method foo is private
```

We need to add processing for `module_function` [here](https://github.com/Shopify/ruby-lsp/blob/ae70ebc6348febe6064789890c221944463597ba/lib/ruby_indexer/lib/ruby_indexer/declaration_listener.rb#L291). Essentially, when we find an invocation to `module_function`, we want to:

1. Get the method entry that was already inserted in the index. Update the visibility to private
2. Add a new method entry for the same method, but in this case the owner is the singleton class

IgnacioFan:
Hi team,

I am taking a look at this issue and will push something a few days later.

vinistock:
Hello! Sounds good, I have assigned the issue to you. If you hit any blockers or have any doubts, please do not hesitate to ask for assistance here and the team can discuss the implementation.

IgnacioFan:
@vinistock, thanks for the pseudo code. I am now writing a test, but I need to figure out how to test the singleton entry.

Right now, the following code is a piece of tests placed in method_test.rb.
```ruby
def test_visibility_tracking_with_module_function
  index(<<~RUBY)
    module Foo
      def foo; end
      module_function :foo
    end
    class Bar
      include Foo
    end
    Bar.new.foo
  RUB

  assert_entry("foo", Entry::Method, "/fake/path/foo.rb:1-2:1-14", visibility: Entry::Visibility::PRIVATE)
  # How can I make another assertion to test singleton entry?
end
```

IgnacioFan:
Hi @vinistock, thank you! I think I am ready to push code for reviews, but I got permission denied 😅

vinistock:
You need to fork the repository in order to propose PRs in open source repositories you don't have permissions. This guide explains it in more detail https://docs.github.com/en/get-started/exploring-projects-on-github/contributing-to-a-project.

fnordfish:
Hi all, and thanks for your work. I found this issue because the following example is not working for me. Not sure how easy it is to implement, because it requires some form of state - all method definitions after a bare `module_function` become module functions.

```ruby
module Foo
  def a; end # not a module function

  module_function

  def foo; end # This is now a module function

[...]

PR #2733, opened 2024-10-17
Support module_function in the indexer

IgnacioFan:
### Motivation

Closes https://github.com/Shopify/ruby-lsp/issues/2653 

The `module_function` invocation creates both a singleton method and a private method when using `include` in a class. The indexer needs to account for this behavior by mutating the method's visibility and handling both method types.

### Implementation

When the indexer encounters a `module_function`, it updates the corresponding method entry:
  - Changes the instance method’s visibility to private.
  - Adds the singleton method to the indexer (public).
  
### Automated Tests

Add a new test

### DIFF
--- a/lib/ruby_indexer/lib/ruby_indexer/declaration_listener.rb
+++ b/lib/ruby_indexer/lib/ruby_indexer/declaration_listener.rb
@@ -758,33 +758,39 @@ module RubyIndexer
       arguments_node = node.arguments
       return unless arguments_node
 
-      name_argument = arguments_node.arguments.first
-      return unless name_argument.is_a?(Prism::SymbolNode)
-
-      method_name = name_argument.value.to_s
       owner_name = @owner_stack.last&.name
       return unless owner_name
 
-      entries = @index.resolve_method(method_name, owner_name)
-      return unless entries
+      arguments_node.arguments.each do |argument|
+        method_name = case argument
+        when Prism::StringNode
+          argument.content
+        when Prism::SymbolNode
+          argument.value
+        end
+        next unless method_name
 
-      entries.each do |entry|
-        entry_owner_name = entry.owner&.name
-        next unless entry_owner_name
+        entries = @index.resolve_method(method_name, owner_name)
+        next unless entries
 
-        entry.visibility = Entry::Visibility::PRIVATE
+        entries.each do |entry|
+          entry_owner_name = entry.owner&.name
+          next unless entry_owner_name
 
-        singleton = @index.existing_or_new_singleton_class(entry_owner_name)
-        @index.add(Entry::Method.new(
-          method_name,
-          @file_path,
-          Location.from_prism_location(node.location, @code_units_cache),
-          Location.from_prism_location(T.must(node.message_loc), @code_units_cache),
-          collect_comments(node),
-          [],
-          Entry::Visibility::PUBLIC,
-          singleton,
-        ))
+          entry.visibility = Entry::Visibility::PRIVATE
+
+          singleton = @index.existing_or_new_singleton_class(entry_owner_name)
+          @index.add(Entry::Method.new(
+            method_name,
+            @file_path,
+            Location.from_prism_location(node.location, @code_units_cache),
+            Location.from_prism_location(T.must(node.message_loc), @code_units_cache),
+            collect_comments(node),
+            entry.signatures,
+            Entry::Visibility::PUBLIC,
+            singleton,
+          ))
+        end
       end
     end
 
--- a/lib/ruby_indexer/lib/ruby_indexer/declaration_listener.rb
+++ b/lib/ruby_indexer/lib/ruby_indexer/declaration_listener.rb
@@ -780,11 +780,12 @@ module RubyIndexer
           entry.visibility = Entry::Visibility::PRIVATE
 
           singleton = @index.existing_or_new_singleton_class(entry_owner_name)
+          location = Location.from_prism_location(argument.location, @code_units_cache)
           @index.add(Entry::Method.new(
             method_name,
             @file_path,
-            Location.from_prism_location(node.location, @code_units_cache),
-            Location.from_prism_location(T.must(node.message_loc), @code_units_cache),
+            location,
+            location,
             collect_comments(node),
             entry.signatures,
             Entry::Visibility::PUBLIC,
--- a/lib/ruby_indexer/lib/ruby_indexer/declaration_listener.rb
+++ b/lib/ruby_indexer/lib/ruby_indexer/declaration_listener.rb
@@ -312,6 +312,29 @@ module RubyIndexer
         @visibility_stack.push(Entry::Visibility::PROTECTED)
       when :private
         @visibility_stack.push(Entry::Visibility::PRIVATE)
+      when :module_function
+        name_argument = node.arguments.arguments.first
+        return unless name_argument
+
+        entries = @index.resolve_method(name_argument.value, @owner_stack.last.name)
+        return unless entries
+
+        entries.each do |e|
+          e.visibility = Entry::Visibility::PRIVATE
+
+          singleton = @index.existing_or_new_singleton_class(e.owner.name)
+
+          @index.add(Entry::Method.new(
+            name_argument.value,
+            @file_path,
+            Location.from_prism_location(node.location, @code_units_cache),
+            Location.from_prism_location(node.message_loc, @code_units_cache),
+            collect_comments(node),
+            [],
+            Entry::Visibility::PUBLIC,
+            singleton,
+          ))
+        end
       end
 
       @enhancements.each do |enhancement|
--- a/lib/ruby_indexer/lib/ruby_indexer/declaration_listener.rb
+++ b/lib/ruby_indexer/lib/ruby_indexer/declaration_listener.rb
@@ -313,28 +313,7 @@ module RubyIndexer
       when :private
         @visibility_stack.push(Entry::Visibility::PRIVATE)
       when :module_function
-        name_argument = node.arguments.arguments.first
-        return unless name_argument
-
-        entries = @index.resolve_method(name_argument.value, @owner_stack.last.name)
-        return unless entries
-
-        entries.each do |e|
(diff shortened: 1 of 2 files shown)


lib/ruby_indexer/lib/ruby_indexer/declaration_listener.rb
```diff
@@ -312,6 +312,29 @@ def on_call_node_enter(node)
         @visibility_stack.push(Entry::Visibility::PROTECTED)
       when :private
         @visibility_stack.push(Entry::Visibility::PRIVATE)
+      when :module_function
+        name_argument = node.arguments.arguments.first
```
andyw8, line 316: Let's extract this into `handle_module_function`.

lib/ruby_indexer/lib/ruby_indexer/declaration_listener.rb
```diff
@@ -312,6 +312,29 @@ def on_call_node_enter(node)
...
+        return unless name_argument
+
+        entries = @index.resolve_method(name_argument.value, @owner_stack.last.name)
+        return unless entries
+
+        entries.each do |e|
```
andyw8, line 322: (nit) We usually avoid single-letter variable names, let's call this `entry`.

IgnacioFan:
> There are some typecheck failures.

Hi @andyw8, Thank you for reviewing! I pushed a new commit, fixed the type check failures, and signed off on the CLA. Do I have to commit "I have signed the CLA!" to rerun the CI? 

image

lib/ruby_indexer/lib/ruby_indexer/declaration_listener.rb
```diff
@@ -751,6 +753,41 @@ def handle_module_operation(node, operation)
       end
     end
 
+    sig { params(node: Prism::CallNode).void }
+    def handle_module_function(node)
+      arguments_node = node.arguments
+      return unless arguments_node
+
+      name_argument = arguments_node.arguments.first
+      return unless name_argument.is_a?(Prism::SymbolNode)
```
vinistock, line 762: Question: does Ruby allow you to use a string as the argument to `module_function` or does it only accept a symbol? If it accepts a string, we will want to handle it here as well.

lib/ruby_indexer/lib/ruby_indexer/declaration_listener.rb
```diff
@@ -751,6 +753,41 @@ def handle_module_operation(node, operation)
...
+        entry.visibility = Entry::Visibility::PRIVATE
+
+        singleton = @index.existing_or_new_singleton_class(entry_owner_name)
+        @index.add(Entry::Method.new(
+          method_name,
+          @file_path,
+          Location.from_prism_location(node.location, @code_units_cache),
+          Location.from_prism_location(T.must(node.message_loc), @code_units_cache),
+          collect_comments(node),
+          [],
```
vinistock, line 784: Here we need to pass the same signatures as the original method, otherwise features like signature help and hover will not show the correct signature.

```suggestion
          entry.signatures,
```

lib/ruby_indexer/lib/ruby_indexer/declaration_listener.rb
```diff
@@ -751,6 +753,41 @@ def handle_module_operation(node, operation)
...
+      return unless entries
+
+      entries.each do |entry|
+        entry_owner_name = entry.owner&.name
+        next unless entry_owner_name
+
+        entry.visibility = Entry::Visibility::PRIVATE
+
+        singleton = @index.existing_or_new_singleton_class(entry_owner_name)
+        @index.add(Entry::Method.new(
```
vinistock, line 778: This `@index.add` should happen only once outside of this loop. The `resolve_method` method returns an array to account for methods that are overridden, but there's no need to create multiple entries for the singleton version of the method.

Let's move it after the loop. This loop should only set all entries' visibilities to private and then we add the singleton version to the index.

To get the singleton class, you can use the same `owner_name` you used to resolve the method:

```ruby
singleton = @index.existing_or_new_singleton_class(owner_name)
```

lib/ruby_indexer/test/method_test.rb
```diff
@@ -123,6 +123,33 @@ def baz; end
...
+    def test_visibility_tracking_with_module_function
+      index(<<~RUBY)
+        module Foo
+          def foo; end
+          module_function :foo
+        end
+
+        class Bar
+          include Foo
+        end
```
vinistock, line 135: I don't think we need this class for this test.

lib/ruby_indexer/lib/ruby_indexer/declaration_listener.rb
IgnacioFan: I am confused, as I moved the `@index.add` out of the loop. Where can I get the correct `signatures` and pass it to the `@index.add`?

vinistock: That's a great question. My original comment was not correct.

Imagine we have this scenario where a method is overridden with an incompatible signature. We can't know which override wins, because there's no way to confidently determine the load order of the files (due to autoload and dynamic requires).

```ruby
# file1.rb
# If this file is loaded last, then bar accepts one parameter named `a`
class Foo
  def bar(a)
  end
end

# file2.rb
# If this file is loaded last, then bar accepts two parameters named `b` and `c`
class Foo
  def bar(b, c)
  end
end

# In this scenario, invoking `@index.resolve_method("bar", "Foo")` will return two
# entries. One for each `bar` declaration
```

Since we cannot know which override wins, I think the best decision is to add singleton entries for each override (i.e.: move the `index.add` inside of the loop and use the same signature from the original instance method).

lib/ruby_indexer/lib/ruby_indexer/declaration_listener.rb
vinistock: Ignore this. See https://github.com/Shopify/ruby-lsp/pull/2733#discussion_r1808866464

lib/ruby_indexer/lib/ruby_indexer/declaration_listener.rb
IgnacioFan: Hey, Thank you for bringing this up. I've never considered this case, but I'm glad we know it could happen.

lib/ruby_indexer/lib/ruby_indexer/declaration_listener.rb
```diff
@@ -751,6 +753,47 @@ def handle_module_operation(node, operation)
...
+        entries = @index.resolve_method(method_name, owner_name)
+        next unless entries
+
+        entries.each do |entry|
+          entry_owner_name = entry.owner&.name
+          next unless entry_owner_name
+
+          entry.visibility = Entry::Visibility::PRIVATE
+
+          singleton = @index.existing_or_new_singleton_class(entry_owner_name)
```
vinistock, line 782: Since the singleton method is being defined when invoking `module_function`, let's use the argument's location to introduce the new entry in the index.

That way, if you jump to definition on an occurrence of the method, you land on the argument of the `module_function` invocation.

```suggestion
          singleton = @index.existing_or_new_singleton_class(entry_owner_name)
          location = Location.from_prism_location(argument.location, @code_units_cache)
```

lib/ruby_indexer/lib/ruby_indexer/declaration_listener.rb
```diff
@@ -751,6 +753,47 @@ def handle_module_operation(node, operation)
...
+          @index.add(Entry::Method.new(
+            method_name,
+            @file_path,
+            Location.from_prism_location(node.location, @code_units_cache),
+            Location.from_prism_location(T.must(node.message_loc), @code_units_cache),
```
vinistock, line 787:
```suggestion
            location,
            location,
```

lib/ruby_indexer/lib/ruby_indexer/declaration_listener.rb
```diff
@@ -751,6 +753,47 @@ def handle_module_operation(node, operation)
...
+            collect_comments(node),
```
vinistock, line 788: Typically, developers will document the original method and not the call to `module_function`. For example:

```ruby
module Foo
  # documentation
  def bar; end

  module_function :bar
end
```

By collecting comments on the call node for `module_function`, we are not bringing the documentation associated with the original method.

I think here we want to concatenated the documentation for the existing entry with any other comments that exist on this call node.

```suggestion
            collect_comments(node).concat(entry.comments),
```

vinistock (commented):
This is looking great!

andyw8:
Let's also add a fixture for `module_function` for easier manual verification/testing?

(I was surprised to see that Prism [doesn't have one already](https://github.com/search?q=repo%3Aruby%2Fprism%20module_function&type=code)).

lib/ruby_indexer/lib/ruby_indexer/declaration_listener.rb
IgnacioFan: @vinistock, Upon adding the concat, I changed the test by giving a doc to `foo`. I expected only to receive "This is foo" for the `foo` entry method. However, I also got the "This is foo" comment for the `bar` method entry. Still trying to figure out what's wrong happening!

```ruby
index(<<~RUBY)
  module Test
    # This is foo
    def foo; end
    def bar; end
    module_function :foo, "bar"
  end
RUBY

entries = T.must(@index["foo"])
first_entry, second_entry = *entries
first_entry.comments  # This is foo
second_entry.comments # This is foo

# Not what I expected
entries = T.must(@index["bar"])
first_entry, second_entry = *entries
first_entry.comments  # This is foo
second_entry.comments # This is foo
```

vinistock: There's a chance this is an existing bug. If you add a line break between the `foo` and `bar` declaration, does the issue go away?

vinistock: Please do! You discovered it.

vinistock (approved):

Issue #2692, opened 2024-10-08
FormattingExtension 'Ruby LSP' is configured as formatter but it cannot format 'Ruby'-files


Skoti:
#### VS Code Version

1.93.1

#### Ruby LSP Extension Version

0.8.3

#### Ruby LSP Server Version

0.19.1

#### Ruby Version

3.3.2

#### Ruby Version Manager

rbenv

#### Installed Extensions

<details>
<summary>Click to expand</summary>

- circleci (2.8.4)
- doxdocgen (1.4.0)
- githistory (0.6.20)
- bats (0.1.10)
- Kotlin (1.7.1)
- vscode-docker (1.29.3)
- vscode-dotnet-runtime (2.1.7)
- debugpy (2024.10.0)
- python (2024.14.1)
- vscode-pylance (2024.10.1)
- remote-containers (0.388.0)
- remote-ssh (0.115.0)
- remote-ssh-edit (0.87.0)
- remote-wsl (0.88.4)
- cmake-tools (1.19.52)
- cpptools (1.21.6)
- cpptools-extension-pack (1.3.0)
- cpptools-themes (2.0.0)
- remote-explorer (0.4.3)
- vscode-pbkit (0.0.8)
- platformio-ide (3.3.3)
- vscode-yaml (1.15.0)
- ruby-lsp (0.8.3)
- code-spell-checker (3.0.1)
- io (3.19.0)
- cmake (0.0.17)
- vscode-lldb (1.11.0)
- vscode-gradle (3.16.4)
- grammarly (0.24.0)
- vscode-proto3 (0.5.5)
</details>

#### Ruby LSP Settings

<details>
<summary>Click to expand</summary>

##### Workspace

```json
{
  "formatter": "rubocop"
}
```

##### User

```json
{
  "enableExperimentalFeatures": false,
  "enabledFeatures": {
    "codeActions": true,
    "diagnostics": true,
    "documentHighlights": true,
    "documentLink": true,
    "documentSymbols": true,
    "foldingRanges": true,
    "formatting": true,
    "hover": true,
    "inlayHint": true,
    "onTypeFormatting": true,
    "selectionRanges": true,
    "semanticHighlighting": true,
    "completion": true,
    "codeLens": true,
    "definition": true,
    "workspaceSymbol": true,
    "signatureHelp": true,
    "typeHierarchy": true
  },
  "featuresConfiguration": {},
  "addonSettings": {},
  "rubyVersionManager": {
[...]

PR #2745, opened 2024-10-21
Only restart client if the contents of the watched files have changed

vinistock:
### Motivation

Related to #2692

I don't know what in the user's setup is causing `.rubocop.yml` to be touched repeatedly, but we had issues in the past when the lockfile was continuously getting updated and we got into a restart loop.

I think we should protect our file watchers from restarting when a file was touched, but the contents are still the same.

### Implementation

Since the restart watchers are based on glob patterns, the idea is to let the first restart happen and remember the SHA of the content when we performed the restart. In subsequent file watch hits, we only restart if the contents were modified, otherwise we do nothing.

### DIFF
--- a/vscode/src/workspace.ts
+++ b/vscode/src/workspace.ts
@@ -1,3 +1,5 @@
+import { createHash } from "crypto";
+
 import * as vscode from "vscode";
 import { CodeLens, State } from "vscode-languageclient/node";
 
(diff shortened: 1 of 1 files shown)


andyw8 (approved):
I'd be hesitant about doing this _only_ to target https://github.com/Shopify/ruby-lsp/issues/2692, since it's only impacting one user, but as it will have a side-benefit of reducing unnecessary restarts then let's go ahead.

lolychee:
It worked for me. 👍

PR #2751, opened 2024-10-22
Add launcher mode for gracefully handling missing dependencies

vinistock:
### Motivation

One of the biggest pain points of using the Ruby LSP is that we're tightly tied to the project's bundle. There are several reasons for why this is the case:

- We need to be able to require the project's linter/formatter and all of its extensions
- We need to be able to auto-detect and require add-ons
- Add-ons need to be able to require code from the gems they are related to
- We need to be able to auto-detect every dependency that has to be indexed for definition, completion, hover, signature help, workspace symbol and so on

Not being able to hook into the project's bundle would mean significant losses in terms of DX. However, this also means that if even a single gem failed to install, we are unable to launch the server - and that is also bad DX.

Even worse, `bundle exec` immediately exits the process if any gems are missing, which violates the LSP life cycle (the server needs to inform the editor about an initialization failure by printing a JSON back and the process should never exit prematurely).

This PR proposes a new way of launching the Ruby LSP, which should allow us to have a bit more control over missing dependencies.

### Implementation

I recommend reviewing per commit.

1. Started accepting the initialize request and a setup error as part of instantiating the server. This allows us report errors that occurred during `Bundler.setup` back to the user. It also allows us to read what is the workspace URI before invoking `Bundler.setup`, which allows us to fix the longstanding bug of not setting up the bundle in the right directory for editors that launch the server process outside of the workspace directory
2. Added a new `ruby-lsp-launcher` executable. I will explain below the reason why we need this
3. Added a new setting to control usage of the launcher in VS Code

#### Why do we need a new executable

We need to be able to:

1. Set up the composed bundle with all of the dependencies we need
2. Then control when `Bundler.setup` is invoked, so that we can fail gracefully and report errors back to the user

The problem is that when a gem's executable is invoked, you don't actually invoke the executable. You invoke Rubygems binstubs, which automatically activate all of the gem's dependencies.

That means that when `ruby-lsp` is executed, Rubygems will eagerly activate whatever latest versions of our dependencies are present. If the project's bundle is locked to any other version other than the ones eagerly activated, then `Bundler.setup` is guaranteed to fail complaining that another version of that gem has already been activated.

To work around this, we replace the current process with `exec`, invoking the `ruby-lsp-launcher` executable directly (without going through the Rubygems binstub) to avoid having our own dependencies eagerly activated and polluting the process before we had a chance to invoke `Bundler.setup`.

This approach means that no gems are activated until we finished setting up the composed bundle and manually invoke `Bundler.setup` ourselves, which ensures that no conflicts can occur and let's us handle the errors in an editor friendly way.

### Automated Tests

I will follow up with some integration tests.

### DIFF
--- a/vscode/package.json
+++ b/vscode/package.json
@@ -473,6 +473,11 @@
           "description": "Enable ERB support. This can only work with server versions v0.17.5 or above",
           "type": "boolean",
           "default": true
+        },
+        "rubyLsp.useLauncher": {
+          "description": "[EXPERIMENTAL] Uses server launcher for gracefully handling missing dependencies.",
+          "type": "boolean",
+          "default": false
         }
       }
     },
--- a/vscode/src/client.ts
+++ b/vscode/src/client.ts
@@ -62,6 +62,7 @@ function getLspExecutables(
   const customBundleGemfile: string = config.get("bundleGemfile")!;
   const useBundlerCompose: boolean = config.get("useBundlerCompose")!;
   const bypassTypechecker: boolean = config.get("bypassTypechecker")!;
+  const useLauncher: boolean = config.get("useLauncher")!;
 
   const executableOptions: ExecutableOptions = {
     cwd: workspaceFolder.uri.fsPath,
@@ -98,17 +99,25 @@ function getLspExecutables(
       options: executableOptions,
     };
   } else {
-    const argsWithBranch = branch.length > 0 ? ["--branch", branch] : [];
+    const args = [];
+
+    if (branch.length > 0) {
+      args.push("--branch", branch);
+    }
+
+    if (useLauncher) {
+      args.push("--use-launcher");
+    }
 
     run = {
       command: "ruby-lsp",
-      args: argsWithBranch,
+      args,
       options: executableOptions,
     };
 
     debug = {
       command: "ruby-lsp",
-      args: argsWithBranch.concat(["--debug"]),
+      args: args.concat(["--debug"]),
       options: executableOptions,
     };
   }


lib/ruby_lsp/server.rb
```diff
@@ -290,6 +290,18 @@ def run_initialize(message)
       begin_progress("indexing-progress", "Ruby LSP: indexing files")
 
       global_state_notifications.each { |notification| send_message(notification) }
+
+      if @setup_error
+        message = <<~MESSAGE
+          An error occurred while setting up Bundler. This may be due to a failure when installing dependencies.
```
andyw8, line 296: May be worth linking to somewhere in the docs where we can give a full explanation?

lib/ruby_lsp/server.rb
```diff
@@ -290,6 +290,18 @@ def run_initialize(message)
...
+          The Ruby LSP will continue to run, but features related to the missing dependencies will not be available.
```
andyw8, line 297:
```suggestion
          The Ruby LSP will continue to run, but features related to the missing dependencies will be limited.
```
(otherwise it may sound like the feature gets completely disabled in all places).

exe/ruby-lsp
```diff
@@ -33,6 +33,10 @@ parser = OptionParser.new do |opts|
     options[:doctor] = true
   end
 
+  opts.on("--use-launcher", "Use launcher mechanism to handle missing dependencies gracefully") do
```
andyw8, line 36: Should we label it experimental here too?

exe/ruby-lsp-launcher
```diff
@@ -0,0 +1,105 @@
...
+
+# Composed the Ruby LSP bundle in a forked process so that we can require gems without polluting the main process
+# `$LOAD_PATH` and `Gem.loaded_specs`
+pid = fork do
+  require_relative "../lib/ruby_lsp/setup_bundler"
+  require "json"
+  require "uri"
+  require "core_ext/uri"
+
+  initialize_request = JSON.parse(raw_initialize, symbolize_names: true)
```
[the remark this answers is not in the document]
st0012: That's AWESOME 😄

PR #2765, opened 2024-10-23
Prevent passing version to `bin/bundle`

### LINKED ISSUE #2731: LSP fails to start: Could not find command "_2.5.22_".
This was introduced recently by the following change in #2658 (which resolved #2532):

https://github.com/Shopify/ruby-lsp/blob/72f019a9185da16c349eb57395b65f7a4aab6f2c/lib/ruby_lsp/setup_bundler.rb#L194-L203

Now when trying to activate the LSP in Visual Studio Code I see the following error message:
Image

Log excerpts:

<details>
<summary>
Log output
</summary>

```
2024-10-17 02:42:10.644 [info] (baw-server) Checking if chruby is available on the path with command: /bin/sh -i -c 'chruby --version'
2024-10-17 02:42:10.650 [info] (baw-server) Checking if rbenv is available on the path with command: /bin/sh -i -c 'rbenv --version'
2024-10-17 02:42:10.655 [info] (baw-server) Checking if rvm is available on the path with command: /bin/sh -i -c 'rvm --version'
2024-10-17 02:42:10.662 [info] (baw-server) Checking if asdf is available on the path with command: /bin/sh -i -c 'asdf --version'
2024-10-17 02:42:10.716 [info] (baw-server) Discovered version manager none
2024-10-17 02:42:10.716 [info] (baw-server) Running command: `ruby -W0 -rjson -e 'STDERR.print("RUBY_LSP_ACTIVATION_SEPARATOR" + { env: ENV.to_h, yjit: !!defined?(RubyVM:: YJIT), version: RUBY_VERSION }.to_json + "RUBY_LSP_ACTIVATION_SEPARATOR")'` in /home/baw_web/baw-server using shell: /bin/bash
2024-10-17 02:42:11.461 [info] (baw-server) Ruby LSP> Skipping custom bundle setup since LSP dependencies are already in /home/baw_web/baw-server/Gemfile

2024-10-17 02:42:11.559 [info] (baw-server) Ruby LSP> Running bundle install for the custom bundle. This may take a while...
Ruby LSP> Command: (bundle _2.5.22_ check || bundle _2.5.22_ install) 1>&2

2024-10-17 02:42:11.721 [info] (baw-server) Could not find command "_2.5.22_".

2024-10-17 02:42:11.876 [info] (baw-server) Could not find command "_2.5.22_".

2024-10-17 02:42:12.006 [info] (baw-server) Could not find command "_2.5.22_".

2024-10-17 02:42:12.008 [info] (baw-server) [Error - 2:42:12 AM] Server process exited with code 15.
2024-10-17 02:42:12.008 [info] (baw-server) [Error - 2:42:12 AM] Server initialization failed.
2024-10-17 02:42:12.009 [info] (baw-server)   Message: Pending response rejected since connection got disposed
  Code: -32097 
2024-10-17 02:42:12.009 [info] (baw-server) [Error - 2:42:12 AM] Ruby LSP client: couldn't create connection to server.
2024-10-17 02:42:12.009 [info] (baw-server)   Message: Pending response rejected since connection got disposed
  Code: -32097 
2024-10-17 02:42:12.009 [error] (baw-server) Error starting the server: Pending response rejected since connection got disposed
2024-10-17 02:42:47.436 [info] (baw-server) [Error - 2:42:47 AM] Connection to server got closed. Server will not be restarted.
```

</details>

I tried:

- updating bundler
- installing the lsp gems in my gem file

To avoid running the custom bundle workflow but nothing worked.

I don't understand how inserting `_<version>_` between `bundler` and the command is meant to form a valid bundler command?

---

#### VS Code Version

1.94.2

0.8.5

#### Ruby LSP Server Version

Unknown

#### Ruby Version

3.2.2

#### Ruby Version Manager

none

#### Installed Extensions

<details>

- solargraph (0.24.1)
- gitlens (15.6.1)
- EditorConfig (0.16.4)
- copilot (1.239.0)
- copilot-chat (0.21.2)
- vscode-pull-request-github (0.98.0)
- vscode-test-explorer (2.22.1)
- vscode-rdbg (0.2.2)
- test-adapter-converter (0.2.0)
- ruby-lsp (0.8.5)
- code-spell-checker (3.0.1)
</details>

#### Ruby LSP Settings

<details>

##### Workspace

```json
{}
```

##### User

```json
{
  "enableExperimentalFeatures": false,
  "enabledFeatures": {
    "codeActions": true,
    "diagnostics": true,
    "documentHighlights": true,
    "documentLink": true,
    "documentSymbols": true,
    "foldingRanges": true,
    "formatting": true,
    "hover": true,
    "inlayHint": true,
    "onTypeFormatting": true,
    "selectionRanges": true,
    "semanticHighlighting": true,
    "completion": true,
    "codeLens": true,
    "definition": true,
    "workspaceSymbol": true,
    "signatureHelp": true,
    "typeHierarchy": true
  },
  "featuresConfiguration": {},
  "addonSettings": {},
  "rubyVersionManager": {
    "identifier": "auto"
  },
  "customRubyCommand": "",
  "formatter": "auto",
  "linters": null,
  "bundleGemfile": "",
  "testTimeout": 30,
  "branch": "",
  "pullDiagnosticsOn": "both",
  "useBundlerCompose": false,
  "bypassTypechecker": false,
  "rubyExecutablePath": "",
  "indexing": {},
  "erbSupport": true
}
```

</details>

vinistock:
### Motivation

Closes #2731

If the user has their local `bin` directory as part of the `PATH` and a `bin/bundle` binstub exists, we accidentally invoke the binstub rather than invoking `bundle exec`.

This is problematic because the binstub doesn't accept the same arguments as the bundle executable. Namely, we can't pass the version using `_2.5.12_` to ensure that the locked Bundler version is used and no restarts happen.

I'd argue that overriding the `bundle` executable with something that doesn't support the same arguments is not great as it prevents any tools from invoking Bundler directly, which is what they may legitimately want to do (this is the Ruby LSP's case).

However, it appears that this affects a considerable number of users, so let's put in a fix to prevent this error from happening.

### Implementation

If there's a `bin/bundle` and any parts of the path are pointing to the local `bin` directory, then we favour using the binstub with no version.

Otherwise, we continue doing the same as before. I'm hesitant to expand this to always use the binstub if present since people could have modified their binstubs and that may lead to other errors.

### Automated Tests

I added two new tests.

### DIFF
--- a/lib/ruby_lsp/setup_bundler.rb
+++ b/lib/ruby_lsp/setup_bundler.rb
@@ -186,15 +186,7 @@ module RubyLsp
         env["BUNDLE_PATH"] = File.expand_path(env["BUNDLE_PATH"], @project_path)
       end
 
-      # If there's a Bundler version locked, then we need to use that one to run bundle commands, so that the composed
-      # lockfile is also locked to the same version. This avoids Bundler restarts on version mismatches
-      base_bundle = if @bundler_version
-        env["BUNDLER_VERSION"] = @bundler_version.to_s
-        install_bundler_if_needed
-        "bundle _#{@bundler_version}_"
-      else
-        "bundle"
-      end
+      base_bundle = base_bundle_command(env)
 
       # If `ruby-lsp` and `debug` (and potentially `ruby-lsp-rails`) are already in the Gemfile, then we shouldn't try
       # to upgrade them or else we'll produce undesired source control changes. If the custom bundle was just created
@@ -325,6 +317,34 @@ module RubyLsp
       return false unless application_contents
 
       /class .* < (::)?Rails::Application/.match?(application_contents)
+    end
+
+    # Returns the base bundle command we should use for this project, which will be:
+    # - `bundle` if there's no locked Bundler version and no `bin/bundle` binstub in the $PATH
+    # - `bundle _<version>_` if there's a locked Bundler version
+    # - `bin/bundle` if there's a `bin/bundle` binstub in the $PATH
+    sig { params(env: T::Hash[String, String]).returns(String) }
+    def base_bundle_command(env)
+      path_parts = if Gem.win_platform?
+        ENV["Path"] || ENV["PATH"] || ENV["path"] || ""
+      else
+        ENV["PATH"] || ""
+      end.split(File::PATH_SEPARATOR)
+
+      bin_dir = File.expand_path("bin", @project_path)
+      bundle_binstub = File.join(@project_path, "bin", "bundle")
+
+      if File.exist?(bundle_binstub) && path_parts.any? { |path| File.expand_path(path, @project_path) == bin_dir }
+        return bundle_binstub
+      end
+
+      if @bundler_version
+        env["BUNDLER_VERSION"] = @bundler_version.to_s
+        install_bundler_if_needed
+        return "bundle _#{@bundler_version}_"
+      end
+
+      "bundle"
     end
   end
 end
--- a/test/integration_test.rb
+++ b/test/integration_test.rb
@@ -4,11 +4,9 @@
 require "test_helper"
 
 class IntegrationTest < Minitest::Test
-  def setup
-    skip("CI only") unless ENV["CI"]
-  end
-
   def test_ruby_lsp_doctor_works
+    skip("CI only") unless ENV["CI"]
+
     in_isolation do
       system("bundle exec ruby-lsp --doctor")
       assert_equal(0, $CHILD_STATUS)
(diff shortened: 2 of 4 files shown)


test/integration_test.rb
```diff
@@ -4,24 +4,84 @@
...
+
+        Bundler.with_unbundled_env do
+          capture_subprocess_io do
+            system("bundle install")
+
+            Object.any_instance.expects(:exec).with do |env, command|
+              assert(env.key?("BUNDLE_GEMFILE"))
+              assert(env.key?("BUNDLER_VERSION"))
+              assert_match(/bundle _[\d\.]+_ exec ruby-lsp/, command)
+            end.once.raises(StandardError.new("stop"))
```
andyw8, line 41: From the [docs](https://mocha.jamesmead.org/Mocha/Expectation.html#with-instance_method), would this be more idiomatic? Seems it verifies the block returns a truthy value.

```ruby
Object.any_instance.expects(:exec).with do |env, command|
  env.key?("BUNDLE_GEMFILE") &&
  env.key?("BUNDLER_VERSION") &&
  command.mach(/bundle _[\d\.]+_ exec ruby-lsp/)
end
```

test/integration_test.rb
```diff
@@ -4,24 +4,84 @@
...
+          capture_subprocess_io do
+            system("bundle install")
+
+            Object.any_instance.expects(:exec).with do |env, command|
+              assert(env.key?("BUNDLE_GEMFILE"))
+              refute(env.key?("BUNDLER_VERSION"))
+              assert_equal("bundle exec ruby-lsp", command)
+            end.once.raises(StandardError.new("stop"))
+
+            ENV["PATH"] = "./bin#{File::PATH_SEPARATOR}#{ENV["PATH"]}"
```
andyw8, line 74:
```suggestion
            ENV["PATH"] = ".#{File::PATH_SEPARATOR}bin#{File::PATH_SEPARATOR}#{ENV["PATH"]}"
```
?

test/integration_test.rb
vinistock: Good catch. Asserting in the `with` block is not a good practice.

lib/ruby_lsp/setup_bundler.rb
```diff
@@ -186,9 +186,22 @@ def run_bundle_install(bundle_gemfile = @gemfile)
...
+      bundle_binstub = File.join(@project_path, "bin", "bundle")
+
       # If there's a Bundler version locked, then we need to use that one to run bundle commands, so that the composed
       # lockfile is also locked to the same version. This avoids Bundler restarts on version mismatches
-      base_bundle = if @bundler_version
+      base_bundle = if File.exist?(bundle_binstub) && path_parts.any? do |path|
+        File.expand_path(path, @project_path) == bin_directory
+      end
+        bundle_binstub
+      elsif @bundler_version
```
alexcrocha, line 204: It might just be me not being used to inline conditional assignment, but I found this a little hard to read.

What do you think of this:

```ruby
base_bundle = setup_bundle

def setup_bundle
  if bundle_binstub_exists_and_matches?
    bundle_binstub
  elsif @bundler_version
    env["BUNDLER_VERSION"] = @bundler_version.to_s
    install_bundler_if_needed
    "bundle _#{@bundler_version}_"
  else
    "bundle"
  end
end

def bundle_binstub_exists_and_matches?
  File.exist?(bundle_binstub) && path_parts.any? do |path|
    File.expand_path(path, @project_path) == bin_directory
  end
end
```

I am also curious, other than personal preference, are there any tangible drawbacks to splitting it into multiple methods?

vinistock: I extracted the `base_bundle_command` as a method. Regarding drawbacks, it will mostly depend on the performance requirements of the code you are writing.

Dispatching a method in Ruby is always more expensive than not doing it. So, for example, if you have a high iteration loop that's critical for performance, it might be a good move to reduce the number of method dispatches that happen inside to maximize performance.

In this particular case, where the method is invoked once for the whole life cycle of the language server, it makes absolutely no difference. And it does improve readability, thanks for pointing that out!

PR #2774, opened 2024-10-24
Use Bundler CLI directly and send errors to telemetry

vinistock:
### Motivation

This PR starts invoking the Bundler CLI objects for install and update directly on the new launcher strategy instead of shelling out.

This will provide us with two benefits: one is performance since we don't have to start a whole new process to start running Bundler commands. The second is that we can now rescue the errors and send them to telemetry, so that we can understand what are the common failures to bundle install and start making it more robust.

### Implementation

The idea is to invoke `Bundler::CLI::{Install,Update}` as needed, rescue errors and report them to the editor as telemetry.

I'm also enabling YJIT in the child process that sets up the composed bundle, since it speeds up both install and update.

### Automated Tests

Added tests.

### DIFF
--- a/lib/ruby_lsp/setup_bundler.rb
+++ b/lib/ruby_lsp/setup_bundler.rb
@@ -3,6 +3,9 @@
 
 require "sorbet-runtime"
 require "bundler"
+require "bundler/cli"
+require "bundler/cli/install"
+require "bundler/cli/update"
 require "fileutils"
 require "pathname"
 require "digest"
@@ -49,6 +52,7 @@ module RubyLsp
       @custom_lockfile = T.let(@custom_dir + (@lockfile&.basename || "Gemfile.lock"), Pathname)
       @lockfile_hash_path = T.let(@custom_dir + "main_lockfile_hash", Pathname)
       @last_updated_path = T.let(@custom_dir + "last_updated", Pathname)
+      @error_path = T.let(@custom_dir + "install_error", Pathname)
 
       dependencies, bundler_version = load_dependencies
       @dependencies = T.let(dependencies, T::Hash[String, T.untyped])
@@ -108,17 +112,19 @@ module RubyLsp
     def custom_bundle_dependencies
       @custom_bundle_dependencies ||= T.let(
         begin
+          original_bundle_gemfile = ENV["BUNDLE_GEMFILE"]
+
           if @custom_lockfile.exist?
             ENV["BUNDLE_GEMFILE"] = @custom_gemfile.to_s
             Bundler::LockfileParser.new(@custom_lockfile.read).dependencies
           else
             {}
           end
+        ensure
+          ENV["BUNDLE_GEMFILE"] = original_bundle_gemfile
         end,
         T.nilable(T::Hash[String, T.untyped]),
       )
-    ensure
-      ENV.delete("BUNDLE_GEMFILE")
     end
 
     sig { void }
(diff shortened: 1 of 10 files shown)


test/setup_bundler_test.rb
```diff
@@ -658,6 +658,62 @@ def test_sets_bundler_version_to_avoid_reloads
...
+            system("bundle install")
+
+            # Run the script once to generate a custom bundle
+            run_script(dir)
+          end
+        end
+
+        capture_subprocess_io do
+          Bundler.with_unbundled_env do
+            mock_object = mock("update")
```
andyw8, line 703:
```suggestion
            mock_update = mock("update")
```

lib/ruby_lsp/setup_bundler.rb
```diff
@@ -188,6 +194,54 @@ def run_bundle_install(bundle_gemfile = @gemfile)
...
+
+      unless should_bundle_update?
+        Bundler::CLI::Install.new({}).run
+        correct_relative_remote_paths if @custom_lockfile.exist?
+        return env
+      end
+
+      # Try to auto upgrade the gems we depend on, unless they are in the Gemfile as that would result in undesired
+      # source control changes
+      gems = ["ruby-lsp", "debug", "prism"].filter_map do |dep|
```
andyw8, line 232: Equivalent?
`gems = ["ruby-lsp", "debug", "prism"].reject { |dep| @dependencies[dep] }`

test/setup_bundler_test.rb
```diff
@@ -658,6 +658,62 @@ def test_sets_bundler_version_to_avoid_reloads
...
+        File.write(File.join(dir, "gems.rb"), <<~GEMFILE)
+          source "https://rubygems.org"
+          gem "irb"
+        GEMFILE
+
+        Bundler.with_unbundled_env do
+          capture_subprocess_io do
+            system("bundle install")
+
+            mock_object = mock("install")
```
andyw8, line 673:
```suggestion
            mock_install = mock("install")
```

test/setup_bundler_test.rb
```diff
@@ -658,6 +658,62 @@ def test_sets_bundler_version_to_avoid_reloads
...
+    Dir.mktmpdir do |dir|
+      Dir.chdir(dir) do
+        File.write(File.join(dir, "gems.rb"), <<~GEMFILE)
+          source "https://rubygems.org"
+          gem "irb"
+        GEMFILE
+
+        Bundler.with_unbundled_env do
+          capture_subprocess_io do
+            system("bundle install")
```
andyw8, line 671: Let's pass `exception: true` to the `system` calls to catch any problems early?

lib/ruby_lsp/setup_bundler.rb
```diff
@@ -188,6 +194,54 @@ def run_bundle_install(bundle_gemfile = @gemfile)
...
+        dep unless @dependencies[dep]
+      end
+      gems << "ruby-lsp-rails" if @rails_app && !@dependencies["ruby-lsp-rails"]
+
+      Bundler::CLI::Update.new({ conservative: true }, gems).run
```
andyw8, line 237: Why `conservative` here? I don't think we pass that with the current approach?

test/setup_bundler_test.rb
vinistock: I'm not opposed to doing that, but we use this same technique in every other tests, so we would need to revisit everything.

lib/ruby_lsp/setup_bundler.rb
vinistock: I'm afraid we were accidentally bumping other gems with the previous approach. I want to try to be more precise in this mode.

andyw8 (approved):

PR #2779, opened 2024-10-25
Highlight definition and control block structures

### LINKED ISSUE #2494: Add block highlights to the usual highlights

### Use case

It would be great to be able to see where some particular block starts and ends, this would be useful to understand with some messy files what are we looking into or when defining some test with describes to understand where some particular describe ends, can be also done with folding ranges but i feel is useful to have this one. Left some images with examples, i think this can be an improvement on the document highlighting 

Image

Image

Image

Image

Image

### Description

Something similar to document highlighting but for the blocks.
When the cursor is over this "block starters" to highlight the corresponding end.

### Implementation

_No response_

rogancodes:
### Motivation

Closes #2494

### Implementation

If the cursor is on a definition or control node, both the start and end of the node will be highlighted.

### Automated Tests

Amended existing test case.

### Manual Tests

Place the cursor on one of the definition or control nodes.

[Screencast from 25-10-24 08:44:06 AM IST.webm](https://github.com/user-attachments/assets/e7899805-0244-4853-ae31-587236463691)

### DIFF
--- a/lib/ruby_lsp/listeners/document_highlight.rb
+++ b/lib/ruby_lsp/listeners/document_highlight.rb
@@ -86,6 +86,22 @@ module RubyLsp
         T::Array[T.class_of(Prism::Node)],
       )
 
+      DEFINITION_AND_CONTROL_NODES = T.let(
+        [
+          Prism::ModuleNode,
+          Prism::ClassNode,
+          Prism::SingletonClassNode,
+          Prism::DefNode,
+          Prism::CaseNode,
+          Prism::WhileNode,
+          Prism::UntilNode,
+          Prism::ForNode,
+          Prism::IfNode,
+          Prism::UnlessNode,
+        ],
+        T::Array[T.class_of(Prism::Node)],
+      )
+
       sig do
         params(
           response_builder: ResponseBuilders::CollectionResponseBuilder[Interface::DocumentHighlight],
@@ -98,6 +114,12 @@ module RubyLsp
         @response_builder = response_builder
 
         return unless target && parent
+
+        if DEFINITION_AND_CONTROL_NODES.include?(target.class)
+          add_highlight_for_definition_and_control_node(target)
+
+          return
+        end
 
         highlight_target =
           case target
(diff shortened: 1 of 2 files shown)


lib/ruby_lsp/listeners/document_highlight.rb
```diff
@@ -99,6 +115,12 @@ def initialize(response_builder, target, parent, dispatcher)
 
         return unless target && parent
 
+        if DEFINITION_AND_CONTROL_NODES.include?(target.class)
+          add_highlight_for_definition_and_control_node(target)
+
+          return
+        end
```
vinistock, line 122: This approach doesn't quite fit the listener architecture and it's a bit odd to be returning early in the initialize with a response.

We will need to ensure that we have listener events for all of the desired nodes and then in each one of them we add the highlight if applicable.

vinistock (changes requested):

lib/ruby_lsp/listeners/document_highlight.rb
vinistock: No, we wouldn't want to highlight all of them. Only the one under the cursor. It would need to be something like this:

```ruby
def on_class_node_enter(node)
  if @target == node && position_covers_def_keyword
    # highlight `def` and `end`
  end
end
```

Currently, the document highlight listener doesn't have access to the position, but we can pass that in the initializer.

rogancodes: ok, I’ll proceed with this approach.

lib/ruby_lsp/listeners/document_highlight.rb
```diff
@@ -252,13 +273,34 @@ def on_required_parameter_node_enter(node)
 
       sig { params(node: Prism::ClassNode).void }
       def on_class_node_enter(node)
+        if @target.is_a?(Prism::ClassNode)
+          class_keyword_loc = node.class_keyword_loc
+          end_keyword_loc = node_closing_keyword_loc(node)
+          if end_keyword_loc && (covers_target_position?(class_keyword_loc) || covers_target_position?(end_keyword_loc))
+            add_highlight(Constant::DocumentHighlightKind::TEXT, class_keyword_loc)
+            add_highlight(Constant::DocumentHighlightKind::TEXT, end_keyword_loc)
```
vinistock, line 281: This logic seems to be repeated for all nodes. Can we extract this into a method that accepts both locations and then add the highlights if necessary?

lib/ruby_lsp/listeners/document_highlight.rb
```diff
@@ -550,6 +683,34 @@ def node_value(node)
...
+            node.end_keyword_loc
+          when Prism::WhileNode, Prism::UntilNode
+            node.closing_loc
+          end
+        )
+        return if location.nil?
+
+        # Sometimes, the closing keyword may not be present.
+        location.length.positive? ? location : nil
+      end
```
vinistock, line 702: This method is always invoked from places where we already know the node types, which is a bit odd.

Instead of doing that, let's invoke either `end_key_loc` or `closing_loc` directly on the event handlers and let's make the method extracted to add the highlights handle the possibility of the location being `nil`. For example, something like this

```ruby
def on_until_node_enter(node)
  if @target.is_a?(Prism::UntilNode)
    add_matching_end_highlights(node.keyword_loc, node.closing_loc)
  end  
end

def add_matching_end_highlights(keyword_loc, end_loc)
  return unless keyword_loc && end_loc
  
  return unless (covers_target_position?(until_keyword_loc) || covers_target_position?(end_keyword_loc))

  add_highlight(Constant::DocumentHighlightKind::TEXT, keyword_loc)
  add_highlight(Constant::DocumentHighlightKind::TEXT, end_loc)
end
```

vinistock (commented):
This is looking good!

vinistock (approved):

waynehoover:
Are there any `editor.semanticTokenColorCustomizations` rules we need to add to make this visible? I'm not seeing it using the github light color theme in vscode.

waynehoover:
Thanks, my `editor.occurrencesHighlight` was sot to off, which is why it wasn't showing. 

Is there anyway to make the def/end highlighting continue to highlight when I'm inside the method? Like what Zed does, here is an example:

https://share.zight.com/yAugGqED

jbigler:
Here's a visual example of my comment above:
image
My cursor is on the `end` statement, but it is confused by the `class:` argument and so isn't highlighting the corresponding `do` statement.

PR #2787, opened 2024-10-27
Add references support for instance variables

### LINKED ISSUE #2641: Add references support for instance variables
Build on top of #2632 to add support for finding instance variable references.

monkeyWzr:
### Motivation

Closes #2641 

### Implementation

Add references support for instance variables.

### Manual Tests

Kapture 2024-10-27 at 21 57 23

### DIFF
--- a/lib/ruby_indexer/lib/ruby_indexer/reference_finder.rb
+++ b/lib/ruby_indexer/lib/ruby_indexer/reference_finder.rb
@@ -283,32 +283,32 @@ module RubyIndexer
 
     sig { params(node: Prism::InstanceVariableReadNode).void }
     def on_instance_variable_read_node_enter(node)
-      collect_instance_variable_references(node.name.to_s, node.location)
+      collect_instance_variable_references(node.name.to_s, node.location, false)
     end
 
     sig { params(node: Prism::InstanceVariableWriteNode).void }
     def on_instance_variable_write_node_enter(node)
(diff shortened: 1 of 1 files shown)


monkeyWzr:
I have signed the CLA!

vinistock (approved):
This looks and we can ship. For some reason though, CI seems to be stuck and won't run. Could you please try rebasing this branch to trigger CI again?

Regarding the discussion in the issue, ideally we would use our type inferrer to narrow instance variable references only to the ones that actually refer to the same one. Would you be interested in tackling that in another PR? If not, I'll create an issue for it.

We'd essentially need to require the type inferrer to be passed to the references finder and then we need to ensure that the receiver type of the instance variables match the one the user clicked on.

monkeyWzr:
@vinistock 

>Regarding the discussion in the issue, ideally we would use our type inferrer to narrow instance variable references only to the ones that actually refer to the same one. Would you be interested in tackling that in another PR? If not, I'll create an issue for it.

Yeah I'd love to give it a shot!

vinistock:
> Yeah I'd love to give it a shot!

Awesome! It should be something along these lines:

1. We need the instance variable target to remember what type it belongs to, so that we can compare with other occurrences
2. We need to pass the type inferrer instance all the way to the reference finder, so that it can be used for instance variables
3. We can only consider it a valid reference if both name and receiver type match between the target and the occurrence

Don't hesitate to reach out if you have doubts!

PR #2800, opened 2024-10-29
Show that server is running on degraded mode

vinistock:
### Motivation

Let's show in the language status item if the server is running in degraded mode, so that it's easier to realize why full functionality may not be available.

### Implementation

In essence, if there was an install or a setup error, then we are in degraded mode. We return that as part of the initialize result and use that to change the status entry message.

### Automated Tests

Added a test.

### DIFF
--- a/vscode/src/status.ts
+++ b/vscode/src/status.ts
@@ -92,8 +92,15 @@ export class ServerStatus extends StatusItem {
         this.item.text = workspace.lspClient.serverVersion
           ? `Ruby LSP v${workspace.lspClient.serverVersion}: Running`
           : "Ruby LSP: Running";
+
+        if (workspace.lspClient.degraded) {
+          this.item.text += " (degraded)";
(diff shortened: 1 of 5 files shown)


vscode/src/status.ts
```diff
@@ -109,6 +109,12 @@ export class ServerStatus extends StatusItem {
         break;
       }
     }
+
+    if (workspace.lspClient.degraded) {
```
andyw8, line 113: Do you have an expectation for how this should behave in conjunction with the Start/Stop/Restart commands?

st0012: Should this be put under the `State.Running` case instead?

vinistock: Yeah, I think that makes sense. Updated the code to only append `(degraded)` and change the status.

andyw8 (approved):

PR #2801, opened 2024-10-29
Print Bundler progress to stderr when invoking CLI directly

vinistock:
### Motivation

In launcher mode #2778, we don't see the Bundler output printed because we're setting the UI to silent. If we make the Thor shell used by Bundler print to stderr instead of stdout, then we can see the output normally.

### Implementation

Started prepending a patch to the Thor shell to guarantee we can only print to stderr. This is only used in launch mode.

### Automated Tests

Added a test.

### DIFF
--- a/lib/ruby_lsp/setup_bundler.rb
+++ b/lib/ruby_lsp/setup_bundler.rb
@@ -32,6 +32,7 @@ module RubyLsp
       @project_path = project_path
       @branch = T.let(options[:branch], T.nilable(String))
       @launcher = T.let(options[:launcher], T.nilable(T::Boolean))
+      patch_thor_to_print_progress_to_stderr! if @launcher
 
       # Regular bundle paths
       @gemfile = T.let(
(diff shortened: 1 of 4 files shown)


lib/ruby_lsp/setup_bundler.rb
```diff
@@ -32,6 +32,7 @@ def initialize(project_path, **options)
       @project_path = project_path
       @branch = T.let(options[:branch], T.nilable(String))
       @launcher = T.let(options[:launcher], T.nilable(T::Boolean))
+      print_progress_to_stderr! if @launcher
```
andyw8, line 35:
```suggestion
      patch_thor_to_print_progress_to_stderr! if @launcher
```

lib/ruby_lsp/setup_bundler.rb
```diff
@@ -402,5 +403,19 @@ def base_bundle_command(env)
 
       "bundle"
     end
+
+    sig { void }
+    def print_progress_to_stderr!
+      Bundler::Thor::Shell::Basic.prepend(Module.new do
```
andyw8, line 409: Are we confident this will work in older Bundler versions? (e.g. has this part of Bundler been stable for a long time?)

andyw8 (approved):

PR #2802, opened 2024-10-30
Fix a typo in the activation script

Earlopain:
I for sure thought it would change meaning but turns out this space doesn't matter and the ast is the same.


andyw8:
Interesting!

```
irb(main):010> defined?(RubyVM::YJIT)
=> "constant"
irb(main):011> defined?(RubyVM:: YJIT)
=> "constant"
irb(main):012> defined?(RubyVM ::YJIT)
=> nil
irb(main):013> defined?(RubyVM :: YJIT)
=> nil
```

Earlopain:
You can add as many spaces to the right as you want. Might be a ruby bug, not sure, `defined?` _is_ pretty weird.

PR #2803, opened 2024-10-30
Handle responses for show window request and delegate to add-ons

vinistock:
### Motivation

Add-ons won't have a lot of flexibility when it comes to UI elements since the LSP specification doesn't prescribe that. However, we can still allow for a certain degree of interactions by delegating responses of the `window/showMessageRequest` request.

This request made by the server allows LSPs to show a dialog with buttons to the user. The response includes which button was clicked and any other attributes (as long as the client supports additional properties).

We can use this to allow for more interactivity for add-ons. The current use case we will use this for is letting the Rails add-on inform the user there's a pending migration with a button to run it.

### Implementation

Started handling responses for `window/showMessageRequest`. We enforce the add-ons must provide their name as part of the original request in order to receive the response delegated appropriately.

Also, started remembering if the client supports additional properties. Add-ons can only use this mechanism in such cases because otherwise they don't have the means to inform us of their name for delegation.

### Automated Tests

Added a test.

### DIFF
--- a/lib/ruby_lsp/addon.rb
+++ b/lib/ruby_lsp/addon.rb
@@ -194,6 +194,13 @@ module RubyLsp
     sig { abstract.returns(String) }
     def version; end
 
+    # Handle a response from a window/showMessageRequest request. Add-ons must include the addon_name as part of the
+    # original request so that the response is delegated to the correct add-on and must override this method to handle
+    # the response
+    # https://microsoft.github.io/language-server-protocol/specification#window_showMessageRequest
+    sig { overridable.params(title: String).void }
+    def handle_window_show_message_response(title); end
+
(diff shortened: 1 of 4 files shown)


st0012:
How can we trigger the request in the first place? Through the VS Code extension?

vinistock:
No, the server sends a request to the client and it returns a response once something was clicked.

It's the [window/showMessageRequest](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#window_showMessageRequest). Essentially, you would do this:

```ruby
# Send the request to the client
outgoing_queue << Request.new(
  id: @current_request_id,
  method: "window/showMessageRequest",
  params: ...
)

# Then once the user clicks on one of the buttons, we receive the response and delegate it
```

PR #2805, opened 2024-10-30
Refactor client capabilities into its own class

vinistock:
### Motivation

As pointed out by @andyw8, our implementation of the global state was growing a little too much and so was the body of `apply_options`. This PR extracts a new object to store client capabilities, so that we can organize concerns a bit better.

### Implementation

Extracted the client capabilities into an object and adapted the code that was using them for the new structure.

### Automated Tests

Updated existing tests.


lib/ruby_lsp/requests/rename.rb
```diff
@@ -72,7 +72,7 @@ def perform
 
         # If the client doesn't support resource operations, such as renaming files, then we can only return the basic
         # text changes
-        unless @global_state.supported_resource_operations.include?("rename")
+        unless @global_state.client_capabilities.supported_resource_operations.include?("rename")
```
andyw8, line 75: (nit) Could add a method `ClientCapabilities#supports_rename?` so there's one less level to drill through.

andyw8 (approved):

lib/ruby_lsp/requests/rename.rb
vinistock: Good idea. Added that method.

PR #2810, opened 2024-10-31
Prevent composed bundle from failing with SSH git sources

vinistock:


vinistock:
* **#2810** <a href="https://app.graphite.dev/github/pr/Shopify/ruby-lsp/2810?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

This stack of pull requests is managed by Graphite. <a href="https://stacking.dev/?utm_source=stack-comment">Learn more about stacking.</a>
<h2></h2>

 Join @vinistock and the rest of your teammates on <a href="https://graphite.dev?utm-source=stack-comment">Graphite <b>Graphite</b></a>

KaanOzkan (approved):
I guess it's fine to not cover cases where exception is due to an faulty source

vinistock:
> I guess it's fine to not cover cases where exception is due to a faulty source

Yeah, in that scenario, bundle install will fail. With the new experimental launcher mode, we will still be able to launch the server despite that happening and report the problem to the user, but there's not much else that we can do.

PR #2811, opened 2024-10-31
Add supports_watching_files to GlobalState

vinistock:
### Motivation

To prevent breaking the activation of the Standard add-on, which relies on the old API and doesn't declare version constraints using our new depends_on API.

### Implementation

Added the method they rely back, delegating to client capabilities.

### Automated Tests

Added a test.

### DIFF
--- a/lib/ruby_lsp/global_state.rb
+++ b/lib/ruby_lsp/global_state.rb
@@ -159,6 +159,11 @@ module RubyLsp
       end
     end
 
+    sig { returns(T::Boolean) }
+    def supports_watching_files
+      @client_capabilities.supports_watching_files
+    end
+
     private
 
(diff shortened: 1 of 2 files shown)


st0012 (approved):

lib/ruby_lsp/global_state.rb
```diff
@@ -159,6 +159,11 @@ def encoding_name
       end
     end
 
+    sig { returns(T::Boolean) }
```
andyw8, line 162: (nit) maybe worth a comment saying this is deprecated, and only for backwards compatibility.

andyw8 (approved):

lib/ruby_lsp/global_state.rb
vinistock: Let's get this out first to prevent breaking people's setups.

PR #2814, opened 2024-10-31
Ensure we're handling client responses for window show message

vinistock:
### Motivation

When the server makes a request to the client, the response doesn't include the requested method by default. We can make it so that add-ons need to include it as part of the message properties, but normally only the ID can be used to reconcile request and response.

We need to let add-ons include the method name as part of the properties so that we can delegate appropriately.

### Implementation

Our regular `process_message` would just ignore any responses because `message[:method]` is always `nil` in a response. We need to extract it from `result` and define that add-ons must include the `method` as part of it.

### Automated Tests

Adapted the test.

### DIFF
--- a/lib/ruby_lsp/server.rb
+++ b/lib/ruby_lsp/server.rb
@@ -81,8 +81,6 @@ module RubyLsp
         workspace_did_change_watched_files(message)
       when "workspace/symbol"
         workspace_symbol(message)
-      when "window/showMessageRequest"
-        window_show_message_request(message)
       when "rubyLsp/textDocument/showSyntaxTree"
         text_document_show_syntax_tree(message)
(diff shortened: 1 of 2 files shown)


vinistock:
* **#2814** <a href="https://app.graphite.dev/github/pr/Shopify/ruby-lsp/2814?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

This stack of pull requests is managed by Graphite. <a href="https://stacking.dev/?utm_source=stack-comment">Learn more about stacking.</a>
<h2></h2>

 Join @vinistock and the rest of your teammates on <a href="https://graphite.dev?utm-source=stack-comment">Graphite <b>Graphite</b></a>

PR #2828, opened 2024-11-05
Docs/add rubocop minimum version

### LINKED ISSUE #1795: Mention minimum required dependent gem versions
When working with an existing project which has rubocop in its Gemfile, it needs to be a compatible version, or a user gets this error message:

```
Error running diagnostics: Incompatible RuboCop version. Ruby LSP requires >= 1.4.0
```

It looks like this comes from [ruby-lsp](https://github.com/Shopify/ruby-lsp/blob/1ac69c9b4187bcae056b577122f7f8db2ff96346/lib/ruby_lsp/requests/support/rubocop_runner.rb#L13) but it doesn't mention it in readme/docs directly either.

vinibispo:
### Motivation

Resolves #1795 

### Implementation

- I've created a new subsection under the usage section called implicit dependencies to let users know the required dependencies that we need them to use
- I've added a comment reminder to update the documentation when changing the rubocop dependency version
### Automated Tests

I didn't run, because I've only changed the documentation
### Manual Tests
I've served Jekyll and went to the browser page to see how it looks
image

### DIFF
--- a/jekyll/index.markdown
+++ b/jekyll/index.markdown
@@ -23,7 +23,7 @@ Want to discuss Ruby developer experience? Consider joining the public
     - [With VS Code](#with-vs-code)
     - [With other editors](#with-other-editors)
     - [Composed Ruby LSP bundle](#composed-ruby-lsp-bundle)
-    - [Implicit Dependencies](#implicit-dependencies)
+    - [Dependency Compatibility](#dependency-compatibility)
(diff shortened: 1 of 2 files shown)


vinibispo:
@vinistock @andyw8 I've just signed the Individual Contributor License Agreement. Can you rerun the CLA?

PR #2833, opened 2024-11-06
Move content from VS Code extension README into the main docs site

andyw8:
Let's unify the documentation. I cleaned up a little redundancy (see commits) but there may be more work to do to remove overlap or inconsistencies.


andyw8:
(fixing the unintentional Prism changes)

andyw8:
TODO: restore the original README since it's needed for the VS Code extension marketplace

andyw8:
This got stale, see https://github.com/Shopify/ruby-lsp/pull/3103 instead.

PR #2848, opened 2024-11-14
Adds support for attr_* method references

tommycrumrine:

### DIFF
--- a/lib/ruby_indexer/lib/ruby_indexer/reference_finder.rb
+++ b/lib/ruby_indexer/lib/ruby_indexer/reference_finder.rb
@@ -264,12 +264,27 @@ module RubyIndexer
 
     sig { params(node: Prism::CallNode).void }
     def on_call_node_enter(node)
-      if @target.is_a?(MethodTarget) && (name = node.name.to_s) == @target.method_name
+      return unless  @target.is_a?(MethodTarget)
+
+      if (name = node.name.to_s) == @target.method_name
         @references << Reference.new(name, T.must(node.message_loc), declaration: false)
+      elsif node.name == :attr_writer && attr_writer_matches_method(node.arguments.arguments, @target.method_name)
+        @references << Reference.new(@target.method_name, T.must(node.message_loc), declaration: true)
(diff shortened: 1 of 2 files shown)


lib/ruby_indexer/test/reference_finder_test.rb
```diff
@@ -143,6 +143,152 @@ def baz
       assert_equal(9, refs[1].location.start_line)
     end
 
+    def test_matches_attr_writer_with_call_node_argument
+      refs = find_method_references("foo=", <<~RUBY)
+        class Bar
+          attr_reader :foo, bar
```
vinistock, line 149: Should this be `attr_writer` or `attr_accessor`?

lib/ruby_indexer/lib/ruby_indexer/reference_finder.rb
```diff
@@ -264,13 +264,51 @@ def on_def_node_leave(node)
...
+      when :attr_accessor
+        attr_accessor_references?(unescaped_argument_names(node))
+      else
+        false
+      end
+    end
+
+    sig { params(node: Prism::CallNode).returns(T::Array[String]) }
+    def unescaped_argument_names(node)
+      node.arguments.arguments.select { |arg| arg.respond_to?(:unescaped) }.map(&:unescaped)
```
vinistock, line 294: In the middle of typing, we may find empty arguments, so we have to handle that. Also, we need to account for symbol and string nodes as arguments.

```suggestion
      arguments = node.arguments.arguments
      return [] unless arguments

      arguments.filter_map do |arg|
        case arg
        when Prism::StringNode
          arg.unescaped
        when Prism::SymbolNode
          arg.value
        end
      end
```

vinistock (commented):
Just a small change, but this looks great.

lib/ruby_indexer/test/reference_finder_test.rb
tommycrumrine: Thanks for the feedback @vinistock and apologies for the delay!! Heads up, I re-opened the PR [here](https://github.com/Shopify/ruby-lsp/pull/3706) including fixes for both your comments.

PR #2850, opened 2024-11-14
Display output channel when addons are clicked

st0012:
This way, when users see that an addon is errored, they can quickly navigate to the output channel to see the error.

### Motivation

Closes https://github.com/Shopify/team-ruby-dx/issues/1324

### Implementation

Since the Output tab is the only place we display information about addons now, I made the addon quick pick items clickable to navigate to it.

Currently clicking any addon will always go to the same `Ruby LSP` output channel. But if we started creating new channels for individual addons, we can provide a better experience by pointing to addon-specific channels too.

### DIFF
--- a/vscode/src/rubyLsp.ts
+++ b/vscode/src/rubyLsp.ts
@@ -287,14 +287,7 @@ export class RubyLsp {
         }
 
         const options: vscode.QuickPickItem[] = client.addons
-          .sort((addon) => {
-            // Display errored addons last
-            if (addon.errored) {
-              return 1;
-            }
-
-            return -1;
-          })
+          .sort((addon) => (addon.errored ? 1 : -1))
           .map((addon) => {
             const icon = addon.errored ? "$(error)" : "$(pass)";
             return {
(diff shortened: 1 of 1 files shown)


vscode/src/rubyLsp.ts
```diff
@@ -287,24 +287,32 @@ export class RubyLsp {
...
 
-        await vscode.window.showQuickPick(options, {
-          placeHolder: "Addons (readonly)",
+        const quickPick = vscode.window.createQuickPick();
+        quickPick.items = options;
+        quickPick.placeholder = "Addons (click to view output)";
+
+        quickPick.onDidAccept(() => {
+          const selected = quickPick.selectedItems[0];
+          // Ideally, we should display information that's specific to the selected addon
```
vinistock, line 304: If we provide an API in the server that tags logs with the add-on name, then you might be able to reveal the output channel with a filter applied, so that you only see the output related to that add-on.

vinistock (approved):

vscode/src/rubyLsp.ts
st0012: Yeah that will be great too 👍

PR #2855, opened 2024-11-14
Add support for `extend self` to `handle_module_operation`

### LINKED ISSUE #2782: Add support for `extend self` in the indexer
If you invoke `extend self` in a module, every instance method added to that module becomes available as singleton methods as well, all with public visibility.

```ruby
module Foo
  def bar
  end

  extend self

  def baz
  end
end

# both `bar` and `baz` are public singleton and instance methods
```

Handling this a bit tricky because you can invoke `extend self` anywhere in the body of the module. I think the right way to handle this is like this:

- When we find an `extend self`, we take every instance method the module defines and add their singleton versions to the index as well
- Additionally, we mark a new attribute on the module to remember that it has extended `self`
- Then we always check if the module extended self and, if it did, any new instance method also becomes a singleton method

Since modules can be re-opened in multiple files, I think that's the only way to guarantee that we end up with an accurate representation.

tlemburg:
By adding the instance methods as class methods through the singleton mixin operations, we are able to get all instance methods "copied" over into the class methods implicitly.

I don't think it's necessary to test whether this extend self applies when the module is opened after being closed again, or re-opened in another file, but we can add that as well if preferred.

### Motivation

Closes #2782

### Implementation

I simply allowed a `SelfNode` as an argument to enter the `handle_module_operation` method and then utilized
an `Include` entry to include the methods onto the singleton.

### Automated Tests

Added tests checking that the method resolves as an instance method to the class as well as a class method.

### Manual Tests

Tried a test file and ensured that methods were represented as class methods:

Screenshot 2024-11-14 at 3 50 52 PM


lib/ruby_indexer/test/index_test.rb
```diff
@@ -1672,6 +1672,26 @@ def test_linearizing_singleton_object
...
+          def baz
+          end
+        end
+      RUBY
+
+      ["bar", "baz"].product(["Foo", "Foo::<Class:Foo>"]).each do |method, receiver|
+        entry = @index.resolve_method(method, receiver)&.first
+        refute_nil(entry)
+        assert_equal(method, T.must(entry).name)
+      end
```
vinistock, line 1692: Let's also add an assertion for the expected linearized ancestors, which should include `Foo` itself after the singleton class. This will make it easier to understand what the expected inheritance structure is.

```suggestion
      end

      assert_equal(
        [
          "Foo::<Class:Foo>",
          "Foo",
          "Module",
          "Object",
          "Kernel",
          "BasicObject",
        ],
        @index.linearized_ancestors_of("Foo::<Class:Foo>"),
      )
```

vinistock (approved):
This solution is much better than what I had originally mentioned in the issue. The code looks good to me, let's just add one more assertion and we can ship.

lib/ruby_indexer/test/index_test.rb
tlemburg: Will do!

tlemburg:
> Thanks for your contribution!

You're welcome! Great you meet you and @andyw8 at the conf!

PR #2858, opened 2024-11-14
Support `private_class_method` in the indexer

### LINKED ISSUE #2781: Add support for `private_class_method` in the indexer
Very similar to #2653 and how we handle private constants. We need to start taking `private_class_method` into consideration in the indexer.

When not using the `class << self` syntax, the `private` method has no impact whatsoever on the visibility of methods and the only way to mark them as private is by invoking `private_class_method`.

```ruby
class Foo
  private

  # this method is public
  def self.bar
  end

  # this makes it private
  private_class_method(:bar)
end
```

Essentially, we need to:

- Add a new branch [here](https://github.com/Shopify/ruby-lsp/blob/d21d40f621d976dcce470ea330ead91c555e4b85/lib/ruby_indexer/lib/ruby_indexer/declaration_listener.rb#L315) to handle the `private_class_method`
- Grab the entry that was already inserted in the index for that method and owner combination
- Mark the method as private

kcdragon:
### Motivation

Closes #2781

`private_class_method` can be used to declare class methods as private. Prior to this PR, this method was not setting the visibility to private for the methods in the arguments.

### Implementation

My implementation is similar to the `module_function` implementation which was mentioned in the issue.
I handled the case where the arguments are `String`s, `Symbol`s or an `Array` of those. They are mentioned as the valid arguments in [the Ruby docs](https://ruby-doc.org/3.3.6/Module.html#method-i-private_class_method).

### Automated Tests

I added tests to cover `private_class_method` being passed a combination or `String`s and `Symbol`s or an `Array` of `String`s and `Symbol`s.

### Manual Tests

In this example, 
```
class Test
  def self.foo
  end

  def self.bar
  end

  private_class_method(:foo)
end
```
`foo` does not get suggested but `bar` does.

Screenshot 2024-11-14 at 5 03 50 PM
Screenshot 2024-11-14 at 5 04 00 PM


kcdragon:
I have signed the CLA!

lib/ruby_indexer/test/method_test.rb
```diff
@@ -149,6 +149,48 @@ def bar; end
...
+
+          private_class_method("foo", :bar)
+        end
+      RUBY
+
+      ["foo", "bar"].each do |keyword|
+        entries = T.must(@index[keyword])
+        assert_equal(1, entries.size)
+        entry = entries.first
+        assert_equal(Entry::Visibility::PRIVATE, entry.visibility)
```
andyw8, line 169:
```suggestion
        assert_predicate(entry.visibility, :private?)
```
(I updated some other occurrences in https://github.com/Shopify/ruby-lsp/pull/2859)

andyw8:
Looks great!

Another way that `private_class_method` can be used is a a prefix to `def`, e.g.:

```ruby
private_class_method def foo
  ...
end
```

Can we add a test to verify this doesn't cause anything to break?

lib/ruby_indexer/test/method_test.rb
```diff
@@ -149,6 +149,48 @@ def bar; end
...
+
+          private_class_method(["foo", :bar])
+        end
+      RUBY
+
+      ["foo", "bar"].each do |keyword|
+        entries = T.must(@index[keyword])
+        assert_equal(1, entries.size)
+        entry = entries.first
+        assert_equal(Entry::Visibility::PRIVATE, entry.visibility)
```
andyw8, line 190:
```suggestion
        assert_predicate(entry.visibility, :private?)
```

andyw8:
I pushed a commit to the branch below illustrating how an inline `private_class_method def self.foo` could be handled:

https://github.com/Shopify/ruby-lsp/tree/andyw8/def-private-class-method

andyw8:
@vinistock we do:

https://github.com/Shopify/ruby-lsp/blob/592495eace9286f6845d2dbafc003d399e689000/lib/ruby_indexer/lib/ruby_indexer/declaration_listener.rb#L333-L337

(and I verified manually)

kcdragon:
Thanks for the feedback! I'll update the PR over the next couple of days.

lib/ruby_indexer/lib/ruby_indexer/declaration_listener.rb
```diff
@@ -805,6 +808,45 @@ def handle_module_function(node)
...
+          [argument]
+        when Prism::ArrayNode
+          argument.elements
+        else
+          []
+        end
+
+        unless string_or_symbol_nodes.empty?
+          # pop the visibility off since there isn't a method definition following `private_class_method`
+          @visibility_stack.pop
```
kcdragon, line 825: I needed to pop this off the stack otherwise subsequent class method definitions would also get the private visibility.

andyw8 (commented):
Looks great! Once that conflict is resolved then we should be good to ship this.

andyw8 (approved):

kcdragon:
@andyw8 @vinistock Thanks for the help. This was a great project to work on for RubyConf Hack Day.

PR #2861, opened 2024-11-16
Eagerly compute SHA's for watched files

vinistock:
### Motivation

This PR fixes two issues related to automated restarts. The first one is preventing undesired restarts immediately after the server finishes booting.

The second one is preventing unwanted restarts if a create event is fired for an existing file. In theory, this shouldn't happen, but our tests tell a different story and we consistently see a create event getting fired for updates.

### Implementation

The two fixes are:

1. Unified all of our watchers to only check the top level configuration files and then started eagerly computing the SHAs for all of them (if they are present). This will prevent unwanted restarts if these files are touched immediately after boot
2. Started using the hash check for `didCreate` as well. If the file being updated fires a create event and the content SHA matches

### Automated Tests

Added a new test and reduced some of the sleeps on the other test, which is taking a bit too long.

### DIFF
--- a/vscode/src/workspace.ts
+++ b/vscode/src/workspace.ts
@@ -14,6 +14,13 @@ import {
 } from "./common";
 import { WorkspaceChannel } from "./workspaceChannel";
 
+const WATCHED_FILES = [
+  "Gemfile.lock",
+  "gems.locked",
+  ".rubocop.yml",
+  ".rubocop",
+];
+
 export class Workspace implements WorkspaceInterface {
   public lspClient?: Client;
   public readonly ruby: Ruby;
@@ -58,8 +65,6 @@ export class Workspace implements WorkspaceInterface {
     this.createTestItems = createTestItems;
     this.isMainWorkspace = isMainWorkspace;
     this.virtualDocuments = virtualDocuments;
-
-    this.registerRestarts(context);
   }
 
   // Activate this workspace. This method is intended to be invoked only once, unlikely `start` which may be invoked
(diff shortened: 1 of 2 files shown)


vinistock:
* **#2861** <a href="https://app.graphite.dev/github/pr/Shopify/ruby-lsp/2861?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/Shopify/ruby-lsp/2861?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`

<h2></h2>

vscode/src/workspace.ts
```diff
@@ -342,26 +361,14 @@ export class Workspace implements WorkspaceInterface {
...
       vscode.workspace.onDidChangeConfiguration(async (event) => {
         if (event.affectsConfiguration("rubyLsp")) {
-          // Re-activate Ruby if the version manager changed
-          if (
-            event.affectsConfiguration("rubyLsp.rubyVersionManager") ||
-            event.affectsConfiguration("rubyLsp.bundleGemfile") ||
-            event.affectsConfiguration("rubyLsp.customRubyCommand")
-          ) {
-            await this.ruby.activateRuby();
-          }
```
vinistock, line 363: This already happens on `start`, so this code was actually causing duplicate Ruby env activation.

vscode/src/workspace.ts
```diff
@@ -342,67 +363,49 @@ export class Workspace implements WorkspaceInterface {
...
-
-          this.outputChannel.info(
-            "Restarting the Ruby LSP because configuration changed",
-          );
-          await this.restart();
+          await this.debouncedRestart("configuration changed");
```
vinistock, line 374: We should stop invoking `restart` directly. The debounced version guarantees that we won't perform a double restart if two independent reasons trigger them.

vscode/src/workspace.ts
```diff
@@ -14,6 +14,13 @@ import {
 } from "./common";
 import { WorkspaceChannel } from "./workspaceChannel";
 
+const WATCHED_FILES = [
+  "Gemfile.lock",
+  "gems.locked",
+  ".rubocop.yml",
```
andyw8, line 20: It looks like we were previously matching in `.rubocop.yml` in any location, but now only in the root?

vinistock:
### Merge activity

* **Nov 19, 1:17 PM EST**: A user merged this pull request with [Graphite](https://app.graphite.dev/github/pr/Shopify/ruby-lsp/2861).

PR #2864, opened 2024-11-17
Prevent full indexing (`index_all`) from being called multiple times

andyw8:
During the RubyConf 2024 hackday, while exploring some indexing behaviour using `irb`, there was some confusion because the index was referring to code that had been deleted.

In this PR, I've added a check that raises on attempting to called `index_all` if it already contains entries.

We _could_ use a more technical approach such as adding a `reindex` method, or clearing the index each time when calling `index_all`. But since this is a relative rare usage, I think this protection is enough for the time being.

### DIFF
--- a/lib/ruby_indexer/lib/ruby_indexer/index.rb
+++ b/lib/ruby_indexer/lib/ruby_indexer/index.rb
@@ -7,6 +7,7 @@ module RubyIndexer
 
     class UnresolvableAliasError < StandardError; end
     class NonExistingNamespaceError < StandardError; end
+    class IndexNotEmptyError < StandardError; end
 
     # The minimum Jaro-Winkler similarity score for an entry to be considered a match for a given fuzzy search query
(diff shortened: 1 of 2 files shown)


lib/ruby_indexer/lib/ruby_indexer/index.rb
```diff
@@ -360,6 +361,14 @@ def resolve(name, nesting, seen_names = [])
       ).void
     end
     def index_all(indexable_paths: @configuration.indexables, &block)
+      # When troubleshooting an indexing issue, e.g. through irb, it's not obvious that `index_all` will augment the
+      # existing index values, meaning it may contain 'stale' entries. This check ensures that the user is aware of this
+      # behavior and can take appropriate action.
+      # binding.break
+      if @entries.any?
+        raise IndexNotEmptyError, "The index is not empty. Create a new instance before re-indexing"
+      end
```
vinistock, line 370: Isn't this going to fail every single time when using the LSP itself because we always index RBS first?

Issue #2856, opened 2024-11-14
`Document `bypassTypechecker" for contributors


andyw8:

pkondzior:
I stumbled upon this problem yesterday while trying to fix another bug in `ruby-lsp` using the Zed editor.  
At first, I thought something was seriously wrong with the LSP itself, but after quite a bit of debugging, I realized that my Zed configuration doesn’t have the Sorbet LSP set up by default, while `ruby-lsp` still expects it to be present.

Unfortunately, Zed doesn’t allow passing environment variables, so the existing bypass mechanism wasn’t available.  
To address this, I introduced this as `initializationOption` so that `bypassTypechecker` is now available for editors other than VS Code.


cc @egiurleo

PR #2871, opened 2024-11-18
Add Sorbet troubleshooting note

andyw8:
### Motivation

Document this implicit Ruby LSP behavior for the benefit of users trying to set up or troubleshoot their editor.

### Implementation

Add a note to the docs.

### Automated Tests

n/a

### Manual Tests

n/a


vinistock (approved):
We should also check if the user has the Sorbet extension and, if not, enforce that our features will be enabled.

andyw8:
> We should also check if the user has the Sorbet extension and, if not, enforce that our features will be enabled.

Yes: https://github.com/Shopify/ruby-lsp/issues/2856

Issue #2822, opened 2024-11-04
command not found: rbenv


getninjaN:
### Description

Instantly when starting VSCode and extensions are starting up I get an error telling me that the rbenv command can't be found.
It worked on Friday, but today it's not and I'm not sure what could be wrong. ¯\\_(ツ)_/¯

`rbenv` was installed via Homebrew and is up to date.
All my extensions is up to date.

![Image](https://github.com/user-attachments/assets/7a05bc85-f4a7-425f-9fd4-f213a59e527a)

### Ruby LSP Information

The `Collect Ruby LSP Information for Issue Reporting` is not working. Nothing happens.

### Reproduction steps

1. Start VSCode with Ruby LSP installed.
2. Get error.

**Code snippet or error message**
```bash
Automatic Ruby environment activation with rbenv failed:
Command failed: rbenv exec ruby -W0 -rjson -e 'STDERR.print("RUBY_LSP_ACTIVATION_SEPARATOR" + { env: ENV.to_h, yjit: !!defined?(RubyVM::YJIT), version: RUBY_VERSION, gemPath: Gem.path }.to_json + "RUBY_LSP_ACTIVATION_SEPARATOR")'
zsh:1: command not found: rbenv
```

slhck:
I've got the same issue. `rbenv` is loaded in `.zshrc` with:

```
if [ -d "$HOME/.rbenv" ]; then
  export PATH="$HOME/.rbenv/bin:$HOME/.rbenv/plugins/ruby-build/bin:$PATH"
  if which rbenv > /dev/null; then eval "$(rbenv init - zsh)"; fi
fi
```

And it's working in a regular VS Code terminal window (i.e., `rbenv version` returns the right Ruby version for this project).

vinistock:
Thank you for the bug report! We haven't changed anything related to `rbenv` in a while. Did you change anything in your shell configurations recently? I'm asking because we've seen similar issues, but haven't fully understood what causes it.

> And it's working in a regular VS Code terminal window (i.e., rbenv version returns the right Ruby version for this project).

This is a common source of confusion for users, which I've been struggling to make clearer. The zsh configurations working in your terminal is expected. The terminal launches the shell, which sources your configuration scripts and most of it expects to be run in a shell environment.

[...]

PR #2875, opened 2024-11-18
Allow configuring rbenv executable path

aurelien-reeves-scalingo:
### Motivation

Ref https://github.com/Shopify/ruby-lsp/issues/2822#issuecomment-2483246117

Allow configuring the rbenv executable path if it's not installed in the default location and cannot be found automatically.

### Implementation

Added a new setting under version manager. If the value is set to something, we use that as the rbenv path, otherwise we fallback to the default installation path.

Implementation also makes sure that the specified path exists.

This has been inspired by https://github.com/Shopify/ruby-lsp/pull/1914

### Automated Tests

Added a test to verify that we use the configured value to invoke rbenv.

### DIFF
--- a/vscode/src/ruby/rbenv.ts
+++ b/vscode/src/ruby/rbenv.ts
@@ -32,7 +32,10 @@ export class Rbenv extends VersionManager {
       return this.ensureRbenvExistsAt(configuredRbenvPath);
     } else {
       return this.findExec(
-        [vscode.Uri.file("/opt/homebrew/bin"), vscode.Uri.file("/usr/local/bin")],
+        [
+          vscode.Uri.file("/opt/homebrew/bin"),
+          vscode.Uri.file("/usr/local/bin"),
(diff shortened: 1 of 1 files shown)


vscode/src/test/suite/ruby/rbenv.test.ts
```diff
@@ -59,6 +60,67 @@ suite("Rbenv", () => {
     execStub.restore();
   });
 
+  test("Allows configuring where rbenv is installed", async () => {
+    // eslint-disable-next-line no-process-env
```
vinistock, line 64: Is this needed? There's no `process.env` reference below.

vinistock:
You can fix the linting errors with `cd vscode && yarn run format`.

aurelien-reeves-scalingo:
> Thanks for your contribution!

Thanks to you for your guidance! :)

PR #2879, opened 2024-11-19
Show multi-root workspace warning as progress instead

vinistock:
### Motivation

We noticed that the warning for multi-root workspaces is sometimes ignored by users. The problem is that if you don't click anything, the LSP will never start because the promise is never resolved.

Let's show the warning under a timer and auto-dismiss so that the behaviour is less confusing.

### Implementation

I put the same warning inside a progress notification that auto-dismisses under 5 seconds.

### DIFF
--- a/vscode/src/rubyLsp.ts
+++ b/vscode/src/rubyLsp.ts
@@ -175,7 +175,6 @@ export class RubyLsp {
     workspaceFolder: vscode.WorkspaceFolder,
     eager: boolean,
   ) {
-    const workspaceDir = workspaceFolder.uri.fsPath;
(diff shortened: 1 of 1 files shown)


vinistock:
* **#2879** <a href="https://app.graphite.dev/github/pr/Shopify/ruby-lsp/2879?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/Shopify/ruby-lsp/2879?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`

<h2></h2>

PR #2893, opened 2024-11-21
Inhibit restart while git operation in progress

andyw8:
I need to test, but I think this would help reduces errors causes by unwanted restarts, e.g. while pulling updates from the git remote. Worth considering?

Idea is from https://stackoverflow.com/questions/47901986/how-to-tell-if-a-git-operation-is-in-progress

### DIFF
--- a/vscode/src/workspace.ts
+++ b/vscode/src/workspace.ts
@@ -85,7 +85,7 @@ export class Workspace implements WorkspaceInterface {
 
     this.registerCreateDeleteWatcher(
       rootGitUri,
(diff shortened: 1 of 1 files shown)


vinistock:
> We want to prevent Ruby LSP from repeatedly restarting when doing a long operation, such as pulling a branch with many new commits. (Normally we restart on changes to Gemfile.lock).
> But I believe this will prevent Ruby LSP attempting to boot while there's an incomplete Gemfile.lock.

Can you explain your train of thought here? When you say incomplete Gemfile.lock, when does that happen? Even if you pull many commits, aren't the files update in a single operation at the end?

PR #2895, opened 2024-11-21
Declare type in initializer for `RubyIndexer::Enhancement`

andyw8:
This will allow us to remove the shim from addons which use Sorbet, such as ruby-lsp-rails.

### DIFF
--- a/lib/ruby_indexer/lib/ruby_indexer/enhancement.rb
+++ b/lib/ruby_indexer/lib/ruby_indexer/enhancement.rb
@@ -33,7 +33,7 @@ module RubyIndexer
 
     sig { params(listener: DeclarationListener).void }
     def initialize(listener)
-      @listener = listener
+      @listener = T.let(listener, RubyIndexer::DeclarationListener)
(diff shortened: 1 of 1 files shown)


lib/ruby_indexer/lib/ruby_indexer/enhancement.rb
```diff
@@ -33,7 +33,7 @@ def clear
 
     sig { params(listener: DeclarationListener).void }
     def initialize(listener)
-      @listener = listener
+      @listener = T.let(listener, RubyIndexer::DeclarationListener)
```
vinistock, line 36: How will this eliminate the need for the shim? Tapioca doesn't carry instance variable declarations to the RBIs AFAIK.

Also, Sorbet already knows the right type for `@listener` thanks to the method signature.

vinistock: We can try, but I'm not sure if Tapioca picks up instance variable types.

PR #2912, opened 2024-11-26
Fallback to closest Ruby if we can't find an installation for the requested version

### LINKED ISSUE #2793: Improve chruby handling for missing `.ruby-version` and missing Ruby installations
There are some corner cases for our chruby integration that aren't very user friendly at the moment.

## Missing .ruby-version file

There are two ways to define a "global" Ruby version with chruby. The first one is to set the version via a shellscript, which doesn't do anything in the editor since the NodeJS process running the extension is not sourcing those files. It only impacts the terminal.

The second one is to define a `.ruby-version` in a parent directory. Chruby goes up directories trying to find a `.ruby-version` file. I had assumed that this was quite common and the standard way of using chruby, but based on our telemetry this is clearly not the case.

The question then becomes what to do if we can't find a `.ruby-version` in any of the parent directories. We can try to launch with whatever is the latest Ruby installed, but that may conflict with the Ruby constraints defined in the `Gemfile`, which will fail to launch the language server. We could first check if the Gemfile has that constraint, but that may also mean having to parse the constraints ourselves (e.g.: `ruby "~> 3.1"`).

We can also try to better explain that adding a `.ruby-version` in parent directories serves as the fallback, but that will still require manual intervention from the user, even if it's just clicking a button to create the new file.

Whatever we end up deciding here, it's really important to show in the UI what happened, so that it's clear why the LSP is running with a specific Ruby version.

## Missing rubies

The second part of the problem is when there is a `.ruby-version` file, but the version specified is not installed. This is in part related to #2173.

We could also try to launch with an alternative Ruby, but we face similar limitations as the ones described previously with Gemfile constraints.

We also don't know how to install rubies for `chruby` since the tool doesn't come with Ruby installation. People can use whatever they want to install Ruby (Homebrew, ruby-install, ruby-build) and then use `chruby` only for switching versions, which limits what we can offer as part of the experience here.

It might be possible, although a higher maintenance cost, to introduce integration with Ruby installers in addition to version "switchers". That way, for tools like `chruby` that only support switching versions, we can try to discover what the user has available to install rubies and offer to use that.

vinistock:
### Motivation

Closes #2793

This PR starts adds a fallback for when the specified Ruby in `.ruby-version` is not found/not installed.

The UX is similar to the fallback for missing `.ruby-version` configurations. We show a temporary progress warning the user we're going to fallback and give the chance of cancelling, which allows them to manually configure the right version.

### Implementation

The implementation is very similar to the fallback for missing `.ruby-version`. The main distinction is that, when we know the version specified, we can try to fallback to the closest one available.

There are a few decisions I baked into the implementation:

1. We do not try to approximate major versions. For example, if the `.ruby-version` file specifies Ruby 4.0.0, we would not try to fallback to 3.0.0. I think this is a fair limitation and since Ruby is so conservative about major versions, it shouldn't be too painful
2. To approximate, we search for the Ruby installation with the smallest difference in minor version using the highest patch version as the tie breaker (this is done with a sort). For example, if the requested version is `3.3.5` and we have `3.1.0` and `3.2.2` installed, we want to pick `3.2.2` because that's closest to `3.3.5`. And as another example, if both `3.2.0` and `3.2.2` are installed, we would pick `3.2.2` because that's the highest patch

### Automated Tests

Added tests.

### DIFF
--- a/vscode/src/ruby/chruby.ts
+++ b/vscode/src/ruby/chruby.ts
@@ -17,7 +17,7 @@ interface RubyVersion {
   version: string;
 }
 
-class RubyVersionCancellationError extends Error {}
+class RubyActivationCancellationError extends Error {}
 
 // A tool to change the current Ruby version
 // Learn more: https://github.com/postmodern/chruby
@@ -48,12 +48,14 @@ export class Chruby extends VersionManager {
 
   async activate(): Promise<ActivationResult> {
     let versionInfo = await this.discoverRubyVersion();
-    let rubyUri: vscode.Uri;
+    let rubyUri: vscode.Uri | undefined;
 
-    if (versionInfo) {
-      rubyUri = await this.findRubyUri(versionInfo);
-    } else {
-      try {
+    // If the version informed is available, try to find the Ruby installation. Otherwise, try to fall back to an
+    // existing version
+    try {
+      if (versionInfo) {
+        rubyUri = await this.findRubyUri(versionInfo);
+      } else {
         const fallback = await this.fallbackWithCancellation(
           "No .ruby-version file found. Trying to fall back to latest installed Ruby in 10 seconds",
           "You can create a .ruby-version file in a parent directory to configure a fallback",
@@ -63,14 +65,39 @@ export class Chruby extends VersionManager {
 
         versionInfo = fallback.rubyVersion;
         rubyUri = fallback.uri;
-      } catch (error: any) {
-        if (error instanceof RubyVersionCancellationError) {
-          // Try to re-activate if the user has configured a fallback during cancellation
-          return this.activate();
-        }
-
-        throw error;
       }
+    } catch (error: any) {
+      if (error instanceof RubyActivationCancellationError) {
+        // Try to re-activate if the user has configured a fallback during cancellation
+        return this.activate();
+      }
+
+      throw error;
+    }
+
+    // If we couldn't find a Ruby installation, that means there's a `.ruby-version` file, but that Ruby is not
+    // installed. In this case, we fallback to a closest installation of Ruby - preferably only varying in patch
+    try {
+      if (!rubyUri) {
+        const currentVersionInfo = { ...versionInfo };
+
+        const fallback = await this.fallbackWithCancellation(
+          `Couldn't find installation for ${versionInfo.version}. Trying to fall back to other Ruby in 10 seconds`,
+          "You can cancel this fallback and install the required Ruby",
+          async () => this.findClosestRubyInstallation(currentVersionInfo),
+          () => this.missingRubyError(currentVersionInfo.version),
+        );
+
+        versionInfo = fallback.rubyVersion;
+        rubyUri = fallback.uri;
+      }
+    } catch (error: any) {
+      if (error instanceof RubyActivationCancellationError) {
+        // Try to re-activate if the user has configured a fallback during cancellation
+        return this.activate();
+      }
+
+      throw error;
     }
 
     this.outputChannel.info(
@@ -106,7 +133,9 @@ export class Chruby extends VersionManager {
   }
 
   // Returns the full URI to the Ruby executable
-  protected async findRubyUri(rubyVersion: RubyVersion): Promise<vscode.Uri> {
+  protected async findRubyUri(
+    rubyVersion: RubyVersion,
+  ): Promise<vscode.Uri | undefined> {
     const possibleVersionNames = rubyVersion.engine
       ? [`${rubyVersion.engine}-${rubyVersion.version}`, rubyVersion.version]
       : [rubyVersion.version, `ruby-${rubyVersion.version}`];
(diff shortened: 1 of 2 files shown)


vscode/src/ruby/chruby.ts
```diff
@@ -17,7 +17,7 @@ interface RubyVersion {
   version: string;
 }
 
-class RubyVersionCancellationError extends Error {}
+class RubyCancellationError extends Error {}
```
andyw8, line 20:
```suggestion
class RubyInstallationCancellationError extends Error {}
```
(or similar, to make it a little more specific)

vscode/src/ruby/chruby.ts
```diff
@@ -48,12 +48,14 @@ export class Chruby extends VersionManager {
...
   async activate(): Promise<ActivationResult> {
     let versionInfo = await this.discoverRubyVersion();
-    let rubyUri: vscode.Uri;
+    let rubyUri: vscode.Uri | undefined;
 
-    if (versionInfo) {
-      rubyUri = await this.findRubyUri(versionInfo);
-    } else {
-      try {
+    // If the version informed is available, try to find the Ruby installation. Otherwise, try to fallback to an
```
andyw8, line 53:
```suggestion
    // If the version informed is available, try to find the Ruby installation. Otherwise, try to fall back to an
```

andyw8, line 53:
```suggestion
    // If the version specified is available, try to find the Ruby installation. Otherwise, try to fallback to an
```

vscode/src/ruby/chruby.ts
```diff
@@ -63,14 +65,39 @@ export class Chruby extends VersionManager {
...
+      if (error instanceof RubyCancellationError) {
+        // Try to re-activate if the user has configured a fallback during cancellation
+        return this.activate();
+      }
+
+      throw error;
+    }
+
+    // If we couldn't find a Ruby installation, that means there's a `.ruby-version` file, but that Ruby is not
+    // installed. In this case, we fallback to a closest installation of Ruby - preferrably only varying in patch
```
andyw8, line 79:
```suggestion
    // installed. In this case, we fallback to a closest installation of Ruby - preferably only varying in patch
```
(I'll look into why the spellcheck config missed this)

vscode/src/test/suite/ruby/chruby.test.ts
```diff
@@ -318,4 +318,36 @@ suite("Chruby", () => {
...
+    chruby.rubyInstallationUris = [
+      vscode.Uri.file(path.join(rootPath, "opt", "rubies")),
+    ];
+
+    const { env, version, yjit } = await chruby.activate();
+
+    assert.match(env.GEM_PATH!, new RegExp(`ruby/${VERSION_REGEX}`));
+    assert.match(env.GEM_PATH!, new RegExp(`lib/ruby/gems/${VERSION_REGEX}`));
+    assert.strictEqual(version, RUBY_VERSION);
+    assert.notStrictEqual(yjit, undefined);
```
andyw8, line 335: Is `yjit` relevant here?

vscode/src/test/suite/ruby/chruby.test.ts
```diff
@@ -318,4 +318,36 @@ suite("Chruby", () => {
       return chruby.activate();
     });
   });
+
+  test("Uses closest Ruby if the version specified in .ruby-version is not installed (patch difference)", async () => {
+    fs.writeFileSync(path.join(workspacePath, ".ruby-version"), "ruby '3.3.3'");
+
+    const chruby = new Chruby(workspaceFolder, outputChannel, async () => {});
```
andyw8, line 325: We repeat some very similar setup in many tests in this file, perhaps we can extract out some common behaviour?

vscode/src/test/suite/ruby/chruby.test.ts
```diff
@@ -318,4 +318,36 @@ suite("Chruby", () => {
...
+      vscode.Uri.file(path.join(rootPath, "opt", "rubies")),
+    ];
+
+    const { env, version, yjit } = await chruby.activate();
+
+    assert.match(env.GEM_PATH!, new RegExp(`ruby/${VERSION_REGEX}`));
+    assert.match(env.GEM_PATH!, new RegExp(`lib/ruby/gems/${VERSION_REGEX}`));
+    assert.strictEqual(version, RUBY_VERSION);
+    assert.notStrictEqual(yjit, undefined);
+  }).timeout(20000);
```
andyw8, line 352: How about a test to verify what would happen for pre-release Rubies?

vscode/src/test/suite/ruby/chruby.test.ts
vinistock: Yes, we want to check that YJIT was set to something.

vscode/src/test/suite/ruby/chruby.test.ts
vinistock: I extracted the assertions into a function.

vscode/src/test/suite/ruby/chruby.test.ts
```diff
@@ -71,6 +72,15 @@ suite("Chruby", () => {
     fs.rmSync(rootPath, { recursive: true, force: true });
   });
 
+  function assertActivatedRuby(activationResult: ActivationResult) {
```
andyw8, line 75: (nit) I would move this to after the tests.

andyw8 (approved):

vinistock:
### Merge activity

* **Nov 29, 10:39 AM EST**: A user merged this pull request with [Graphite](https://app.graphite.dev/github/pr/Shopify/ruby-lsp/2912).

PR #2915, opened 2024-11-26
Move URI extension to indexer

vinistock:
### Motivation

Move our URI extension to the indexer, so that we can get rid of `IndexablePath` and use URIs only (which is the first step to address #1908).

### Implementation

This PR just moves the extension to the indexer.


vinistock:
* **#2916** <a href="https://app.graphite.dev/github/pr/Shopify/ruby-lsp/2916?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#2915** <a href="https://app.graphite.dev/github/pr/Shopify/ruby-lsp/2915?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/Shopify/ruby-lsp/2915?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`

<h2></h2>

lib/ruby_indexer/ruby_indexer.rb
```diff
@@ -4,6 +4,7 @@
 require "yaml"
 require "did_you_mean"
 
+require "ruby_indexer/lib/ruby_indexer/uri"
```
andyw8, line 7: As we near a separate release of the indexer, I expet we'll need to be careful not to patch Core classes a way which could affect other parts of someone codebase.

andyw8 (approved):

lib/ruby_indexer/ruby_indexer.rb
vinistock: If the Ruby LSP is being added as part of the composed bundle, there's no way it could impact someone's codebase because it's not a part of the main Gemfile.

And even if they do have it in the Gemfile, the default require for `gem "ruby-lsp"` doesn't do anything. You need to manually `require "ruby_lsp/internal"`, exactly for this reason (avoiding polluting the application's namespace).

PR #2916, opened 2024-11-26
Replace IndexablePath with URI

vinistock:
### Motivation

Second step for #1908

This PR gets rid of `IndexablePath` and standardizes on using URI. This is necessary because `IndexablePath` is coupled with file paths on disk, which is incompatible with the URI scheme for unsaved files (`untitled:Untitled-1`) - a scenario where the files don't exist on disk.

### Implementation

The important changes are just in the URI extension file. Everything else is essentially just renaming things from `IndexablePath` to URI.

**Note**: this PR takes a slightly different approach than my previous attempt. I previously tried to create a separate URI class for the indexer, fully disconnected from the original `URI::Generic` class. That turned out to be a terrible idea that lead to an enormous amount of duplication, so the approach here is just to extend `URI::Generic` a bit more to handle require paths.

### Automated Tests

Added a couple of new tests for the URI extension.


vinistock:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.dev/github/pr/Shopify/ruby-lsp/2916?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* **#2916** <a href="https://app.graphite.dev/github/pr/Shopify/ruby-lsp/2916?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/Shopify/ruby-lsp/2916?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#2915** <a href="https://app.graphite.dev/github/pr/Shopify/ruby-lsp/2915?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`

<h2></h2>

exe/ruby-lsp
```diff
@@ -139,9 +139,9 @@ if options[:doctor]
 
   puts "Globbing for indexable files"
 
-  index.configuration.indexables.each do |indexable|
-    puts "indexing: #{indexable.full_path}"
-    index.index_single(indexable)
+  index.configuration.indexables.each do |uri|
```
andyw8, line 142: Not a blocker for this, but I think `indexable_uris` would be clearer. (`indexable` sounds like a module/interface).

vinistock: https://github.com/Shopify/ruby-lsp/pull/2924

Issue #2223, opened 2024-06-21
Connecting to LSP with docker-in-docker setups.


paul:
### Description

We've got a Rails app with some expensive and complicated dependencies (libarrow, among other things), so we run the app in docker-compose. Additionally, for most of our devs using macs, we've started using gitpod to develop the app. They run their own gidpod container on their own infrastructure, and expose the docker socket to that container, and we can docker-compose our app from there. Then vscode can connect to the gitpod docker infrastructure and interact with the files, etc...

I'm trying to figure out how to convince the vscode that connects to the gitpod container to run the ruby-lsp command within our dev container. I can write a wrapper script that would do something like `docker compose run -it dev ruby_lsp`, but I'm not sure how to get the extension to run my wrapper. Also, its gonna try to boot rails, so I'd need to wrap that command, too.

Not a bug report exactly, but maybe a feature request, to allow for a setting in the vscode extension that would let me override the command used to invoke ruby_lsp. (and another to control how it boots the rails app). Alternatively, there's other issues on here to expose a socket or port for vscode to connect to, so that might work if we can start ruby-lsp independently as part of docker-compose, and configure vscode to connect there.

### Reproduction steps

1. Have an app with a complicated set of dependencies, such that it can't `bundle install` or boot on a local machine or in a "normal" container, only within a purpose-built container with the deps needed.
2. Get vscode to run ruby_lsp inside that container.

willtcarey:
We've got a similar setup to what you're describing, except we remote into linux machines from our macOS machines (using the VS Code Remote SSH extensino) and then run our app code in docker.

Ruby LSP appears to have support for everything running inside of a devcontainer, but not this case where you are running VS Code outside of the container which is running your app code/LSP instance.

I managed to make the VS Code extension run my custom wrapper script by running the `Ruby LSP: Select ruby version manager` -> `Change manual ruby configuration` then picking my wrapper script. However, it's a bit odd that the extension only stores that in a `workspaceState` rather than a configuration option. I'd like to share that configuration with my team, but I don't know a way to do that with workspaceState.

https://github.com/Shopify/ruby-lsp/blob/caa4a4038f951f85c05a5a8bb363804065adddc1/vscode/src/ruby.ts#L106-L136

The second thing that's odd here is that I could avoid going the `workspaceState` route at all if I could set my version manager to `none` and set a custom `rubyExecutablePath`. But the custom ruby path is not passed to the `none` version manager on the first pass.

[...]

PR #2919, opened 2024-11-27
Add docker compose as a version manager

d-lebed:
### Motivation

Many developers working on complex projects rely on Docker Compose to set up their development environments. However, it is not always feasible to use VS Code Dev Containers due to their limited functionality and lack of compatibility with custom workflows. To address this limitation, I have implemented a Docker Compose version manager for the Ruby LSP VS Code extension. This change allows developers to use the LSP server seamlessly within their existing Docker Compose setups, including setups enhanced with tools like Mutagen or even "dip". This change aims to improve compatibility and flexibility for developers using diverse development environments.

This change should also solve https://github.com/Shopify/ruby-lsp/issues/2223

### Implementation

I had to make several significant changes:

- **Process Execution:** To retrieve Ruby information, I replaced the use of `exec` with `spawn`, passing a script for execution. This change was made to avoid issues with escaping and to address various problems with container entrypoint implementations.

- **Enhanced Manager Functionality:** Since commands like `bundle install`, Rails generators, and others need to be executed inside Docker Compose, the `manager` object now has expanded functionality and is available within the `Ruby` class.

- **URI Converters:** I implemented `uriConverters` methods (`code2Protocol` and `protocol2Code`) to map paths between the local file system and the container. I also applied the converter logic in relevant parts of the code that I was able to identify.

- **Docker Compose Configuration Parsing:** I added automatic parsing of the Docker Compose configuration. Initially, path mapping was intended to be configured manually, but I found it straightforward to extract this information from the configuration itself. 

   - **Bonus Features:**
      - Added the ability to select a Docker Compose service via the UI and save it in Workspace settings.
      - Included parsing for Mutagen configurations, which is particularly useful for environments running on remote servers.

I did not check the Sorbet extension or others that rely on `process.env`. They will not work, obviously.

### Automated Tests

I updated all the existing tests and added a couple of new ones. If the proposed changes are approved, I will add sufficient new tests to cover the new functionality.

### Manual Tests

I tested the extension locally on macOS using both Docker Compose and Mutagen Compose. Additionally, I verified compatibility with `rvm` and `shadowenv`.

To test the functionality, you can use a similar `docker-compose.yml` configuration:

```yaml
services:
  backend:
    build:
      context: "."
    volumes:
      - .:/app
      - bundle:/usr/local/bundle

volumes:
  bundle:

x-mutagen:
  sync:  
    bundle:
      alpha: "volume://bundle"
      beta: "../bundle/myapp"
      mode: "one-way-replica"
```


d-lebed:
I have signed the CLA!

Splines:
> However, it is not always feasible to use VS Code Dev Containers due to their limited functionality and lack of compatibility with custom workflows.

I second this. We have a fully-functioning Docker setup but don't develop directly in this Docker container and instead mount our codebase into the container (see for example [here](https://github.com/MaMpf-HD/mampf/blob/2bf97423bfc9fe8bd8ac736582663ba67386280d/docker/development/docker-compose.yml#L135)). I do think that this is [a "common" development setup](https://shopify.github.io/ruby-lsp/design-and-roadmap.html#favoring-common-development-setups).

While the VSCode Dev Containers seem very promising, trying to take the first steps to migrate our whole setup to it revealed that this would be a big undertaking, not feasible for our small team at the moment. (And the motivation is not clear if our current setup also works fine). Instead, we have bundler installed locally on the developers machine and Ruby LSP knows nothing about our Docker Container. <sub>This means we cannot use something like the Ruby LSP Rails addon as this would require Ruby LSP to run in the docker container, but that's ok for us as Ruby LSP itself already provides a plethora of nice functionality.</sub>

The only place where this setup does not work is for RSpec tests. That's why we currently use the [`Ruby Test Explorer`](https://marketplace.visualstudio.com/items?itemName=connorshea.vscode-ruby-test-adapter) VSCode extension. We then specify a custom `rspecCommand` and point it to a Python script, which runs the test in our test docker container and reports the results back. While a bit hacky, it works fine at the moment. See [my gist](https://gist.github.com/Splines/c5ec77d70025893870f917f6442853d4) to see how it works.

https://github.com/user-attachments/assets/a0e91b18-bfd9-4e82-9475-12da1148f51f


> If the language server is booted from the container on a port and the extension connects to it, will that do the trick?

This approach does indeed sound promising. I hope that this would not result in large performance penalties when formatting documents.

vinistock:
Container related discussions were happening in a few different places, so I created https://github.com/Shopify/ruby-lsp/discussions/3738 to try to centralize and move the discussion forward.

Issue #2877, opened 2024-11-19
Make the extension's test suite less dependent on individual setups


vinistock:
Currently, the extension's test suite makes several assumptions about what is going to be available in the user's machine. For example, the Shadowenv test assumes that it will be available and the chruby test assumes rubies will be available in certain locations.

This is not friendly to external contributors who may not have their environments configured similarly to our team. Let's ensure that the test suite skips tests that cannot be executed. Let's also try to ensure that we can run as many tests as possible locally.

PR #2928, opened 2024-11-28
Test for presence of shadowenv binary before running Shadowenv suite

adam12:
Not all non-Shopify developer environments have a version of shadowenv installed. In those circumstances, skip the suite with a console warning.

Part of #2877

### DIFF
--- a/vscode/src/test/suite/ruby/shadowenv.test.ts
+++ b/vscode/src/test/suite/ruby/shadowenv.test.ts
@@ -3,6 +3,7 @@ import fs from "fs";
 import assert from "assert";
 import path from "path";
 import os from "os";
+import { execSync } from "child_process";
 
 import { beforeEach, afterEach } from "mocha";
 import * as vscode from "vscode";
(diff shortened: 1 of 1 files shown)


vscode/src/test/suite/ruby/shadowenv.test.ts
```diff
@@ -110,7 +118,9 @@ suite("Shadowenv", () => {
   });
 
   afterEach(() => {
-    fs.rmSync(rootPath, { recursive: true, force: true });
+    if (rootPath) {
+      fs.rmSync(rootPath, { recursive: true, force: true });
+    }
```
adam12, line 123: `rootPath` is sometimes undefined.

Truthfully, I have no idea why :joy: Open to suggestions if anyone has any idea.

vinistock: I _suspect_ that may be related to using async exec. If you switch from `asyncExec` to execSync to check for Shadowenv, do you still see the issue?

adam12: I think you were right on this.

vinistock (approved):
Nice. Thanks for your contribution!

PR #2930, opened 2024-11-28
Enable spellcheck for TypeScript files

andyw8:


project-words
```diff
@@ -1,36 +1,40 @@
 activestorage
 autocorrect
+Autocorrect
```
andyw8, line 3: I regerated the list with:

```
cspell --words-only --unique --no-progress
```
which distinguishes lowercase and capitalized.

vinistock (approved):

PR #2931, opened 2024-11-28
Allow opting into specific feature flags

vinistock:
### Motivation

In the implementation of feature flags, I forgot to check if the user opted into a specific flag. Currently, the code doesn't check for that, which means you can only opt into all of them.

### Implementation

We just need to check if the user opted into that specific flag.

### Automated Tests

Added a test.


KaanOzkan (approved):

vinistock:
### Merge activity

* **Nov 28, 11:22 AM EST**: A user merged this pull request with [Graphite](https://app.graphite.dev/github/pr/Shopify/ruby-lsp/2931).

PR #2932, opened 2024-11-28
Chruby test suite looks for ruby binary in multiple locations

adam12:
Support both `/opt/rubies` and `~/.rubies` for installation paths, checking both the `engine-version` and `version` formats.

Part of #2877


vinistock (approved):
This is definitely the right change as `chruby` users might have rubies installed in all of those possible locations.

We will also need to consider how to make this test pass for non `chruby` users. For example, `rbenv` users will have their rubies installed under `~/.rbenv/rubies` or something like that, so this test still won't pass.

Maybe it can check for all of those possible locations and if none are available, skip?

PR #2934, opened 2024-11-28
Start accepting any URI scheme in the index

vinistock:
### Motivation

Another step towards #1908

After the recent changes, we can now make the index treat entries coming from any URIs appropriately.

### Implementation

The main aspect of change is changing the `@files_to_entries` variable into `@uris_to_entries`. Since entries can now be discovered in URIs not backed by a file on disk, we need to be able to add, remove and handle changes based on URIs.

Additionally, for synchronizing changes in the index, we need to accept the source now, which will be kept in memory for unsaved files.

**Note**: this is a breaking change because it means that the `entries_for` API now _needs_ to accept a URI instead of a file path.

### Automated Tests

Added a test that verifies all of the important functionality for entries discovered in unsaved URIs.

### DIFF
--- a/lib/ruby_indexer/lib/ruby_indexer/entry.rb
+++ b/lib/ruby_indexer/lib/ruby_indexer/entry.rb
@@ -60,19 +60,24 @@ module RubyIndexer
 
     sig { returns(String) }
     def file_name
-      File.basename(file_path)
+      if @uri.scheme == "untitled"
+        T.must(@uri.opaque)
+      else
+        File.basename(T.must(file_path))
+      end
     end
 
-    sig { returns(String) }
+    sig { returns(T.nilable(String)) }
     def file_path
-      T.must(@uri.full_path)
+      @uri.full_path
     end
 
     sig { returns(String) }
     def comments
       @comments ||= begin
         # Parse only the comments based on the file path, which is much faster than parsing the entire file
-        parsed_comments = Prism.parse_file_comments(file_path)
+        path = file_path
+        parsed_comments = path ? Prism.parse_file_comments(path) : []
 
         # Group comments based on whether they belong to a single block of comments
         grouped = parsed_comments.slice_when do |left, right|
--- a/lib/ruby_indexer/lib/ruby_indexer/index.rb
+++ b/lib/ruby_indexer/lib/ruby_indexer/index.rb
@@ -29,10 +29,11 @@ module RubyIndexer
 
       # Holds references to where entries where discovered so that we can easily delete them
       # {
-      #  "/my/project/foo.rb" => [#<Entry::Class>, #<Entry::Class>],
-      #  "/my/project/bar.rb" => [#<Entry::Class>],
+      #  "file:///my/project/foo.rb" => [#<Entry::Class>, #<Entry::Class>],
+      #  "file:///my/project/bar.rb" => [#<Entry::Class>],
+      #  "untitled:Untitled-1" => [#<Entry::Class>],
       # }
-      @files_to_entries = T.let({}, T::Hash[String, T::Array[Entry]])
+      @uris_to_entries = T.let({}, T::Hash[String, T::Array[Entry]])
 
       # Holds all require paths for every indexed item so that we can provide autocomplete for requires
       @require_paths_tree = T.let(PrefixTree[URI::Generic].new, PrefixTree[URI::Generic])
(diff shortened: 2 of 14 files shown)


vinistock:
* **#2934** <a href="https://app.graphite.dev/github/pr/Shopify/ruby-lsp/2934?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/Shopify/ruby-lsp/2934?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`

<h2></h2>

lib/ruby_indexer/lib/ruby_indexer/index.rb
```diff
@@ -380,38 +379,36 @@ def index_all(uris: @configuration.indexable_uris, &block)
...
-      return unless full_path
+      content = case uri.scheme
+      when "file"
+        full_path = uri.full_path
+        source || ((full_path = uri.full_path) && File.read(full_path))
+      when "untitled"
+        source
+      end
+
+      unless content
```
andyw8, line 390: Could this move to before the `when`?

lib/ruby_indexer/lib/ruby_indexer/index.rb
```diff
@@ -380,38 +379,36 @@ def index_all(uris: @configuration.indexable_uris, &block)
 
     sig { params(uri: URI::Generic, source: T.nilable(String), collect_comments: T::Boolean).void }
     def index_single(uri, source = nil, collect_comments: true)
-      full_path = uri.full_path
-      return unless full_path
+      content = case uri.scheme
+      when "file"
+        full_path = uri.full_path
+        source || ((full_path = uri.full_path) && File.read(full_path))
```
andyw8, line 385: Why are we re-assigning `full_path`?

lib/ruby_indexer/lib/ruby_indexer/index.rb
```diff
@@ -380,38 +379,36 @@ def index_all(uris: @configuration.indexable_uris, &block)
...
+      when "untitled"
+        source
+      end
+
+      unless content
+        $stderr.puts("No content specified for #{uri}. This is likely a bug")
```
andyw8, line 391:
```suggestion
        $stderr.puts("No content specified for #{uri}. This is likely a bug, please report it.")
```

lib/ruby_indexer/lib/ruby_indexer/index.rb
vinistock: I refactored this method to require the source to be passed in, so that we don't have to handle this here.

lib/ruby_indexer/lib/ruby_indexer/entry.rb
```diff
@@ -60,19 +60,24 @@ def private?
...
-      T.must(@uri.full_path)
+      @uri.full_path
     end
 
     sig { returns(String) }
     def comments
       @comments ||= begin
         # Parse only the comments based on the file path, which is much faster than parsing the entire file
-        parsed_comments = Prism.parse_file_comments(file_path)
+        path = file_path
```
andyw8, line 79: (nit) do we need this local?

lib/ruby_indexer/lib/ruby_indexer/index.rb
```diff
@@ -374,49 +373,40 @@ def index_all(uris: @configuration.indexable_uris, &block)
...
       if e.backtrace&.first&.include?("prism")
-        $stderr.puts "Prism error indexing #{T.must(full_path)}: #{e.message}"
+        $stderr.puts "Prism error indexing #{uri}: #{e.message}"
       else
         raise
       end
     end
 
+    # Indexes a File URI by reading the contents from disk
+    sig { params(uri: URI::Generic, collect_comments: T::Boolean).void }
```
andyw8, line 402: (nit) could this be a `URI::File` so we don't accidentally pass a non-file?

andyw8 (approved):

lib/ruby_indexer/lib/ruby_indexer/entry.rb
vinistock: Yes, because `file_path` returns a nilable string and we need to narrow the type inside this scope to pass it to `Prism.parse_file_comments`.

PR #2941, opened 2024-11-29
Index documents on modification

vinistock:


lib/ruby_indexer/lib/ruby_indexer/index.rb
```diff
@@ -48,6 +48,8 @@ def initialize
       )
 
       @configuration = T.let(RubyIndexer::Configuration.new, Configuration)
+
+      @initial_indexing_complete = T.let(false, T::Boolean)
```
st0012, line 52: nit: I think it should be `@initial_indexing_completed`

lib/ruby_indexer/lib/ruby_indexer/index.rb
```diff
@@ -358,11 +361,13 @@ def index_all(uris: @configuration.indexable_uris, &block)
       # existing index values, meaning it may contain 'stale' entries. This check ensures that the user is aware of this
       # behavior and can take appropriate action.
       # binding.break
```
st0012, line 363: Is this old debugging code?

lib/ruby_indexer/lib/ruby_indexer/index.rb
```diff
@@ -56,8 +58,8 @@ def register_included_hook(module_name, &hook)
       (@included_hooks[module_name] ||= []) << hook
     end
 
-    sig { params(uri: URI::Generic).void }
-    def delete(uri)
+    sig { params(uri: URI::Generic, skip_require_tree: T::Boolean).void }
+    def delete(uri, skip_require_tree: false)
```
st0012, line 62: nit: I prefer naming it `skip_require_paths_tree` so when searching for `require_paths_tree`, this flag jumps out too.

lib/ruby_indexer/test/index_test.rb
```diff
@@ -2061,7 +2061,8 @@ def test_build_non_redundant_name
     end
 
     def test_prevents_multiple_calls_to_index_all
-      # For this test class, `index_all` is already called once in `setup`.
+      @index.index_all
```
st0012, line 2064: Why is this change needed?

vinistock:
@andyw8 good catch. I updated the code to handle ancestor invalidation too.

andyw8:
Let's also update the docs to remove the part that begins "There is currently a technical limitation that code is only indexed when saved."

PR #2942, opened 2024-11-29
Only reindex documents if declarations are being modified

vinistock:
### Motivation

This PR tries to limit the amount of times we pay the price of reindexing the current document being modified by analyzing the context behind the last edit made to it.

### Implementation

The idea of this approach is to remember the position of the last edit made to a document and the nature of the edit (insert, replace or delete).

Then, depending on the type of edit and what nodes exist under the position where the edit was made, we decide if reindexing is relevant or not.

To determine that, we look at the node under the cursor and, if it matches a node we use for indexing in the `DeclarationListener`, then we consider that we need to reindex.

It's not perfect, so let me know if you have other ideas on how to determine this in a way that is:

1. Fast
2. Accurate

Necessarily in this order because the whole point is to reduce the performance impact of having to reindex. If detecting whether we need to do it or not is equally as slow, then there's no point in doing it.

### Automated Tests

Added some tests, but we should probably add more.

### DIFF
--- a/lib/ruby_indexer/lib/ruby_indexer/declaration_listener.rb
+++ b/lib/ruby_indexer/lib/ruby_indexer/declaration_listener.rb
@@ -64,7 +64,6 @@ module RubyIndexer
         :on_constant_path_or_write_node_enter,
         :on_constant_path_operator_write_node_enter,
         :on_constant_path_and_write_node_enter,
-        :on_constant_or_write_node_enter,
         :on_constant_write_node_enter,
         :on_constant_or_write_node_enter,
         :on_constant_and_write_node_enter,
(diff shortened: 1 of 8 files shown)


lib/ruby_indexer/lib/ruby_indexer/declaration_listener.rb
```diff
@@ -64,7 +64,6 @@ def initialize(index, dispatcher, parse_result, uri, collect_comments: false)
         :on_constant_path_or_write_node_enter,
         :on_constant_path_operator_write_node_enter,
         :on_constant_path_and_write_node_enter,
-        :on_constant_or_write_node_enter,
```
vinistock, line 67: This event was duplicated.

PR #2943, opened 2024-11-30
VSCode extension checks multiple install locations for `mise` binary

### LINKED ISSUE #2878: Update autodetection of Mise in VSCode extension to support multiple locations
Homebrew installs Mise in a different location than the location on the website, which is the one that the Ruby LSP extension attempts to detect. Homebrew is a common installation method for Mise, so the Ruby LSP should detect the path via this installation method if possible.

The path Homebrew uses on Apple M hardware is: `/opt/homebrew/bin/mise`

adam12:
### Motivation

The installation path of Mise differs through the installation method chosen. Using Homebrew results in a path under `/opt/homebrew` where as the installation method from the Mise website is typically under `$HOME/.local`.

### Implementation

When a custom mise path is not defined, 2 potential installation locations are tried instead.

### Automated Tests

Skipping for now per discussion in PR.

### Manual Tests

Extension manually started and observed Mise path of `/opt/homebrew` detected in logs.

```
2024-12-12 16:29:05.413 [info] (manager) Discovered version manager mise
2024-12-12 16:29:05.414 [info] (manager) Running command: `/opt/homebrew/bin/mise x -- ruby -W0 -rjson -e 'STDERR.print("RUBY_LSP_ACTIVATION_SEPARATOR" + { env: ENV.to_h, yjit: !!defined?(RubyVM::YJIT), version: RUBY_VERSION, gemPath: Gem.path }.to_json + "RUBY_LSP_ACTIVATION_SEPARATOR")'` in /Users/adam/code/github.com/midstack/manager using shell: /opt/homebrew/bin/fish
```

Closes #2878

### DIFF
--- a/vscode/src/ruby/mise.ts
+++ b/vscode/src/ruby/mise.ts
@@ -30,24 +30,52 @@ export class Mise extends VersionManager {
     const misePath = config.get<string | undefined>(
       "rubyVersionManager.miseExecutablePath",
     );
-    const miseUri = misePath
-      ? vscode.Uri.file(misePath)
-      : vscode.Uri.joinPath(
-          vscode.Uri.file(os.homedir()),
-          ".local",
-          "bin",
-          "mise",
-        );
 
-    try {
-      await vscode.workspace.fs.stat(miseUri);
-      return miseUri;
-    } catch (error: any) {
-      // Couldn't find it
+    if (misePath) {
+      const configuredPath = vscode.Uri.file(misePath);
+
+      try {
+        await vscode.workspace.fs.stat(configuredPath);
+        return configuredPath;
+      } catch (error: any) {
+        throw new Error(
+          `Mise executable configured as ${configuredPath}, but that file doesn't exist`,
+        );
+      }
+    }
+
+    // Possible mise installation paths
+    //
+    // 1. Installation from curl | sh (per mise.jdx.dev Getting Started)
+    // 2. Homebrew M series
+    const possiblePaths = [
+      vscode.Uri.joinPath(
+        vscode.Uri.file(os.homedir()),
+        ".local",
+        "bin",
+        "mise",
+      ),
+      vscode.Uri.joinPath(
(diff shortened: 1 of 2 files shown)


vinistock:
Yeah, this is currently mocked in other tests. I think it should be fine to leave this one without tests since it's the same approach of checking directories we're already using in other manager integrations.

adam12:
Poking around another file, I noticed that there is another instance where mise is checked in a specific path. Likely should be updated as well.

https://github.com/Shopify/ruby-lsp/blob/896617a0c5f7a22ebe12912a481bf1b59db14c12/vscode/src/ruby.ts#L411-L421

vscode/src/ruby/mise.ts
```diff
@@ -30,24 +30,52 @@ export class Mise extends VersionManager {
...
+        return possiblePath;
+      } catch (error: any) {
+        // Continue looking
+      }
     }
 
     throw new Error(
-      `The Ruby LSP version manager is configured to be Mise, but ${miseUri.fsPath} does not exist`,
+      `The Ruby LSP version manager is configured to be Mise, but could not find Mise installation. Searched in ` +
+        `${possiblePaths.join(", ")}`,
```
adam12, line 78: This is kind of silly but I couldn't appease the linter any other way.

```
  77:7  error  Unexpected string concatenation  prefer-template
```

vinistock:
> Poking around another file, I noticed that there is another instance where mise is checked in a specific path. Likely should be updated as well.

Yeah, we need to refactor that manager detection logic at some point. Each version manager integration should have some function that will return true if it's detect to be installed and then we don't need to duplicate the logic. For now, feel free to duplicate it.

vscode/src/ruby/mise.ts
vinistock: You can also just break the line

```typescript
throw new Error(
  `blah blah
   ${possiblePaths.join(", ")}`
```

PR #2952, opened 2024-12-04
Enhancement - Move node range caluclation out of loop in completion listener

rogancodes:
### Motivation

In the Completion listener, we calculate the node range within the each loop for both constant completion candidates and instance variable completion candidates.

### Implementation

Moved the range calculation outside the loop, as the node is already accessible outside the loop.

### Automated Tests

Not applicable

### Manual Tests

Not applicable


st0012 (approved):
Thank you!

PR #2956, opened 2024-12-04
Update links for reporting issues

andyw8:
The link https://github.com/Shopify/ruby-lsp/issues/new?labels=bug&template=bug_template.yml wasn't working because we split the template as VS Code / non-VS Code.

I think it's sufficient to link to the [choose](https://github.com/Shopify/ruby-lsp/issues/new/choose) page where users can select the appropriate issue type.


st0012 (approved):
Should it be a footnote link in the `index` and `troubleshooting` pages?

Issue #2837, opened 2024-11-06
Block parameters prefixed with underscore are incorrectly highlighted


vinistock:
Our [grammar file](https://github.com/Shopify/ruby-lsp/blob/main/vscode/grammars/ruby.cson.json) is not properly handling block parameters that are prefixed with underscore. For example:

```ruby
# `a` looks ok
# `b` doesn't
[].each do |a, _b|
end
```

We might be missing `_` as a valid character in the regex.

pepetorres1998:
Can I work on this issue?

pepetorres1998:
I'm sorry I did not send a message before, life happened and couldn't work on this 🙏

andyw8:
Completed in https://github.com/Shopify/ruby-lsp/pull/2972

PR #2972, opened 2024-12-11
Github issue: https://github.com/Shopify/ruby-lsp/issues/2837

willyamacaroni:
### Motivation

https://github.com/Shopify/ruby-lsp/issues/2837

What:
Removes the distinction between block variables prefixed with underscore and those not prefixed with underscore.

Why:
As per the github issue: the leading underscore is just a convention. From Ruby's perspective, it's still a parameter and even referring to it still works.

### Implementation

The regex was making 3 capture groups, one of which was specifically for variables prefixed with an underescore. I removed that group and modified the other variable group to include the underscore prefix. 

### Automated Tests

I saw there was `grammars.test.ts` but I'm running into a lot of errors trying to run the tests. It's a work in progress

### Manual Tests

With changes
with changes
ges:

Without changes:
without changes

### DIFF
--- a/vscode/grammars/ruby.cson.json
+++ b/vscode/grammars/ruby.cson.json
@@ -2260,16 +2260,13 @@
           "end": "(?=,|\\|\\s*)",
           "patterns": [
             {
-              "match": "\\G([&*]?)([a-zA-Z][\\w_]*)|(_[\\w_]*)",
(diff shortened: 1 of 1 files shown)


willyamacaroni:
image

andyw8:
(I changed the PR title so the changelog for the next release is easier to read)

PR #2973, opened 2024-12-12
Add glyph for 'Run' code lenses

andyw8:
Inspired by https://github.com/rust-lang/rust-analyzer

Screenshot 2024-12-12 at 10 44 08 AM


andyw8:
(if this is accepted then we should also update `ruby-lsp-rails` to match).

Issue #2446, opened 2024-08-15
100% CPU usage


tsvallender:
### Description

I am fairly regularly (probably about once a day at a full-time job) having ruby-lsp hit 100% CPU and stay there until I restart the process. I’m using the latest ruby-lsp with the Rails extension, in Neovim. I’ve seen this behaviour in multiple Ruby versions but am mostly in 3.3.4 currently.

I’m more than happy to do some work trying to figure out what’s causing this but am unsure where to start, any pointers appreciated.

andyw8:
(and are you using ruby-lsp-rails, or only ruby-lsp?)

tsvallender:
> @tsvallender if you are able to the LSP logs from just before it causes the CPU to peak, then that may help us to diagnose the problem.
> 
> Also, are you are on the latest release? (v0.17.13)

There hasn't been anything extra in the logs, and yes latest release. A colleague suggested stracing the process next time it happens so will see if that suggests anything. Always the way it doesn't happen when you're waiting for it...

Earlopain:
I've had a hang somewhat recently (like maybe 1 month ago) as well but `strace` didn't show a thing. Process just kept doing things even after vscode was gone. Seems like this happens relativly often for you. How about you try adding the following to the ruby-lsp script:

```rb
Signal.trap("SIGUSR1") do
  output = +""
  Thread.list.each do |thr|
    output << ("-" * 40) << "\n"
    output << thr.inspect << "\n"
    output << thr.backtrace.join("\n") << "\n"
  end
  File.write("ruby-lsp-trace.log", output)
end if Signal.list.include?("USR1")
```

You can then do `kill -USR1 pid` where `pid` is somewhere in the output from `ps -aux | grep ruby-lsp`. I'm not sure how useful the output would actually be, its just something I used elsewhere. This should create a file with backtraces in the root of the project as long as you are not on windows.

This is all a bit awkward with the server auto-updating and all that, I wonder if this would be a generally useful addition for debugging things like this. There have been a few other issues with CPU pegging as their topic.

Earlopain:
Certainly looks like something that could happen there, thought the only way I can think of is when edits are received out-of-order.

I coincidentally had another case today myself, though I failed to capture a stacktrace (oops). I'm going to add some logging to that method should we be running into the same thing.

vinistock:
[...]

PR #2976, opened 2024-12-12
Locate targets under mutex lock

vinistock:
### Motivation

I'm hoping this is the solution for #2446

My hypothesis is that we're seeing a concurrency issue where threads switch in the middle of locating targets and then the document gets mutated. That would render the location we're currently searching for invalid, which can then lead to infinite loops.

### Implementation

My idea is to move our mutex into the global state so that we can use it in more places. Then we lock the mutex while locating targets to avoid having any document edits be applied in the middle.

**Note**: I experimented with passing the mutex down to the document instances, but it was a bit messier.

### DIFF
--- a/lib/ruby_lsp/requests/prepare_rename.rb
+++ b/lib/ruby_lsp/requests/prepare_rename.rb
@@ -24,7 +24,7 @@ module RubyLsp
 
       sig { override.returns(T.nilable(Interface::Range)) }
       def perform
-        char_position = @document.create_scanner.find_char_position(@position)
+        char_position, _ = @document.find_index_by_position(@position)
 
         node_context = RubyDocument.locate(
           @document.parse_result.value,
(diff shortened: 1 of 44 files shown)


lib/ruby_lsp/requests/prepare_rename.rb
```diff
@@ -12,19 +12,23 @@ class PrepareRename < Request
...
+        @global_state = global_state
         @document = document
         @position = T.let(position, T::Hash[Symbol, Integer])
       end
 
       sig { override.returns(T.nilable(Interface::Range)) }
       def perform
-        char_position = @document.create_scanner.find_char_position(@position)
+        char_position = @global_state.synchronize do
+          @document.create_scanner.find_char_position(@position)
```
st0012, line 30: Should we find a way to make the mutex mandatory when calling scanner? Like:

```rb
@document.create_scanner(@global_state.mutex).find_char_position(@position)
```

And the scanner will do the mutax wrapping around `find_char_position`. This will making the requirement clear but also simplify the usage.

st0012 (approved):
LGTM 👍

PR #2981, opened 2024-12-13
feat: Handle instance variable target nodes in document symbol

### LINKED ISSUE #2896: Handle instance variable target nodes in document symbol
Similar to #2834, we need to handle instance variable target nodes in document symbols too.

```ruby
class Foo
  def initialize
    @a, @b = [1, 2] # << should produce two symbols, one for `a` and one for `b`
  end
end
```

Hungle2911:
### Motivation

Resolve #2896 
 Currently, when using destructuring assignment for instance variables only a single document symbol is created. This change ensures each instance variable gets its own document symbol. 

### Implementation

I've created new listener method to fetch instance_variable_target_node

### Automated Tests

Added new test case `test_instance_variable_parallel_assignment` that verifies:
* Basic destructuring assignment of instance variables
* Correct symbol creation and naming for each instance variable
* Proper ordering of symbols

### DIFF
--- a/lib/ruby_lsp/listeners/document_symbol.rb
+++ b/lib/ruby_lsp/listeners/document_symbol.rb
@@ -41,6 +41,7 @@ module RubyLsp
           :on_module_node_enter,
           :on_module_node_leave,
           :on_instance_variable_write_node_enter,
+          :on_instance_variable_target_node_enter,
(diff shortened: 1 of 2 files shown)


Hungle2911:
I have signed the CLA!

vinistock (approved):
Thanks for your contribution!

PR #2982, opened 2024-12-13
Move RuboCop file watching to the server

### LINKED ISSUE #1457: Migrate `.rubocop.yml` watching to the server

vinistock:
### Motivation

Closes #1457

This was originally attempted in #1921 and #2510. This PR moves watching `.rubocop.yml` to the server, so that we can avoid a full server restart and instead just create a new instance of the RuboCop runner.

This PR still suffers from the same issue reported in both previous attempts: trying to clear existing diagnostics after changing `.rubocop.yml` simply does not work. I believe this must be some sort of bug in the client or some assumption we're not satisfying. Clearing diagnostics works normally when closing files on `textDocument/didClose` notifications, so I see no reason why it wouldn't work while processing a did change watched files notification.

Despite the bug, since we're trying to improve stability of the LSP, I think it's still worth moving forward. Launching the server is still the most critical part in our entire life cycle and reducing the amount of times we restart helps prevent unnecessary pain for developers. Additionally, simply editing the file with the stale diagnostics already makes them go away immediately, so the bug is not that terrible.

I created an issue to discuss the bug with the maintainers of the protocol https://github.com/microsoft/vscode-languageserver-node/issues/1594.

### Implementation

Stopped restarting the server on `.rubocop.yml` and `.rubocop` modifications. Instead, we watch the files in the server and simply re-create the instance of the runner.

Note that this has the added benefit of working on any editor.

### DIFF
--- a/lib/ruby_lsp/server.rb
+++ b/lib/ruby_lsp/server.rb
@@ -298,29 +298,11 @@ module RubyLsp
 
       # Not every client supports dynamic registration or file watching
       if @global_state.client_capabilities.supports_watching_files
-        send_message(
-          Request.new(
-            id: @current_request_id,
-            method: "client/registerCapability",
-            params: Interface::RegistrationParams.new(
-              registrations: [
-                # Register watching Ruby files
-                Interface::Registration.new(
-                  id: "workspace/didChangeWatchedFiles",
-                  method: "workspace/didChangeWatchedFiles",
-                  register_options: Interface::DidChangeWatchedFilesRegistrationOptions.new(
-                    watchers: [
-                      Interface::FileSystemWatcher.new(
-                        glob_pattern: "**/*.rb",
-                        kind: Constant::WatchKind::CREATE | Constant::WatchKind::CHANGE | Constant::WatchKind::DELETE,
-                      ),
-                    ],
-                  ),
-                ),
-              ],
-            ),
-          ),
-        )
+        send_message(Request.register_watched_files(@current_request_id, "**/*.rb"))
+        send_message(Request.register_watched_files(
+          @current_request_id,
+          Interface::RelativePattern.new(base_uri: @global_state.workspace_uri.to_s, pattern: ".rubocop.yml"),
+        ))
       end
 
       process_indexing_configuration(options.dig(:initializationOptions, :indexing))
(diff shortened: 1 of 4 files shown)


lib/ruby_lsp/server.rb
```diff
@@ -1011,26 +988,55 @@ def workspace_did_change_watched_files(message)
...
+      return unless defined?(Requests::Support::RuboCopFormatter)
+
+      send_log_message("Reloading RuboCop since #{uri} changed")
+      @global_state.register_formatter("rubocop", Requests::Support::RuboCopFormatter.new)
+
+      # Clear all existing diagnostics since the config changed. This has to happen under a mutex because the `state`
+      # hash cannot be mutated during iteration or that will throw an error
+      @mutex.synchronize do
+        @store.each do |uri, _document|
+          send_message(Notification.publish_diagnostics(uri, []))
```
andyw8, line 1035: `Notification.clear_diagnostics(uri)` would be nice to have.

andyw8 (approved):

Issue #2966, opened 2024-12-09
Allow navigation between related code, e.g. from test to implementation


andyw8:
### I have checked that this feature is not already implemented

- [x] This feature does not exist

### Use case

See below

### Description

It's convenient for developers to jump between a test and its corresponding implementation.

(in Vim, Projectionist is a popular plugin for this)

### Implementation

A planned VS Code feature may help with this: https://github.com/microsoft/vscode/issues/126932

But it would good to have it available for all editors.

vinistock:
How is this one different than #2665? Can we close one of the two and keep a single issue?

andyw8:
Ah, I didn't find that when looking.

Let's keep this one as it describes it in a more general way (e.g. for Rails, this the same mechanism that could power jumping from a controller action to its view).

andyw8:
> We should report this as a possible use case in https://github.com/microsoft/vscode/issues/126932 and suggest making go to related code more generic. I might be wrong, but based on the issue's title it looks like they are planning it more as a testing thing rather than a general feature.

https://github.com/microsoft/vscode/issues/126932#issuecomment-2532116867

PR #2985, opened 2024-12-13
Add GotoTest and GotoSource functionality

jenny-codes:
### Motivation

A stab at https://github.com/Shopify/ruby-lsp/issues/2966

### Implementation

Implementing it as a [custom request](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#requestMessage) for now until vscode starts supporting it officially.

The path finding algorithm is a rephrase of https://github.com/jasonrudolph/significant-other-vscode/blob/main/matchmaker.js

https://github.com/shopify/dev-test-runner seems to have a more robust implementation though. I'm also going to explore some other options too before deciding on the final implementation.

### Automated Tests

Not yet! 

### Manual Tests

https://github.com/user-attachments/assets/a0f23d6c-59ad-4f34-b76c-c61d0af2b567


lib/ruby_lsp/requests/goto_relevant_file.rb
```diff
@@ -0,0 +1,74 @@
...
+
+        find_most_similar_with_jacaard(relpath, candidate_paths).map { File.join(workspace_path, _1) }
+      end
+
+      sig { params(path: String).returns(String) }
+      def relevant_filename_pattern(path)
+        input_basename = File.basename(path, File.extname(path))
+
+        relevant_basename_pattern =
+          if input_basename.end_with?("test") || input_basename.end_with?("spec")
```
st0012, line 44: If we're adding the file's extension name to `input_basename`, under what circumstances will it end with just `test` or `spec`? Can we have test cases for them?

st0012:
Sorry for the delay in reviewing. Can you rebase this against the `main` branch?

lib/ruby_lsp/requests/goto_relevant_file.rb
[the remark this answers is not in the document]
jenny-codes: Good catch. Thanks!

test/requests/goto_relevant_file_test.rb
```diff
@@ -0,0 +1,158 @@
+# typed: true
+# frozen_string_literal: true
+
+require "test_helper"
+
+class GotoRelevantFileTest < Minitest::Test
+  def setup
+    Dir.stubs(:pwd).returns("/workspace")
```
andyw8, line 8: I'm a little wary of stubbing `pwd` since if some test helper, or minitest itself, uses is then it may break.

jenny-codes: Can you expand on that?

lib/ruby_lsp/requests/goto_relevant_file.rb
```diff
@@ -0,0 +1,83 @@
...
+        find_most_similar_with_jacaard(relative_path, candidate_paths).map { File.join(workspace_path, _1) }
+      end
+
+      sig { params(path: String).returns(String) }
+      def relevant_filename_pattern(path)
+        input_basename = File.basename(path, File.extname(path))
+
+        test_prefix_pattern = /^(#{TEST_KEYWORDS.join("_|")}_)/
+        test_suffix_pattern = /(_#{TEST_KEYWORDS.join("|_")})$/
+        test_pattern = /#{test_prefix_pattern}|#{test_suffix_pattern}/
```
st0012, line 47: Maybe we can store these patterns in constants too? Do we need to recompute them for each request?

lib/ruby_lsp/requests/goto_relevant_file.rb
```diff
@@ -0,0 +1,83 @@
...
+
+        relevant_basename_pattern =
+          if input_basename.match?(test_pattern)
+            input_basename.gsub(test_pattern, "")
+          else
+            test_prefix_glob = "#{TEST_KEYWORDS.join("_,")}_"
+            test_suffix_glob = "_#{TEST_KEYWORDS.join(",_")}"
```
st0012, line 54: Same as the above.

lib/ruby_lsp/requests/goto_relevant_file.rb
```diff
@@ -0,0 +1,83 @@
...
+
+            "{{#{test_prefix_glob}}#{input_basename},#{input_basename}{#{test_suffix_glob}}}"
+          end
+
+        "#{relevant_basename_pattern}#{File.extname(path)}"
+      end
+
+      sig { params(path: String, candidates: T::Array[String]).returns(T::Array[String]) }
+      def find_most_similar_with_jacaard(path, candidates)
```
st0012, line 63: I'd appreciate a comment explaining what `jacaard` means 😂

lib/ruby_lsp/requests/goto_relevant_file.rb
```diff
@@ -0,0 +1,83 @@
...
+      sig { override.returns(T::Array[String]) }
+      def perform
+        find_relevant_paths
+      end
+
+      private
+
+      sig { returns(T::Array[String]) }
+      def find_relevant_paths
+        workspace_path = Dir.pwd
```
st0012, line 32: Let's take in `GlobalState` and use its `workspace_path` method to get this value. I think `Requests::Rename` is a good reference.

lib/ruby_lsp/requests/goto_relevant_file.rb
```diff
@@ -0,0 +1,83 @@
...
+    class GotoRelevantFile < Request
+      extend T::Sig
+
+      TEST_KEYWORDS = ["test", "spec", "integration_test"]
+
+      sig { params(path: String).void }
+      def initialize(path)
+        super()
+
+        @workspace_path = T.let(Dir.pwd, String)
```
st0012, line 19: Let's use `GlobalState#workspace_path` instead.

lib/ruby_lsp/requests/goto_relevant_file.rb
```diff
@@ -0,0 +1,83 @@
...
+      def perform
+        find_relevant_paths
+      end
+
+      private
+
+      sig { returns(T::Array[String]) }
+      def find_relevant_paths
+        workspace_path = Dir.pwd
+        relative_path = @path.delete_prefix(workspace_path)
```
st0012, line 33: Isn't `@path` already the value without workspace path prefix?

PR #2991, opened 2024-12-17
Add test helper methods from Rails add-on

andyw8:
These methods are currently in ruby-lsp-rails but we also need them for testing the Tapioca LSP. Once this is released, we can remove them from the other places.

I've verified it [here](https://github.com/Shopify/ruby-lsp-rails/actions/runs/12395982615/job/34602952817). You can see there are some failures but those are releated to changes in the indexing API which we will need to deal with separately.

### DIFF
--- a/lib/ruby_lsp/test_helper.rb
+++ b/lib/ruby_lsp/test_helper.rb
@@ -1,4 +1,4 @@
-# typed: strict
+# typed: true
 # frozen_string_literal: true
 
 # NOTE: This module is intended to be used by addons for writing their own tests, so keep that in mind if changing.
@@ -65,8 +65,6 @@ module RubyLsp
       end
     end
 
-    # TODO: write correct sig
-    sig { params(message_queue: T.untyped, type: T.untyped).returns(T.untyped) }
     def pop_log_notification(message_queue, type)
       log = message_queue.pop
       return log if log.params.type == type
@@ -75,8 +73,6 @@ module RubyLsp
       log
     end
 
-    # TODO: write correct sig
-    sig { params(outgoing_queue: T.untyped, block: T.untyped).returns(T.untyped) }
     def pop_message(outgoing_queue, &block)
       message = outgoing_queue.pop
       return message if block.call(message)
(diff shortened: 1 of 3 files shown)


lib/ruby_lsp/test_helper.rb
```diff
@@ -52,5 +52,37 @@ def with_server(source = nil, uri = Kernel.URI("file:///fake.rb"), stub_no_typec
...
       end
     end
+
+    sig { params(server: RubyLsp::Server).returns(RubyLsp::Result) }
+    def pop_result(server)
+      result = server.pop_response
+      result = server.pop_response until result.is_a?(RubyLsp::Result) || result.is_a?(RubyLsp::Error)
+
+      if result.is_a?(RubyLsp::Error)
+        raise "Failed to execute request #{result.message}"
```
andyw8, line 62: In `ruby-lsp-rails` this uses `refute_instance_of`, I've changed it to to avoid coupling to the test framework.

st0012: I think a common practice is to define a custom exception class under `TestHelper` and raise that exception here. This will allow consumers of this helper to assert/refute this case more easily.

sorbet/rbi/shims/test_helper.rbi
```diff
@@ -0,0 +1,5 @@
+# typed: true
+
+module RubyLsp::TestHelper
+  include Kernel # for `raise`
```
KaanOzkan, line 4: You can do `requires_ancestor { Kernel }` instead in the ruby file.

lib/ruby_lsp/test_helper.rb
andyw8: Yeah, I didn't do so since this isn't implementation code, but I can add.

st0012 (approved):
I marked this as an enhancement as other addon authors will hopefully use it too.

sorbet/rbi/shims/test_helper.rbi
KaanOzkan: Yes it's defined in https://github.com/sorbet/sorbet/blob/363435333ad9080b614e0c6473aa800b717c2d28/gems/sorbet-runtime/lib/types/helpers.rb#L64. Since ruby-lsp depends on `sorbet-runtime`, it'll work in add-ons. The feature flag is for the static checker only.

Issue #2992, opened 2024-12-19
Rubocop errors are not displayed in the VSCode


cc-harashima:
#### VS Code Version
1.96.0

#### Ruby LSP Extension Version
0.8.16

#### Ruby Version Manager
rbenv

#### gem Versions
rubocop 1.68.0
rubocop-rails 2.20.2

### Reproduction steps

1. Start the Ruby LSP using a VSCode
2. Open or save a Specific Ruby file
3. An error notification within VSCode occurs as shown in the image below. Rubocop errors and warnings in that file are not displayed.

![Image](https://github.com/user-attachments/assets/dd8950e8-d3a0-4058-befe-d1715d3297a9)

**Code snippet or error message**
```
Error running diagnostics: An internal error occurred for the Rails/UniqueValidationWithoutIndex cop. Updating to a newer version of RuboCop may solve this. For more details, run RuboCop on the command line.
```

#### Supplement

The same issue occurred with Rubocop 1.41.1 when using Ruby-LSP version 0.8.13. At that time, I upgraded Rubocop to version 1.68.0.

Constantly upgrading the Rubocop version increases the operational burden, which is challenging.

cc-harashima:
@vinistock 

Thank you for the information.
I have confirmed that even when not in the middle of typing, RuboCop runs successfully on the command line, but an error dialog appears in the editor.

The `Rails/UniqueValidationWithoutIndex` cop should not be executed when `db/schema.rb` does not exist.
https://docs.rubocop.org/rubocop-rails/cops_rails.html#railsuniquevalidationwithoutindex

It is strange that this error is showing up on the LSP side, as `db/schema.rb` is not present in the current Rails project.

PR #2993, opened 2024-12-19
Improve RuboCop internal error message

andyw8:
In https://github.com/Shopify/ruby-lsp/issues/2992, the error is likely from `rubocop-rails` rather than RuboCop itself.

### DIFF
--- a/lib/ruby_lsp/requests/support/rubocop_runner.rb
+++ b/lib/ruby_lsp/requests/support/rubocop_runner.rb
@@ -36,8 +36,8 @@ module RubyLsp
 
         MESSAGE = <<~EOS
           An internal error occurred %s.
-          Updating to a newer version of RuboCop may solve this.
-          For more details, run RuboCop on the command line.
+          Updating to a newer version of RuboCop (and its extensions) may solve this.
+          For more details, run `bundle exec rubocop` on the command line.
         EOS
 
         sig { params(rubocop_error: T.any(RuboCop::ErrorWithAnalyzedFileLocation, StandardError)).void }


lib/ruby_lsp/requests/support/rubocop_runner.rb
```diff
@@ -36,8 +36,8 @@ class InternalRuboCopError < StandardError
 
         MESSAGE = <<~EOS
           An internal error occurred %s.
-          Updating to a newer version of RuboCop may solve this.
-          For more details, run RuboCop on the command line.
+          Updating to a newer version of RuboCop (and its extensions) may solve this.
+          For more details, run `bundle exec rubocop` on the command line.
```
andyw8, line 40: (recommend `bundle exec` to ensure gem versions match what Ruby LSP uses).

st0012:
I think this is a good improvement, but we should probably print the exception to `OUTPUT` to make it easier to debug too?

This should work:

```diff
diff --git a/lib/ruby_lsp/server.rb b/lib/ruby_lsp/server.rb
index f975e065..34c15fcc 100644
--- a/lib/ruby_lsp/server.rb
+++ b/lib/ruby_lsp/server.rb
@@ -651,6 +651,8 @@ module RubyLsp
       ))
       send_empty_response(message[:id])
     rescue StandardError, LoadError => error
+      send_log_message(error.full_message)
+
       send_message(Notification.window_show_message(
         "Formatting error: #{error.message}",
         type: Constant::MessageType::ERROR,
```

PR #3003, opened 2025-01-02
Upgrade to Ruby 3.4.1

vinistock:
### Motivation

Starting the new year always means the same thing: new rubies! Upgrade to Ruby 3.4.1.

There were some tiny differences that I had to fix:

1. The output of hashes changed from the old rocket syntax to the new. We can't simply update the expectations because then tests will fail on <3.4
2. The locations of some syntax error diagnostics have changed. I switched to a hand written test to make it easy to account for different behaviours, but I kept the fixture so that non-raise expectations continue to run
3. The last one I'm unsure if it's a bug or an intentional change. Based on our document symbol test and implementation, it seems that in <3.4 `empty?` returns `true` if the contents of the string are just a line break. This doesn't seem to be the case anymore
4. Many default gems now became bundled gems. We now need to add `csv` and `base64` to the Jekyll bundle since the gem doesn't declare an explicit dependency on them, but they have been removed


vinistock:
* **#3003** <a href="https://app.graphite.dev/github/pr/Shopify/ruby-lsp/3003?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/Shopify/ruby-lsp/3003?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

KaanOzkan (approved):

PR #3018, opened 2025-01-07
Handle module_function with no arguments

### LINKED ISSUE #2653: Support `module_function` in the indexer

vinistock:
### Motivation

Closes #2653

This PR handles the other missing part of `module_function`, which is when it gets invoked with no argument. After looking into the Ruby source code, it seems that `module_func` is a flag considered as part of the visibility scope.

Invoking `module_function` will both:

1. Start marking new methods as singleton methods
2. Push `private` into the stack

### Implementation

I created a new `VisibilityScope` object to help us encapsulate all aspects of visibility, so that we don't forget to handle `module_func` where necessary.

Then I started handling the case of `module_function` with no arguments, which essentially pushes a new scope into the stack with `module_func: true` and visibility private.

### Automated Tests

Added a few tests.

### DIFF
--- a/lib/ruby_indexer/lib/ruby_indexer/declaration_listener.rb
+++ b/lib/ruby_indexer/lib/ruby_indexer/declaration_listener.rb
@@ -24,7 +24,7 @@ module RubyIndexer
       @index = index
       @uri = uri
       @enhancements = T.let(Enhancement.all(self), T::Array[Enhancement])
-      @visibility_stack = T.let([Entry::Visibility::PUBLIC], T::Array[Entry::Visibility])
+      @visibility_stack = T.let([VisibilityScope.public_scope], T::Array[VisibilityScope])
       @comments_by_line = T.let(
         parse_result.comments.to_h do |c|
           [c.location.start_line, c]
@@ -138,7 +138,7 @@ module RubyIndexer
 
     sig { params(node: Prism::SingletonClassNode).void }
     def on_singleton_class_node_enter(node)
-      @visibility_stack.push(Entry::Visibility::PUBLIC)
+      @visibility_stack.push(VisibilityScope.public_scope)
 
       current_owner = @owner_stack.last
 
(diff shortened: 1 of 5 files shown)


vinistock:
* **#3018** <a href="https://app.graphite.dev/github/pr/Shopify/ruby-lsp/3018?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/Shopify/ruby-lsp/3018?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

lib/ruby_indexer/lib/ruby_indexer/declaration_listener.rb
```diff
@@ -24,7 +24,7 @@ def initialize(index, dispatcher, parse_result, uri, collect_comments: false)
       @index = index
       @uri = uri
       @enhancements = T.let(Enhancement.all(self), T::Array[Enhancement])
-      @visibility_stack = T.let([Entry::Visibility::PUBLIC], T::Array[Entry::Visibility])
+      @visibility_stack = T.let([VisibilityScope.new], T::Array[VisibilityScope])
```
andyw8, line 27: Although it defaults to public, I think it would be better to be expclicit, e.g. `VisibilityScope.public` could a be factory method.

lib/ruby_indexer/lib/ruby_indexer/declaration_listener.rb
```diff
@@ -904,11 +925,20 @@ def handle_module_operation(node, operation)
...
+
       arguments_node = node.arguments
-      return unless arguments_node
 
-      owner_name = @owner_stack.last&.name
-      return unless owner_name
+      # If `module_function` is invoked without arguments, all methods defined after it become singleton methods and the
+      # visibility for instance methods changes to private
+      unless arguments_node
+        @visibility_stack.push(VisibilityScope.new(visibility: Entry::Visibility::PRIVATE, module_func: true))
```
andyw8, line 937: Related to my previous comment, this could be `VisibilityScope.module_func`.

We could also make `new` private to prevent invalid combinations arguments.

andyw8 (approved):

lib/ruby_indexer/lib/ruby_indexer/declaration_listener.rb
vinistock: Created another factory method.

PR #3031, opened 2025-01-09
WIP: Use index to detect test library

### LINKED ISSUE #1334: Use ancestors to determine test framework in code lens
~Depends on #1333~ (complete)

We currently try to determine the test framework based on dependencies. This is an okay approach, but once we have ancestors available we can be a lot more accurate.

We can check if a test inherits from `Minitest::Test` or `Test::Unit::TestCase` and then decide based on that - ideally with a fallback if we can't determine the framework for whatever reason.

andyw8:
Closes #1334

Co-authored with @st0012

### DIFF
--- a/lib/ruby_lsp/requests/code_lens.rb
+++ b/lib/ruby_lsp/requests/code_lens.rb
@@ -26,16 +26,17 @@ module RubyLsp
         params(
           global_state: GlobalState,
           uri: URI::Generic,
+          test_library: String,
           dispatcher: Prism::Dispatcher,
         ).void
       end
-      def initialize(global_state, uri, dispatcher)
+      def initialize(global_state, uri, test_library, dispatcher)
         @response_builder = T.let(
           ResponseBuilders::CollectionResponseBuilder[Interface::CodeLens].new,
           ResponseBuilders::CollectionResponseBuilder[Interface::CodeLens],
         )
         super()
-        Listeners::CodeLens.new(@response_builder, global_state, uri, dispatcher)
+        Listeners::CodeLens.new(@response_builder, global_state, test_library, uri, dispatcher)
 
         Addon.addons.each do |addon|
           addon.create_code_lens_listener(@response_builder, uri, dispatcher)
--- a/lib/ruby_lsp/ruby_document.rb
+++ b/lib/ruby_lsp/ruby_document.rb
@@ -240,5 +240,25 @@ module RubyLsp
         node_types: node_types,
       )
     end
+
+    sig { returns(String) }
+    def test_library
+      entries = @global_state.index.entries_for(@uri.to_s)
+      return "none" unless entries
+
+      # TODO: consider performance hit
+      ancestors = entries.map { @global_state.index.linearized_ancestors_of(_1.name) }.flatten
+
+      # ActiveSupport::TestCase is a subclass of Minitest::Test so we must check for it first
+      if ancestors.include?("ActiveSupport::TestCase")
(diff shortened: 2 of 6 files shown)


lib/ruby_lsp/requests/code_lens.rb
```diff
@@ -24,18 +24,18 @@ def provider
...
         @response_builder = T.let(
           ResponseBuilders::CollectionResponseBuilder[Interface::CodeLens].new,
           ResponseBuilders::CollectionResponseBuilder[Interface::CodeLens],
         )
         super()
-        Listeners::CodeLens.new(@response_builder, global_state, uri, dispatcher)
+        Listeners::CodeLens.new(@response_builder, test_library, uri, dispatcher)
 
         Addon.addons.each do |addon|
           addon.create_code_lens_listener(@response_builder, uri, dispatcher)
```
andyw8, line 41: @st0012 @vinistock I'm wondering how we should update this since currently ruby-lsp-rails checks `@global_state.test_library == "rails"`. What if we make `test_library` be an optional keyword argument? That would prevent breaking the API for other addons.

st0012: This should be considered a breaking change anyway so I think changing args is expected?

lib/ruby_lsp/server.rb
```diff
@@ -491,7 +491,7 @@ def run_combined_requests(message)
       document.cache_set("textDocument/foldingRange", folding_range.perform)
       document.cache_set("textDocument/documentSymbol", document_symbol.perform)
       document.cache_set("textDocument/documentLink", document_link.perform)
-      document.cache_set("textDocument/codeLens", code_lens.perform)
+      document.cache_set("textDocument/codeLens", code_lens.perform) if code_lens
```
andyw8, line 494: `if` is temporary

lib/ruby_lsp/ruby_document.rb
```diff
@@ -257,6 +257,37 @@ def locate_node(position, node_types: [])
       )
     end
 
+    sig { returns(String) }
+    def test_library
```
vinistock, line 261: Part of the advantage of looking at the ancestors is that we become more accurate even in very uncommon setups, like mixing test frameworks - which is something that came up a few times.

Considering that test files can have multiple test classes and they can inherit from different parents, is this approach flexible enough to account for that?

What was the idea behind putting it in document versus doing it in the code lens implementation?

lib/ruby_lsp/ruby_document.rb
```diff
@@ -257,6 +257,37 @@ def locate_node(position, node_types: [])
...
+
+      # ActiveSupport::TestCase is a subclass of Minitest::Test so we must check for it first
+      if ancestors.include?("ActiveSupport::TestCase")
+        "rails"
+      elsif ancestors.include?("Minitest::Test")
+        "minitest"
+      elsif ancestors.include?("Test::Unit::TestCase")
+        "test-unit"
+      else
+        "unknown"
```
vinistock, line 281: We should become a bit more specific now that we're checking for ancestors. For example, as [Alex mentioned here](https://github.com/Shopify/ruby-lsp/issues/1334#issuecomment-2173793731), the test framework isn't really "rails" and is not tied to `ActiveSupport::TestCase`.

We can check for `ActiveSupport::Testing::Declarative` and cover a broader range of cases.

Same thing for Minitest and Test Unit. We can now do a better job at detecting the spec style syntax.

PR #3032, opened 2025-01-09
Fix visibility stack handling for private_class_method

vinistock:
### Motivation

Found this bug while testing out things in Core. The handling of the visibility stack for `private_class_method` was incorrect. It was popping the stack once for each argument. With 3 or more arguments, that would fully exhaust the stack and cause `NoMethodError` for invocations on a `nil`.

### Implementation

I reworked the implementation because it was a bit difficult to follow. We now:

1. Only push to the visibility stack if there's a method definition as an argument. Popping happens when we leave the call node
2. Building a single list of names and process them all in one go, without nested loops

### Automated Tests

Added a test that fails on main.

### DIFF
--- a/lib/ruby_indexer/lib/ruby_indexer/declaration_listener.rb
+++ b/lib/ruby_indexer/lib/ruby_indexer/declaration_listener.rb
@@ -287,7 +287,6 @@ module RubyIndexer
       when :module_function
         handle_module_function(node)
       when :private_class_method
-        @visibility_stack.push(VisibilityScope.new(visibility: Entry::Visibility::PRIVATE))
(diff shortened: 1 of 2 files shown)


vinistock:
* **#3032** <a href="https://app.graphite.dev/github/pr/Shopify/ruby-lsp/3032?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/Shopify/ruby-lsp/3032?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

andyw8 (approved):

PR #3033, opened 2025-01-09
Avoid reporting the same Bundler install error twice

vinistock:
### Motivation

If bundle install fails and one of the gems ends up not being present, then `Bundler.setup` is guaranteed to fail with `Bundler::GemNotFound`, but that's expected. The root cause of the problem is the fact that installation failed.

Let's not set `setup_error` in that particular case to avoid reporting the same issue twice to telemetry.

### Implementation

Started not setting `setup_error` if there was an `install_error` and the error during setup matches the expected class.

### DIFF
--- a/exe/ruby-lsp-launcher
+++ b/exe/ruby-lsp-launcher
@@ -45,6 +45,12 @@ rescue Errno::ECHILD
   # In theory, the child process can finish before we even get to the wait call, but that is not an error
 end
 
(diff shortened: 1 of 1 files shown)


vinistock:
### Merge activity

* **Jan 9, 1:50 PM EST**: A user merged this pull request with [Graphite](https://app.graphite.dev/github/pr/Shopify/ruby-lsp/3033).

PR #3041, opened 2025-01-10
Request diagnostic refresh when RuboCop configs change

vinistock:
### Motivation

The explanation in https://github.com/microsoft/vscode-languageserver-node/issues/1594#issuecomment-2582033204 solved our issue clearing diagnostics. Push and pull diagnostics are two different collections, so you cannot clear diagnostics acquired via pull using a push with an empty array.

The correct solution is to request a diagnostic refresh to the editor, which will then trigger pull diagnostics for all open documents.

### Implementation

Changed the implementation to clear our cache and then request a diagnostic refresh, which now correctly recomputes the diagnostics using the new RuboCop configs.

### Automated Tests

Added a test to verify that we're request the refresh.

### Manual Tests

https://github.com/user-attachments/assets/2a29ce82-4b23-47cf-8af0-8267980feec4

### DIFF
--- a/lib/ruby_lsp/client_capabilities.rb
+++ b/lib/ruby_lsp/client_capabilities.rb
@@ -11,7 +11,8 @@ module RubyLsp
     attr_reader :supports_watching_files,
       :supports_request_delegation,
       :window_show_message_supports_extra_properties,
-      :supports_progress
+      :supports_progress,
+      :supports_diagnostic_refresh
 
     sig { void }
     def initialize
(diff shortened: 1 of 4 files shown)


lib/ruby_lsp/client_capabilities.rb
```diff
@@ -32,6 +33,9 @@ def initialize
 
       # The editor supports displaying progress requests
       @supports_progress = T.let(false, T::Boolean)
+
+      # The editor supports server initiated refresh for diagnostics
+      @diagnostic_refresh_support = T.let(false, T::Boolean)
```
andyw8, line 38:
```suggestion
      @supports_diagnostic_refresh = T.let(false, T::Boolean)
```
to better match `@supports_progress`?

andyw8 (approved):

vinistock:
### Merge activity

* **Jan 16, 10:52 AM EST**: A user merged this pull request with [Graphite](https://app.graphite.dev/github/pr/Shopify/ruby-lsp/3041).

PR #3058, opened 2025-01-13
Replace interpolated string pattern with regexp

st0012:
### Motivation

To suppress this warning: `Unexpected template string expression`

### Implementation

Use regexp instead.

### DIFF
--- a/vscode/src/rubyLsp.ts
+++ b/vscode/src/rubyLsp.ts
@@ -626,7 +626,7 @@ export class RubyLsp {
                 delete newConfig.useBundler;
 
                 const command = (newConfig.command || "").replace(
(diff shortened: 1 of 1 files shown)


vscode/src/rubyLsp.ts
```diff
@@ -626,7 +626,7 @@ export class RubyLsp {
                 delete newConfig.useBundler;
 
                 const command = (newConfig.command || "").replace(
-                  "${workspaceRoot}/",
+                  /\$\{workspaceRoot\}\//,
```
paracycle, line 629: A simpler change might be:
```suggestion
                  `\${workspaceRoot\}`,
```

PR #3061, opened 2025-01-13
add highlight for do...end blocks

### LINKED ISSUE #2929: Highlight block in ERB file
When highlighting the control block structure within an ERB file, if there is a block defining word used as a method argument keyword inside the block, it will match the `end` statement with that keyword instead of the control block.

With a `class:` argument in a method call (with or without parentheses):
Image

Without the `class:` keyword, it works fine:
Image

Calling `class=` as an HTML attribute is fine:
Image

Within a normal Ruby .rb file, this does not occur:
Image

vicocamacho:
### Motivation

Closes #2929

As I was investigating the problem with the issue above I discovered that the highlight functionality for highlight do blocks was not implemented.

### Implementation

This PR enhances the existing Highlight feature to add support for highlighting blocks.

### Automated Tests

Added new expectations using the existing "do block" fixture files.

### Manual Tests

Go to the start of a `do` block, it should highlight its matching `end`

### DIFF
--- a/lib/ruby_lsp/listeners/document_highlight.rb
+++ b/lib/ruby_lsp/listeners/document_highlight.rb
@@ -119,7 +119,8 @@ module RubyLsp
             Prism::RequiredParameterNode, Prism::RestParameterNode
             [target, node_value(target)]
           when Prism::ModuleNode, Prism::ClassNode, Prism::SingletonClassNode, Prism::DefNode, Prism::CaseNode,
-            Prism::WhileNode, Prism::UntilNode, Prism::ForNode, Prism::IfNode, Prism::UnlessNode
+            Prism::WhileNode, Prism::UntilNode, Prism::ForNode, Prism::IfNode, Prism::UnlessNode, Prism::BlockNode,
+            Prism::LambdaNode
             target
           end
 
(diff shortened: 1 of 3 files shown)


vicocamacho:
I have signed the CLA!

jbigler:
Great!

Wow, this reveals a different problem I was having then. When I tested your code the correct matching term was highlighted along with the wrong one. So I was actually using a legacy regex based highlighter without realizing it. :roll_eyes:

vinistock:
@vicocamacho I'm confused about this one. Is this PR related to #2929? Or does it just implement highlighting the matching `do` and `end` tokens for blocks?

The description has the `closes` keyword, but my understanding is that the issue was related to tree sitter.

Let me know if you want to bring this PR back.

vicocamacho:
hey @vinistock,

While I was investigating the error reported in #2929, I discovered that `do ... end` were not being highlighted. I implemented  support for them with the intention of fixing the reported issue. 

It turns out the issue was actually caused by Regex-based highlighting in Neovim.

So while technically this PR does not solve that bug report, it does add a "missing" feature. 

I'm happy to bring back the PR if you think it is worth having.

vinistock:
Yes! Let's do it. We should highlight both matching `do..end`, but also matching `if|unless|while|until|for|def|class|module` and their respective `end` tokens.

Issue #3050, opened 2025-01-11
Go-to definition for singleton methods defined via `class << self`


jesseduffield:
### I have checked that this feature is not already implemented

- [x] This feature does not exist

### Use case

I would like to be able to go-to definition for singleton methods defined via the `class << self` syntax.

### Description

Specifically, If I have:
```ruby
module Foo
  class << self
    def bar
      # do something
    end
  end
end

...
# in another file
Foo.bar
```
I want to be able to click on the `bar` in `Foo.bar` and go to the definition in the Foo module.

Currently you can go to definition for singleton methods defined with the `self.foo` syntax but not with the `class << self` syntax.

jesseduffield:
Hmm, I've opened the LSP's codebase in vscode and opened the dev container to confirm.

A strange issue I see is that if I hit the keybinding for going to definition when at the end of a word, it doesn't work, but if I'm within the word, then it does work:

https://github.com/user-attachments/assets/981f997d-0415-4b51-aa63-e3d50efdfdf8

I've now confirmed in my own codebase that go-to definition works the same, however I've noticed an additional issue: if I have a module that's defined like so:
```ruby
module Foo::Bar
  class << self
    def baz; end
  end
end
```
Then if I try to jump to definition for `Foo::Bar.baz`, it won't work. On the other hand if the module is structured like so:
```ruby
module Foo
  module Bar
    class << self
      def baz; end
    end
  end
end
```

It will.

I believe the former syntax is now the preferred approach in rails applications, and I use that approach in my codebase.

vinistock:
Thank you for the extra details, that's helpful. Not finding the method inside compact namespaces is an indexing bug, which will be fixed by #3062.

Concerning going to definition in the end of the word, I'll need to do a bit more research to understand what other language servers are doing. When your cursor is at the end of the word, it's position is actually one character _after_ the word, which is why we don't find any targets.

But checking the behaviour on TypeScript, you can indeed jump to definition even if the position is off by one. Maybe the expected implementation is that the server will check if there's any content under the position and if not adjust the position by minus one? I'll need to understand this better.

PR #3062, opened 2025-01-14
Fix singleton indexing under compact namespaces

vinistock:
### Motivation

Fixes one of the bugs reported in #3050

We were assigning incorrect names to singleton classes defined inside compact namespaces. The issue is that we were simply taking the last entry in the stack, but when using compact namespaces that entry may include multiple names.

### Implementation

Started ensuring we use only the last part of the name to define the singleton.

I also noticed that we were pushing and popping from the name stack when entering a singleton method defined with `def self.foo`. That's actually unnecessary. The name stack is only used for constants and inside a method definition we'll only find instance variables (which use the owner stack rather than the name stack).

I removed the pushing and popping in that case.

### Automated Tests

Added a test.

### DIFF
--- a/lib/ruby_indexer/lib/ruby_indexer/declaration_listener.rb
+++ b/lib/ruby_indexer/lib/ruby_indexer/declaration_listener.rb
@@ -143,7 +143,7 @@ module RubyIndexer
 
       if current_owner
         expression = node.expression
-        name = (expression.is_a?(Prism::SelfNode) ? "<Class:#{@stack.last}>" : "<Class:#{expression.slice}>")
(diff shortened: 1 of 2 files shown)


vinistock:
* **#3062** <a href="https://app.graphite.dev/github/pr/Shopify/ruby-lsp/3062?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/Shopify/ruby-lsp/3062?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

jesseduffield:
Thanks for making this @vinistock! When might I expect to see this in a release?

PR #3065, opened 2025-01-15
Use local executables when working on the Ruby LSP

vinistock:
### Motivation

I was trying to make some more improvements to our launcher and I realized we were missing something we should have probably done about 2 years ago.

When working on the LSP itself, we don't use the local version of the launcher. We use the local version of the server, but only after the bundle has been composed. Composing the bundle uses whatever latest version is globally installed.

This is detrimental to us catching bugs in the launch process before releasing it to users. We need to change that.

### Implementation

To ensure that the local version of the executable is always chosen when working on the LSP, I started mutating the `PATH` to include the `exe` directory.

That way, whenever `ruby-lsp` is invoked, the first thing that will be found is the local executable.

### Automated Tests

Added a test to make sure the `PATH` is being mutated.

### DIFF
--- a/vscode/src/ruby.ts
+++ b/vscode/src/ruby.ts
@@ -186,6 +186,20 @@ export class Ruby implements RubyInterface {
     if (!this.error) {
       this.fetchRubyVersionInfo();
       await this.setupBundlePath();
+
+      // When working on the Ruby LSP itself, we want to use the local version of our executables rather than the ones
+      // globally installed. That allows us to catch mistakes made in the launch process before they are released
+      if (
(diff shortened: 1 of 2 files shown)


vinistock:
* **#3065** <a href="https://app.graphite.dev/github/pr/Shopify/ruby-lsp/3065?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/Shopify/ruby-lsp/3065?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

andyw8 (approved):

Issue #1458, opened 2024-03-05
Migrate lockfile watching to the server


vinistock:
We currently watch for changes in `Gemfile.lock` or `gems.locked` from the [extension](https://github.com/Shopify/ruby-lsp/blob/c27ef7f053e74cf87753e7b7646d2650268c147f/vscode/src/workspace.ts#L231). It would be nice if we could move that to the server, so that other editors can benefit from it.

I don't know if this would actually work, but I think we can implement it as follows. If the lockfiles are modified, we need to:

1. Shutdown the existing server without returning anything to the client, so that it's transparent. This means closing pipes, terminating threads, clearing the store, etc.
2. Then we need to run the custom bundle logic, to ensure that we reflect the changes from the lockfile
3. Finally, we re-boot the server using `exec`. This may be very tricky. Since the server is restarting in a way that's transparent to the client, we wouldn't receive an `initialize` request or `initialized` notification, which means the new server would be missing indexing information, encoding, formatter, etc. Maybe there's a way that we can pass the information from the old server to the new one?

vinistock:
I don't think this is feasible and we alleviated some of the related problems by composing the bundle asynchronously on a lockfile change and only restarting if the result is valid.

PR #3066, opened 2025-01-15
Check if bundle is valid before restarting

vinistock:
### Motivation

The more we restart the server, the higher the risk that something related to Bundler may fail and prevent users from having a smooth experience.

I think we can be a bit smarter about when we restart the server by checking if composing the bundle is successful before we try. If we know that composing fails, it's better to continue running the current version of the server than crashing.

### Implementation

The proposal is to add a custom request that allows the client to ask the server if composing the bundle succeeds. That way, if there's a lockfile update, we can first check if the update produces a valid bundle.

If we couldn't compose the bundle, then we avoid restarting. Otherwise, we trigger the restart, but skip composing the bundle since we just did that ahead of time - saving time for the user and allow them to skip directly to indexing.

### Note

This solution doesn't fully solve #1458, but it does move the needle a little bit at least for VS Code. The big challenge in moving lockfile watching to the server is how to transfer the state of the server currently running to the new instance we launch.

For example, when we receive a notification that the lockfile was changed, we would need to:

1. Compose the new bundle
2. Replace the current process with a new one, using the new bundle (via `exec`)
3. Transfer the state of the previous process to the new one. We need all configurations that were applied during initialize (think the global state), the state of all current documents (the store object) and other pieces of state that exist in the version of the server running before the relaunch

I have not yet figured out how to achieve 3.

### Automated Tests

Added tests.

### DIFF
--- a/exe/ruby-lsp-launcher
+++ b/exe/ruby-lsp-launcher
@@ -8,14 +8,24 @@
 
 setup_error = nil
 install_error = nil
+reboot = false
 
-# Read the initialize request before even starting the server. We need to do this to figure out the workspace URI.
-# Editors are not required to spawn the language server process on the same directory as the workspace URI, so we need
-# to ensure that we're setting up the bundle in the right place
-$stdin.binmode
-headers = $stdin.gets("\r\n\r\n")
-content_length = headers[/Content-Length: (\d+)/i, 1].to_i
-raw_initialize = $stdin.read(content_length)
+workspace_uri = ARGV.first
+
+raw_initialize = if workspace_uri && !workspace_uri.start_with?("--")
+  # If there's an argument without `--`, then it's the server asking to compose the bundle and passing to this
+  # executable the workspace URI. We can't require gems at this point, so we built a fake initialize request manually
+  reboot = true
+  "{\"params\":{\"workspaceFolders\":[{\"uri\":\"#{workspace_uri}\"}]}}"
+else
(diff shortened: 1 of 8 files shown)


vinistock:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.dev/github/pr/Shopify/ruby-lsp/3066?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* **#3066** <a href="https://app.graphite.dev/github/pr/Shopify/ruby-lsp/3066?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/Shopify/ruby-lsp/3066?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#3065** <a href="https://app.graphite.dev/github/pr/Shopify/ruby-lsp/3065?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

andyw8 (approved):

vinistock:
### Merge activity

* **Jan 23, 9:45 AM EST**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/Shopify/ruby-lsp/3066).

Issue #3046, opened 2025-01-10
RuboCop v1.70 incompatibilities and new RuboCop add-on


vinistock:
RuboCop v1.70 added an [add-on to provide linting and formatting via the Ruby LSP](https://github.com/rubocop/rubocop/pull/13628). This is a great move long term, as it allows RuboCop maintainers to improve the integrations and ensure compatibility across versions.

Unfortunately, this first iteration of the add-on is conflicting with our internal integrations with RuboCop and may cause some issues. Namely, two things may happen:

1. Namespace conflicts like #3044
2. Linter registration conflicts since it's using the same identifier as the one we use internally. This means that the LSP may end up using the add-on version and some features can potentially be missing or other features may break (like watching .rubocop.yml)

### Transition plan

We need a transition plan since people will still rely on our internal integrations while they are upgrading RuboCop. It may take a while until the community catches up to the newest versions.

Steps

- [ ] Decide identifier change. If we change ours from `rubocop` to `rubocop_internal`, it will avoid the conflict, but it also means that anyone who has configured to use `rubocop` will try to use the add-on. If the application is on an older version of RuboCop where the add-on is not available, that will fail and no formatting will be available. We can add logic to check for this scenario and try to do the right thing depending on what's available
- [ ] Our automatic detection for linters will need to change as well. We will now need to take the add-on into account if it's available and then decide whether to use the internal or the add-on version
- [ ] Map all feature differences between our internal integrations and the RuboCop add-on, so that we can ensure feature completeness for the community

vinistock:
#3045 fixed the namespace conflicts.

Earlopain:
Do you still think this should be reverted for now? I haven't tried it out myself yet, so no idea what the impact is.

Most of the PR is good, just removing the addon itself to give more time shouldn't be too big of a ask I think.

vinistock:
No longer necessary after #3045. It seems like things are working normally, although we still need to iron out the remaining details. I updated my comment on the RuboCop PR.

vinistock:
The releases were cut today https://github.com/Shopify/ruby-lsp/releases/tag/v0.23.6 and https://github.com/Shopify/ruby-lsp/releases/tag/vscode-ruby-lsp-v0.8.18.

mtrepanier:
Yes it just appeared in the VsCode marketplace.  But even by choosing `rubocop_internal` instead of `auto` or `rubocop` I have same error

[...]

PR #3067, opened 2025-01-15
Change our internal RuboCop integration identifier

vinistock:
### Motivation

This PR addresses point 1 and 2 from #3046

Now that the RuboCop add-on is using the `rubocop` identifier, we need to change ours to prevent conflicts, while still ensuring that users who are on older versions of RuboCop are getting the right behaviour.

### Implementation

Essentially, we are changing our identifier to `rubocop_internal`.

The only extra logic is falling back when users have explicitly configured their formatter or linter as `rubocop`. Since that is now owned by RuboCop, but only present on versions higher than v1.70, we need to check the version and ensure that it's actually available or fallback to our internal integration.

### Automated Tests

Added tests.

### DIFF
--- a/lib/ruby_lsp/global_state.rb
+++ b/lib/ruby_lsp/global_state.rb
@@ -94,12 +94,19 @@ module RubyLsp
       @workspace_uri = URI(workspace_uri) if workspace_uri
 
       specified_formatter = options.dig(:initializationOptions, :formatter)
+      rubocop_has_addon = Gem::Requirement.new(">= 1.70.0").satisfied_by?(Gem::Version.new(::RuboCop::Version::STRING))
 
       if specified_formatter
         @formatter = specified_formatter
 
         if specified_formatter != "auto"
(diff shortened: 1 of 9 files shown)


lib/ruby_lsp/global_state.rb
```diff
@@ -109,6 +116,23 @@ def apply_options(options)
       end
 
       specified_linters = options.dig(:initializationOptions, :linters)
+
+      if specified_formatter == "rubocop" || specified_linters&.include?("rubocop")
+        notifications << Notification.window_log_message(<<~MESSAGE, type: Constant::MessageType::WARNING)
+          Formatter is configured to be `rubocop`. As of RuboCop v1.70.0, this identifier activates the add-on
+          implemented in the rubocop gem itself instead of the internal integrations provided by the Ruby LSP.
```
andyw8, line 123:
```suggestion
          implemented in the rubocop gem itself instead of the internal integration provided by the Ruby LSP.
```

PR #3068, opened 2025-01-15
Update `GUESSED_TYPES_URL`

csjh:
Switches `GUESSED_TYPES_URL` from dead https://shopify.github.io/ruby-lsp/design-and-roadmap.html#guessed-types to not dead https://shopify.github.io/ruby-lsp/#guessed-types


vinistock (approved):
Thank you!

PR #3074, opened 2025-01-17
Remove server side document selectors

vinistock:
### Motivation

I finally understood the mysterious requests we were receiving from URIs we didn't care about or the duplicate requests thanks to https://github.com/microsoft/vscode-languageserver-node/issues/1487#issuecomment-2597590940.

We were overriding the document selectors we define in the client in the server, with a much more simplistic approach. In the client, we take steps to ensure that we're handling only certain schemes with specific patterns and not duplicating client handling for the `untitled` scheme.

All of that configuration was lost for many requests because the server was simply broadcasting the document selector as `{ language: "ruby" }`, which is not scoped to any scheme or pattern.

This results in every language server running in the current VS Code window advertise that they handle any Ruby file, which is why we kept receiving duplicate requests and why we kept receiving requests for schemes we are not interested in handling - like the `vscode-chat-code-block://` scheme.

### Implementation

Since our document selector logic is so complex and requires knowledge of default gem paths, bundled gem paths, workspace paths and so on, we should not override it from the server. Trying to implement the same document selector logic on both ends would probably lead to unnecessary headache.

All of our requests already specify in `server.rb` if they need to return early for a certain document type, so there's no risk of receiving undesired requests that we don't know how to handle.

Considering that, I just set all document selectors in requests to `nil`, which [makes the editor prefer the one configured by the client](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#textDocumentRegistrationOptions).

### DIFF
--- a/lib/ruby_lsp/requests/code_actions.rb
+++ b/lib/ruby_lsp/requests/code_actions.rb
@@ -19,7 +19,7 @@ module RubyLsp
         sig { returns(Interface::CodeActionRegistrationOptions) }
         def provider
           Interface::CodeActionRegistrationOptions.new(
-            document_selector: [Interface::DocumentFilter.new(language: "ruby")],
+            document_selector: nil,
             resolve_provider: true,
           )
         end
--- a/lib/ruby_lsp/requests/diagnostics.rb
+++ b/lib/ruby_lsp/requests/diagnostics.rb
@@ -15,7 +15,7 @@ module RubyLsp
         sig { returns(Interface::DiagnosticRegistrationOptions) }
         def provider
           Interface::DiagnosticRegistrationOptions.new(
-            document_selector: [Interface::DocumentFilter.new(language: "ruby")],
+            document_selector: nil,
             inter_file_dependencies: false,
             workspace_diagnostics: false,
           )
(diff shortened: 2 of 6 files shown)


vinistock:
* **#3074** <a href="https://app.graphite.dev/github/pr/Shopify/ruby-lsp/3074?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/Shopify/ruby-lsp/3074?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

andyw8 (approved):
Worth asking someone using an editor other than VS Code to help verify?

vinistock:
> Worth asking someone using an editor other than VS Code to help verify?

In other editors, you usually specify the document selectors are part of configuring the language server. For example, in NeoVim: https://github.com/neovim/nvim-lspconfig/blob/339ccc81e08793c3af9b83882a6ebd90c9cc0d3b/lua/lspconfig/configs/ruby_lsp.lua#L6.

The user has more control over the document selectors. It's mostly for VS Code that it gets hidden away in the client implementation. So now the server will respect whatever people configure in other editors.

Did you have a particular editor in mind for testing?

andyw8:
Probably Neovim since it's seems most popular after VS Code.

PR #3077, opened 2025-01-17
Check if lockfile is valid before composing bundle

vinistock:
### Motivation

After #3066 moves forward, we can start getting smarter about our checks of when it's valid to restart. For example, we don't need to go all the way to composing the bundle, if we already know the lockfile's syntax is invalid.

### Implementation

Started parsing the lockfile before composing the bundle, so that we can more quickly reject a restart request if there's an issue.

### Automated Tests

Added a test.


vinistock:
### Merge activity

* **Jan 23, 9:45 AM EST**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/Shopify/ruby-lsp/3077).

PR #3078, opened 2025-01-17
Ignore unrecoverable Bundler errors for telemetry

vinistock:


vinistock:
I'm actually going to hold off on this one. I think #3066 may greatly reduce the need for this.

PR #3098, opened 2025-01-27
Avoid collecting duplicates in reference finder

### LINKED ISSUE #3088: textDocument/references Overreporting when class has multiple definitions in Neovim
### Description

Hi!

I noticed the following while I was working on a Ruby on Rails project that has classes that are re-opened in different parts of the codebase.

Something along the lines of:

```ruby
# app/models/user.rb
class User < ApplicationRecord
  include Validations
end

# app/models/concerns/user/validations.rb
class User
  module Validations
    extend ActiveSupport::Concern

    def method
    end
  end
end

# app/controllers/users_controller.rb
class UsersController < ApplicationController
    def show
      @user = User.find(params[:id])
    end
end
```

if I trigger the `textDocument/references` action while hovering on `User` I see the results twice for the reference in `users_controller.rb`. I also confirmed that this doesn't happen in VSCode.

I would to help with solving the issue once it is confirmed I would just need some guidance.

Thanks!

vicocamacho:
### Motivation

Closes #3088 

As I was playing around with the gem in Neovim I discovered some over reporting of references when a project has classes that are re-opened throughout the codebase.

### Implementation

Stop adding references on class and node module enter events and enhance the way we collect constant references to add additional logic to determine whether a reference is a constant or not.

### Automated Tests

Added one new test case to reference_finder_test.rb that covers for the scenario presented above.

### DIFF
--- a/lib/ruby_indexer/lib/ruby_indexer/location.rb
+++ b/lib/ruby_indexer/lib/ruby_indexer/location.rb
@@ -47,14 +47,14 @@ module RubyIndexer
 
     sig do
       params(
-        another_location: T.any(Location, Prism::Location)
+        other: T.any(Location, Prism::Location),
       ).returns(T::Boolean)
     end
-    def ==(another_location)
-      start_line == another_location.start_line &&
-        end_line == another_location.end_line &&
-        start_column == another_location.start_column &&
-        end_column == another_location.end_column
+    def ==(other)
+      start_line == other.start_line &&
+        end_line == other.end_line &&
+        start_column == other.start_column &&
+        end_column == other.end_column
     end
   end
 end


vinistock (commented):
Thank you for the PR!

I think we can avoid looking backwards in previously collected locations.

The reason for the duplicates is the fact that a class/module node will match both `on_class/module_node_enter` and `on_constant_*` because there's a constant node inside of the declaration.

Instead of looking backwards, let's stop pushing references in the `on_class_node_enter` and `on_module_node_enter` events. This will already get rid of the duplication.

The only piece left in the fix will be determining if a constant is a declaration or not. In `collect_constant_references`, we need to check if there's a declaration for the constant we just resolved, so that we can then consider it as `declaration: true`.

Something like this:

```ruby
entries = @index.resolve(name, @stack)
return unless entries

# Get the fully qualified name of the constant we just resolved
full_constant_name = T.must(entries.first).name

# Check if this constant is a namespace declaration
# (class or module)
declaration = @index[full_constant_name].any? { |e| e.is_a?(Entry::Namespace) }
```

vinistock:
It's almost there. I think this is what we need to do. There are a couple of pieces missing, like implementing `Location#==` and passing the file URI to the reference finder.

```ruby
def collect_constant_references(name, location)
  return unless @target.is_a?(ConstTarget)

  entries = @index.resolve(name, @stack)
  return unless entries

  # Filter down to all constant declarations that match the expected name and type
  matching_entries = entries.select do |e|
    [Entry::Namespace, Entry::Constant, Entry::ConstantAlias, Entry::UnresolvedConstantAlias].include?(e.class) &&
      e.name == @target.fully_qualified_name
  end

  # If any of the matching entries have the same location as the constant and were
  # defined in the same file, then it is that constant's declaration
  #
  # Note: we will need to implement `==` on RubyIndexer::Location to accept either
  # another `RubyIndexer::Location` or a `Prism::Location`, so that we can compare with
  # both.
  #
  # Additionally, we don't currently pass the `@uri` to the reference finder, but we will
  # need it so that we can check if the declaration was made in the same file
  declaration = matching_entries.any? do |e|
    e.name_location == location && e.uri == @uri
  end

  @references << Reference.new(name, location, declaration: declaration)
end
```

lib/ruby_indexer/lib/ruby_indexer/location.rb
```diff
@@ -44,5 +44,17 @@ def initialize(start_line, end_line, start_column, end_column)
...
+      params(
+        another_location: T.any(Location, Prism::Location)
+      ).returns(T::Boolean)
+    end
+    def ==(another_location)
+      start_line == another_location.start_line &&
+        end_line == another_location.end_line &&
+        start_column == another_location.start_column &&
+        end_column == another_location.end_column
+    end
```
vicocamacho, line 58: not sure if this implementation is good enough or we should this classes' `from_prism_location` method.

Also wanted to mention that I was not able to find a unit test file for this class and I was wondering if I should create it.

vinistock: Implementation looks good. We didn't have a test because the implementation of this class is generally very simple and ends up being tested by the logic that uses it.

For example, in the reference finder tests, this will be tested by verifying that we do not add duplicate references for a class/module declaration.

lib/ruby_indexer/lib/ruby_indexer/reference_finder.rb
```diff
@@ -341,17 +327,27 @@ def collect_constant_references(name, location)
...
-        next unless entry.name == @target.fully_qualified_name
+      # Filter down to all constant declarations that match the expected name and type
+      matching_entries = entries.select do |e|
+        [
+          Entry::Namespace,
+          Entry::Constant,
+          Entry::ConstantAlias,
+          Entry::UnresolvedConstantAlias,
+          Entry::Class
+        ].include?(e.class) &&
```
vinistock, line 338: Sorry, I just realized we should be using `is_a?` since some of these entries inherit from a common parent class. Then you can get rid of `Class` since that is a `Namespace`.

```suggestion
        [
          Entry::Namespace,
          Entry::Constant,
          Entry::ConstantAlias,
          Entry::UnresolvedConstantAlias,
        ].any? { |klass| e.is_a?(klass) } &&
```

vinistock (commented):
Just one minor modification, but this looks good

lib/ruby_indexer/lib/ruby_indexer/location.rb
vicocamacho: cool, that makes sense to me. Thanks for the clarification.

vicocamacho:
@vinistock Just incorporated the change. I think it might be good be go

vinistock:
The changes look good to me. There are still some linting errors. You can run linting locally with `bundle exec rubocop`.

Also, there's a conflict that needs to be resolved.

PR #3101, opened 2025-01-28
(docs) Specify dimensions for images to mitigate anchor bug

andyw8:
See https://github.com/just-the-docs/just-the-docs/issues/1574#issuecomment-2572588061

I haven't added dimensions for _all_ images since it's a bit tedious to do for all the ones with the different dimensions, but I think it will help.

Fort the syntax, see the [Kramdown docs for images](https://kramdown.gettalong.org/syntax.html#images)

### DIFF
--- a/jekyll/index.markdown
+++ b/jekyll/index.markdown
@@ -193,40 +193,40 @@ lens](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.
 request requires specific commands to be implemented in the editor in order to work. For VS Code, this is included with the Ruby LSP extension. If you are using a different editor, please check the editor's documentation on how to
 define the required commands.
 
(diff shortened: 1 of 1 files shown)


vinistock:
> The maintainer said it "didn't entirely resolve this problem". We can try it first to see though, maybe it's good enough.

That should be easy to verify. Does it fix the problem when you run `bundle exec jekyll serve`?

andyw8:
> That should be easy to verify. Does it fix the problem when you run bundle exec jekyll serve?

I haven't seen it happen, but it's hard to know for sure since it happens intermittently (and maybe not at all when running locally).

Let's deploy https://github.com/Shopify/ruby-lsp/pull/3102 and keep and eye on it?

PR #3102, opened 2025-01-28
(docs) Specify scroll behaviour to mitigate anchor bug

andyw8:
Alternative to https://github.com/Shopify/ruby-lsp/pull/3101


vinistock (approved):
Test it locally. You can reproduce the behaviour reliably by clicking on any item of the table of contents. The smooth scroll is removed with this

PR #3103, opened 2025-01-28
Copy VS Code README into docs

andyw8:
### Motivation

Copy the VS Code README into the docs for easier discovery.

(We need to keep the original as it's used for the VS Code extension page)

### Implementation

Copy file and images, fix syntax.

### Automated Tests

n/a


vinistock (approved):
Let's also change the link [here](https://shopify.github.io/ruby-lsp/#with-vs-code) to point to the Jekyll version.

PR #3109, opened 2025-01-29
Allow snippets for Cursor editor

vinistock:
### Motivation

Fixes the behaviour described in https://github.com/Shopify/ruby-lsp/discussions/3094#discussioncomment-11997299.

While the LSP spec doesn't standardize snippet support, we use a custom middleware to allow moving the user's cursor around when applying on type formatting. We were checking to apply this only in VS Code, but not Cursor.

### Implementation

Added Cursor to the mix.

### Automated Tests

Added a test.

### DIFF
--- a/lib/ruby_lsp/requests/on_type_formatting.rb
+++ b/lib/ruby_lsp/requests/on_type_formatting.rb
@@ -170,7 +170,7 @@ module RubyLsp
 
       sig { params(line: Integer, character: Integer).void }
       def move_cursor_to(line, character)
-        return unless @client_name.start_with?("Visual Studio Code")
+        return unless /Visual Studio Code|Cursor/.match?(@client_name)
 
         position = Interface::Position.new(
           line: line,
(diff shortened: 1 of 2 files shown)


lib/ruby_lsp/requests/on_type_formatting.rb
```diff
@@ -170,7 +170,7 @@ def add_edit_with_text(text, position = @position)
 
       sig { params(line: Integer, character: Integer).void }
       def move_cursor_to(line, character)
-        return unless @client_name.start_with?("Visual Studio Code")
+        return unless /(Visual Studio Code|Cursor)/.match?(@client_name)
```
andyw8, line 173:
```suggestion
        return unless /Visual Studio Code|Cursor/.match?(@client_name)
```

andyw8:
I wonder if we'll end up in a situation like browser user agents where other names are included for compatibility.

vinistock:
### Merge activity

* **Jan 29, 4:56 PM EST**: The merge label 'graphite-merge' was detected. This PR will be added to the [Graphite merge queue](https://app.graphite.dev/merges?org=Shopify&repo=ruby-lsp) once it meets the requirements.

PR #3119, opened 2025-01-31
Allow for under development feature flags

vinistock:
### Motivation

In addition to being able to roll out feature flags gradually, it would be nice if we could have flags that are only on if explicitly enabled. That allows us to hide under development functionality that's not ready even for beta testers.

This PR adds support for under development flags.

### Implementation

Started considering -1 as the indication that a flag is under development. If the value is -1, that flag is not enabled based on the `all` setting and can only be turned on by explicitly opting in.

### Automated Tests

Added a test.


st0012 (approved):

vinistock:
### Merge activity

* **Feb 3, 8:51 AM EST**: The merge label 'graphite-merge' was detected. This PR will be added to the [Graphite merge queue](https://app.graphite.dev/merges?org=Shopify&repo=ruby-lsp) once it meets the requirements.

PR #3120, opened 2025-01-31
Start discovering all available tests lazily

vinistock:
### Motivation

This is the first step in addressing several test explorer related issues that we currently have open.

This PR implements lazy test discovery up until the file level.

### Implementation

The VS Code API for the explorer supports lazily discovering tests by implementing the `resolveHandler`. I essentially only implement two resolutions:

1. If there's only one workspace folder, we eager discover all test files in the workspace
2. If there are many workspace folders, then the first level in the hierarchy are the workspaces themselves. If you open any of them, then it triggers resolution and we discover all available tests in that workspace

The tests are discovered using the VS Code `findFiles` API, which is very efficient and will avoid having us implement an indexing-like mechanism in the server. Since the test explorer is VS Code only anyway, this should be fine.

I also considered two levels of hierarchy for tests (in addition to the workspace when there are many). The firsts level is the `test/spec/features` directory. Since there may be many of them in an application split by components, we need to have that level of grouping.

The second level is the immediate subdirectory of the `test/spec/features` directory, which in a Rails app normally accounts for `models`, `controllers` and so on. This hierarchy is useful because it will allow users to take actions like running all tests in a workspace, test directory or all tests of a kind (e.g.: models).

### Notes

This PR doesn't yet implement the resolution of the test files themselves. The approach we will take for that is asking the server for which tests are available in a file, which gives us the opportunity to include add-on listeners in the discovery.

Since the implementation is still partial, I put a new under development feature flag.

### Automated Tests

Added tests.

### Manual tests

Enable this under development flag
```json
{
  "rubyLsp.featureFlags": {
    "newTestExperience": true
  }
}
```
Then start the extension on this branch and navigate tests

https://github.com/user-attachments/assets/1aaf145a-7b8f-48bf-b30b-61f0eef3e039

### DIFF
--- a/vscode/package.json
+++ b/vscode/package.json
@@ -502,6 +502,10 @@
               "description": "Opt-in/out of the Tapioca add-on",
               "type": "boolean"
             },
+            "fullTestDiscovery": {
+              "description": "UNDER DEVEOPMENT. Opt-in/out of the full test discovery experience",
+              "type": "boolean"
+            },
             "launcher": {
               "description": "Opt-in/out of the new launcher mode",
               "type": "boolean"
--- a/vscode/src/common.ts
+++ b/vscode/src/common.ts
@@ -86,6 +86,7 @@ export const SUPPORTED_LANGUAGE_IDS = ["ruby", "erb"];
 export const FEATURE_FLAGS = {
   tapiocaAddon: 1.0,
   launcher: 0.1,
+  fullTestDiscovery: -1,
 };
 
 type FeatureFlagConfigurationKey = keyof typeof FEATURE_FLAGS | "all";
--- a/vscode/src/testController.ts
+++ b/vscode/src/testController.ts
@@ -1,10 +1,12 @@
 import { exec } from "child_process";
 import { promisify } from "util";
+import path from "path";
 
 import * as vscode from "vscode";
 import { CodeLens } from "vscode-languageclient/node";
 
 import { Workspace } from "./workspace";
+import { featureEnabled } from "./common";
 
 const asyncExec = promisify(exec);
 
@@ -16,12 +18,16 @@ interface CodeLensData {
   kind: string;
 }
 
+const WORKSPACE_TAG = new vscode.TestTag("workspace");
+const TEST_DIR_TAG = new vscode.TestTag("test_dir");
+const DEBUG_TAG = new vscode.TestTag("debug");
+
 export class TestController {
-  private readonly testController: vscode.TestController;
+  // Only public for testing
+  readonly testController: vscode.TestController;
   private readonly testCommands: WeakMap<vscode.TestItem, string>;
   private readonly testRunProfile: vscode.TestRunProfile;
   private readonly testDebugProfile: vscode.TestRunProfile;
-  private readonly debugTag: vscode.TestTag = new vscode.TestTag("debug");
   private terminal: vscode.Terminal | undefined;
   private readonly telemetry: vscode.TelemetryLogger;
   // We allow the timeout to be configured in seconds, but exec expects it in milliseconds
@@ -30,6 +36,7 @@ export class TestController {
     .get("testTimeout") as number;
 
   private readonly currentWorkspace: () => Workspace | undefined;
+  private readonly fullDiscovery = featureEnabled("fullTestDiscovery");
 
   constructor(
     context: vscode.ExtensionContext,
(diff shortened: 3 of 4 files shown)


vinistock:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.dev/github/pr/Shopify/ruby-lsp/3120?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* **#3120** <a href="https://app.graphite.dev/github/pr/Shopify/ruby-lsp/3120?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/Shopify/ruby-lsp/3120?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#3119** <a href="https://app.graphite.dev/github/pr/Shopify/ruby-lsp/3119?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`
<hr />

You must have a Graphite account in order to use the merge queue. Sign up using [this link](https://app.graphite.dev/invite/github/Shopify?ref=merge-queue-instructions-in-stack-comment&prId=3120).

<h2></h2>
<sub>

</sub>

<sub>

</sub>

<h2></h2>

vscode/src/testController.ts
```diff
@@ -400,4 +416,148 @@ export class TestController {
...
+    }
+  }
+
+  private async gatherWorkspaceTests(
+    workspaceFolder: vscode.WorkspaceFolder,
+    item: vscode.TestItem | undefined,
+  ) {
+    const pattern = new vscode.RelativePattern(
+      workspaceFolder,
+      "**/{test,spec,features}/**/{*_test.rb,test_*.rb,*_spec.rb,*.feature}",
```
st0012, line 473: Why is this pattern different then the one on 445? Should we create a helper function for this?

vscode/package.json
```diff
@@ -497,6 +497,10 @@
               "description": "Opt-in/out of the Tapioca add-on",
               "type": "boolean"
             },
+            "newTestExperience": {
+              "description": "UNDER DEVEOPMENT. Opt-in/out of the new test experience",
+              "type": "boolean"
```
st0012, line 502: Does `all` cover this?


```json
    "rubyLsp.featureFlags": {
        "all": true,
    },
```

st0012 (commented):
Is it possible to expand `test/controllers/foo_controller_test.rb` in the test explorer when I opened `foo_controller_test.rb` in the editor? Similar to how VS Code would expand the file explorer based on the currently opened file.
Not a blocker for the PR but it'd be a very useful feature IMO.

vscode/src/testController.ts
vinistock: That was a mistake. Created a function to avoid mismatches.

vscode/package.json
vinistock: It doesn't after #3119 goes in. This branch depends on that one.

vscode/src/common.ts
```diff
@@ -85,6 +85,7 @@ export const SUPPORTED_LANGUAGE_IDS = ["ruby", "erb"];
 export const FEATURE_FLAGS = {
   tapiocaAddon: 0.3,
   launcher: 0.05,
+  newTestExperience: -1,
```
andyw8, line 88: Can we give a more descriptive name here, maybe `lazyTestDiscovery`? "new" becomes a less useful distinction as time goes on.

vinistock: Changed the name to `fullDiscovery`. The intention is to have everyone migrate to the new experience, so the flag is temporary.

vinistock:
> Is it possible to expand test/controllers/foo_controller_test.rb in the test explorer when I opened foo_controller_test.rb in the editor?

Yes! Added that. Note that you need to have clicked on the explorer first, to lazily populate the tests. We don't force the explorer to reveal, but if you're already there, we focus on the current test.

st0012 (approved):
Tested it in Core, IRB, RDoc, and ruby-lsp-rspec. All worked as expected 👍 

One non-blocking suggestion, I think the experience would be better if it can automatically expand results to the first level having multiple results.

For example, in IRB I needed to expand `test` and then `irb` even though both of them are the only item on their level.

Screenshot 2025-02-07 at 21 32 50

vinistock:
### Merge activity

* **Feb 7, 4:55 PM EST**: The merge label 'graphite-merge' was detected. This PR will be added to the [Graphite merge queue](https://app.graphite.dev/merges?org=Shopify&repo=ruby-lsp) once it meets the requirements.

Issue #3117, opened 2025-01-31
Adding a "=" or a "#" to an erb tag result in a duplicate tag ending


olivierimmosquare:
#### VS Code Version

1.96.4

#### Ruby LSP Extension Version

0.9.0

#### Ruby LSP Server Version

0.23.7

#### Ruby LSP Add-ons

- Ruby LSP Rails
- RuboCop

#### Ruby Version

3.4.1

#### Ruby Version Manager

rvm

#### Installed Extensions

<details>
<summary>Click to expand</summary>

- RunOnSave (0.2.7)
- auto-rename-tag (0.1.10)
- better-comments (3.0.2)
- copilot (1.261.0)
- copilot-chat (0.23.2)
- debugpy (2024.14.0)
- gitlens (16.2.1)
- highlight-matching-tag (0.11.0)
- indent-rainbow (8.3.1)
- material-palenight-theme (2.0.4)
- partial-diff (1.4.3)
- prettyxml (5.2.0)
- python (2024.22.2)
- rails-db-schema (0.2.9)
- rainbow-csv (3.15.1)
- render-crlf (1.8.5)
- ruby-lsp (0.9.0)
- vscode-icons (12.11.0)
- vscode-pylance (2024.12.1)
- vscode-rdbg (0.2.2)
- vscode-react-coffee-pug (0.0.1)
- vscode-theme-onedark (2.3.0)
- vscode-theme-onelight (2.3.0)
- vscode-xml (0.27.2)
- vscode-yaml (1.15.0)
</details>

#### Ruby LSP Settings

<details>
<summary>Click to expand</summary>

##### Workspace

```json
{}
```

##### User

```json
{
  "enabledFeatures": {
    "codeActions": true,
    "diagnostics": true,
    "documentHighlights": true,
    "documentLink": true,
    "documentSymbols": true,
    "foldingRanges": true,
    "formatting": false,
    "hover": true,
    "inlayHint": true,
    "onTypeFormatting": true,
    "selectionRanges": true,
    "semanticHighlighting": true,
    "completion": true,
    "codeLens": true,
    "definition": true,
    "workspaceSymbol": true,
    "signatureHelp": true,
    "typeHierarchy": true
  },
  "featuresConfiguration": {},
  "addonSettings": {},
  "rubyVersionManager": {
    "identifier": "auto"
  },
  "customRubyCommand": "",
  "formatter": "none",
  "linters": null,
  "bundleGemfile": "",
[...]

PR #3121, opened 2025-01-31
fix duplicate ending tag  on erb.html files(#3117)

jules-w2:
The issue manifests in two problematic behaviors: https://github.com/Shopify/ruby-lsp/issues/3117

When modifying an existing ERB tag `<% Time.now %>` by adding `=` or `#`, the editor incorrectly inserts an additional closing `%>`, resulting in malformed ERB syntax.The situation arises very often when you want to comment on a line that already exists in your views.

```ruby
<%#%> Time.now %>
```

Similarly, when adding a comment symbol `=`, the editor produces invalid nested ERB syntax:
```ruby
<%=%> Time.now %>
```

This behavior is caused by the autoClosingPairs configuration in the ERB language settings, which automatically adds closing tags in situations where they should not be added.


andyw8:
@jules-w2 please sign the CLA.

PR #3123, opened 2025-01-31
Add jekyll as a VS Code workspace

andyw8:
This is its own project within the repo.


vinistock (commented):
You also need to exclude jekyll in `files.exclude` otherwise, it will appear twice in the UI

andyw8:
Updated

vinistock (approved):

PR #3136, opened 2025-02-04
Extract `constant_name` method

andyw8:
This came up while pairing on https://github.com/Shopify/ruby-lsp/pull/3031

@vinistock you initially suggested extracting to `GlobalState` but it's also used `ReferenceFinder`, so I think having on `Index` makes more sense?


vinistock:
Ah, actually, let's keep only the one in `common.rb`. Otherwise, this is a breaking change as the Rails add-on depends on `constant_name`.

But let's leave it just to avoid the breaking change, we can use `Index.constant_name` directly in our implementation.

vinistock (approved):

PR #3139, opened 2025-02-05
Ignore untrusted workspace error for telemetry

vinistock:
### Motivation

In efforts to make our telemetry as useful as possible, I think we should ignore errors that we cannot recover from. When the version manager is Shadowenv and the current workspace is untrusted, there's nothing we can do.

Auto-trusting a workspace is a security issue because it would execute arbitrary Shadow Lisp files, so that's not an option.

Since only the user can decide to trust a workspace, there really is nothing we can do if they dismissed our offer to trust it.

### Implementation

Started ignoring untrusted workspace errors for telemetry. Also, added the workspace name as an attribute, which is useful when trying to discover specific types of errors.

### Automated Tests

Added a test.

### DIFF
--- a/vscode/src/ruby.ts
+++ b/vscode/src/ruby.ts
@@ -6,7 +6,7 @@ import * as vscode from "vscode";
 
 import { asyncExec, RubyInterface } from "./common";
 import { WorkspaceChannel } from "./workspaceChannel";
-import { Shadowenv } from "./ruby/shadowenv";
+import { Shadowenv, UntrustedWorkspaceError } from "./ruby/shadowenv";
 import { Chruby } from "./ruby/chruby";
 import { VersionManager } from "./ruby/versionManager";
 import { Mise } from "./ruby/mise";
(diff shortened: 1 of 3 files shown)


andyw8 (approved):

vinistock:
### Merge activity

* **Feb 5, 10:08 AM EST**: The merge label 'graphite-merge' was detected. This PR will be added to the [Graphite merge queue](https://app.graphite.dev/merges?org=Shopify&repo=ruby-lsp) once it meets the requirements.

vscode/src/ruby.ts
```diff
@@ -151,9 +151,14 @@ export class Ruby implements RubyInterface {
       try {
         await this.runManagerActivation();
       } catch (error: any) {
-        this.telemetry.logError(error, {
-          versionManager: this.versionManager.identifier,
-        });
+        if (!(error instanceof UntrustedWorkspaceError)) {
```
andyw8, line 154: Would a little easier to read if we use an early return.

PR #3142, opened 2025-02-05
Ensure disposables are being tracked by the entities that own them

vinistock:


vscode/src/workspace.ts
```diff
@@ -189,7 +189,7 @@ export class Workspace implements WorkspaceInterface {
   }
 
   async stop() {
```
st0012, line 191: Should this be called `dispose` instead 🤔

vscode/src/workspace.ts
```diff
@@ -373,7 +373,7 @@ export class Workspace implements WorkspaceInterface {
 
     // If a configuration that affects the Ruby LSP has changed, update the client options using the latest
     // configuration and restart the server
-    this.context.subscriptions.push(
+    this.subscriptions.push(
```
st0012, line 376: Noob question: why don't we push these to the client too and then let its `dispose` handle them?

vscode/src/workspace.ts
vinistock: We also have a dispose, but they're not quite the same thing. When we stop a workspace, it means stopping the LSP client (which we do dispose), but we don't dispose of the workspace itself. That remains active until the extension is deactivated.

vscode/src/workspace.ts
vinistock: Because you can restart the LSP client as many times as you like in the same workspace, but we never re-create the file watchers (disposables). We can only dispose of them when the workspace is being removed/deactivated.

vinistock: Yes. Added some comments, let me know if you think it's clear.

PR #3144, opened 2025-02-06
Use unique IDs for registering watchers

### LINKED ISSUE #3137: Notify file events failed - Client is not running
#### VS Code Version

1.96.4

0.8.20

#### Ruby LSP Server Version

0.23.6

#### Ruby LSP Add-ons

- RuboCop

#### Ruby Version

3.2.2

#### Ruby Version Manager

rbenv

#### Installed Extensions

<details>

- copilot (1.263.0)
- copilot-chat (0.23.2)
- debugpy (2024.14.0)
- less-aggressive-spinel-theme (0.0.3)
- markdown-preview-enhanced (0.8.15)
- ms-teams-vscode-extension (5.12.1)
- python (2024.23.2025012801)
- ruby-lsp (0.8.20)
- vscode-pylance (2025.1.102)
- vscode-yaml (1.15.0)
</details>

#### Ruby LSP Settings

<details>

##### Workspace

```json
{
  "formatter": "auto"
}
```

##### User

```json
{
  "enabledFeatures": {
    "codeActions": true,
    "diagnostics": true,
    "documentHighlights": true,
    "documentLink": true,
    "documentSymbols": true,
    "foldingRanges": true,
    "formatting": true,
    "hover": true,
    "inlayHint": true,
    "onTypeFormatting": true,
    "selectionRanges": true,
    "semanticHighlighting": true,
    "completion": true,
    "codeLens": true,
    "definition": true,
    "workspaceSymbol": true,
    "signatureHelp": true,
    "typeHierarchy": true
  },
  "featuresConfiguration": {},
  "addonSettings": {},
  "rubyVersionManager": {
    "identifier": "auto"
  },
  "customRubyCommand": "",
  "formatter": "auto",
  "linters": null,
  "bundleGemfile": "",
  "testTimeout": 30,
  "branch": "",
  "pullDiagnosticsOn": "both",
  "useBundlerCompose": false,
  "bypassTypechecker": false,
  "rubyExecutablePath": "",
  "indexing": {},
  "erbSupport": true,
  "featureFlags": {}
}
```

</details>

### Reproduction steps

1. Restart LSP Server
2. Use autocompletion (this works fine)
3. Save file
4. I keep getting the error below and I need to restard the server

```ruby
2025-02-05 09:14:03.324 [info] (MortarApp) [Error - 09:14:03] Notify file events failed.
2025-02-05 09:14:03.324 [info] (MortarApp)   Message: Client is not running
  Code: -32096 
```

vinistock:
### Motivation

Closes #3137

We were using the same registration ID for two different file watchers. That made VS Code override the entry, but one of them became a dangling object that didn't get appropriately cleaned up when the LSP shuts down.

Because that old file watcher was still active, the LSP client itself was not cleared either as a reference was still being held and then notifications for `didChangeWatchedFiles` were sent to a client that was already stopped, resulting in the `Notify file events failed` message.

We should always use unique IDs when registering file watchers.

### Implementation

I did 3 small things:

1. Ensured unique IDs for registering the watchers
2. Started watching `.rubocop`, which we watched before but I missed in my refactor PR
3. Started ensuring that we're only processing changes once

The last point is necessary because add-ons may register for watching the same patterns as one another or as the Ruby LSP and that results in duplicate changes being sent to the server.

The most optimal solution is #3145, but that is going to require more effort, so let's just fix the issue for now.

### Automated Tests

Added a test and verified manually that the issue is gone.

### DIFF
--- a/lib/ruby_lsp/server.rb
+++ b/lib/ruby_lsp/server.rb
@@ -301,10 +301,19 @@ module RubyLsp
 
       # Not every client supports dynamic registration or file watching
       if @global_state.client_capabilities.supports_watching_files
-        send_message(Request.register_watched_files(@current_request_id, "**/*.rb"))
         send_message(Request.register_watched_files(
           @current_request_id,
-          Interface::RelativePattern.new(base_uri: @global_state.workspace_uri.to_s, pattern: ".rubocop.yml"),
+          "**/*.rb",
+          registration_id: "workspace-watcher",
+        ))
+
+        send_message(Request.register_watched_files(
+          @current_request_id,
+          Interface::RelativePattern.new(
+            base_uri: @global_state.workspace_uri.to_s,
+            pattern: "{.rubocop.yml,.rubocop}",
+          ),
+          registration_id: "rubocop-watcher",
         ))
       end
 
@@ -999,6 +1008,11 @@ module RubyLsp
     sig { params(message: T::Hash[Symbol, T.untyped]).void }
     def workspace_did_change_watched_files(message)
       changes = message.dig(:params, :changes)
+      # We allow add-ons to register for watching files and we have no restrictions for what they register for. If the
+      # same pattern is registered more than once, the LSP will receive duplicate change notifications. Receiving them
+      # is fine, but we shouldn't process the same file changes more than once
+      changes.uniq!
+
       index = @global_state.index
       changes.each do |change|
         # File change events include folders, but we're only interested in files
--- a/lib/ruby_lsp/internal.rb
+++ b/lib/ruby_lsp/internal.rb
@@ -23,6 +23,7 @@ require "language_server-protocol"
 require "rbs"
 require "fileutils"
 require "open3"
+require "securerandom"
 
 require "ruby-lsp"
 require "ruby_lsp/base_server"
--- a/lib/ruby_lsp/utils.rb
+++ b/lib/ruby_lsp/utils.rb
@@ -176,11 +176,19 @@ module RubyLsp
     class << self
       extend T::Sig
 
-      sig { params(id: Integer, pattern: T.any(Interface::RelativePattern, String), kind: Integer).returns(Request) }
+      sig do
+        params(
+          id: Integer,
+          pattern: T.any(Interface::RelativePattern, String),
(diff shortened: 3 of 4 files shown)


lib/ruby_lsp/server.rb
```diff
@@ -301,10 +301,14 @@ def run_initialize(message)
 
       # Not every client supports dynamic registration or file watching
       if @global_state.client_capabilities.supports_watching_files
-        send_message(Request.register_watched_files(@current_request_id, "**/*.rb"))
+        send_message(Request.register_watched_files(@current_request_id, "didChangeWatchedFilesWorkspace", "**/*.rb"))
         send_message(Request.register_watched_files(
           @current_request_id,
-          Interface::RelativePattern.new(base_uri: @global_state.workspace_uri.to_s, pattern: ".rubocop.yml"),
+          "didChangeWatchedFilesRuboCopCofig",
```
andyw8, line 307: (typo)
```suggestion
          "didChangeWatchedFilesRuboCopConfig",
```

lib/ruby_lsp/server.rb
```diff
@@ -301,10 +301,14 @@ def run_initialize(message)
 
       # Not every client supports dynamic registration or file watching
       if @global_state.client_capabilities.supports_watching_files
-        send_message(Request.register_watched_files(@current_request_id, "**/*.rb"))
+        send_message(Request.register_watched_files(@current_request_id, "didChangeWatchedFilesWorkspace", "**/*.rb"))
```
andyw8, line 304: The full name seems a little redundant since all will begin `didChangeWatchFiles`, how about `workspace` (or `:workspace`)?

Alternatively, could we auto-generate this ID, e.g. by taking a hash of the pattern argument?

andyw8:
It probably has minimal impact, but don't know what private or under-development might be using it. Your call.

vinistock:
Alright, I made the API compatible with the old style to avoid having to cut a breaking change just for this.

I checked every add-on published to rubygems with the `ruby-lsp` prefix and no one was using it, but better to not make unannounced breaking changes.

andyw8:
Nice solution.

PR #3187, opened 2025-02-12
Add JSON test reporter for minitest

andyw8:
Addresses https://github.com/Shopify/ruby-lsp/issues/3176

This adds a custom reporter for use with the upcoming changes planned for the Test Explorer UI. It emits the test progress as a series of events, for easy consumption by the VS Code extension.

To test the runner output locally:
```
RUBYOPT="-r./lib/ruby_lsp/ruby_lsp_reporter_plugin.rb" bundle exec ruby -Itest test/fixtures/minitest_example.rb
```

### DIFF
--- a/test/fixtures/minitest_example.rb
+++ b/test/fixtures/minitest_example.rb
@@ -1,6 +1,7 @@
 # frozen_string_literal: true
 
-require "test_helper"
+require "minitest/autorun"
+require_relative "../../lib/ruby_indexer/lib/ruby_indexer/uri"
 
 # We are only testing the output of the runner, there's no need for to be random.
 Minitest::Test.i_suck_and_my_tests_are_order_dependent!
--- a/test/fixtures/minitest_example.rb
+++ b/test/fixtures/minitest_example.rb
@@ -5,7 +5,7 @@ require "test_helper"
 # We are only testing the output of the runner, there's no need for to be random.
 Minitest::Test.i_suck_and_my_tests_are_order_dependent!
 
-class Sample < Minitest::Test
+class SampleTest < Minitest::Test
   def test_that_passes
     assert_equal(1, 1)
     assert_equal(2, 2)
--- a/test/fixtures/minitest_example.rb
+++ b/test/fixtures/minitest_example.rb
@@ -1,0 +1,25 @@
+# frozen_string_literal: true
+
+require "test_helper"
+
+# We are only testing the output of the runner, there's no need for to be random.
+Minitest::Test.i_suck_and_my_tests_are_order_dependent!
+
+class Sample < Minitest::Test
+  def test_that_passes
+    assert_equal(1, 1)
+    assert_equal(2, 2)
+  end
+
+  def test_that_fails
+    assert_equal(1, 2)
+  end
+
+  def test_that_is_pending
+    skip("pending test")
+  end
+
+  def test_that_raises
+    raise "oops"
+  end
+end
--- a/test/minitest_test_runner_test.rb
+++ b/test/minitest_test_runner_test.rb
@@ -19,7 +19,7 @@ module RubyLsp
       stdout.binmode
       stdout.sync = true
 
-      actual = parse_output(stdout)
+      actual = parse_json_api_stream(stdout)
 
       uri = URI::Generic.from_path(path: "#{Dir.pwd}/test/fixtures/minitest_example.rb").to_s
       expected = [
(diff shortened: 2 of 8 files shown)


lib/minitest/reporters/ruby_lsp_reporter.rb
```diff
@@ -0,0 +1,95 @@
...
+      end
+
+      sig { params(result: Minitest::Result).void }
+      def record_fail(result)
+        info = {
+          class_name: result.klass,
+          test_name: result.name,
+          type: result.failure.class.name,
+          message: result.failure.message,
+          file: result.source_location[0],
```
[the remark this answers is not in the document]
andyw8: There's a `before_suite` which takes a `Minitest::Reporters::Suite`, but that doesn't have file information:

https://github.com/minitest-reporters/minitest-reporters/blob/265ff4b40d5827e84d7e902b808fbee860b61221/lib/minitest/reporters/base_reporter.rb#L3C11-L3C16

We could propose adding this in `minitest-reporters`.

lib/ruby_lsp/test_reporting.rb
```diff
@@ -0,0 +1,75 @@
...
+        file: file,
+      }
+      puts result.to_json
+    end
+
+    sig do
+      params(
+        class_name: String,
+        test_name: String,
+        type: T.untyped, # TODO: what type should this be?
```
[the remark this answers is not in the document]
andyw8: I've seen two variants so far:

* A `Minitest::Assertion` failure
* A `Minitest::UnexpectedError`, e.g. if something in the test raises

I thought it may be useful to distingish these in the UI, similar to the progress reporter shows `E` vs `F`.

lib/ruby_lsp/test_reporting.rb
```diff
@@ -0,0 +1,75 @@
+# typed: strict
+# frozen_string_literal: true
+
+require "json"
+
+module RubyLsp
+  class TestReporting
```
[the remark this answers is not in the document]
andyw8: I've changed the name.

Also I am now passing in an `IO` instance to allow for easier testing and flexibility.

lib/ruby_lsp/test_reporting.rb
```diff
@@ -0,0 +1,80 @@
+# typed: strict
+# frozen_string_literal: true
+
+require "json"
```
st0012, line 4: Do we need to require `sorbet-runtime` to make sure `T::Sig` is available?

lib/minitest/reporters/ruby_lsp_reporter.rb
```diff
@@ -0,0 +1,96 @@
...
+      end
+
+      sig { params(result: Minitest::Result).returns(String) }
+      def id_from_result(result)
+        [result.name, result.klass].join("#")
+      end
+
+      sig { params(test: Minitest::Test).returns(String) }
+      def file_for_class_name(test)
+        T.must(Kernel.const_source_location(test.class_name)).first
```
st0012, line 92: Just a headsup that even in minitest/test-unit it's possible to define test classes dynamically, which make their source location path a bit different, like:

```rb
class Foo
  class_eval <<~RUBY
    class FooTest < Test::Unit::TestCase
      def test_foo
        puts "========================================="
        puts Kernel.const_source_location("Foo::FooTest")
        puts "========================================="
        fail
      end
    end
  RUBY
end
```

```
=========================================
(eval at /Users/hung-wulo/src/github.com/Shopify/rdoc/test/rdoc/test_rdoc_comment.rb:483)
1
=========================================
```

lib/ruby_lsp/test_reporting.rb
andyw8: Yes, added in `ruby_lsp_reporter.rb`.

lib/minitest/reporters/ruby_lsp_reporter.rb
andyw8: Pushed an update to handle that. Still need to think about what we do in the UI.

test/fixtures/minitest_example.rb
```diff
@@ -0,0 +1,25 @@
+# frozen_string_literal: true
+
+require "test_helper"
+
+# We are only testing the output of the runner, there's no need for to be random.
+Minitest::Test.i_suck_and_my_tests_are_order_dependent!
```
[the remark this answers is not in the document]
andyw8: It does matter because `assert_equal` uses `==` to compare the arrays.

https://ruby-doc.org/stdlib-3.0.0/libdoc/minitest/rdoc/Minitest/Assertions.html#method-i-assert_equal

test/minitest_test_runner_test.rb
```diff
@@ -0,0 +1,104 @@
...
+      assert_equal(expected, actual)
+    end
+
+    private
+
+    def error_location
+      ruby_version = Gem::Version.new(RUBY_VERSION)
+
+      # TODO: confirm when this behavior changed
+      # Also note the difference in the initial character.
```
vinistock, line 96: ✂️

vinistock (approved):

PR #3188, opened 2025-02-12
Ensure we index non test files under test directories

vinistock:
### Motivation

I noticed something with #3180. We can only accurately linearize ancestors for test files if we index files under the test directories that aren't tests, but support implementations.

For example, in our own codebase, we have `lib/ruby_indexer/test/test_case.rb`, which is not a test, but implements the parent class for all tests in the indexer. If we never index that parent class, it's impossible to linearize the ancestors of any of the tests and we can't discover them properly.

### Implementation

I propose we start indexing non test files inside the test directory. In addition to allowing our improved test functionality to work, users often refer to these support declarations when writing the tests themselves.

I also simplified the code, which was hard to follow, while still trying to maintain the original performance gains applied by recent changes.

One small thing is that I couldn't think of a decent way to not exclude `test_case` and `test_helper` from indexing, since both of them are common names that may declare test parent classes and match our file exclusion glob. I created a special case to handle them.

I benchmarked this and it makes no substantial difference in indexing time in Core.

### Automated Tests

Added a test verifying the behaviour changes.

### DIFF
--- a/lib/ruby_indexer/lib/ruby_indexer/configuration.rb
+++ b/lib/ruby_indexer/lib/ruby_indexer/configuration.rb
@@ -31,11 +31,8 @@ module RubyIndexer
 
       @excluded_patterns = T.let(
         [
-          File.join("**", "*_test.rb"),
-          File.join("node_modules", "**", "*"),
-          File.join("spec", "**", "*"),
-          File.join("test", "**", "*"),
-          File.join("tmp", "**", "*"),
+          "**/{test,spec}/**/{*_test.rb,test_*.rb,*_spec.rb}",
+          "**/fixtures/**/*",
         ],
         T::Array[String],
       )
@@ -44,10 +41,13 @@ module RubyIndexer
       if path
         # Substitute Windows backslashes into forward slashes, which are used in glob patterns
         glob = path.gsub(/[\\]+/, "/")
-        @excluded_patterns << File.join(glob, "**", "*.rb")
+        glob.delete_suffix!("/")
+        @excluded_patterns << "#{glob}/**/*.rb"
       end
 
-      @included_patterns = T.let([File.join("**", "*.rb")], T::Array[String])
+      # We start the included patterns with only the non excluded directories so that we can avoid paying the price of
+      # traversing large directories that don't include Ruby files like `node_modules`
+      @included_patterns = T.let(["{#{top_level_directories.join(",")}}/**/*.rb", "*.rb"], T::Array[String])
       @excluded_magic_comments = T.let(
         [
           "frozen_string_literal:",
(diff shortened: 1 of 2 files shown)


vinistock:
* **#3188** <a href="https://app.graphite.dev/github/pr/Shopify/ruby-lsp/3188?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/Shopify/ruby-lsp/3188?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`
<hr />

You must have a Graphite account in order to use the merge queue. Sign up using [this link](https://app.graphite.dev/invite/github/Shopify?ref=merge-queue-instructions-in-stack-comment&prId=3188).

<h2></h2>
<sub>

</sub>

<sub>

</sub>

<h2></h2>

lib/ruby_indexer/lib/ruby_indexer/configuration.rb
```diff
@@ -149,11 +110,23 @@ def indexable_uris
...
       # Remove user specified patterns
+      bundle_path = Bundler.settings["path"]&.gsub(/[\\]+/, "/")
       uris.reject! do |indexable|
-        excluded_patterns.any? do |pattern|
-          File.fnmatch?(pattern, T.must(indexable.full_path), File::FNM_PATHNAME | File::FNM_EXTGLOB)
+        path = T.must(indexable.full_path)
+
+        # Don't exclude the special files, unless they are under fixtures or inside a gem installed inside of the
+        # workspace
+        if excluded_from_ignore_test_files.include?(File.basename(path)) &&
```
andyw8, line 124: This is a complex conditional, I think we can improve it by extracting out some helper methods or explaining variables.

lib/ruby_indexer/lib/ruby_indexer/configuration.rb
```diff
@@ -31,11 +31,8 @@ def initialize
 
       @excluded_patterns = T.let(
         [
-          File.join("**", "*_test.rb"),
-          File.join("node_modules", "**", "*"),
-          File.join("spec", "**", "*"),
-          File.join("test", "**", "*"),
-          File.join("tmp", "**", "*"),
+          "**/{test,spec,features}/**/{*_test.rb,test_*.rb,*_spec.rb,*.feature}",
```
andyw8, line 34: `.feature` files don't contain Ruby code, is there any need to list them here?

lib/ruby_indexer/lib/ruby_indexer/configuration.rb
```diff
@@ -44,10 +41,21 @@ def initialize
...
         # Substitute Windows backslashes into forward slashes, which are used in glob patterns
         glob = path.gsub(/[\\]+/, "/")
-        @excluded_patterns << File.join(glob, "**", "*.rb")
+        glob.delete_suffix!("/")
+        @excluded_patterns << "#{glob}/**/*.rb"
       end
 
-      @included_patterns = T.let([File.join("**", "*.rb")], T::Array[String])
+      excluded_directories = ["tmp", "node_modules", "sorbet"]
+      top_level_directories = Dir.glob("#{Dir.pwd}/*").filter_map do |path|
```
andyw8, line 49: We could extract `top_level_directories` to a method to avoid making `initializer` longer.

PR #3197, opened 2025-02-13
Avoid focusing on test items on document switch

vinistock:
### Motivation

VS Code doesn't provide an API for us to know which test item is currently selected in the UI.

This means that, if you have a specific test example selected, it gets unselected when you switch documents due to the callback we added.

### Implementation

It's unfortunate, but I don't think we're able to auto-select the items correctly based on which document is opened.

### DIFF
--- a/vscode/src/testController.ts
+++ b/vscode/src/testController.ts
@@ -92,24 +92,6 @@ export class TestController {
       vscode.window.onDidCloseTerminal((terminal: vscode.Terminal): void => {
         if (terminal === this.terminal) this.terminal = undefined;
       }),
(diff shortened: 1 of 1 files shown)


amomchilov (approved):
feelsbadman

How is this basic feature so underserved by VSCode's API? :(

vinistock:
### Merge activity

* **Feb 14, 8:59 AM EST**: The merge label 'graphite-merge' was detected. This PR will be added to the [Graphite merge queue](https://app.graphite.dev/merges?org=Shopify&repo=ruby-lsp) once it meets the requirements.

Morriar (approved):
Should we open an RFC upstream?

Issue #3200, opened 2025-02-14
Encoding::UndefinedConversionError when activating VSCode extension


webmatze:
### Ruby LSP Information

The command `Collect Ruby LSP information for issue reporting` returned no result for me but here are some information:

**VS Code Version**
1.98.0-insider

**Ruby LSP Extension Version**
0.9.4

**Ruby LSP Server Version**
0.23.10

**Ruby LSP Add-ons**

- Ruby LSP Rails
- RuboCop

**Ruby Version**
3.2.6

**Ruby Version Manager**
asdf

### Reproduction steps

1. Start the Ruby LSP using a VSCoder Insider Edition
2. error message is shown in popup

**Code snippet or error message**
```bash
Automatic Ruby environment activation with asdf failed: Command failed: . /opt/homebrew/opt/asdf/libexec/asdf.sh && asdf exec ruby -EUTF-8:UTF-8 '/Users/mathiaskarstaedt/.vscode-insiders/extensions/shopify.ruby-lsp-0.9.4/activation.rb'
/Users/mathiaskarstaedt/.vscode-insiders/extensions/shopify.ruby-lsp-0.9.4/activation.rb:3:in `write': "\xE2" from ASCII-8BIT to UTF-8 (Encoding::UndefinedConversionError)
	from /Users/mathiaskarstaedt/.vscode-insiders/extensions/shopify.ruby-lsp-0.9.4/activation.rb:3:in `print'
	from /Users/mathiaskarstaedt/.vscode-insiders/extensions/shopify.ruby-lsp-0.9.4/activation.rb:3:in `<main>'
```

**Hint** 
I managed to fix the problem by changing **line 3** of the **activation.rb** to:

```ruby
STDERR.print("RUBY_LSP_ACTIVATION_SEPARATOR#{env.join("RUBY_LSP_FS")}RUBY_LSP_ACTIVATION_SEPARATOR".encode('UTF-8', invalid: :replace, undef: :replace))
```

rynonl:
Confirm that I am seeing this as well (Ruby 3.4.1). Thanks for the workaround in the meantime.

webmatze:
@vinistock Thank you for the fast reaction. Sadly I am still getting the same error after updating to v0.9.5 on my system.
When using my dirty fix from above it works again.

mckeed:
[...]

PR #3214, opened 2025-02-19
Fix Encoding::UndefinedConversionError in activation.rb

mckeed:
### Motivation

Another fix for #3200. The [last one](https://github.com/Shopify/ruby-lsp/issues/3161) only checked for malformed UTF-8, not correct UTF-8 with ASCII encoding, which is the issue.

### Implementation

I changed it so it converts everything to UTF-8 to avoid the Encoding::UndefinedConversionError error.

### Automated Tests

I updated the test so it fails on the previous fix.

### Manual Tests

On macOS, add a unicode character to your system shell prompt and launch vscode using the .app, not from the command line.

### DIFF
--- a/test/integration_test.rb
+++ b/test/integration_test.rb
@@ -29,7 +29,7 @@ class IntegrationTest < Minitest::Test
   def test_activation_script_succeeds_even_on_binary_encoding
     ENV["LC_ALL"] = "C"
     ENV["LANG"] = "C"
-    ENV["PS1"] = "\xE2".b
+    ENV["PS1"] = "\xE2\x96\xB7".b
 
     _stdout, stderr, status = Open3.capture3(
       "ruby",
@@ -39,7 +39,7 @@ class IntegrationTest < Minitest::Test
 
     assert_equal(0, status.exitstatus, stderr)
 
-    activation_string = T.must(/RUBY_LSP_ACTIVATION_SEPARATOR(.*)RUBY_LSP_ACTIVATION_SEPARATOR/.match(stderr))[1]
+    activation_string = T.must(/RUBY_LSP_ACTIVATION_SEPARATOR(.*)RUBY_LSP_ACTIVATION_SEPARATOR/m.match(stderr))[1]
     version, gem_path, yjit, *fields = T.must(activation_string).split("RUBY_LSP_FS")
 
     assert_equal(RUBY_VERSION, version)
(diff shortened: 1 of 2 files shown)


test/integration_test.rb
```diff
@@ -29,7 +29,7 @@ def test_ruby_lsp_check_works
   def test_activation_script_succeeds_even_on_binary_encoding
     ENV["LC_ALL"] = "C"
     ENV["LANG"] = "C"
-    ENV["PS1"] = "\xE2".b
+    ENV["PS1"] = "\xE2\x96\xB7".b
```
vinistock, line 32: We still want to verify that we're rejecting values that cannot be converted to UTF-8. Can you please duplicate this test so that we have examples for both?

vscode/activation.rb
```diff
@@ -1,3 +1,3 @@
-env = ENV.filter_map { |k, v| "#{k}RUBY_LSP_VS#{v}" if v.dup.force_encoding(Encoding::UTF_8).valid_encoding? }
+env = ENV.filter_map { |k, v| v = v.dup.force_encoding(Encoding::UTF_8); "#{k}RUBY_LSP_VS#{v}" if v.valid_encoding? }
```
vinistock, line 1: I guess we can simplify this and just try to encode straight away and rescue the conversion error.

```suggestion
env = ENV.filter_map do |k, v|
 "#{k}RUBY_LSP_VS#{v.encode(Encoding::UTF_8)}"
rescue Encoding::UndefinedConversionError
  nil
end
```

vinistock (commented):
Thanks for the PR. Let's get this fixed

test/integration_test.rb
mckeed: Done. I'm not sure how the result of this is used by the extension – do we also want this test to assert that valid unicode in ENV is successfully returned from the activation script?

vinistock: I tried creating an end to end test using the extension, but there was no easy way to get the `LANG=C` to apply to the shell invoked by NodeJS when running our activation script.

For now, it's fine to just keep these integration tests.

vscode/activation.rb
vinistock: Interesting. I didn't expect `encode` to raise even when it's possible to force the encoding to UTF-8. You're right, let's go with the original then.

```ruby
env = ENV.filter_map do |k, v|
  utf_8_value = v.dup.force_encoding(Encoding::UTF_8)
  "#{k}RUBY_LSP_VS#{utf_8_value}" if utf_8_value.valid_encoding?
end
```

PR #3220, opened 2025-02-20
Exclude RBS comments from on-type formatting

st0012:
### Motivation

This allows users to start typing method definitions right after typing the signature.

### Implementation

Exclude `#:` from on type completion's comment matching regexp.


lib/ruby_lsp/requests/on_type_formatting.rb
```diff
@@ -60,7 +60,7 @@ def perform
         when "|"
           handle_pipe if @document.syntax_error?
         when "\n"
-          if (comment_match = @previous_line.match(/^#(\s*)/))
+          if (comment_match = @previous_line.match(/^#(?!:)(\s*)/))
```
andyw8, line 63: Let's add a comment here about the RBS syntax?

st0012: Added 👍

test/requests/on_type_formatting_test.rb
```diff
@@ -292,6 +292,32 @@ def test_comment_continuation
...
+      source: +"",
+      version: 1,
+      uri: URI("file:///fake.rb"),
+      global_state: @global_state,
+    )
+
+    document.push_edits(
+      [{
+        range: { start: { line: 0, character: 0 }, end: { line: 0, character: 0 } },
+        text: "    #: (String) -> String",
```
Morriar, line 306: Let's maybe also add an example with a trailing `#:` comments to ensure we don't accidentally change the regexp?

This should not trigger:
```rb
a = 42 #: Integer
```

Morriar (approved):

test/requests/on_type_formatting_test.rb
st0012: Added

st0012:
### Merge activity

* **Feb 20, 5:16 PM EST**: The merge label 'graphite-merge' was detected. This PR will be added to the [Graphite merge queue](https://app.graphite.dev/merges?org=Shopify&repo=ruby-lsp) once it meets the requirements.

PR #3221, opened 2025-02-20
Fix incorrect RBS highlighting in Ruby code

st0012:
### Motivation

The `#:` grammar rule should only be applied to comments, which means the capture group should only match if the `#:` is the first non-whitespace character in the line.

Otherwise, the grammar will incorrectly apply to `#:` in Ruby code as well.

This commit fixes the grammar rule so that it only applies to comments.

#### Before

Screenshot 2025-02-21 at 18 32 08

#### After

Screenshot 2025-02-21 at 18 31 53

You can see that it's still not able to handle the comments inside heredoc. But I think it's the best we can do with just regexp.

### DIFF
--- a/vscode/src/test/suite/grammars.test.ts
+++ b/vscode/src/test/suite/grammars.test.ts
@@ -344,91 +344,27 @@ suite("Grammars", () => {
       test("inline method signature", () => {
         const ruby = "#: (String) -> (String | nil)";
         const expectedTokens = [
-          [
-            "#:",
-            [
-              "rbs-comment.injection",
-              "meta.type.signature.rbs",
-              "comment.line.number-sign.rbs",
-            ],
-          ],
-          [" ", ["rbs-comment.injection", "meta.type.signature.rbs"]],
-          [
-            "(",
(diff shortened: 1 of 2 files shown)


vscode/src/test/suite/grammars.test.ts
```diff
@@ -814,6 +814,39 @@ suite("Grammars", () => {
...
+            [
+              "rbs-comment.injection",
+              "meta.type.signature.rbs",
+              "variable.other.constant.rbs",
+            ],
+          ],
+        ];
+        const actualTokens = tokenizeRBS(ruby);
+        assert.deepStrictEqual(actualTokens, expectedTokens);
+      });
```
Morriar, line 849: Let's also add this (which I'm still not sure if it should be accepted or not):

```rb
# You can use sig comments on methods like this: #: -> void
x = 42 # you can also add inline annotations #: Integer
```

st0012: Updated the grammar rule and tests for these cases.

PR #3223, opened 2025-02-20
Add test discovery support for Minitest spec style syntax

alexcrocha:
### Motivation

This PR addresses and closes #3178 by adding support for Minitest spec style syntax. Previously, our support was limited to the traditional test style syntax. 

By adding support for the spec style, we enhance the discoverability of tests in VS Code's Test Explorer panel, improving the developer experience.

### Implementation

- Implemented a listener to create test items for spec style syntax `describe` and `it` methods
- Ensured compatibility with traditional test syntax (`def test_`)
- Future PR will refactor to reduce code duplication between listeners

### Tests

- Added tests and fixtures

PR #3228, opened 2025-02-21
Add support for `specify` in Code Lens

andyw8:
This came up while pairing with @alexcrocha on https://github.com/Shopify/ruby-lsp/pull/3223

Ref: https://rubydoc.info/gems/minitest/4.1.0/MiniTest/Spec#it-class_method

### DIFF
--- a/lib/ruby_lsp/listeners/code_lens.rb
+++ b/lib/ruby_lsp/listeners/code_lens.rb
@@ -20,8 +20,6 @@ module RubyLsp
       )
       ACCESS_MODIFIERS = T.let([:public, :private, :protected], T::Array[Symbol])
       SUPPORTED_TEST_LIBRARIES = T.let(["minitest", "test-unit"], T::Array[String])
-      DESCRIBE_KEYWORD = T.let(:describe, Symbol)
-      IT_KEYWORD = T.let(:it, Symbol)
       DYNAMIC_REFERENCE_MARKER = T.let("<dynamic_reference>", String)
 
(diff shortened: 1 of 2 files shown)


test/requests/code_lens_expectations_test.rb
```diff
@@ -51,7 +51,7 @@ def test_bar; end
   def test_command_generation_for_minitest_spec
     stub_test_library("minitest")
     source = <<~RUBY
-      class FooTest < MiniTest::Test
+      class FooTest < MiniTest::Spec
```
andyw8, line 54: (noticed this was incorrect)

lib/ruby_lsp/listeners/code_lens.rb
```diff
@@ -151,23 +149,21 @@ def on_call_node_enter(node)
...
-            @group_id += 1
-          when IT_KEYWORD
-            add_spec_code_lens(node, kind: :example)
-          end
+        case name
+        when :describe
+          add_spec_code_lens(node, kind: :group)
+          @group_id_stack.push(@group_id)
+          @group_id += 1
+        when :it, :specify # `specify` is an alias for `it`
```
vinistock, line 157: Do we need the comment? It doesn't add much that the code doesn't tell us already.

andyw8: I think it's helpful. The code shows that we treat them the same, but the the comment explains _why_.

PR #3236, opened 2025-02-24
Fix range and workspace processing for test discovery updates

vinistock:
### Motivation

I noticed a small batch of improvements we need to make to the test controller.

### Implementation

1. We should invalidate test results when the test file is updated
2. We were always passing the first workspace folder, when in fact we should use the one associated with the document being saved
3. We were incorrectly trying to read `column` from the server response, but the field is actually `character`

### DIFF
--- a/vscode/src/testController.ts
+++ b/vscode/src/testController.ts
@@ -609,18 +609,21 @@ export class TestController {
     }
 
     let initialCollection = this.testController.items;
+    let workspaceFolder = workspaceFolders[0];
 
     // If there's more than one workspace folder, then the first level is the workspace
     if (workspaceFolders.length > 1) {
+      workspaceFolder = vscode.workspace.getWorkspaceFolder(uri)!;
+
       initialCollection = initialCollection.get(
-        vscode.workspace.getWorkspaceFolder(uri)!.uri.toString(),
+        workspaceFolder.uri.toString(),
       )!.children;
     }
 
     // There's always a first level, but not always a second level
     const { firstLevelUri, secondLevelUri } = await this.directoryLevelUris(
       uri,
-      workspaceFolders[0],
+      workspaceFolder,
     );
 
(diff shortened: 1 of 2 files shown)


andyw8:
> We should invalidate test results when the test file is updated

Are there any reference implementations of the Test Explorer, and if so, do they do this?

I'm hesitant for a couple of reasons:

* I can imagine a workflow where the developer runs a full test file, then sees one test fail, due to a typo in a test. Then then might prefer to just re-run that single test, and see everything turn green, than re-run everything.
* Changing the implementation could also mean a test is no longer correct. Since we can't easily map back to the test to invalidate it, this introduces a potential inconsistency in behaviour.

vscode/src/testController.ts
```diff
@@ -677,8 +681,8 @@ export class TestController {
       const end = item.range.end;
 
       testItem.range = new vscode.Range(
-        new vscode.Position(start.line, start.column),
-        new vscode.Position(end.line, end.column),
+        new vscode.Position(start.line, start.character),
```
andyw8, line 684: Should these lines also change?
https://github.com/Shopify/ruby-lsp/pull/3236/files#diff-4970e950e007c1318a00bbe05a2e39b92ab1466ebff44ed740a8953a02dc6a4dL144-L145

vinistock: No, because that code is the old implementation that doesn't use the same request.

vinistock:
I don't think there are any concerns in respect to not invalidating the results when the implementation is changed. It's easy to know if the test has been modified, but detecting which tests are impacted by an implementation change is absolutely not trivial. I don't believe it would degrade the experience to invalidate results when we know for sure they have to be invalidated.

We can figure out exactly which test items were updated by diffing the existing children against the new ones, which then prevents the first problem you brought up. Since that is a more involved change, I just removed the invalidation for now to move forward with the fixes.

andyw8 (approved):

PR #3244, opened 2025-02-26
update sample code at addon page to fix uninitialized constant error

tomitahisaki:
### Motivation

it cause uninitialized constant error, when using sample code.
So, I would like to update sample code to resolve error

### Manual Tests

- `bundle exec jekyll serve`
- verify doc where I changed

### DIFF
--- a/jekyll/add-ons.markdown
+++ b/jekyll/add-ons.markdown
@@ -190,7 +190,6 @@ module RubyLsp
     class Addon < ::RubyLsp::Addon
       def activate(global_state, message_queue)
         @message_queue = message_queue
-        @config = SomeConfiguration.new
       end
 
       def deactivate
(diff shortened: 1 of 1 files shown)


tomitahisaki:
I have signed the CLA!

andyw8 (changes requested):
`@config` is to illustrate how configuration can be passed along from the add-on to it requests. To make that clearer, we could define an basic `SomeConfiguration` class as part of the sample.

tomitahisaki:
> `@config` is to illustrate how configuration can be passed along from the add-on to its request handlers. To make that clearer, we could define a basic `SomeConfiguration` class as part of the sample.

Thanks 
I do not understand what `@config` does.
I colse this PR

PR #3246, opened 2025-02-26
Enable RuboCop in `vscode` & `jekyll` directories

sambostock:
### Motivation

The `vscode` and `jekyll` directories are not being checked by RuboCop.

- `vscode/activation.rb` should be checked, as it is Ruby
- Both directories contain Ruby within Markdown code snippets. If we care about linting those in the main directory, then we should also lint them here.

### Implementation

The `Include` config is updated so the additional directories are linted.

This identifies a number of offenses, which have been corrected.

Notably, `rubocop-md` treats Markdown code blocks without a language specified as Ruby by default. We could change this behaviour, but it's clearer to explicitly mark the snippets as `txt` when there is no matching language.

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -38,6 +38,10 @@ RubyLsp/UseLanguageServerAliases:
   Exclude:
     - "test/**/*.rb"
 
+RubyLsp/UseRegisterWithHandlerMethod:
+  Exclude:
+    - "jekyll/add-ons.markdown" # this cop doesn't apply to enhancements
+
 Sorbet/FalseSigil:
   Enabled: false
 


vscode/.rubocop.yml
```diff
@@ -0,0 +1,4 @@
+inherit_from: ../.rubocop.yml
+
+Sorbet:
+  Enabled: false
```
sambostock, line 4: RuboCop was complaining that `vscode/activation.rb` should be

```ruby
# typed: strict
```

which enforces `sig`s on all methods. The file currently has no method definitions, but if it ever gained any, we couldn't attach any `sig`s without `sorbet-runtime`, so `strict` is clearly too high.

We _could_ use

```ruby
# typed: true
```

to simply benefit from type checking, but then Sorbet complains about

```ruby
env = ENV.filter_map do |k, v|
  utf_8_value = v.dup.force_encoding(Encoding::UTF_8)
  "#{k}RUBY_LSP_VS#{utf_8_value}" if utf_8_value.valid_encoding?
end
env.unshift(RUBY_VERSION, Gem.path.join(","), !!defined?(RubyVM::YJIT))
#    Expected `String` but found `T::Boolean` ^^^^^^^^^^^^^^^^^^^^^^^^^ 
$stderr.print("RUBY_LSP_ACTIVATION_SEPARATOR#{env.join("RUBY_LSP_FS")}RUBY_LSP_ACTIVATION_SEPARATOR")
```

because it thinks `env` must be a `T::Array[String]`. We could do

```ruby
env.unshift(RUBY_VERSION, Gem.path.join(","), defined?(RubyVM::YJIT) ? "true" : "false")
```

but technically, we don't _need_ to (interpolating a boolean works fine), so it seemed simplest to just disable the Sorbet cops.

sambostock (commented):
I don't know what to do about

```
jekyll/add-ons.markdown:274:3: C: RubyLsp/UseRegisterWithHandlerMethod: Created a handler without registering the associated on_call_node_enter event.
  def on_call_node_enter(node) ...
  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
jekyll/add-ons.markdown:301:3: C: RubyLsp/UseRegisterWithHandlerMethod: Created a handler without registering the associated on_call_node_leave event.
  def on_call_node_leave(node); end
  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

but it's the repo's own cops failing, so I'm guessing we should be addressing them?

vscode/.rubocop.yml
vinistock: Yeah, all we have under `vscode` for Ruby are the activation scripts. There's no point in adopting Sorbet, let's please keep it disabled.

PR #3254, opened 2025-02-27
Adjust Sorbet/TrueSigil cop config

andyw8:
The current config allows any file in `lib/` to be set as `typed: true` instead of `typed: strict`. It's easy to miss the `typed:` level when reviewing newly added files, so let's change the config so that exclusions need to be explicitly listed under `Sorbet/StrictSigil`.

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -52,7 +52,6 @@ Sorbet/TrueSigil:
     - "lib/ruby_lsp/scripts/compose_bundle.rb"
   Exclude:
     - "**/*.rake"
-    - "lib/**/*.rb"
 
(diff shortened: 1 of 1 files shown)


vinistock:
I'm a bit confused. We have this in our configs:

```yaml
Sorbet/TrueSigil:
  Enabled: true
  Exclude:
    - "lib/**/*.rb"

Sorbet/StrictSigil:
  Enabled: true
  Include:
    - "lib/**/*.rb"
```

Doesn't that mean that strict is enforced in all lib files, but true is not? If I change any sigils to `true` under lib I immediately get an error.

PR #3274, opened 2025-03-05
Move JSON API parsing to test helper

andyw8:
So that it can be shared by the minitest and test/unit reporters.


lib/ruby_lsp/test_helper.rb
```diff
@@ -11,6 +11,7 @@ class TestError < StandardError; end
     extend T::Helpers
 
     requires_ancestor { Kernel }
+    requires_ancestor { Minitest::Test }
```
andyw8, line 14: for `flunk`

lib/ruby_lsp/test_helper.rb
```diff
@@ -76,5 +77,16 @@ def pop_message(outgoing_queue, &block)
       message = outgoing_queue.pop until block.call(message)
       message
     end
+
+    #: (IO output) -> Array[Hash[Symbol, untyped]]
+    def parse_json_api_stream(output)
+      output.each_line("\r\n\r\n").map do |headers|
```
andyw8, line 83: Slight re-write to avoid the intermediate `result`.

lib/ruby_lsp/test_helper.rb
vinistock: This would be tying our test helpers to only Minitest, but people can use whatever framework to write add-ons.

Let's not use `flunk` and instead just raise a specific error.

lib/ruby_lsp/test_helper.rb
```diff
@@ -76,5 +77,16 @@ def pop_message(outgoing_queue, &block)
       message = outgoing_queue.pop until block.call(message)
       message
     end
+
+    #: (IO output) -> Array[Hash[Symbol, untyped]]
```
vinistock, line 81: Not sure how type checking is not failing, but the return type here is not correct.

The hash keys are strings since we're not using `symbolize_names: true`.

lib/ruby_lsp/test_helper.rb
```diff
@@ -76,5 +77,16 @@ def pop_message(outgoing_queue, &block)
...
+    def parse_json_api_stream(output)
```
vinistock, line 82: If we're going to provide this as an API for all add-ons, would it be more helpful if we could instead provide a general API for `read_message` that reads a single JSON RPC message from a given IO object?

andyw8: Maybe. There might be slightly different needs for something used as part of the production code vs something only used by the tests. Let's do this test part first then re-consider?

vinistock: In that case, let's put this in our test helper and not in the API exposed to add-ons.

However, we should be opinionated about how people test add-ons and the closest to how it works in prod the better.

vinistock (approved):

PR #3286, opened 2025-03-10
Capture test output for Minitest reporter

andyw8:


lib/ruby_lsp/test_reporter.rb
```diff
@@ -67,5 +76,28 @@ def send_message(method_name, params)
         $stdout.write("Content-Length: #{json_message.bytesize}\r\n\r\n#{json_message}")
       end
     end
+
+    class IOWrapper < SimpleDelegator
+      #: (Array[String]) -> void
```
vinistock, line 81: Splats shouldn't include the `Array`. You only declare the expected type of the items ([docs](https://sorbet.org/docs/sigs#rest-parameters))
```suggestion
      #: (String) -> void
```

lib/ruby_lsp/test_reporter.rb
```diff
@@ -67,5 +76,28 @@ def send_message(method_name, params)
...
+      def puts(*args)
+        args.each { |arg| log("#{arg}\n") }
+      end
+
+      #: (Array[String]) -> void
```
vinistock, line 86:
```suggestion
      #: (String) -> void
```

lib/ruby_lsp/test_reporter.rb
```diff
@@ -67,5 +76,28 @@ def send_message(method_name, params)
...
+      def log(message)
+        TestReporter.append_output(message: message)
+      end
+    end
   end
 end
+
+# We wrap the default output stream so that we can capture anything written to stdout and emit it as part of
+# the JSON event stream.
+$> = RubyLsp::TestReporter::IOWrapper.new($stdout)
```
vinistock, line 103: We still need to keep a handle for the original stdout and replace it with the same wrapper.

lib/ruby_lsp/test_reporter.rb
```diff
@@ -67,5 +76,28 @@ def send_message(method_name, params)
         $stdout.write("Content-Length: #{json_message.bytesize}\r\n\r\n#{json_message}")
       end
     end
+
+    class IOWrapper < SimpleDelegator
+      #: (String) -> void
+      def puts(*args)
+        args.each { |arg| log("#{arg}\n") }
```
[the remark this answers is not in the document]
andyw8: Updated.

vinistock (approved):

Issue #3284, opened 2025-03-10
Intermittent CI failure for `repeated rebase steps don't trigger multiple restarts`


andyw8:
I've seen this fail on a couple of jobs:

```
  1) Workspace
       repeated rebase steps don't trigger multiple restarts:

      AssertionError [ERR_ASSERTION]: Expected values to be strictly equal:

0 !== 1

      + expected - actual

      -0
      +1
```
e.g. https://github.com/Shopify/ruby-lsp/actions/runs/13767768592/job/38498470626?pr=3283

PR #3287, opened 2025-03-10
Increase timeout for rebase test

andyw8:
Trying to solve https://github.com/Shopify/ruby-lsp/issues/3284


vinistock (approved):
Let's give it a try, but I'm not super confident it's just a matter of time

PR #3288, opened 2025-03-10
Extract shared behaviour of test discovery listeners

### LINKED ISSUE #3241: Extract common parts of `TestStyle` and `SpecStyle` into a shared parent class/module
There's a certain amount of shared behaviour between our `TestStyle` and `SpecStyle` listeners. For example, both need to keep track of nesting, handle dynamic references and so on.

Let's see if we can extract some of the shared behaviour either into a parent class or a module, so that it can be reused by both.

alexcrocha:
### Motivation

Closes #3241 

There's a certain amount of shared behaviour between the `TestStyle` and `SpecStyle` listeners. Both need to track class/module nesting, handle dynamic references, manage visibility stacks, and process node enter/leave events. By extracting this common functionality into a shared parent class, we can reduce code duplication and improve maintainability.

### Implementation

This PR introduces a new abstract `DiscoverTests` class that both `TestStyle` and `SpecStyle` inherit from. The parent class handles:

- Tracking class/module nesting
- Visibility stack management
- Node enter/leave handlers for classes and modules
- Dynamic reference handling
- Ancestry resolution
- Fully qualified name calculation

The specialized test discovery logic remains in the child classes, while common state tracking and node handling is centralized in the parent. 

### Automated Tests

Do we need to add more automated tests or do the existing tests suffice?

### DIFF
--- a/lib/ruby_lsp/listeners/spec_style.rb
+++ b/lib/ruby_lsp/listeners/spec_style.rb
@@ -3,7 +3,7 @@
 
 module RubyLsp
   module Listeners
-    class SpecStyle < DiscoverTests
+    class SpecStyle < TestDiscovery
       extend T::Sig
 
       #: (response_builder: ResponseBuilders::TestCollection, global_state: GlobalState, dispatcher: Prism::Dispatcher, uri: URI::Generic) -> void
--- a/lib/ruby_lsp/listeners/test_discovery.rb
+++ b/lib/ruby_lsp/listeners/test_discovery.rb
@@ -3,7 +3,7 @@
 
 module RubyLsp
   module Listeners
-    class DiscoverTests
+    class TestDiscovery
       extend T::Helpers
       abstract!
 
--- a/lib/ruby_lsp/listeners/test_style.rb
+++ b/lib/ruby_lsp/listeners/test_style.rb
@@ -3,7 +3,7 @@
 
 module RubyLsp
   module Listeners
-    class TestStyle < DiscoverTests
+    class TestStyle < TestDiscovery
       class << self
         # Resolves the minimal set of commands required to execute the requested tests
         #: (Array[Hash[Symbol, untyped]]) -> Array[String]
@@ -78,7 +78,7 @@ module RubyLsp
           regexes = groups_and_examples.flat_map do |group, info|
             examples = info[:examples]
             group_regex = Shellwords.escape(group).gsub(
-              Shellwords.escape(DiscoverTests::DYNAMIC_REFERENCE_MARKER),
+              Shellwords.escape(TestDiscovery::DYNAMIC_REFERENCE_MARKER),
               ".*",
             )
             if examples.empty?
@@ -104,7 +104,7 @@ module RubyLsp
           groups_and_examples.map do |group, info|
             examples = info[:examples]
             group_regex = Shellwords.escape(group).gsub(
-              Shellwords.escape(DiscoverTests::DYNAMIC_REFERENCE_MARKER),
+              Shellwords.escape(TestDiscovery::DYNAMIC_REFERENCE_MARKER),
               ".*",
             )
             command = +"#{BASE_COMMAND} -Itest #{file_path} --testcase \"/^#{group_regex}$/\""
--- a/lib/ruby_lsp/requests/discover_tests.rb
+++ b/lib/ruby_lsp/requests/discover_tests.rb
@@ -1,7 +1,7 @@
 # typed: strict
 # frozen_string_literal: true
 
-require "ruby_lsp/listeners/discover_tests"
+require "ruby_lsp/listeners/test_discovery"
 require "ruby_lsp/listeners/test_style"
 require "ruby_lsp/listeners/spec_style"
 


lib/ruby_lsp/listeners/discover_tests.rb
```diff
@@ -0,0 +1,87 @@
+# typed: strict
+# frozen_string_literal: true
+
+module RubyLsp
+  module Listeners
+    class DiscoverTests
+      extend T::Sig
```
andyw8, line 7: We don't need this any more with the RBS syntax.

lib/ruby_lsp/listeners/discover_tests.rb
```diff
@@ -0,0 +1,89 @@
+# typed: strict
+# frozen_string_literal: true
+
+module RubyLsp
+  module Listeners
+    class DiscoverTests
+      extend T::Helpers
+      abstract!
+
+      include Requests::Support::Common
```
andyw8, line 10: Since its now in the parent, we can remove this `include` from the child classes.

lib/ruby_lsp/listeners/discover_tests.rb
```diff
@@ -0,0 +1,83 @@
...
+
+      DYNAMIC_REFERENCE_MARKER = "<dynamic_reference>"
```
andyw8, line 12: If https://github.com/Shopify/ruby-lsp/pull/3298 is merged then we can remove this.

lib/ruby_lsp/listeners/discover_tests.rb
```diff
@@ -0,0 +1,83 @@
...
+
+      #: (ResponseBuilders::TestCollection response_builder, GlobalState global_state, Prism::Dispatcher dispatcher, URI::Generic uri) -> void
+      def initialize(response_builder, global_state, dispatcher, uri)
+        @response_builder = response_builder
+        @uri = uri
+        @index = T.let(global_state.index, RubyIndexer::Index)
+        @visibility_stack = T.let([:public], T::Array[Symbol])
+        @nesting = T.let([], T::Array[String])
+
+        @fully_qualified_name = T.let("", String)
```
andyw8, line 22: It looks like `@fully_qualified_name` is only used for `TestStyle`, can we push it down to there?

lib/ruby_lsp/listeners/spec_style.rb
```diff
@@ -3,75 +3,41 @@
...
         @describe_block_nesting = T.let([], T::Array[String])
         @spec_class_stack = T.let([], T::Array[T::Boolean])
 
         dispatcher.register(
           self,
           :on_class_node_enter,
           :on_class_node_leave,
-          :on_module_node_enter,
-          :on_module_node_leave,
+          :on_module_node_enter, # rubocop:disable RubyLsp/UseRegisterWithHandlerMethod
```
andyw8, line 22: Can't we remove these lines?

lib/ruby_lsp/listeners/discover_tests.rb
```diff
@@ -0,0 +1,83 @@
...
+        @attached_ancestors = T.let([], T::Array[String])
+      end
+
+      #: (Prism::ClassNode node) -> void
+      def on_class_node_enter(node) # rubocop:disable RubyLsp/UseRegisterWithHandlerMethod
```
andyw8, line 27: (the need to disable RoboCop shows that this cop is far from ideal. Hopefully we can merge https://github.com/Shopify/ruby-lsp/pull/3291 and eliminate it).

lib/ruby_lsp/listeners/discover_tests.rb
andyw8: (and also `calc_fully_qualified_name`).

andyw8: Copilot claims a few others but I didn't confirm:

Screenshot 2025-03-12 at 1 49 14 PM

lib/ruby_lsp/listeners/spec_style.rb
andyw8: (if not, we can just have a method like `callbacks_to_register` in each subclass, which the parent initializer can call)

alexcrocha: It sounds like a better approach. I will give this a try.

alexcrocha: > could we registere the common ones in the parent class, and then the more specific ones in the subclasses?

This is working now. 

I added a comment in the subclasses to highlight there are more handlers registered in the parent class. But maybe that comment is not necessary and I just need to get comfortable with it.

lib/ruby_lsp/listeners/spec_style.rb
```diff
@@ -3,75 +3,36 @@
...
         @describe_block_nesting = T.let([], T::Array[String])
         @spec_class_stack = T.let([], T::Array[T::Boolean])
 
         dispatcher.register(
           self,
-          :on_class_node_enter,
-          :on_class_node_leave,
-          :on_module_node_enter,
-          :on_module_node_leave,
+          # Common handlers registered in parent class
```
alexcrocha, line 18: Is this comment helpful, or unnecessary?

lib/ruby_lsp/listeners/discover_tests.rb
```diff
@@ -0,0 +1,91 @@
+# typed: strict
+# frozen_string_literal: true
+
+module RubyLsp
+  module Listeners
+    class DiscoverTests
```
vinistock, line 6: I wonder if this could potentially be confusing since it's the same name of the request. That said, I don't actually have any suggestions for a better name.

lib/ruby_lsp/listeners/discover_tests.rb
```diff
@@ -0,0 +1,91 @@
...
+          self,
+          :on_class_node_enter,
+          :on_class_node_leave,
+          :on_module_node_enter,
+          :on_module_node_leave,
+        )
+      end
+
+      #: (Prism::ClassNode node) -> void
+      def on_class_node_enter(node)
```
vinistock, line 35: I like using the events for module_enter/leave and class_leave. That means that the child class doesn't even need to define those, which is very nice.

For `class_enter`, I wonder if it would be better to just expose a help method since each listeners has slight variations on how they work. That would avoid you needing to remember the ancestors as part of the state for example.

It would also avoid the awkwardness of defining the `on_class_node_enter` event without registering it because the parent already does it.

Something like this:

```ruby
class Parent
  def with_test_ancestor_tracking(node, &block)
    @visibility_stack << :public
    name = constant_name(node.constant_path)
    name ||= name_with_dynamic_reference(node.constant_path)

    fully_qualified_name = calc_fully_qualified_name(name)
    attached_ancestors = calc_attached_ancestors(node, fully_qualified_name)

    # Let the caller decide how to handle the
    # fully qualified name + linearized ancestors
    block.call(fully_qualified_name, attached_ancestors)

    @nesting << name
  end
end

class Child < Parent
  def on_class_node_enter(node)
    with_test_ancestor_tracking(node) do |name, ancestors|
      # add test item if desired
      # set some other state if desired
    end
  end
end
```

What do you think?

vinistock (commented):
This is looking good. I think by making the class handling a helper method, we can get an even cleaner implementation and then ship 🚢

lib/ruby_lsp/listeners/discover_tests.rb
alexcrocha: > What do you think?

It wasn't very clear to me at first, but after implementing it I can see how this is a much nicer solution.

lib/ruby_lsp/listeners/discover_tests.rb
alexcrocha: What are your thoughts on renaming it to `TestDiscovery`? (can revert the commit if it's a no)

PR #3289, opened 2025-03-10
Ensure `URI.from_path` is available to minitest runner

andyw8:
When starting to verify this with Rails, I discovered the we need to ensure `URI.from_path` is available (it's already there for test-unit), since we use it in `uri_from_test_class`.

### DIFF
--- a/lib/ruby_lsp/ruby_lsp_reporter_plugin.rb
+++ b/lib/ruby_lsp/ruby_lsp_reporter_plugin.rb
@@ -2,6 +2,7 @@
 # frozen_string_literal: true
 
 require_relative "test_reporter"
+require "ruby_indexer/lib/ruby_indexer/uri"
 
 require "minitest"
 
--- a/project-words
+++ b/project-words
@@ -74,6 +74,7 @@ quickfixes
 quux
 quxx
 rdbg
+rbundler
 readlines
 realpath
 reparsing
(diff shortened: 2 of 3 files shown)


andyw8:
I modified the test to show the underlying failure:

```
Failure:
RubyLsp::MinitestTestRunnerTest#test_minitest_output [/home/runner/work/ruby-lsp/ruby-lsp/test/minitest_test_runner_test.rb:19]
Minitest::Assertion: command failed: <internal:/opt/hostedtoolcache/Ruby/3.1.6/x64/lib/ruby/3.1.0/rubygems/core_ext/kernel_require.rb>:85:in `require': cannot load such file -- ruby_indexer/lib/ruby_indexer/uri (LoadError)
	from <internal:/opt/hostedtoolcache/Ruby/3.1.6/x64/lib/ruby/3.1.0/rubygems/core_ext/kernel_require.rb>:85:in `require'
	from /home/runner/work/ruby-lsp/ruby-lsp/lib/ruby_lsp/ruby_lsp_reporter_plugin.rb:5:in `<top (required)>'
	from <internal:/opt/hostedtoolcache/Ruby/3.1.6/x64/lib/ruby/3.1.0/rubygems/core_ext/kernel_require.rb>:85:in `require'
	from <internal:/opt/hostedtoolcache/Ruby/3.1.6/x64/lib/ruby/3.1.0/rubygems/core_ext/kernel_require.rb>:85:in `require'
```

It seems there is some issue with the load path. Since it only affects a Ruby version which is EOL in 2 weeks, and is only in our test code, I'm not to spend much time on this. The latest commit works around it.

PR #3292, opened 2025-03-11
Add support for ERB syntax highlighting in Heredoc sections

### LINKED ISSUE #2863: Syntax highlight for ERB in heredoc

### Use case

I'm writing a lot of my ViewComponent with inline `erb` as heredoc and it would be great if the syntax highlight works for it as well.

### Description

Based on the current grammar, using `ERB` in heredoc won't get syntax highlighted like the other supported language in this [file](https://github.com/Shopify/ruby-lsp/blob/9e24cff30b825855254aeaebbce025ac104d60e4/vscode/grammars/ruby.cson.json).

I tried with `HTML` and it worked only with the non Ruby part.

It's quite odd that there are support for JS, GRAPHQL, and other languages but not ERB.

### Implementation

We only need to add a configuration for ERB to the grammar [file](https://github.com/Shopify/ruby-lsp/blob/9e24cff30b825855254aeaebbce025ac104d60e4/vscode/grammars/ruby.cson.json).

domingo2000:
### Motivation

Closes #2863

### Implementation

This is a follow up of this PR #2903

Added `.erb` grammar to the vscode grammar extension.

### Automated Tests

Added test for erb grammar in the grammars.test.ts

### Manual Tests

I run the extension locally. The grammar is taking the ERB block in the heredoc. But for some reason, reusing the `text.html.erb` grammar of the RubyLSP extension gives different syntax highlighting than in a example.erb file.

`example.rb`
image
`example.erb`
image

### DIFF
--- a/vscode/grammars/ruby.cson.json
+++ b/vscode/grammars/ruby.cson.json
@@ -1692,6 +1692,44 @@
       ]
     },
     {
+      "begin": "(?=(?><<[-~]([\"'`]?)((?:[_\\w]+_|)ERB)\\b\\1))",
+      "comment": "Heredoc with embedded ERB",
+      "end": "(?!\\G)",
+      "name": "meta.embedded.block.erb",
+      "patterns": [
+        {
+          "begin": "(?><<[-~]([\"'`]?)((?:[_\\w]+_|)ERB)\\b\\1)",
+          "beginCaptures": {
+            "0": {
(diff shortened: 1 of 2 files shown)


vinistock:
Yeah, unfortunately there's something still weird with the ERB blocks. They are not being highlighted as expected.

PR #3303, opened 2025-03-16
Retain editor focus after Run In Terminal

andyw8:
I discovered that you can trigger a Code Lens with a keybinding, which is useful for Run In Terminal.

But currently the focus stays with the terminal so you need to focus back to the editor after doing so. This PR changes it to switch back automatically.

This does impact the current flow. For example, if someone is used to triggering a breakpoint with `binding.b`, they will now need to focus on the Terminal panel before doing anything. I expect the majority of uses for Run In Terminal are likely not doing that though, so overall it should be an improvement. 

(Note: There is another change needed after this for the keyboard functionality to fully work).

### DIFF
--- a/vscode/src/testController.ts
+++ b/vscode/src/testController.ts
@@ -184,8 +184,18 @@ export class TestController {
       this.terminal = this.getTerminal();
     }
 
+    const activeEditor = vscode.window.activeTextEditor;
     this.terminal.show();
     this.terminal.sendText(command);
+    if (activeEditor) {
+      setTimeout(() => {
+        vscode.window.showTextDocument(
+          activeEditor.document,
+          activeEditor.viewColumn,
+          false,
+        );
+      }, 100);
+    }
 
     this.telemetry.logUsage("ruby_lsp.code_lens", {
       type: "counter",


vscode/src/testController.ts
```diff
@@ -184,8 +184,18 @@ export class TestController {
       this.terminal = this.getTerminal();
     }
 
+    const activeEditor = vscode.window.activeTextEditor;
     this.terminal.show();
     this.terminal.sendText(command);
+    if (activeEditor) {
+      setTimeout(() => {
```
andyw8, line 191: Without a timeout here it fails. I'm unsure what would be a better way.

st0012:
Do you mind recording demo vids for before/after comparison? I ran this branch but the behaviour seems to be the same (focus moves and stays at terminal).

vscode/src/testController.ts
vinistock: The `show` method on `Terminal` supports a `preserveFocus` argument https://code.visualstudio.com/api/references/vscode-api#Terminal.

Let's pass the argument to `show` and get rid of the rest.

andyw8: Updated

vinistock: Did you push the changes? I'm still seeing the old code with `setTimeout`.

andyw8: That's odd, it's showing correctly on my branch here:

https://github.com/andyw8/ruby-lsp/tree/andyw8/retain-editor-focus

vinistock: Yes, but the commits are different for some reason. That branch is on 2e0331d while this PR is on 14026c6.

andyw8:
Re-opened as https://github.com/Shopify/ruby-lsp/pull/3679

PR #3321, opened 2025-03-19
Use map compact instead of filter_map in activation script

vinistock:
### Motivation

The activation script is failing when no Shadowenv environment is activated because it then falls back to system Ruby (2.6) and `filter_map` doesn't exist in it.

There are better long term solutions to this, like trying to fallback to chruby when Shadowenv fails or offering other options, but for now I would really like to remove these errors from telemetry since we finally managed to fix the encoding related problems too.

### Implementation

Switched to map/compact instead of `filter_map`.

### DIFF
--- a/vscode/activation.rb
+++ b/vscode/activation.rb
@@ -1,6 +1,7 @@
-env = ENV.filter_map do |k, v|
+# Using .map.compact just so that it doesn't crash immediately on Ruby 2.6
+env = ENV.map do |k, v|
   utf_8_value = v.dup.force_encoding(Encoding::UTF_8)
   "#{k}RUBY_LSP_VS#{utf_8_value}" if utf_8_value.valid_encoding?
-end
+end.compact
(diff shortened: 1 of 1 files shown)


andyw8 (approved):
Could we do something in CI to run this against Ruby 2.6 to prevent regressions in future?

(alternatively, we could add a comment warning to avoid anything not available in Ruby 2.6)

vinistock:
Ideally, I think we'd go with the fallback I mentioned, using an alternative version manager when we hit this case. You can't use the Ruby LSP with Ruby 2.6 anyway so this script succeeding makes no functional difference for the user.

It would be better if we could do something like

1. Try Shadowenv. It fails because it didn't find any rubies and landed on the system one
2. If we detect the user has chruby, then we re-activate using it

Ultimately, this is a result of Shadowenv not allowing for fallbacks in its configuration. We already provide users with manual fallback configurations, but maybe we can remove more friction by adding an automatic fallback attempt.

PR #3326, opened 2025-03-19
Introduce test file watcher

### LINKED ISSUE #3174: Add created and deleted file watcher to ensure items are properly updated
If test files are created or deleted, we need the VS Code extension to update the test explorer collection automatically.

Let's add a file watcher for created or deleted using the same glob pattern, so that we can maintain the list up to date.

alexcrocha:
### Motivation

Closes #3174 

Currently, when test files are created or deleted, the test explorer doesn't automatically update to reflect these changes, requiring manual intervention or workspace reload to see the updates.

We need to make sure that, when there are changes to our files, the test explorer is aware of these changes automatically.

### Implementation

Leveraged file system watchers to monitor test file changes:

- Uses the existing glob pattern `**/{test,spec,features}/**/{*_test.rb,test_*.rb,*_spec.rb,*.feature}`
- Handles four events:
  - `testFileWatcher.onDidCreate`: Adds new test files, nested appropriately, in the test explorer
  - `testFileWatcher.onDidChange`: Refreshes test items when a test file is modified
  - `testFileWatcher.onDidDelete`: Removes deleted test files from the test explorer
  - `nestedTestDirWatcher.onDidDelete`: Removes deleted test directories from the test explorer

Refactored the test hierarchy detection logic:
  - Extracted `detectHierarchyLevels` to analyze and return hierarchy information without modifying collections
  - Added `getOrCreateHierarchyLevels` to ensure test items exist based on the detected hierarchy
  - Added `addTestFileItem`: Handles individual test file item creation
  - Added `shouldSkipTestFile` to filter out helper files and fixtures

Open to suggestions on method naming 😄

### DIFF
--- a/vscode/src/testController.ts
+++ b/vscode/src/testController.ts
@@ -121,6 +121,7 @@ export class TestController {
         if (terminal === this.terminal) this.terminal = undefined;
       }),
       testFileWatcher,
+      nestedTestDirWatcher,
       testFileWatcher.onDidCreate(async (uri) => {
         const workspace = vscode.workspace.getWorkspaceFolder(uri);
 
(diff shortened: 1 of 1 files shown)


vscode/src/testController.ts
```diff
@@ -121,6 +152,24 @@ export class TestController {
           }
         }
       }),
+      nestedTestDirWatcher.onDidDelete(async (uri) => {
+        if (uri.path.includes("/.git/")) {
```
vinistock, line 156: This may not work on Windows because it could use backslashes to represent paths. It's safer to first split the path into directories and then check for inclusion.

Also, always use `fsPath` instead of `path`.

```suggestion
        const pathParts = uri.fsPath.split(path.sep);
        if (pathParts.includes(".git")) {
```

vinistock (approved):
Great work

PR #3344, opened 2025-03-22
Improve support for hovering over a gem in Gemfile

andyw8:
### Motivation

Currently you can hover over an entry in the `Gemfile` to reveal information about a gem, but this works only for the `gem` call and not the string containing the gem name.

### Implementation

Added behaviour to the `on_string_node_enter` callback.

### Automated Tests

Included

### Manual Tests

Can be verified with Ruby LSP's own Gemfile, using this branch.

### DIFF
--- a/test/requests/hover_expectations_test.rb
+++ b/test/requests/hover_expectations_test.rb
@@ -273,13 +273,12 @@ class HoverExpectationsTest < ExpectationsTestRunner
     end
   end
 
-  def test_hovering_over_gemfile_dependency
+  def test_hovering_over_gemfile_dependency_using_gem_call
     source = <<~RUBY
       gem 'rake'
     RUBY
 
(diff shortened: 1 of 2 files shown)


test/requests/hover_expectations_test.rb
```diff
@@ -273,13 +273,12 @@ class A
     end
   end
 
-  def test_hovering_over_gemfile_dependency
+  def test_hovering_over_gemfile_dependency_using_gem_call
     source = <<~RUBY
       gem 'rake'
     RUBY
 
-    # We need to pretend that Sorbet is not a dependency or else we can't properly test
```
andyw8, line 281: Not needed, I think these were leftover from a copy and paste.

vinistock (approved):
Nice. Thanks!

andyw8:
np, I've rebased.

PR #3349, opened 2025-03-24
Auto resolve first test file of each top level directory

vinistock:
### Motivation

We identified a challenge when resolving test commands for entire test directories where not every file has been expanded. How do we know what add-on should resolve the command if we haven't discovered the examples inside?

For example, if an application has a `spec` directory, how do we know if we're dealing with RSpec or Minitest spec style? They require different commands to execute the tests inside the directory.

One possible solution would be to discover all tests inside the requested directory, but that would be very slow. A decent compromise proposed by @st0012 was to resolve only 1 test file inside a top level directory and then assume that all tests inside of it belong to the same framework. I think this is reasonable given that mixing test frameworks inside the same directory is not a very common usecase.

This PR starts auto resolving 1 of the test files discovered inside each top level test directory and then associates everything nested under that directory with the same framework tag, so that add-ons can resolve commands only for the items they are interested in.

### Implementation

The idea is to keep track of the framework tag during test file discovery. Whenever we find a new top level test dir, we reset the tag to discover it again.

Then we resolve the specific test file, so that the server can return what framework that file is associated with, which we associate to all items under that directory.

### Automated Tests

There were some duplicate examples in our tests and quite a bit of duplicate code. I cleaned them up a little bit and adjusted some expectations.

### DIFF
--- a/vscode/src/testController.ts
+++ b/vscode/src/testController.ts
@@ -818,9 +818,43 @@ export class TestController {
   ) {
     const initialCollection = item ? item.children : this.testController.items;
     const pattern = this.testPattern(workspaceFolder);
+    let frameworkTag: vscode.TestTag | undefined;
+    let previousFirstLevel: vscode.TestItem | undefined;
 
     for (const uri of await vscode.workspace.findFiles(pattern)) {
-      await this.addTestItemsForFile(uri, workspaceFolder, initialCollection);
+      const itemParts = await this.addTestItemsForFile(
+        uri,
+        workspaceFolder,
+        initialCollection,
+      );
+
+      if (!itemParts) {
+        continue;
+      }
+
+      const { firstLevel, secondLevel, testItem } = itemParts;
+
+      if (!previousFirstLevel || previousFirstLevel.id !== firstLevel.id) {
+        await this.resolveHandler(testItem);
+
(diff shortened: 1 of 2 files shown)


vinistock:
* **#3349** <a href="https://app.graphite.dev/github/pr/Shopify/ruby-lsp/3349?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/Shopify/ruby-lsp/3349?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`
<hr />

You must have a Graphite account in order to use the merge queue. Sign up using [this link](https://app.graphite.dev/invite/github/Shopify?ref=merge-queue-instructions-in-stack-comment&prId=3349).

<h2></h2>
<sub>

</sub>

<sub>

</sub>

<h2></h2>

vscode/src/testController.ts
```diff
@@ -900,6 +923,22 @@ export class TestController {
         this.addDiscoveredItems(item.children, testItem);
       }
     });
+
+    if (!testItems[0]) {
+      return;
```
st0012, line 928: Should this check be placed even before we try to loop through items?

st0012 (approved):

PR #3351, opened 2025-03-24
Allow add-ons to resolve test commands

### LINKED ISSUE #3177: Ensure that the extension is getting the command to run from the new custom request and that it is populating the results based on the JSON output

vinistock:
### Motivation

Closes #3177

Add-ons need to be able to both discover test items and resolve the commands needed to execute them. This PR adds support for add-ons to resolve commands.

### Implementation

The idea is to have the add-ons only handle test items that were tagged by them, ignoring everything else. In this scenario, we should be able to simply concatenate all responses and avoid conflicts.

### Automated Tests

Added tests.

### DIFF
--- a/test/requests/resolve_test_commands_test.rb
+++ b/test/requests/resolve_test_commands_test.rb
@@ -1132,5 +1132,64 @@ module RubyLsp
         assert_empty(result[:commands])
       end
     end
+
+    def test_addons_can_support_other_frameworks
+      Class.new(RubyLsp::Addon) do
+        def activate(global_state, outgoing_queue); end
+
+        def name
+          "Rails add-on"
+        end
+
+        def deactivate; end
+
+        def version
+          "0.1.0"
+        end
+
+        def resolve_test_commands(items)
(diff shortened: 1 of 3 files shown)


alexcrocha (approved):

test/requests/resolve_test_commands_test.rb
```diff
@@ -1132,5 +1132,64 @@ def test_ignores_items_tagged_by_other_frameworks
...
+        end
+
+        def deactivate; end
+
+        def version
+          "0.1.0"
+        end
+
+        def resolve_test_commands(items)
+          example = items.dig(0, :children, 0)
```
st0012, line 1151: Are addons expected to go through all items and all their children (and their children) to filter out unrelated items, and then generate commands from the remaining ones?

vinistock: Yes. However, after #3349, the top level items will already be tagged with the framework they are related to, so add-ons can exclude them before going through the children.

egiurleo (approved):

PR #3354, opened 2025-03-25
Implement debug handler based on command resolution

### LINKED ISSUE #3311: Review debug profile implementation in the test explorer
The implementation of the debug profile might still be tied to the old behaviour of the test explorer. Double-check if this is the case and migrate its implementation to our new approach.

Implementation notes:

- Check if it's possible to allow for debugging an arbitrary collection of tests. This would involved asking the server for which commands to run and then delegating to the debugger to run them. I _believe_ this should work
- If the former doesn't work properly, we can limit that UI debugging is only available for specific examples

vinistock:
### Motivation

Closes #3311

This PR implements a new handler for the debug profile based on the test command resolution mechanism we created.

### Implementation

Which profile is triggering the execution of the test explorer is included as part of the request parameter. This allows us to share the logic all the way after resolving test commands.

After that, we need to launch the debugger with the given commands instead of running them in the background like we do for the run profile.

### Note

We unfortunately cannot use our JSON reporters to update the explorer's UI in this case, because the debugger needs to use the stdio pipes to communicate with the editor. We can't try to create a JSON RPC connection with the underlying process as that breaks the debug adapter.

### Automated Tests

Added a test.

### Manual Tests

1. Launch the extension on this branch
2. Open the test explorer
3. Run some tests in debug mode
4. Verify the debugger is launched

### DIFF
--- a/vscode/src/client.ts
+++ b/vscode/src/client.ts
@@ -342,6 +342,11 @@ class ExperimentalCapabilities implements StaticFeature {
   clear(): void {}
 }
 
+export interface ResolvedCommands {
+  commands: string[];
+  reporterPaths: string[] | undefined;
+}
+
 export default class Client extends LanguageClient implements ClientInterface {
   public readonly ruby: Ruby;
   public serverVersion?: string;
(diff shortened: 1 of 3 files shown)


vinistock:
That's a great idea. I created https://github.com/Shopify/ruby-lsp/issues/3359 so that we can explore that in more detail. It will impact all of the communication with the explorer panel, so I will still ship this PR as is.

I hope we can indeed use sockets or a TCP server to communicate results, because then the debugging experience will be fully complete.

PR #3356, opened 2025-03-26
Extract long conditions in `CodeLens#generate_test_command` into methods

st0012:
I think this extraction makes the code easier to read and change in the future.


st0012:
@vinistock Yeah it's fine. This is just something LLM proposed during my pairing with @alexcrocha  that I thought was fine to upstream too 😂

Issue #1800, opened 2024-03-19
Add support for test coverage using VS Code's new API


vinistock:
VS Code is looking to stabilize their new test coverage API in the March release https://github.com/microsoft/vscode/issues/123713. We can use the new API to provide test coverage for Ruby projects.

Docs https://code.visualstudio.com/api/extension-guides/testing#test-coverage.

- [ ] Add docs about how to use the coverage test profile

andyw8:
There is also an experimental Ruby API for coverage: https://rubyapi.org/o/coverage

PR #3360, opened 2025-03-27
Add coverage data collection to the test reporter

vinistock:
### Motivation

First step for #1800

This PR adds the capability to our test reporter to collect coverage using Ruby's built-in coverage API.

### Implementation

The coverage API is pretty convenient. You invoke `Coverage.start` to begin collecting and then `Coverage.result` returns the data.

The implementation is essentially just that. The rest is simply transforming the coverage data that comes from the Ruby API into the format that VS Code expects, so that we don't need any handling on the extension side.

### Automated Tests

Added a test.

### DIFF
--- a/lib/ruby_lsp/test_reporter.rb
+++ b/lib/ruby_lsp/test_reporter.rb
@@ -67,6 +67,85 @@ module RubyLsp
         send_message("append_output", params)
       end
 
+      # Gather the results returned by Coverage.result and format like the VS Code test explorer expects
+      #
+      # Coverage result format:
+      #
+      # Lines are reported in order as an array where each number is the number of times it was executed. For example,
+      # the following says that line 0 was executed 1 time and line 1 executed 3 times: [1, 3].
+      # Nil values represent lines for which coverage is not available, like empty lines, comments or keywords like
+      # `else`
+      #
+      # Branches are a hash containing the name of the branch and the location where it is found in tuples with the
+      # following elements: [NAME, ID, START_LINE, START_COLUMN, END_LINE, END_COLUMN] as the keys and the value is the
+      # number of times it was executed
+      #
+      # Methods are a similar hash [ClassName, :method_name, START_LINE, START_COLUMN, END_LINE, END_COLUMN] => NUMBER
+      # OF EXECUTIONS
+      #
+      # Example:
+      # {
+      #   "file_path" => {
+      #     "lines" => [1, 2, 3, nil],
+      #     "branches" => {
+      #       ["&.", 0, 6, 21, 6, 65] => { [:then, 1, 6, 21, 6, 65] => 0, [:else, 5, 7, 0, 7, 87] => 1 }
+      #     },
+      #     "methods" => {
+      #       ["Foo", :bar, 6, 21, 6, 65] => 0
+      #     }
+      # }
+      #: () -> Hash[String, StatementCoverage]
+      def gather_coverage_results
+        # Ignore coverage results inside dependencies
+        bundle_path = Bundler.bundle_path.to_s
+        default_gems_path = File.dirname(RbConfig::CONFIG["rubylibdir"])
+
+        result = Coverage.result.reject do |file_path, _coverage_info|
+          file_path.start_with?(bundle_path) ||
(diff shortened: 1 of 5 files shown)


lib/ruby_lsp/test_reporter.rb
```diff
@@ -67,6 +67,80 @@ def append_output(message:)
...
+      #     "methods" => {
+      #       ["Foo", :bar, 6, 21, 6, 65] => 0
+      #     }
+      # }
+      #: () -> Hash[String, untyped]
+      def gather_coverage_results
+        # Ignore coverage results inside dependencies
+        result = Coverage.result.reject do |file_path, _coverage_info|
+          file_path.start_with?(Bundler.bundle_path.to_s) ||
+            file_path.start_with?(File.dirname(RbConfig::CONFIG["rubylibdir"])) ||
```
amomchilov, line 102: Is it worth hoisting this `File.dirname(RbConfig::CONFIG["rubylibdir"])` rather than recalculating for each file?

lib/ruby_lsp/test_reporter.rb
```diff
@@ -67,6 +67,80 @@ def append_output(message:)
...
+                groupingLine: branch[2],
+                executed: execution_count,
+                location: {
+                  start: { line: start_line, character: start_column },
+                  end: { line: end_line, character: end_column },
+                },
+                label: "#{branch_name} #{name}",
+              }
+            end
+          end.group_by { |branch| branch[:groupingLine] }
```
amomchilov, line 126: Can you replace the `to_h` above with this `group_by`?

```suggestion
        result.to_h do |file_path, coverage_info|
          # Format the branch coverage information as VS Code expects it and then group it based on the start line of
          # the conditional that causes the branching. We need to match each line coverage data with the branches that
          # spawn from that line
          branch_by_line = coverage_info[:branches].flat_map do |branch, data|
            branch_name = branch[0]

            data.map do |then_or_else, execution_count|
              name, _id, start_line, start_column, end_line, end_column = then_or_else


              grouping_line: branch[2]
              coverage_result = {
                groupingLine: grouping_line, # TODO: do we still need this here?
                executed: execution_count,
                location: {
                  start: { line: start_line, character: start_column },
                  end: { line: end_line, character: end_column },
                },
                label: "#{branch_name} #{name}",
              }
              
              [grouping_line, coverage_result]
            end
          end
```

lib/ruby_lsp/test_reporter.rb
```diff
@@ -67,6 +67,80 @@ def append_output(message:)
...
+            next if execution_count.nil?
+
+            {
+              executed: execution_count,
+              location: { line: line_index, character: 0 },
+              branches: branch_by_line[line_index] || [],
+            }
+          end
+
+          # The expected format is URI => { executed: number_of_times_executed, location: { ... }, branches: [ ... ] }
```
amomchilov, line 139: Can we capture this with a Hash shape in the side instead of `Hash[String, untyped]`?

amomchilov (approved):

lib/ruby_lsp/test_reporter.rb
amomchilov: Oops, I forgot to change the first line, which should be `result.group_by do |file_path, coverage_info|`

The idea is that a `to_h` followed by a `group_by` can be merged into just a `group_by`, similar to like how a `map` followed by a `flatten` can be merged to a single `flat_map`

lib/ruby_lsp/test_reporter.rb
amomchilov: I just figured that if there's going to be a comment to describe the shape, we may as well just formalize it into an RBS comment.

We don't have to describe the shape "all the way down", we can just summarize it to the same level that your comment did, so the method sig above can change like:

```diff
-       #: () -> Hash[String, untyped]
+      #: () -> Hash[String, { executed: Integer, location: Hash[Symbol, untyped], branches: Array }]
```

lib/ruby_lsp/test_reporter.rb
amomchilov: Nevermind, I'm trippin'

amomchilov (approved):

PR #3374, opened 2025-04-01
Strip HTML comments from RBS docs

andyw8:
### Motivation

Ruby LSP supports hovering over a method to show its docs. For Ruby itself, these docs come from RBS, and may include HTML comments which are a kind of metadata.

VS Code ignores these HTML comments, but in other editors, such as Vim and Zed, they are displayed as part of the content, e.g.:

Screenshot 2025-04-01 at 1 16 24 PM

This PR strips them out:

Screenshot 2025-04-01 at 1 42 31 PM

### Implementation

I've added the removal behaviour where the indexing happens, but it may be better to do it at a higher level, e.g. on the hover request, so that the 'raw' comment is still available through the index in case it's needed for other purposes. Any opinions?

### Automated Tests

None yet - I will add once there is a consensus on the approach.

### Manual Tests

Hover over a method definition from RBS, such as `"".length`.

cc @adam12 who [reported](https://ruby-dx.slack.com/archives/C050D17S6FP/p1737144877686379) this in the Ruby DX Slack a while ago.

### DIFF
--- a/lib/ruby_indexer/lib/ruby_indexer/rbs_indexer.rb
+++ b/lib/ruby_indexer/lib/ruby_indexer/rbs_indexer.rb
@@ -293,7 +293,9 @@ module RubyIndexer
 
     #: ((RBS::AST::Declarations::Class | RBS::AST::Declarations::Module | RBS::AST::Declarations::Constant | RBS::AST::Declarations::Global | RBS::AST::Members::MethodDefinition | RBS::AST::Members::Alias) declaration) -> String?
     def comments_to_string(declaration)
-      declaration.comment&.string
+      # HTML comments are not rendered in VS Code, but they are in some other editors, so
+      # we need to strip them out.
+      declaration.comment&.string&.gsub(/<!--.*?-->/m, "")
     end
   end
 end


vinistock:
I understand that seeing this in the editor is unnecessary and can be annoying, but I don't believe that stripping these out is the responsibility of the Ruby LSP.

If you take a look at the [commonmark Markdown specification](https://spec.commonmark.org/0.31.2/), under section 4.6 for HTML blocks, it mentions HTML comments as an acceptable part of the syntax. Most tools I've seen always ignore HTML comments inside Markdown files.

Additionally, the [LSP spec itself](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#markupContentInnerDefinition) mentions that clients have the liberty to decide how they want to sanitize the HTML included in Markdown content.

Finally, servers do not perform any Markdown parsing. Clients, on the other hand, **already need to parse Markdown** for presentation purposes, so they are much better equipped to decide what to strip from the content. I'd suggest reporting this to the editors.

andyw8:
Fixed in Zed v0.186.7: https://github.com/zed-industries/zed/pull/28318

PR #3377, opened 2025-04-04
find instance variable references with matched scopes

monkeyWzr:
### Motivation

https://github.com/Shopify/ruby-lsp/pull/2787#pullrequestreview-2534617160

### Implementation

Find instance variables references only with matched scopes.

### Automated Tests

Added.

### Manual Tests

Right click on an instance variable and select `Find All References`.

### DIFF
--- a/lib/ruby_indexer/lib/ruby_indexer/reference_finder.rb
+++ b/lib/ruby_indexer/lib/ruby_indexer/reference_finder.rb
@@ -36,13 +36,13 @@ module RubyIndexer
       attr_reader :name
 
       #: Array[String]
-      attr_reader :receiver_ancestors
+      attr_reader :owner_ancestors
 
-      #: (String name, Array[String] receiver_ancestors) -> void
-      def initialize(name, receiver_ancestors)
+      #: (String name, Array[String] owner_ancestors) -> void
+      def initialize(name, owner_ancestors)
         super()
         @name = name
-        @receiver_ancestors = receiver_ancestors
+        @owner_ancestors = owner_ancestors
       end
     end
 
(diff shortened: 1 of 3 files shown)


lib/ruby_indexer/lib/ruby_indexer/reference_finder.rb
```diff
@@ -35,10 +35,14 @@ class InstanceVariableTarget < Target
...
-      #: (String name) -> void
-      def initialize(name)
+      #: Array[String]
+      attr_reader :receiver_ancestors
+
+      #: (String name, Array[String] receiver_ancestors) -> void
+      def initialize(name, receiver_ancestors)
         super()
         @name = name
+        @receiver_ancestors = receiver_ancestors
```
vinistock, line 45: Let's call this `owner_ancestors` since instance variables don't technically have a receiver.

lib/ruby_indexer/lib/ruby_indexer/reference_finder.rb
```diff
@@ -321,7 +325,10 @@ def collect_constant_references(name, location)
     def collect_instance_variable_references(name, location, declaration)
       return unless @target.is_a?(InstanceVariableTarget) && name == @target.name
 
-      @references << Reference.new(name, location, declaration: declaration)
+      receiver_type = @stack.join("::")
```
vinistock, line 328: This helper method handles the namespaces considering top level references too.

```ruby
module Foo
  class ::Bar; end
end
```

```suggestion
      receiver_type = Index.actual_nesting(@stack, nil)
```

vinistock (approved):
This is great! I have some minor comments, but looks good to ship

lib/ruby_indexer/lib/ruby_indexer/reference_finder.rb
monkeyWzr: Thanks!

PR #3381, opened 2025-04-08
Ensure colons are escaped when building URIs

### LINKED ISSUE #3285: Go to definition shows duplicate definitions due to URI different between file:///c%3A/ and file:///c:/ on Windows
### Description

On Windows, the "Go to Definition" feature shows duplicate entries for the same definition due to differences in the URI format:

One uses: file:///c%3A/
The other uses: file:///c:/

### Reproduction steps
1. Open a Ruby project in VS Code on Windows.
2. Use the "Go to Definition" feature on a method or class.
3. Observe that duplicate definitions appear in the results.

Image

Image

Image

vinistock:
### Motivation

Closes #3285

The issue reported is that we're duplicating entries on Windows because one URI gets escaped and the other does not. The source of the problem is a difference between how Ruby escapes URIs and VS Code.

VS Code will always escape colon characters as `%3A`, but that's not the default behaviour for `URI::DEFAULT_PARSER.escape`. When we read URIs from disk, we're using the Ruby logic which generates one URI and then, when we receive requests from the editor for the same resource, we receive the escaped version of the URI.

```ruby
# What we get with Ruby
file:///C:/some/path.rb

# What we get from VS Code
file:///C%3A/some/path.rb
```

We need to ensure that we are escaping URIs in a way that's consistent with VS Code, otherwise we risk duplicating resources.

### Implementation

The URI implementation has a regex of unsafe characters, which it uses to decide what to escape and what to not escape.

To fix the problem, I just took that regex and removed colon from the list of safe characters, which then makes we produce the expected URIs during the initial indexing.

### Automated Tests

Added a test related to the bug and another one for unicode characters, which we were missing.

### DIFF
--- a/lib/ruby_indexer/lib/ruby_indexer/uri.rb
+++ b/lib/ruby_indexer/lib/ruby_indexer/uri.rb
@@ -15,14 +15,19 @@ module URI
     class << self
       #: (path: String, ?fragment: String?, ?scheme: String, ?load_path_entry: String?) -> URI::Generic
       def from_path(path:, fragment: nil, scheme: "file", load_path_entry: nil)
+        # This unsafe regex is the same one used in the URI::RFC2396_REGEXP class with the exception of the fact that we
+        # do not include colon as a safe character. VS Code URIs always escape colons and we need to ensure we do the
+        # same to avoid inconsistencies in our URIs, which are used to identify resources
+        unsafe_regex = %r{[^\-_.!~*'()a-zA-Z\d;/?@&=+$,\[\]]}
+
         # On Windows, if the path begins with the disk name, we need to add a leading slash to make it a valid URI
         escaped_path = if /^[A-Z]:/i.match?(path)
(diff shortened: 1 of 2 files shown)


vinistock:
* **#3381** <a href="https://app.graphite.dev/github/pr/Shopify/ruby-lsp/3381?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/Shopify/ruby-lsp/3381?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`
<hr />

You must have a Graphite account in order to use the merge queue. Sign up using [this link](https://app.graphite.dev/invite/github/Shopify?ref=merge-queue-instructions-in-stack-comment&prId=3381).

<h2></h2>
<sub>

</sub>

<sub>

</sub>

<h2></h2>

alexcrocha (approved):

PR #3382, opened 2025-04-09
Update docs for Zed

andyw8:
A few small updates.

### DIFF
--- a/jekyll/editors.markdown
+++ b/jekyll/editors.markdown
@@ -283,11 +283,9 @@ Restart LSP or Sublime Text and `ruby-lsp` will automatically activate when open
 
 ## Zed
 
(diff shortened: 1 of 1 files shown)


jekyll/editors.markdown
```diff
@@ -283,11 +283,13 @@ Restart LSP or Sublime Text and `ruby-lsp` will automatically activate when open
 
 ## Zed
 
-[Setting up Ruby LSP](https://github.com/zed-industries/zed/blob/main/docs/src/languages/ruby.md#setting-up-ruby-lsp)
```
andyw8, line 286: I've updated this below to point to the docs on zed.dev, rather than the GitHub source.

andyw8:
I've pushed an update.

PR #3383, opened 2025-04-09
Introduce Addon::notify to standardize error handling for add-ons

### LINKED ISSUE #3315: Explore preventing add-ons from exiting the process
Currently, if an add-on executes `exit`, the Ruby LSP process will die and cause the editor connection to crash.

I've been wondering if we can prevent that from happening. My idea would be to capture the original `exit` method in a constant (so that we can ourselves use it when shutting down) and then override it to no-op.

Something like this:

```ruby
ORIGINAL_EXIT = Kernel.method(:exit)

module Kernel
  def exit
    # do nothing
  end
end

# when shutting down
ORIGINAL_EXIT.call
```

ryan-robeson:
Wraps calls to add-on actions with error handling so that they don't bring down the server.

### Motivation
Closes #3315 

### Automated Tests
Yes

### DIFF
--- a/lib/ruby_lsp/addon.rb
+++ b/lib/ruby_lsp/addon.rb
@@ -2,7 +2,7 @@
 # frozen_string_literal: true
 
 module RubyLsp
-  # To register an add-on, inherit from this class and implement both `name` and `activate`
+  # To register an add-on, inherit from this class and implement `name`, `activate`, and `deactivate`.
   #
   # # Example
   #
@@ -11,6 +11,10 @@ module RubyLsp
   #   class MyAddon < Addon
   #     def activate
   #       # Perform any relevant initialization
+  #     end
+  #
+  #     def deactivate
+  #       # Clean up resources like file watchers or child processes
   #     end
   #
   #     def name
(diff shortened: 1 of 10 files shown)


vinistock (commented):
Thank you for the PR! Unfortunately, I don't think this will exactly solve the problem.

Those `create_..._listener` methods are only factories to hook new listeners into the AST dispatching process. An add-on is more likely to raise or accidentally try to exit during the dispatch of the AST events, which is controlled by `Prism::Dispatcher`.

Also, we cannot have `send` in the main portions of the automated requests due to performance reasons (dynamic dispatch is significantly slower).

What do you think about adding another branch to our [general rescue](https://github.com/Shopify/ruby-lsp/blob/b8b425e8d9028607d669d15306ed9c76d0e3d9c1/lib/ruby_lsp/server.rb#L127) for system exit and interrupt?

In that spot, we can catch if any features or add-ons are trying to exit and log a warning with the backtrace. Additionally, this won't mess up the LSP's own exiting mechanisms because we call `exit` outside of the `process_message` method.

vinistock:
I opened #3617 to address it.

ryan-robeson:
My bad, been meaning to take another look at this. New PR looks like a better solution to me though

Issue #1782, opened 2023-10-20
Support running the server through a socket


vinistock:
The [specification](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#implementationConsiderations) mentions that language servers should be able to communicate through sockets and the [client package](https://github.com/microsoft/vscode-languageserver-node) support a port option.

Being able to run the server through a port would better accommodate use cases like the one described in Shopify/ruby-lsp#1728.

PR #3386, opened 2025-04-10
VS Code: Add customBinaryPath option

palkan:
### Motivation

This option allows to provide any compatible binary to be used as `ruby-lsp` exec. This is especially useful for containerized development with VS Code running on the host machine (i.e., not using Dev Containers or Remote Containers). A typical use case is the applications using `dip` (and [ruby-on-whales](https://evilmartians.com/chronicles/ruby-on-whales-docker-for-ruby-rails-development)) (see #2919). However, the ability to customize the binary path has many more potential applications.

_Motivation continues..._

I've been using a similar approach with Zed for a while (it already provides this option). The trick is to use a thin Bash wrapper to run `ruby-lsp` anywhere. Mine looks as follows:

```sh
#!/bin/bash

cd $(dirname $0)/..

dip ruby-lsp $@
```

That's it.

With this change, we can configure `rubyLsp.customBinaryPath` to point to a custom executable (say, ".dockerdev/ruby-lsp"), and benefit from LSP features even without having Ruby LSP (and even Ruby) installed locally (let me omit some Docker-specific tricks used to make it work, not relevant to the proposal).

### Implementation

- Added new configuration option to `package.json`: `customBinaryPath`
- Updated `client.ts` to use the new option if present to generate commands configuration.

### Automated Tests

None. Couldn't find any tests for similar functionality (commands configuration). Would love to have some guidance here. 

### Manual Tests

Packaged and installed the forked version locally and verified that LSP works.

### DIFF
--- a/vscode/package.json
+++ b/vscode/package.json
@@ -395,6 +395,10 @@
           "description": "A shell command to activate the right Ruby version or add a custom Ruby bin folder to the PATH. Only used if rubyVersionManager is set to 'custom'",
           "type": "string"
         },
+        "rubyLsp.customBinaryPath": {
+          "description": "A path to a custom Ruby LSP binary executable",
+          "type": "string"
+        },
         "rubyLsp.formatter": {
           "description": "Which tool the Ruby LSP should use for formatting files",
           "type": "string",
--- a/vscode/src/client.ts
+++ b/vscode/src/client.ts
@@ -99,6 +99,7 @@ function getLspExecutables(
   const config = vscode.workspace.getConfiguration("rubyLsp");
   const branch: string = config.get("branch")!;
   const customBundleGemfile: string = config.get("bundleGemfile")!;
+  const customBinPath: string = config.get("customBinaryPath")!;
   const useBundlerCompose: boolean = config.get("useBundlerCompose")!;
   const bypassTypechecker: boolean = config.get("bypassTypechecker")!;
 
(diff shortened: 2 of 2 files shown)


palkan:
I have signed the CLA!

willtcarey:
This should resolve https://github.com/Shopify/ruby-lsp/issues/2223 with the same motivation and ideas here

palkan:
Hey @vinistock,

Sorry for pushing, but it would be nice to at least know if there is something the team considers to be a good addition (and, thus, we can expect it soon) or not at all. Thanks.

vinistock:
Thanks for the PR. Would supporting running the language server through [sockets](https://github.com/Shopify/ruby-lsp/issues/1782) solve this use case?

Connecting to the LSP via sockets is standardized in the spec. If that solves the use cases, I would rather we implemented that rather than allowing customizing the executable. 

The launching of the LSP ends up being the most critical part of our integration because we need to fully replicate the user's environment with the contributions of their version manager + bundler. I'm still not satisfied with the level of robustness of our launching and I'm afraid that allowing people to configure whatever executable they want could potentially build us into a corner where we cannot change the main `ruby-lsp` executable without a major breaking change.

Before committing to a decision, I'd also like to understand how other language servers solve this problem. What does Rust Analyzer do (or any other LSP)? As far as I know, there's no way to change its executable.

palkan:
Hey @vinistock,

Have you had a chance to look at my previous comment?

dsager:
> > ...how other language servers solve this problem
> 
> That seems more about editor plugins then LSPs.

The [VSCode extension for rubocop](https://github.com/rubocop/vscode-rubocop) also allows you to set a custom command, adding the to the following configuration works well:
```json
{
  "rubocop.commandPath": "docker-compose exec my-app-name bundle exec rubocop"
}
```

vinistock:
Container related discussions were happening in a few different places, so I created https://github.com/Shopify/ruby-lsp/discussions/3738 to try to centralize and move the discussion forward.

PR #3396, opened 2025-04-16
Add Experimental MCP Support

st0012:
# Add Model Context Protocol (MCP) support to Ruby LSP

This PR adds support for the Model Context Protocol (MCP), enabling AI assistants to programmatically access Ruby project indexes and file contents that Ruby LSP already collects.

## Connection Mechanism

The implementation uses a hybrid approach:

1. Ruby LSP runs an HTTP-based MCP server powered by WEBrick on a dynamically assigned local port
2. A lightweight stdio-to-HTTP bridge script (`ruby-mcp-bridge`) connects the editor's MCP client to our HTTP server
3. Communication flows bidirectionally between AI assistants and the Ruby project's metadata

## Architecture Diagram

Screenshot 2025-04-14 at 09 43 50

### VS Code Configuration

Add to `.vscode/mcp.json`:
```json
{
  "inputs": [],
  "servers": {
    "rubyMcp": {
      "type": "stdio",
      "command": "${workspaceFolder}/.ruby-lsp/ruby-mcp-bridge"
    }
  }
}
```

See [VS Code MCP documentation](https://code.visualstudio.com/api/extension-guides/model-context-protocol) for more details.

### Cursor Configuration

Add to `.cursor/mcp.json`:
```json
{
  "mcpServers": {
    "rubyMcp": {
      "command": "./.ruby-lsp/ruby-mcp-bridge"
    }
  }
}
```

See [Cursor MCP documentation](https://cursor.sh/blog/mcp) for more details.

## Notes For Manual Testing

This feature requires both the extension and server changes to work. So for testing:

1. Run the extension with the subject project
2. Update the subject project's Gemfile to have `gem "ruby-lsp", github: "Shopify/ruby-lsp", branch: "mcp"`
3. Make sure the server is connected to the respective editors. When connected, the editor should display what tools are available
4. In both editors, MCP servers are only used in "Agent" mode
5. For VS Code, tool usage needs to be selected manually at each prompt

## Known Issues

⚠️ **Cursor-specific issue**: We've identified that Cursor's MCP/LSP client implementation can occasionally cause hangs in the Ruby LSP process. The main thread gets stuck waiting on `gets` in the LSP communication channel. This issue may be specific to Cursor as it doesn't occur (or occurs much less frequently) in VS Code, and more often in certain projects (e.g. `ruby/irb`) than others. The cause may be related to how Cursor handles MCP/LSP requests and requires more investigation.

## Supported Tools (TBD)

The MCP server provides four tools:

1. **get_classes_and_modules** - Browse all indexed classes and modules in your project, with optional query filtering
6. ~~**read_ruby_files** - Access file contents, including from dependencies~~
    - I removed this because clients usually would ask additional permission to read files outside of the workspace, and Ruby MCP should not provide a way to get around that.
8. **get_methods_details** - Get comprehensive information about methods (documentation, parameters, location, visibility)
9. **get_class_module_details** - See class/module details including documentation, ancestry chains, and available methods

These tools enable AI assistants to understand Ruby code structure, navigate dependencies, and access relevant documentation that are not accessible through the editor's simple codebase indexing.


st0012:
We decided not to continue with this approach because we identified a few challenges:
- LSP & MCP have very different expectations on servers' lifecycle and numbers of connections per client application:
    - A local MCP server nowadays usually last as long as the client's connection to it. But a LSP server, especially Ruby LSP, can restart multiple times in the same editor session (e.g. updating bundle or rebasing...etc.)
    - An editor usually only has 1 MCP client but can have multiple LSP clients (in a multi-workspace setup)
- Since LSP servers are connected through stdio of the process, to connect to a MCP client too the server needs to either use a unix or tcp/ip socket, which adds a lot of complexity
- While using http for connection is also possible, it requires adding a http server library to the project, which we want to avoid

And so far we don't see a solution that can overcome all the challenges while providing a easy, robust experience to users. It also won't work for non-editor clients like Claude Code. So after some discussions, we decided that it's not worth continuing the type of integration with LLMs.

That said, an alternative could be to use [VS Code's LLM tooling API](https://code.visualstudio.com/api/references/vscode-api#lm) that directly requests information through Ruby LSP's custom server requests. It'll be a much simpler solution but will only work for VS Code (until Cursor supports the same APIs).

kuboon:
I've tried this with ruby-lsp but it seems fails.
https://github.com/mizchi/lsmcp

PR #3397, opened 2025-04-16
Completion for keyword arguments

### LINKED ISSUE #2913: Provide completion for keyword arguments in method calls
If we know exactly which method the user is currently typing, we can provide completion for the keyword arguments required to invoke that method.

E.g.:

```ruby
class Foo
  def bar(a:, b:)
  end

  def baz
    baz()
      #^^ when the cursor is here we can show `a` and `b` as possible symbol completions
  end
end
```

monkeyWzr:
### Motivation

Closes #2913.

### Implementation

* add trigger characters for completion: `(`, `,`
* find possible candidates from method's signature and existing arguments
* handle nesting method calls

### Automated Tests

Added.

### Manual Tests

Kapture 2025-04-16 at 18 21 51

### DIFF
--- a/lib/ruby_lsp/listeners/completion.rb
+++ b/lib/ruby_lsp/listeners/completion.rb
@@ -50,7 +50,7 @@ module RubyLsp
         "__LINE__",
       ].freeze
 
-      #: (ResponseBuilders::CollectionResponseBuilder[Interface::CompletionItem] response_builder, GlobalState global_state, NodeContext node_context, RubyDocument::SorbetLevel sorbet_level, Prism::Dispatcher dispatcher, URI::Generic uri, String? trigger_character) -> void
+      #: (ResponseBuilders::CollectionResponseBuilder[Interface::CompletionItem] response_builder, GlobalState global_state, NodeContext node_context, RubyDocument::SorbetLevel sorbet_level, Prism::Dispatcher dispatcher, URI::Generic uri, String? trigger_character, Integer char_position) -> void
       def initialize( # rubocop:disable Metrics/ParameterLists
         response_builder,
         global_state,
@@ -58,7 +58,8 @@ module RubyLsp
         sorbet_level,
         dispatcher,
         uri,
-        trigger_character
+        trigger_character,
+        char_position
       )
         @response_builder = response_builder
         @global_state = global_state
@@ -68,6 +69,7 @@ module RubyLsp
         @sorbet_level = sorbet_level
         @uri = uri
         @trigger_character = trigger_character
+        @char_position = char_position
 
         dispatcher.register(
           self,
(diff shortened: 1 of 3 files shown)


lib/ruby_lsp/listeners/completion.rb
```diff
@@ -169,6 +171,11 @@ def on_call_node_enter(node)
           end
         end
 
+        if ["(", ","].include?(@trigger_character)
+          complete_keyword_arguments(node)
+          return
+        end
```
vinistock, line 177: Instead of checking on call nodes if the trigger character matches, we should start handling the event for `Prism::ParametersNode`, so that completion is triggered for parameters.

This should be a matter of 

1. [Registering the event](https://github.com/Shopify/ruby-lsp/blob/23a7084fe44636c140dee3a9d54252f920fcabcb/lib/ruby_lsp/listeners/completion.rb#L94)
2. [Adding the node as a possible target](https://github.com/Shopify/ruby-lsp/blob/23a7084fe44636c140dee3a9d54252f920fcabcb/lib/ruby_lsp/requests/completion.rb#L59)
3. Creating an `on_parameters_node_enter` handler
4. Doing the work to provide keyword completion inside that method

lib/ruby_lsp/listeners/completion.rb
```diff
@@ -525,10 +532,139 @@ def complete_methods(node, name)
...
+        #                 ^
+        # when we type `b`, it triggers completion for b,
+        # so we provide keyword arguments completion for `foo` using `b` as a filter prefix
+        if @node_context.parent.is_a?(Prism::CallNode) && method_name
+          call_node = T.cast(@node_context.parent, Prism::CallNode)
+          candidates = keyword_argument_completion_candidates(call_node, method_name)
+          candidates.each do |param|
+            build_keyword_argument_completion_item(param, range)
+          end
+        end
```
vinistock, line 547: You won't need this check after registering for parameter node events.

lib/ruby_lsp/listeners/completion.rb
```diff
@@ -525,10 +532,139 @@ def complete_methods(node, name)
...
+
+        candidates
+      end
+
+      #: (Prism::CallNode node, String? name) -> (RubyIndexer::Entry::Member | RubyIndexer::Entry::MethodAlias)?
+      def resolve_method(node, name)
+        return unless name
+
+        type = @type_inferrer.send(:infer_receiver_for_call_node, node, @node_context)
+        return unless type
```
vinistock, line 663: Let's not rely on a private API. We can just call `@type_inferrer.infer_receiver_type(@node_context)` and make any necessary modifications to handle the new scenario in the type inferrer.

Issue #2156, opened 2024-06-11
Deprecate `.index.yml`, move configuration to initializer


andyw8:
We had [originally](https://github.com/Shopify/ruby-lsp/pull/1434) considered changing `.index.yml` to be `.ruby-lsp.yml` so that additional could be supported.

However from looking at other language server implementations, the preferred approach for this kind of things seems instead be to pass any configuration as part of the initialization options.

tachyons:
@andyw8 @vinistock  Is it possible to rethink about this approach ? When working with large code bases like GitLab, ruby-lsp takes several Giga bytes of RAM for indexing, So it is important to exclude unnecessary directories  to make the ruby-lsp usable. So Gitlab added [this file by default ](https://gitlab.com/gitlab-org/gitlab/-/merge_requests/146069). 

Passing this directories in the editor is are hard, and difficult to share in the team. Also for a person working on different ruby projects, the directories to include will be different.

vinistock:
@tachyons I brought this up with the team. We are okay with keeping the editor agnostic configuration file under certain conditions:

1. We want to move from YAML to JSON. Easier to handle and then the extension could even contribute completion for the keys and values in VS Code
2. We want to rename the file to `ruby-lsp.json`, so that it captures any Ruby LSP configuration and not just indexing
3. By keeping the configuration file, there will be 3 levels of configuration that will need to be merged. User editor settings (global), workspace editor settings (project-specific) and the `ruby-lsp.json` configuration file (project-specific). Editors already make workspace take precedence over user settings and merge them, but we will also want the configuration file to take precedence over editor settings
4. We will not support global configurations through files like these. For example, a `~/.ruby-lsp.json` file. Otherwise, it would introduce a 4th level of precedence to account for. Global configuration will remain editor only

I'll re-open this issue and change the title, but our team is currently quite busy looking at other things, so we cannot prioritize the work to make these changes right now. If you're interested in contributing it, we're happy to help and provide guidance on what needs to change.

tachyons:
@vinistock Thanks for re opening the issue. I don't have bandwidth to contribute at the moment, I'll try in the future if I get time.

PR #3417, opened 2025-04-27
Replace .index.yml with .ruby-lsp.json for configuration

tachyons:
Change indexing configuration to use JSON instead of YAML, with proper precedence between file and editor settings. Add tests for new functionality.

Addresses https://github.com/Shopify/ruby-lsp/issues/2156


vinistock:
@tachyons is this PR good for a review? Or were you hoping to do some more work on it before?

tachyons:
@vinistock current implementation only supports the indexing configuration in .ruby-lsp.json. Rest of the work is pending, that is why I kept the PR as draft, I haven't got time to wrap this feature.

PR #3424, opened 2025-04-28
Simplify local variable regexes in grammar

amomchilov:


vscode/grammars/ruby.cson.json
```diff
@@ -109,7 +109,7 @@
           "name": "keyword.operator.assignment.augmented.ruby"
         }
       },
-      "match": "^\\s*((?:_|[a-z])([A-Za-z0-9_])*)\\s*((&&|\\|\\|)=)",
+      "match": "^\\s*([_a-z]?[A-Za-z0-9_]*)\\s*((&&|\\|\\|)=)",
```
Morriar, line 112: Aren't you allowing an empty string? since `[_a-z]` is optional and `[A-Za-z0-9_]` can be repeated 0 times?

amomchilov: Ah yes, there should be no `?`.

amomchilov:
Hmmm I think the tests are flaking. Will look into it tomorrow.

PR #3433, opened 2025-04-29
Require `bundler/setup` for running tests

vinistock:
### Motivation

I had already added `-rbundler/setup` to the debug command, but I forgot to do it in the run execution mode.

If we don't set up the bundle, we risk trying to require versions of the Ruby LSP that don't match the application's lockfile and that raises an error, crashing the execution of the test suite.

### DIFF
--- a/vscode/src/testController.ts
+++ b/vscode/src/testController.ts
@@ -579,9 +579,10 @@ export class TestController {
     // Require the custom JSON RPC reporters through RUBYOPT. We cannot use Ruby's `-r` flag because the moment the
     // test framework is loaded, it might change which options are accepted. For example, if we append `-r` after the
     // file path for Minitest, it will fail with unrecognized argument errors
+    const commonOpts = `-rbundler/setup ${response.reporterPaths?.map((path) => `-r${path}`).join(" ")}`;
     const rubyOpt = workspace.ruby.env.RUBYOPT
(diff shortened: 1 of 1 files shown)


vscode/src/testController.ts
```diff
@@ -580,8 +580,8 @@ export class TestController {
     // test framework is loaded, it might change which options are accepted. For example, if we append `-r` after the
     // file path for Minitest, it will fail with unrecognized argument errors
     const rubyOpt = workspace.ruby.env.RUBYOPT
-      ? `${workspace.ruby.env.RUBYOPT} ${response.reporterPaths?.map((path) => `-r${path}`).join(" ")}`
-      : response.reporterPaths?.map((path) => `-r${path}`).join(" ");
+      ? `${workspace.ruby.env.RUBYOPT} -rbundler/setup ${response.reporterPaths?.map((path) => `-r${path}`).join(" ")}`
+      : `-rbundler/setup ${response.reporterPaths?.map((path) => `-r${path}`).join(" ")}`;
```
amomchilov, line 584: Minor nit, but could we extract out the common parts here? It's hard to know what's actually different

```suggestion
    const commonOpts = `-rbundler/setup ${response.reporterPaths?.map((path) => `-r${path}`).join(" ")}`
    
    const rubyOpt = workspace.ruby.env.RUBYOPT ? `${workspace.ruby.env.RUBYOPT} ${commonOpts}` : commonOpts;
```

amomchilov (approved):

PR #3440, opened 2025-04-30
Migrate T.cast to RBS comment syntax

vinistock:


sorbet/rbi/shims/type_aliases.rbi
```diff
@@ -0,0 +1,21 @@
+# typed: true
+
+AnyConstant = T.type_alias do
```
Morriar, line 3: Not having this at runtime will make the signatures raise when we try to generate RBIs for the indexer with Tapioca

PR #3445, opened 2025-05-01
Ensure `finish` event is emitted on test crashes

vinistock:
### Motivation

If the test suite crashes immediately, for example when requiring the code, we will never emit the `finish` event, which makes the extension side hang.

We need to ensure that we emit the `finish` event, both for proper executions and crashes.

### Implementation

I suggest that we start remembering if we already sent a `finish` event. If not, we can send it `at_exit`, which catches the crash case.

### DIFF
--- a/lib/ruby_lsp/test_reporters/lsp_reporter.rb
+++ b/lib/ruby_lsp/test_reporters/lsp_reporter.rb
@@ -9,6 +9,9 @@ module RubyLsp
   class LspReporter
     include Singleton
 
+    #: bool
+    attr_reader :invoked_shutdown
+
     #: -> void
     def initialize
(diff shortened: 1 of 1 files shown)


lib/ruby_lsp/test_reporters/lsp_reporter.rb
```diff
@@ -161,4 +168,10 @@ def send_message(method_name, **params)
     File.write(File.join(".ruby-lsp", "coverage_result.json"), coverage_results.to_json)
     RubyLsp::LspReporter.instance.internal_shutdown
```
st0012, line 169: Should the condition be applied here too?

st0012 (approved):

PR #3452, opened 2025-05-02
Write TCP server port to a tempfile instead of using env var

vinistock:
### Motivation

The strategy of passing the port to LSP reporter via environment variable is a bit weak. For example, if you reload VS Code and then proceed to run tests manually in a terminal that we set up, you will get a connection refused error because we never updated the port variable in that old terminal.

I think we can be a lot more reliable by writing the port to a file.

### Implementation

Every workspace reports test results through the same port, so my idea is to write a temporary file with the port information that the LSP reporter can read on its own.

That way, every time the extension reloads, we update the port in the file and, every time the tests are executed (regardless if it's manual or through the explorer), we will read that value again ensuring that we're using the right port.

Finally, I noticed one more hurdle. When running tests manually through the terminal, we were receiving the connection okay, but there was no `run` object to use for reporting the items' statuses. We need to create a `run` object on demand when not running through the explorer, otherwise the statuses won't appear as expected.

### DIFF
--- a/lib/ruby_lsp/test_reporters/lsp_reporter.rb
+++ b/lib/ruby_lsp/test_reporters/lsp_reporter.rb
@@ -16,7 +16,7 @@ module RubyLsp
     #: -> void
     def initialize
       dir_path = File.join(Dir.tmpdir, "ruby-lsp")
-      FileUtils.mkdir_p(dir_path) unless File.exist?(dir_path)
+      FileUtils.mkdir_p(dir_path)
 
       port_path = File.join(dir_path, "test_reporter_port")
       port = ENV["RUBY_LSP_REPORTER_PORT"]


vscode/src/streamingRunner.ts
```diff
@@ -272,6 +296,11 @@ export class StreamingRunner implements vscode.Disposable {
       this.connection.dispose();
     }
 
+    if (this.run!.name === "on_demand_run_in_terminal") {
+      this.run!.end();
+    }
+    this.run = undefined;
```
amomchilov, line 302: Is it possible that `this.run!` is already `undefined` when this runs the first time?

lib/ruby_lsp/test_reporters/lsp_reporter.rb
```diff
@@ -14,11 +15,24 @@ class LspReporter
 
     #: -> void
     def initialize
+      dir_path = File.join(Dir.tmpdir, "ruby-lsp")
+      FileUtils.mkdir_p(dir_path) unless File.exist?(dir_path)
```
amomchilov, line 19: No need to check first. If you're going to end up doing a syscall and IO op anyway, may as well just be the the `mkdir -p` call

```suggestion
      FileUtils.mkdir_p(dir_path)
```

lib/ruby_lsp/test_reporters/lsp_reporter.rb
```diff
@@ -14,11 +15,24 @@ class LspReporter
 
     #: -> void
     def initialize
+      dir_path = File.join(Dir.tmpdir, "ruby-lsp")
```
amomchilov, line 18: Is anything going to ever clean this up?

lib/ruby_lsp/test_reporters/lsp_reporter.rb
vinistock: Good point

PR #3480, opened 2025-05-12
Lazily discover test framework on execution

vinistock:
### Motivation

When running an entire test file or directory, we need to know what framework is associated with it so that commands can be resolved according to that framework. Our current strategy is to sample the first test file inside directories to assign a framework during test discovery.

Unfortunately, for large projects, that's very slow as we end up having to call the server hundreds of times. It's also wasteful because many of those directories wouldn't even be executed and so the framework discovery is not even used.

This PR stops assigning framework tags during discovery and moves that to test execution, so that we only do the work when we actually need to run the tests. This makes test file discovery nearly instantaneous and does not slow down execution significantly.

### Implementation

1. Removed all of the code related to assigning the framework automatically during discovery
2. Started assigning the framework on execution if necessary

### Automated Tests

Added a test.

### DIFF
--- a/vscode/src/testController.ts
+++ b/vscode/src/testController.ts
@@ -531,6 +531,8 @@ export class TestController {
       this.testController.items.forEach((test) => items.push(test));
     }
 
+    await this.discoverFrameworkTag(items);
+
     const workspaceToTestItems = new Map<
       vscode.WorkspaceFolder,
       vscode.TestItem[]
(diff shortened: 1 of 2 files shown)


vscode/src/testController.ts
```diff
@@ -626,6 +628,50 @@ export class TestController {
...
+
+    for (const item of missingFramework) {
+      let testFileItem = item;
+
+      while (!testFileItem.tags.some((tag) => tag === TEST_FILE_TAG)) {
+        const children: vscode.TestItem[] = [];
+
+        testFileItem.children.forEach((child) => {
+          children.push(child);
+        });
```
amomchilov, line 651: Wish they added more of the collection APIs on this, like `map`

vscode/src/testController.ts
```diff
@@ -626,6 +628,50 @@ export class TestController {
...
+    }
+  }
+
+  private recursivelySetFrameworkTag(item: vscode.TestItem) {
+    const tag = item.tags.find((tag) => tag.id.startsWith("framework"))!;
+    let parent = item.parent;
+
+    while (parent) {
+      if (!parent.tags.some((tag) => tag.id.startsWith("framework"))) {
+        parent.tags = [...parent.tags, tag];
```
amomchilov, line 668: Do we need a new array here?

```suggestion
        parent.tags.push(tag);
```

vscode/src/testController.ts
```diff
@@ -626,6 +628,50 @@ export class TestController {
...
+        testFileItem = children[0];
+      }
+
+      // Handle the case where the test file is empty and has no children
+      await this.resolveHandler(testFileItem);
+      this.recursivelySetFrameworkTag(testFileItem);
+    }
+  }
+
+  private recursivelySetFrameworkTag(item: vscode.TestItem) {
```
amomchilov, line 662: This isn't recursive :)

amomchilov (approved):
Makes sense!

vscode/src/testController.ts
vinistock: Indeed! I updated the name

vscode/src/testController.ts
vinistock: Yeah or at least allowed accessing the first entry.

vscode/src/testController.ts
```diff
@@ -626,6 +628,50 @@ export class TestController {
...
+      let testFileItem = item;
+
+      while (!testFileItem.tags.some((tag) => tag === TEST_FILE_TAG)) {
+        const children: vscode.TestItem[] = [];
+
+        testFileItem.children.forEach((child) => {
+          children.push(child);
+        });
+
+        testFileItem = children[0];
```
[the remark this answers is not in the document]
vinistock: Yeah, let's do that

PR #3485, opened 2025-05-13
Use framework callback instead of at_exit

st0012:
### Motivation

Both minitest and test-unit executes tests in at_exit callback. This means that our own at_exit callback may be invoked before the tests actually finish.

Instead, we should register our finishing callbacks using the frameworks' own callback mechanism. In minitest, we use after_run, and in test-unit, we use at_exit.

This change fixes the issue where test-unit tests would crash with `'IO#write': closed stream (IOError)`, which is caused by the `at_exit` callback being invoked before the tests finish.


vinistock (approved):

st0012:
### Merge activity

* **May 13, 4:26 PM EDT**: The merge label 'graphite-merge' was detected. This PR will be added to the [Graphite merge queue](https://app.graphite.dev/merges?org=Shopify&repo=ruby-lsp) once it meets the requirements.

PR #3490, opened 2025-05-14
Handle spec example name normalization

vinistock:
### Motivation

We were not using the correct IDs for `it` examples in `Minitest::Spec`. Minitest will take the description and format it using the example count ([see here](https://github.com/minitest/minitest/blob/4c7eb3567a3168b3e100ff1ccd345b0f88d0fb4b/lib/minitest/spec.rb#L230C20-L230C23)).

This means that the Minitest reporter will use a different ID than discovery and we'll not be able to show results properly.

### Implementation

Since we discover tests through static analysis, it is not possible to match the exact same numbers Minitest uses from the runtime for its examples.

Instead of trying to match that exactly, we can be more accurate if we use the line number of where we found that test as part of its ID - which is guaranteed to match. For this to work, we must ensure that both discovery and test result reporting are using the same ID formatting, which is what this PR does.

**Note**: there's a decision being made here, which is that dynamically defined anonymous tests cannot be supported (see the doc changes). They are not that common anyway and we have limited support for meta-programming.

### Automated Tests

Added/improved current tests.

### DIFF
--- a/jekyll/test_explorer.markdown
+++ b/jekyll/test_explorer.markdown
@@ -67,6 +67,19 @@ class MyTest < Minitest::Spec
 end
 ```
 
+Dynamically defined anonymous tests are not supported properly because there's no way to accurately reconcile their
+discovery with execution.
+
+```ruby
+class MyTest < Minitest::Spec
+  # Anonymous examples (no description) defined dynamically are not supported
+  5.times do
+    it do
+    end
+  end
+end
+```
+
 ### Tests that accept external parameters
 
 In Ruby, you can write tests that accept external parameters, like environment variables.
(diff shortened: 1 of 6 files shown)


vinistock:
* **#3490** <a href="https://app.graphite.dev/github/pr/Shopify/ruby-lsp/3490?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/Shopify/ruby-lsp/3490?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`
<hr />

You must have a Graphite account in order to use the merge queue. Sign up using [this link](https://app.graphite.dev/invite/github/Shopify?ref=merge-queue-instructions-in-stack-comment&prId=3490).

<h2></h2>
<sub>

</sub>

<sub>

</sub>

<h2></h2>

PR #3505, opened 2025-05-19
Add minitest/spec code lens test command resolution for minitest/spec

domingo2000:
### Motivation

Currently the Minitest Spec style is not beign resolved correctly in the code lens for the new Test Explorer implementation. That happens because the ids of the TestItem generated by `SpecStyle` was not using the `GroupX::GroupY#ExampleZ` syntax required for the vscode extension `runTestCommand` function.

Because of that, even when the code lens for the inner groups of minitest spec style test was beign shown in the editor, they were not working correctly.

### Implementation

1. Change the ids of the TestItem generated by minitest spec to what actually minitest parse underground, generating ids like `describe 1:: describe 2#test_0001_example` this reflects the _real methods_ created by this DSL in the minitest runtime.
2. Change the command test resolution to use the `id` instead of the label. With that we can have the spec syntax nicely formatted in the Vscode pannel, but the id contains all the necesary information for the Minitest Test Style to work. Because of that the command test resolution of minitest spec is the same as the current minitest test resolution.

### Automated Tests

1. Added tests to the correct id parsing of the test items in the minitest spec style.
2. Extended the current test to reflect deeply nested (2 level) describe/example blocks in the minitest style.

### Manual Tests

1. Enable the `"fullTestDiscovery": true`
2. Create a file `some_test.rb` in the `spec/` folder 
3. Test that the CodeLens and Test Pannel works correctly for the spec style

Example File:
```ruby
# ./spec/some_test.rb
# frozen_string_literal: true
require "minitest/autorun"

describe "First Spec" do
  it "test one" do
    assert true
  end

  it "test two" do
    assert true
  end

  specify "test three" do
    assert true
  end

  describe "other block" do

    it "pass" do
      assert true
    end

    it "fails" do
      assert false
    end
  end
end
```

Result:

[Screencast from 18-05-25 20:58:20.webm](https://github.com/user-attachments/assets/390404d9-56db-4505-94fa-7ec6ab6266e7)

### DIFF
--- a/lib/ruby_lsp/listeners/spec_style.rb
+++ b/lib/ruby_lsp/listeners/spec_style.rb
@@ -10,6 +10,7 @@ module RubyLsp
 
         @describe_block_nesting = [] #: Array[String]
         @spec_class_stack = [] #: Array[bool]
+        @spec_id = 0 #: Integer
 
         dispatcher.register(
           self,
@@ -63,6 +64,9 @@ module RubyLsp
 
         return unless in_spec_context?
 
+        # Reset spec_id when entering a new group
+        @spec_id = 0
+
         if @nesting.empty? && @describe_block_nesting.empty?
           test_item = Requests::Support::TestItem.new(
             description,
@@ -89,6 +93,9 @@ module RubyLsp
         description = extract_description(node)
         return unless description
 
+        # Increment spec_id for each example
+        @spec_id += 1
+
         add_to_parent_test_group(description, node)
       end
 
@@ -97,8 +104,15 @@ module RubyLsp
         parent_test_group = find_parent_test_group
         return unless parent_test_group
 
+        id = if node.name == :describe
+          "#{@nesting.dup.concat(@describe_block_nesting).join("::")}::#{description}"
+        else
+          method_name = format("test_%04d_%s", @spec_id, description)
+          "#{@nesting.dup.concat(@describe_block_nesting).join("::")}##{method_name}"
+        end
+
         test_item = Requests::Support::TestItem.new(
-          description,
+          id,
           description,
           @uri,
           range_from_node(node),
@@ -123,7 +137,8 @@ module RubyLsp
         return test_group unless nested_describe_groups
 
         nested_describe_groups.each do |description|
-          test_group = test_group[description]
+          id = "#{test_group.id}::#{description}"
+          test_group = test_group[id]
         end
 
         test_group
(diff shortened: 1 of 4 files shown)


domingo2000:
Just took a look at the merge conflicts and saw that the runtime numbers were implemented last week in #3490.

After manually testing the current implementation against this one, I think we can offer a smoother experience. Here's the difference:

1. This approach infers the runtime numbers assigned by Minitest using static analysis. Since the collector performs a depth-first search on the Abstract Syntax Tree (AST) of the code, the numbers are assigned correctly at runtime in Minitest Spec style.
2. We can decouple the IDs from the labels, which could lead to more consistent naming between the code and what users see in the VS Code test panel.

I've reused the up/down counter implementation during the visits of the nodes, similar to how it's done in the current code lens in lib/ruby_lsp/listeners/code_lens.rb.

@vinistock What are your thoughts on this?  Happy to have feedback on this.

domingo2000:
No, currently because the numbers of the tests created are given by the line number, these numbers are not generated correctly for the spec style. 

Simplifiying the example. For the following ruby file

```ruby
# frozen_string_literal: true
require "minitest/autorun"

describe "Second Spec" do
  it "test one" do
    assert false
  end

  it "test two" do
    assert true
  end
end
```

The commands generated for the `it` cases are
`exe/ruby-lsp-test-exec bundle exec ruby -Itest /home/domingo/ruby/ruby-lsp/spec/some_test.rb --name "/^Second\ Spec#test_0004_test one\$/"`
And
`exe/ruby-lsp-test-exec bundle exec ruby -Itest /home/domingo/ruby/ruby-lsp/spec/some_test.rb --name "/^Second\ Spec#test_0008_test two\$/"`

The numbers `0004` and `0008` based on the line number does not correspond to minitest tests methods created so the commands resolved for the code lens are wrong and does not run any test.

But if we change the strategy for the number assigned from line number to a counter that

1. Goes up for each tests in the group
2. Resets for each group

Just like [minitest here](https://github.com/minitest/minitest/blob/master/lib/minitest/spec.rb#L227).

Because the visitor visits the nodes in the AST in the same order (DFS) that execution run the methods `describe/it` in the DSL, we can match almost all ids generated by minitest spec correctly. 

Minitest suite shuffle the tests _after_ assigning the ids so it does not matter the order of execution of each tests, it only matters the name of the methods generated, and that generation will always have the same order.

Pros:
- In describe/it specs without dynamic tests the ids matches the methods, making test commands work properly.
Cons:
- Dynamically defined tests could mess up the counters in the static analysis. (But dynamically defined tests cases are not supported).

vinistock:
We cannot guarantee that static analysis will discover tests in the same exact order as Minitest and any mismatches will make the reporting of results fail.

I would rather we went with something like this https://github.com/Shopify/ruby-lsp/pull/3514. Will that solve the problem you mentioned?

PR #3510, opened 2025-05-20
Bump new explorer rollout to 15%

vinistock:
### Motivation

We fixed a bunch of bugs and shortcomings of the experience that we identified during the 5% rollout. We should be a in a good place to bump to 15% and start getting more feedback.

### DIFF
--- a/vscode/src/common.ts
+++ b/vscode/src/common.ts
@@ -88,7 +88,7 @@ export const SUPPORTED_LANGUAGE_IDS = ["ruby", "erb"];
 export const FEATURE_FLAGS = {
   tapiocaAddon: 1.0,
   launcher: 0.1,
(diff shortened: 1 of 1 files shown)


Morriar (approved):

vscode/src/common.ts
```diff
@@ -88,7 +88,7 @@ export const SUPPORTED_LANGUAGE_IDS = ["ruby", "erb"];
 export const FEATURE_FLAGS = {
   tapiocaAddon: 1.0,
   launcher: 0.1,
-  fullTestDiscovery: 0.05,
+  fullTestDiscovery: 0.15,
```
Morriar, line 91: Should we be more aggressive or are we concerned by the amount of support / bug report this may create?

PR #3517, opened 2025-05-21
Fix commands for groups inside class in minitest spec style

### LINKED ISSUE #3516: Describe specs in minitest spec style inside class gives incorrect ids for describe blocks.
### Description

The Minitest Spec style is not working when there are describe specs inside a ruby class that inherits from `Minitest::Spec`. 

Editor: VScode
RubyLsp: Latest commit
Project: inside RubyLsp repository

### Reproduction steps

```ruby
# ./spec/some_test.rb
require "minitest/autorun"

class SomeClass < Minitest::Spec
  describe "foo" do
    it "bar" do
      assert true
    end
  end
end
```

The commands generated are 

1. For `describe "foo"`: `exe/ruby-lsp-test-exec bundle exec ruby -Itest /home/domingo/ruby/ruby-lsp/spec/some_test.rb --name "/^foo(#|::)/"` ❌. It should be `/^SomeClass::foo(#|::)/`
3. For `it "bar"`: `exe/ruby-lsp-test-exec bundle exec ruby -Itest /home/domingo/ruby/ruby-lsp/spec/some_test.rb --name "/^SomeClass::foo#test_\d{4}_bar\$/"` ✅ 

The ids are incorrect for the describe spec inside the class because is not prepending the class id.

domingo2000:
### Motivation

Closes #3516

The minitest spec style commands where not resolved correctly inside classes.

### Implementation

Instead of using the `label` of the test groups use the `id` because the id contains all the information.

### Automated Tests

Added 1 tests with the case to run the tests of just a group.

### Manual Tests

Tested against this example

```ruby
# ./spec/some_test.rb
require "minitest/autorun"

class SomeTest < Minitest::Spec
  describe "foo" do
    it "bar" do
      assert true
    end
  end
end
```

And works fine.

### DIFF
--- a/lib/ruby_lsp/listeners/test_style.rb
+++ b/lib/ruby_lsp/listeners/test_style.rb
@@ -44,7 +44,7 @@ module RubyLsp
               # If all of the children of the current test group are other groups, then there's no need to add it to the
               # aggregated examples
               unless children.any? && children.all? { |child| child[:tags].include?("test_group") }
-                aggregated_tests[path][item[:label]] = { tags: tags, examples: [] }
+                aggregated_tests[path][item[:id]] = { tags: tags, examples: [] }
               end
             else
               class_name, method_name = item[:id].split("#")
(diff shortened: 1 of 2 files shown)


vinistock (approved):
Nice. Thank you, Domingo!

PR #3523, opened 2025-05-22
Allow to disable any addon in the addon settings.

### LINKED ISSUE #3418: Allow disabling project addons

### Use case

I want to execute Ruby-LSP as vanilla as possible, without any interference of gems in the project, such as Standard/Rubocop/etc.

### Description

I see that this is [partially implemented on Server](https://github.com/Shopify/ruby-lsp/blob/main/lib/ruby_lsp/server.rb#L176) class, but it's currently [not exposed](https://github.com/Shopify/ruby-lsp/blob/main/lib/ruby_lsp/server.rb#L357) through any interface.

Are there any plans to allow users to permit/block project addons?

### Implementation

I would keep it under an initialization flag.

domingo2000:
### Motivation

Closes #3418 

### Implementation

In addon loading check for the addon settings first. The addon can be disabled using

```json
"rubyLsp.addonSettings": {
    "RuboCop": {
      "enabled": false
    }
  }
```

We make the addons enabled by default.

### Automated Tests

Added tests when the addon is enabled and when the addon is disabled. The other tests are already testing if the setting is not set.

### Manual Tests

Tests in the same repo enabling/disabling a dummy addon.

[Screencast from 21-05-25 22:27:12.webm](https://github.com/user-attachments/assets/4e778adc-bd2b-41c1-8c7e-8094139bd5e5)

### DIFF
--- a/lib/ruby_lsp/addon.rb
+++ b/lib/ruby_lsp/addon.rb
@@ -78,6 +78,12 @@ module RubyLsp
         self.addons = addon_classes.map(&:new)
         self.file_watcher_addons = addons.select { |addon| addon.respond_to?(:workspace_did_change_watched_files) }
 
+        # Disable addons with { "rubyLsp.addonSettings": {"Addon Name": { enabled: false }}}
+        self.addons = addons.filter do |addon|
+          addon_settings = global_state.settings_for_addon(addon.name) || {}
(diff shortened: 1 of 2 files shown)


domingo2000:
I can also write some documentation for this if needed. Just let me know where it should go and what kind of UX writing would work best. 📝

PR #3527, opened 2025-05-22
Avoid shitfting editor focus when using test code lens

vinistock:
### Motivation

We received feedback from users that the shifting of editor focus when running a test in the terminal via code lens is surprising and not the desired behaviour. Let's prevent the focus from changing and let people manually expand the explorer if desired.

I still kept the focus change for when someone uses the `Run` code lens. I think that's sensible since the user will probably want to see the test results and those will only show up there.


st0012 (approved):
Thanks for fixing this 😂

PR #3546, opened 2025-05-27
Update Stray VS Code Workspace Name in Docs

bfad:
When [updating the VS Code workspace file name][Commit], it didn't get fully updated in the docs. This updates one of the places that had the old name.

[Commit]: https://github.com/Shopify/ruby-lsp/commit/2be659e375cf0cee8a8c161651e96ac75a192d56


vinistock (approved):
Thanks!

Issue #3528, opened 2025-05-23
Internal writer methods complete incorrectly


joeldrapper:
#### VS Code Version

1.100.2

#### Ruby LSP Extension Version

0.9.25

#### Ruby LSP Server Version

0.23.17

#### Ruby LSP Add-ons

- RuboCop
- Ruby LSP My Gem

#### Ruby Version

3.4.3

#### Ruby Version Manager

mise

#### Installed Extensions

<details>
<summary>Click to expand</summary>

- ruby-lsp (0.9.25)
</details>

#### Ruby LSP Settings

<details>
<summary>Click to expand</summary>

##### Workspace

```json
{}
```

##### User

```json
{
  "enabledFeatures": {
    "codeActions": true,
    "diagnostics": true,
    "documentHighlights": true,
    "documentLink": true,
    "documentSymbols": true,
    "foldingRanges": true,
    "formatting": true,
    "hover": true,
    "inlayHint": true,
    "onTypeFormatting": true,
    "selectionRanges": true,
    "semanticHighlighting": true,
    "completion": true,
    "codeLens": true,
    "definition": true,
    "workspaceSymbol": true,
    "signatureHelp": true,
    "typeHierarchy": true
  },
  "featuresConfiguration": {},
  "addonSettings": {},
  "rubyVersionManager": {
    "identifier": "auto"
  },
  "customRubyCommand": "",
  "formatter": "auto",
  "linters": null,
  "bundleGemfile": "",
  "testTimeout": 30,
  "branch": "",
  "pullDiagnosticsOn": "both",
  "useBundlerCompose": false,
  "bypassTypechecker": false,
  "rubyExecutablePath": "",
  "indexing": {},
  "erbSupport": true,
  "featureFlags": {},
  "sigOpacityLevel": "1"
}
```

</details>

### Reproduction steps

1. Open any class with a writer method from attr_accessor, attr_writer, or one you defined manually.
2. In an instance method, accept completion for the writer method.
3. It will complete the writer method as `method_name=` instead of `self.method_name = `

vinistock:
Thanks for the report! Indeed, for methods that end with `=` invoked on self we need to include the explicit `self.` in front.

PR #3550, opened 2025-05-28
Complete writer methods with self receiver when receiver is self

antw:
### Motivation

Closes https://github.com/Shopify/ruby-lsp/issues/3528

Setter methods where the receiver is `self` were being completed without the explicit `self.` prefix. This would result in local variable assignment rather than calling the writer method.

For example, when typing `b` in a class method and expecting completion for `baz=`, the LSP would suggest:

```ruby
def bar
  baz= # This assigns to a local variable, not calling the writer method
end
```

Instead of the correct:
```ruby
def bar
  self.baz=
end
```

### Implementation

- Modified the `add_method_completions` method in `completion.rb` to detect when completing setter methods with a `self` receiver.
- The changes only affect files where Sorbet type checking is not enabled; those use Sorbet's completion, which has the same issue, but is outside the scope of Ruby LSP (as far as I know).

The implementation checks `node.receiver.nil?` to detect implicit self and avoid adding `self.` when the user has already typed the receiver.

> [!NOTE]
> This is my first ruby-lsp PR. Feedback on the approach would be appreciated, particularly:
> 
> - I read through [the documentation for CompletionItem](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#completionItem) and it _feels_ like setting the `newText` is the correct option (leaving the `label` and `filterText` as before) but I'd be curious if reviewers feel otherwise.
> - I put spaces around the "=" as suggested in the issue since that's how most people would write this..

### Automated Tests

Added test coverage in `completion_test.rb` with three scenarios:
- Implicit self receiver (`b` in class method) - verifies `self.` prefix is added to completion text
- Explicit self receiver (`self.b` in class method) - verifies no additional `self.` prefix is added 
- External receiver (`foo.b` outside class) - verifies no `self.` prefix is added

### DIFF
--- a/lib/ruby_lsp/listeners/completion.rb
+++ b/lib/ruby_lsp/listeners/completion.rb
@@ -522,7 +522,7 @@ module RubyLsp
             method_name = entry_name.delete_suffix("=")
 
             # For writer methods, format as assignment and prefix "self." when no receiver is specified
-            new_text = node.receiver.nil? ? "self.#{method_name}=" : method_name
+            new_text = node.receiver.nil? ? "self.#{method_name} = " : "#{method_name} = "
           end
 
           label_details = Interface::CompletionItemLabelDetails.new(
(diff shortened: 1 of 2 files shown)


vinistock (approved):
Nice. Thank you!

PR #3553, opened 2025-05-28
Add hover documentation for reserved keywords

mikeygough:
Partially closes https://github.com/Shopify/ruby-lsp/issues/2227

This PR adds hover documentation for the following Ruby reserved keywords:

* `break`
* `case`
* `class`
* `def`
* `defined?`
* `else`
* `ensure`
* `for`
* `module`
* `next`
* `rescue`
* `return`
* `undef`
* `unless`
* `until`
* `when`
* `while`

https://github.com/user-attachments/assets/4408310e-f558-4274-a0c8-737ec17301f1

This documentation was written by AI following [Shopify's Ruby style guide](https://ruby-style-guide.shopify.dev/). It has been reviewed by myself but would appreciate additional, more experienced eyes!

*NOTE* - The following keywords still need definitions which may require modifying their associated `Prism Node`:

* `alias`
* `and`
* `begin`
* `do`
* `elsif`
* `end`
* `false`
* `if`
* `in`
* `nill`
* `not`
* `or`
* `redo`
* `retry`
* `self`
* `super`
* `then`
* `true`

### Motivation

We can deliver a better Ruby programming experience if developers can hover over built-in keywords to learn more about how they work!

### Implementation

This implementation follows a pattern established by the `yield` keyword. It adds a key value pair to the `KEYWORD_DOCS' hash inside `static_docs.rb`, writes a keyword specific markdown file that goes into a deeper explanation and extends `hover.rb` to handle keyword specific hover events.

### Automated Tests

Run `dev t`

### Manual Tests

Start the Ruby LSP. Use some keywords and checkout the tooltips + related documentation!


static_docs/case.md
```diff
@@ -0,0 +1,99 @@
...
+
+case day
+when "Saturday", "Sunday"
+  puts "Weekend!"
+else
+  puts "Weekday"
+end
+```
+
+## Pattern Matching (Ruby 2.7+)
```
vinistock, line 49: No need to mention that it's 2.7+. The Ruby LSP only supports 3.0 or higher and 2.7 has been EOL for ages.

static_docs/class.md
```diff
@@ -0,0 +1,143 @@
...
+  end
+end
+
+dog = Dog.new
+puts dog.speak # Output: Woof!
+```
+
+## Access Control
+
+Ruby provides three levels of method access control: `public`, `private`, and `protected`.
```
vinistock, line 83: Similarly, this may be a better fit for when you hover over `private`, `public` and `protected` as to not make this document too overwhelming.

static_docs/class.md
```diff
@@ -0,0 +1,143 @@
...
+    @balance += amount
+  end
+end
+
+account = BankAccount.new(100)
+puts account.display_balance
+# Output: Current balance: $100
+```
+
+## Class Instance Variables
```
vinistock, line 116: I'd mention attr_reader, attr_writer and attr_accessor here and leave class level stuff for a different doc.

PR #3554, opened 2025-05-29
Create Cop to catch YARD tag annotations for method signatures

bfad:
### Motivation

This creates a cop that flags YARD comments with method signature tags. The type annotations for YARD are not very robust. The only benefit to having them over using RBS-style comments is to document what the parameters are, but most methods should have good names that don't make this useful.

### Implementation

YARD tags can be wrapped over multiple comment lines. This solution creates an Enumerator that generates an array of comment lines for each YARD tag. It then checks if the tag is one of the method annotation tags and, if so, adds an error to the first comment line of the tag (the line with the tag).

### Automated Tests

Yes


bfad:
Closing this as it needs to be a Sorbet Cop in the `rubocop-sorbet` repo.

Issue #3111, opened 2025-01-30
Implement find references and rename for locals


yedhink:
#### VS Code Version

1.97.0-insider

#### Ruby LSP Extension Version

0.8.19

#### Ruby LSP Server Version

0.23.7

#### Ruby LSP Add-ons

- Ruby LSP Rails
- RuboCop

#### Ruby Version

3.4.1

#### Ruby Version Manager

rbenv

#### Installed Extensions

<details>
<summary>Click to expand</summary>

- catppuccin-vsc (3.16.0)
- dotenv (1.0.1)
- prettier-vscode (11.0.0)
- ruby-extensions-pack (0.1.12)
- ruby-lsp (0.8.19)
- sorbet-vscode-extension (0.3.37)
- vim (1.29.0)
- vscode-eslint (3.0.10)
- vscode-mdx (1.8.13)
- vscode-ruby (0.28.0)
</details>

#### Ruby LSP Settings

<details>
<summary>Click to expand</summary>

##### Workspace

```json
{}
```

##### User

```json
{
  "enabledFeatures": {
    "codeActions": true,
    "diagnostics": true,
    "documentHighlights": true,
    "documentLink": true,
    "documentSymbols": true,
    "foldingRanges": true,
    "formatting": true,
    "hover": true,
    "inlayHint": true,
    "onTypeFormatting": true,
    "selectionRanges": true,
    "semanticHighlighting": true,
    "completion": true,
    "codeLens": true,
    "definition": true,
    "workspaceSymbol": true,
    "signatureHelp": true,
    "typeHierarchy": true
  },
  "featuresConfiguration": {},
  "addonSettings": {},
  "rubyVersionManager": {
    "identifier": "auto"
  },
  "customRubyCommand": "",
  "formatter": "auto",
  "linters": null,
  "bundleGemfile": "",
  "testTimeout": 30,
  "branch": "",
  "pullDiagnosticsOn": "both",
  "useBundlerCompose": false,
  "bypassTypechecker": false,
  "rubyExecutablePath": "",
  "indexing": {},
  "erbSupport": true,
  "featureFlags": {}
}
```

</details>

### Reproduction steps

<img width="1512" alt="Image" src="https://github.com/user-attachments/assets/e62fa0a2-20d2-4a6b-a23c-0b1333370ce6" />

Add the following test case in Rails 8 and try to rename the variable `password`:

```rbi
  def test_password_is_not_too_short
[...]

Issue #2642, opened 2024-10-01
Add references support for local variables


vinistock:
Build on top of #2632 to add support for finding local variable references. Note that in order to implement this, the reference finder must now keep track of lexical scopes so that we appropriately identify local variables and their occurrences.

PR #3558, opened 2025-05-29
Add support for local variable references

nunosilva800:
### Motivation

Closes https://github.com/Shopify/ruby-lsp/issues/2642
Closes https://github.com/Shopify/ruby-lsp/issues/3111

### Implementation

1. Made the RubyIndexer::ReferenceFinder register `on_local_variable_**_enter` with the `Prism::Dispatcher`, and collect references to it while doing so.
2. `RubyLsp::Requests::References` now expects `Prism::LocalVariable**Node` and when the target is a `RubyIndexer::ReferenceFinder::LocalVariableTarget`, it collects references in the current ruby document and only under the parent of the target. This should limit the scope.
3. `RubyLsp::Requests::Rename` and `RubyLsp::Requests::PrepareRename` also expects `Prism::LocalVariable**Node` and follows same logic as the Reference request resolver to find and rename the local var.

Notes:

1. the usage of `Prism::RequiredParameterNode` to understand when a local var might be defined in a block argument - there might be better ways of doing this (as well as considering optional parameters?)

2. since the search for local var references is limited to the parent of target (what is under the cursor), when the var is inside a block, any reference to the var outside the block is not found. This is something that perhaps can be improved.

### Automated Tests

Unit tests were added.

### Manual Tests

Cannot use it manually... after following https://shopify.github.io/ruby-lsp/contributing.html#live-debugging I still see:

image

Which means something is missing from this PR 👀

### DIFF
--- a/lib/ruby_lsp/requests/prepare_rename.rb
+++ b/lib/ruby_lsp/requests/prepare_rename.rb
@@ -24,7 +24,18 @@ module RubyLsp
         node_context = RubyDocument.locate(
           @document.parse_result.value,
           char_position,
-          node_types: [Prism::ConstantReadNode, Prism::ConstantPathNode, Prism::ConstantPathTargetNode],
+          node_types: [
+            Prism::ConstantReadNode,
+            Prism::ConstantPathNode,
+            Prism::ConstantPathTargetNode,
+            Prism::LocalVariableAndWriteNode,
+            Prism::LocalVariableOperatorWriteNode,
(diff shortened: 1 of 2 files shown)


nunosilva800:
@vinistock any clues on why I cannot test this with live debugging? 
Also, what other examples of local var references/rename can you think of that should be tested? The current implementation is quite limited, but having other examples might guide me towards a better solution.

PR #3560, opened 2025-05-29
feat: register profile command

### LINKED ISSUE #2273: Add in-editor profiling with Vernier
It would be amazing if we could profile a given Ruby script/executable using [vernier](https://github.com/jhawthorn/vernier) and then show the results inside of the editor.

VS Code doesn't have profiling APIs yet and I'm not sure if this is planned or not. Which means we need to build everything custom. However, we may be able to show the generated webpage that you'd normally view in the Firefox profiler view within a VS Code webview.

maxveldink:
### Motivation

Closes #2273 
Requires this [Vernier PR](https://github.com/jhawthorn/vernier/pull/154) to ship and a new version to be released. ✅

This adds a way to profile the currently open file using Vernier, and display the output in editor (VSCode will also display numbers inline with the file).

### Implementation

Registers a command that uses Vernier under the hood with the active file and open the `.cpuprofile` file.

### Automated Tests

Since this was a straightforward command and I didn't see tests for similar commands, I opted to not add a test here. Happy to extract this to a function and test it if we'd prefer to do so.

### Manual Tests

1. Run the extension, using a project with a Ruby workspace that as a Ruby script that can be profiled.
2. Run the `Profile current file` command (either from the palette or editor menu of active file)
3. See a `.cpuprofile` open. When you click back to the profiled file, VSCode should automatically annotate lines with perf information.

https://github.com/user-attachments/assets/f513d7a5-7a93-4047-9764-a59115db45e2

### DIFF
--- a/vscode/src/rubyLsp.ts
+++ b/vscode/src/rubyLsp.ts
@@ -1,7 +1,8 @@
-import * as vscode from "vscode";
-import { Range } from "vscode-languageclient/node";
 import * as os from "os";
 import * as path from "path";
+
+import * as vscode from "vscode";
+import { Range } from "vscode-languageclient/node";
 
 import DocumentProvider from "./documentProvider";
 import { Workspace } from "./workspace";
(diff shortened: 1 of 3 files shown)


vscode/src/rubyLsp.ts
```diff
@@ -737,6 +740,74 @@ export class RubyLsp {
...
+      vscode.commands.registerCommand(Command.ProfileCurrentFile, async () => {
+        const workspace = this.currentActiveWorkspace();
+
+        if (!workspace) {
+          vscode.window.showInformationMessage("No workspace found");
+          return;
+        }
+
+        try {
+          const { stdout } = await workspace.execute("vernier --version");
```
maxveldink, line 752: This command won't work until https://github.com/jhawthorn/vernier/commit/59d9137d18bc7b512477d6a4fd7886578c28e86d ships

vinistock (approved):
Nice. Thanks!

PR #3563, opened 2025-05-30
Do not collect trailing comments of previous lines

jesse-shopify:
### Motivation

- Comments for ivars were including trailing comments from previous lines. With RBS, these trailing comments are often RBS typing information.
- In general, a trailing comment is assumed to be for the line it is on, not subsequent lines

### Implementation

- When collecting comments, encountering a trailing comment can stop the collection because no other comments will be for the current node

### Automated Tests

- test added

### Manual Tests

- a common case will be ivars in initialize with RBS typing, eg

```ruby
class Foo
  def initialize
    @a = "Hello" #: String
    @b = 123
  end
end
```

- Hoevering over `@b` should not display `:String`


lib/ruby_indexer/test/instance_variables_test.rb
```diff
@@ -236,5 +236,29 @@ def something; end
...
+        end
+      RUBY
+
+      assert_entry("@a", Entry::InstanceVariable, "/fake/path/foo.rb:2-4:2-6")
+      entry = @index["@a"]&.first #: as Entry::InstanceVariable
+      assert_equal("Documentation for @a", entry.comments)
+
+      assert_entry("@b", Entry::InstanceVariable, "/fake/path/foo.rb:3-4:3-6")
+      entry = @index["@b"]&.first #: as Entry::InstanceVariable
+      assert_equal("", entry.comments)
```
vinistock, line 257:
```suggestion
      assert_empty(entry.comments)
```

lib/ruby_indexer/test/instance_variables_test.rb
```diff
@@ -236,5 +236,29 @@ def something; end
...
+
+      assert_entry("@c", Entry::InstanceVariable, "/fake/path/foo.rb:4-4:4-6")
+      entry = @index["@c"]&.first #: as Entry::InstanceVariable
+      assert_equal("", entry.comments)
```
vinistock, line 261:
```suggestion
      assert_empty(entry.comments)
```

vinistock (approved):

PR #3585, opened 2025-06-04
Provide --update-server command for ruby-lsp exec

### LINKED ISSUE #3124: Provide --update-server command from executable
The relationship between the global `ruby-lsp` executable version vs the server version that actually gets run is often a source of confusion to users. Additionally, if some other gem is constraining our upgrades, users sometimes have trouble identifying which gem is holding back the upgrade.

On top of those, users with relative paths in their Bundler settings don't have a very easy way to upgrade because Bundler considers the Gemfile path to be the root of everything. If you have `BUNDLE_PATH=vendor/bundle`, then running this

```shell
BUNDLE_GEMFILE=.ruby-lsp/Gemfile bundle update ruby-lsp
```

Will install gems in `.ruby-lsp/vendor/bundle` because the path becomes relative to the Gemfile. All of this is not great and we can provide better tools for upgrades and diagnosing issues.

Let's add a `ruby-lsp --update-server` command that will simply run `bundle update` using our composed environment, with all Bundler settings turned into absolute paths (which is what we do while regular composition). Furthermore, if a gem is preventing the upgrade, it would be awesome if we could inform the user about which one it is.

vicocamacho:
### Motivation

Closes #3124

This PR adds a `--update-server` option to the `ruby-lsp` command to provide users with the ability to easy upgrade the gem's version and automatically used the correct settings to perform the upgrade.

### Implementation

I leveraged the existing `SetupBundler` class to pull the correct environment configuration and then I use that to run the bundle update command.

### Automated Tests

I added one automated integration test to verify the execution.

### Manual Tests

I have run some manual tests in my local machine.

### DIFF
--- a/exe/ruby-lsp
+++ b/exe/ruby-lsp
@@ -37,6 +37,27 @@ parser = OptionParser.new do |opts|
     options[:launcher] = true
   end
 
+  opts.on("--update-server", "Update the Ruby LSP server") do
+    puts("Ruby LSP> Executing server update...")
+    require_relative "../lib/ruby_lsp/setup_bundler"
+
+    begin
+      env = RubyLsp::SetupBundler.new(Dir.pwd, **options).setup!
+    rescue RubyLsp::SetupBundler::BundleNotLocked
+      warn("Project contains a Gemfile, but no Gemfile.lock. Run `bundle install` to lock gems")
+      exit(78)
+    end
+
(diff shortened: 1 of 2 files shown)


vicocamacho:
@vinistock would you mind taking a look and verifying that this is doing what the PR is asking?

PR #3252, opened 2025-02-26
Use `parse_lex` instead of `parse` for Ruby and ERB documents

vinistock:

PR #3590, opened 2025-06-06
Update `collect_references` to use `Prism::LexResult`

st0012:
### Motivation

In #3252, `RubyDocument#parse_result` becomes `Prism::ParseLexResult`, which's `value` is an array instead of a single node. And we forgot to update `collect_references` to account for this change because the codepath was not covered in the test.

So this commit adds a test for the codepath and updates `collect_references` to use `Prism::LexResult`.


alexcrocha (approved):
Thank you 🙏

PR #3611, opened 2025-06-17
Collect add-ons versions for issue reporting

### LINKED ISSUE #3599: The `Collect Ruby LSP information for issue reporting` command should report addon versions too
https://github.com/Shopify/ruby-lsp/blob/4538672242d2911ce71fed1af3990fc6b08d27a7/vscode/src/infoCollector.ts#L30

snutij:
### Motivation

Closes #3599 

Collect the version for each add-on, with a fallback to "unknown" for those without a specified version.

### Implementation

Nothing too fancy, just using a template string.

### Automated Tests

I do not think there is already testing covering this command; I can handle it if needed.

### Manual Tests

https://github.com/user-attachments/assets/4e2d86c1-1076-4ccf-b52d-ae80ba1eca6e


vinistock (approved):

vinistock:
Thanks!

PR #3632, opened 2025-06-26
Allow users to disable test related code lenses

### LINKED ISSUE #2167: Too many code lens options for running tests
The code lens feature to run tests is great! 🎉 

But right now, there are triggers everywhere:

1. Run all the tests
2. Run all the tests in the file
3. Hover a specific test to run it
4. Gutter trigger for a specific test (same thing as 3)
5. Inline trigger for a test (same thing as 3)

To be fair, 1-2 are different, but 3-5 are the same thing.

I find 5 quite annoying, because it makes the test file hard to read and difficult to work with visually (moving blocks of code around, quickly scanning code, etc), with every few lines being broken with this magic extra line, with controls.

Obviously subjective, but since 4 is just a few pixels away and 3 also easy to reach, I prefer both of them to 5.

### Two suggestions (either would work)

A. Remove 5, and let people use 3 or 4 to run a test.
B. Make it configurable to remove/hide 5.

CleanShot 2024-06-12 at 11 00 47

vinistock:
### Motivation

Closes #2167

This PR allows users to configure their code lens request, giving them the ability to disable test related buttons.

**Question**: I thought providing 3 settings for controlling each one of the lenses (run, run in terminal and debug) could be a bit too much granularity. Do people agree? Or would people prefer being able to customize at that level?

### Implementation

Added another entry to our feature configuration object and started checking that before registering the test code lens listeners.

### Automated Tests

Added a test.

### DIFF
--- a/lib/ruby_lsp/global_state.rb
+++ b/lib/ruby_lsp/global_state.rb
@@ -62,6 +62,10 @@ module RubyLsp
           implicitRescue: false,
           implicitHashValue: false,
         }),
+        codeLens: RequestConfig.new({
+          enableAll: false,
+          enableTestCodeLens: true,
+        }),
       } #: Hash[Symbol, RequestConfig]
     end
 
--- a/lib/ruby_lsp/requests/code_lens.rb
+++ b/lib/ruby_lsp/requests/code_lens.rb
@@ -27,10 +27,14 @@ module RubyLsp
         @document = document
         @test_builder = ResponseBuilders::TestCollection.new #: ResponseBuilders::TestCollection
         uri = document.uri
+        code_lens_config = global_state.feature_configuration(:codeLens)
+        test_lenses_enabled = !code_lens_config || code_lens_config.enabled?(:enableTestCodeLens)
 
         if global_state.enabled_feature?(:fullTestDiscovery)
-          Listeners::TestStyle.new(@test_builder, global_state, dispatcher, uri)
-          Listeners::SpecStyle.new(@test_builder, global_state, dispatcher, uri)
+          if test_lenses_enabled
+            Listeners::TestStyle.new(@test_builder, global_state, dispatcher, uri)
+            Listeners::SpecStyle.new(@test_builder, global_state, dispatcher, uri)
+          end
         else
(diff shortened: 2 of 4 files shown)


vinistock:
* **#3632** <a href="https://app.graphite.dev/github/pr/Shopify/ruby-lsp/3632?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/Shopify/ruby-lsp/3632?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`
<hr />

You must have a Graphite account in order to use the merge queue. Sign up using [this link](https://app.graphite.dev/invite/github/Shopify?ref=merge-queue-instructions-in-stack-comment&prId=3632).

<h2></h2>
<sub>

</sub>

<sub>

</sub>

<h2></h2>

lib/ruby_lsp/global_state.rb
```diff
@@ -62,6 +62,10 @@ def initialize
           implicitRescue: false,
           implicitHashValue: false,
         }),
+        codeLens: RequestConfig.new({
+          enableAll: false,
+          enableTestCodeLens: true,
+        }),
```
alexcrocha, line 68: Can you clarify how `enableAll` works? 

Does `false` here mean that if `enableAll` is set to `true` in the users config, it will not affect this `codeLens` setting?

vinistock: If `enableAll` is `true`, then every setting is considered `true` regardless of their actual value. It forces you to opt-in to everything, so we need to make the default `false`.

However, for the test code lenses, we want the default to be enabled.

alexcrocha: That makes sense, but I was still confused about the implementation. So I checked how we're making use of `enableAll` and it all makes sense now 👍 

```ruby
    def enabled?(feature)
      @configuration[:enableAll] || @configuration[feature]
    end
```

alexcrocha (approved):

sandstrom:
Great! 💯

PR #3634, opened 2025-06-26
Support Windsurf for setting cursor position in snippets

### LINKED ISSUE #3633: Support Windsurf for placing cursor in correct place after auto complete end and |
Creating a block creates `end` for me, but positions the cursor after. Typing a `|` for a block arg types a closing `|` and positions the cursor after.

Typing `class Hotdog<enter>` yields
```rb
class Hotdog
end_
   ^-- cursor
```

Typing `things.each do |` yields
```rb
things.each do ||_
                 ^--- cursor after 2nd |
```

This creates quite a bit of extra chaos in an environment where AI is already trying to wreck your code for you ;)

This was presented in a [recent issue](https://github.com/Shopify/ruby-lsp/issues/3156) related to VSCodium, which was resolved by specifying [that fork in a list of clients able to position the cursor](https://github.com/Shopify/ruby-lsp/pull/3302).

There was also [some conversation](https://github.com/Shopify/ruby-lsp/pull/3302/files#r2000964016) about the whack-a-mole nature of supporting VSCode forks where Vini proposed using predicate methods instead of a string of all the forks under the sun.

I'd appreciate if Windsurf could be made compatible as well, either by way of whitelisting, or a more robust solution.

ttilberg:
### Motivation

Closes #3633 by repeating the work done for supporting VSCodium.

### Implementation

Copies the same pattern fixing a similar issue in VSCodium from https://github.com/Shopify/ruby-lsp/pull/3302

The discussion in that PR noted that there might be a better way, but if it's acceptable to repeat that style of support for now, I've presented a working PR that I've tested locally.

### Automated Tests

Yes, the tests were copied from the previous referenced implementation.

### Manual Tests

I made this change in my local `bundle open ruby-lsp` to check the correct key being `Windsurf`, and Windsurf then operated as expected.

Typing `class Hotdog<enter>` successfully added `end` with the cursor in the body instead of end.

```rb
class Hotdog
  _ #<-- cursor
end

### DIFF
--- a/lib/ruby_lsp/requests/on_type_formatting.rb
+++ b/lib/ruby_lsp/requests/on_type_formatting.rb
@@ -162,7 +162,7 @@ module RubyLsp
 
       #: (Integer line, Integer character) -> void
       def move_cursor_to(line, character)
-        return unless /Visual Studio Code|Cursor|VSCodium/.match?(@client_name)
+        return unless /Visual Studio Code|Cursor|VSCodium|Windsurf/.match?(@client_name)
 
         position = Interface::Position.new(
           line: line,
--- a/test/requests/on_type_formatting_test.rb
+++ b/test/requests/on_type_formatting_test.rb
@@ -1015,6 +1015,46 @@ class OnTypeFormattingTest < Minitest::Test
     assert_equal(expected_edits.to_json, edits.to_json)
   end
 
+  def test_includes_snippets_on_windsurf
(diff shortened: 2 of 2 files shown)


ttilberg:
I've now signed the CLA, but I don't believe I'm able to trigger a rerun without adding a useless commit.

vinistock (approved):
Thanks!

PR #3661, opened 2025-07-14
Use -r instead of RUBYOPT to require LSP reporters

vinistock:
### Motivation

I identified an issue we did not foresee. By requiring our custom reporters through `RUBYOPT`, we end up causing any subprocesses spawned by the user's code to also require them. This can lead to weird issues.

For example, if the user code raises, then our `at_exit` hook that checks for crashes will run and send a duplicate `finish` event to the explorer _after_ we already closed the current run request. Since there's no current run, the explorer then thinks that this is a test being executed from the terminal manually and creates a fresh run with zero results inside.

Note: it is possible that this is the infamous issue where spring prevents us from reporting test execution events, but I'm not 100% confident yet.

### Implementation

Instead of trying to sanitize `RUBYOPT`, which can become tricky very quickly, we should just move to using the `-r` command line argument for Ruby. That simply requires the reporter without cascading down to subprocesses and does what we want.

Note: this is a breaking change because the Rails add-on will now need to add the same require to its command.

### DIFF
--- a/lib/ruby_lsp/listeners/test_style.rb
+++ b/lib/ruby_lsp/listeners/test_style.rb
@@ -75,8 +75,9 @@ module RubyLsp
 
           unless full_files.empty?
             specs, tests = full_files.partition { |path| spec?(path) }
-            commands << "#{BASE_COMMAND} -Itest -e \"ARGV.each { |f| require f }\" #{tests.join(" ")}" if tests.any?
-            commands << "#{BASE_COMMAND} -Ispec -e \"ARGV.each { |f| require f }\" #{specs.join(" ")}" if specs.any?
+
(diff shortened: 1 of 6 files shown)


vinistock:
* **#3661** <a href="https://app.graphite.dev/github/pr/Shopify/ruby-lsp/3661?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/Shopify/ruby-lsp/3661?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`
<hr />

You must have a Graphite account in order to use the merge queue. Sign up using [this link](https://app.graphite.dev/invite/github/Shopify?ref=merge-queue-instructions-in-stack-comment&prId=3661).

<h2></h2>
<sub>

</sub>

<sub>

</sub>

<h2></h2>

st0012 (approved):

PR #3668, opened 2025-07-17
Add `CLAUDE.md`

st0012:
### Motivation

Claude Code's performance could be significantly better in project that have a well written & correct `CLAUDE.md` to provide basic context, which tools like Copilot seem to recognise automatically too. So I think we should add it to make this project easier to work with for AI coding agents.

### Implementation

I used Claude Code's `/init` command to generate the document and fixed/improved it with @vinistock.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -1,0 +1,235 @@
+# CLAUDE.md
+
+This file provides guidance to Claude Code (claude.ai/code) when working with code in this repository.
+
+## Project Overview
+
+Ruby LSP is a Language Server Protocol implementation for Ruby, providing IDE features across different editors. The project consists of:
+
+- Core Ruby LSP gem (language server implementation)
+- Ruby Indexer (code indexing for navigation)
+- VS Code extension (TypeScript frontend)
+- Documentation site (Jekyll-based)
+
+## Development Commands
+
+### Ruby LSP Core
+
+```bash
+# Run all tests
+bundle exec rake
+
+# Run specific test file
+bin/test test/requests/completion_test.rb
+
+# Run tests matching a pattern
+bin/test -n /test_name_pattern/
+
+# Type check with Sorbet
+bundle exec srb tc
+
+# Lint with RuboCop
+bin/rubocop
+
+# Auto-fix RuboCop violations
+bin/rubocop -a
+```
+
+### VS Code Extension
+
+```bash
+cd vscode
+yarn install       # Install dependencies
+yarn run lint      # Lint TypeScript code
+yarn run test      # Run extension tests
+```
+
+### Executables
+
+- `exe/ruby-lsp` - Main language server executable
+- `exe/ruby-lsp-check` - Validation tool
+- `exe/ruby-lsp-launcher` - Experimental launcher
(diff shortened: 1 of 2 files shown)


alexcrocha (approved):
🚀

CLAUDE.md
```diff
@@ -0,0 +1,235 @@
...
+
+```bash
+# Run all tests
+bundle exec rake
+
+# Run specific test file
+bin/test test/requests/completion_test.rb
+
+# Run tests matching a pattern
+bin/test -n /test_name_pattern/
```
vinistock: Not quite. This should be `bin/test path name_pattern`

CLAUDE.md
```diff
@@ -0,0 +1,235 @@
...
+2. Register in `lib/ruby_lsp/executor.rb`
+3. Add tests in `test/requests/`
+4. Update VS Code extension if needed
+
+### Working with the Indexer
+
+The Ruby Indexer (`lib/ruby_indexer/`) handles:
+
+- Building symbol tables
+- Resolving constant references
```
vinistock:
```suggestion
- Resolving constant, instance variable, class variable, global variable and method references
```

CLAUDE.md
```diff
@@ -0,0 +1,235 @@
...
+- Finding definitions across files
```
vinistock, line 184:
```suggestion
- Finding definitions across files
- Providing completion candidates
- Dealing with inheritance and ancestor linearization
```

CLAUDE.md
```diff
@@ -0,0 +1,235 @@
...
+
+# Union and nullable types
+result = nil #: (String | Symbol)?
+```
+
+**Type Syntax Reference:** <https://sorbet.org/docs/rbs-support>
+
+### Debugging
+
+1. Enable debug logging: Set `RUBY_LSP_DEBUG=true`
```
vinistock: I don't think we have this env variable.

CLAUDE.md
```diff
@@ -0,0 +1,235 @@
...
+2. Use VS Code launch configurations in `.vscode/launch.json`
+3. Add `binding.irb` for debugging Ruby code
+
+## Important Files
```
vinistock: This section is duplicate. Can we dedup this with the previous one?

ruby-lsp.code-workspace
```diff
@@ -34,7 +34,7 @@
         "typescript.tsc.autoDetect": "off",
         "github.copilot.chat.codeGeneration.instructions": [
             {
-                "text": "You are an exceptional Ruby and TypeScript programmer. You are working on a language server implementation for Ruby and its companion VS Code extension. Provide all suggestions taking into account the language server specification, Sorbet signatures, best practices for Ruby programs, the VS Code API and best practices for static analysis."
+                "text": "You are an exceptional Ruby and TypeScript programmer. Use the CLAUDE.md file to guide your work."
```
vinistock: How about we instead create the `.github/` instructions file as a symlink that points to `CLAUDE.md`?

vinistock (approved):
🚀

Issue #3279, opened 2025-03-07
Format on save not working in VS Code


graygilmore:
#### VS Code Version

1.96.2

#### Ruby LSP Extension Version

0.9.7

#### Ruby LSP Server Version

0.23.11

#### Ruby LSP Add-ons

- Ruby LSP Rails
- Tapioca
- RuboCop

#### Ruby Version

3.4.1

#### Ruby Version Manager

shadowenv

#### Installed Extensions

<details>
<summary>Click to expand</summary>

- EditorConfig (0.17.1)
- explorer (1.16.0)
- gitlens (16.3.2)
- polaris-for-vscode (0.9.8)
- prettier-vscode (11.0.0)
- remote-ssh (0.113.1)
- remote-ssh-edit (0.87.0)
- ruby-extensions-pack (0.1.13)
- ruby-lsp (0.9.7)
- sorbet-vscode-extension (0.3.40)
- theme-check-vscode (3.6.0)
- vscode-bazel (0.11.0)
- vscode-eslint (3.0.10)
- vscode-graphql (0.13.0)
- vscode-graphql-syntax (1.3.8)
- vscode-rdbg (0.2.2)
- vscode-stylelint (1.4.0)
- vscode-theme-onelight (2.3.0)
</details>

#### Ruby LSP Settings

<details>
<summary>Click to expand</summary>

##### Workspace

```json
{
  "testTimeout": 300,
  "featureFlags": {
    "tapiocaAddon": true
  }
}
```

##### User

```json
{
  "enabledFeatures": {
    "codeActions": true,
    "diagnostics": true,
    "documentHighlights": true,
    "documentLink": true,
    "documentSymbols": true,
    "foldingRanges": true,
    "formatting": true,
    "hover": true,
    "inlayHint": true,
    "onTypeFormatting": false,
    "selectionRanges": true,
    "semanticHighlighting": true,
    "completion": true,
    "codeLens": true,
    "definition": true,
    "workspaceSymbol": true,
    "signatureHelp": true,
    "typeHierarchy": true
  },
  "featuresConfiguration": {},
  "addonSettings": {},
  "rubyVersionManager": {
    "identifier": "auto"
  },
  "customRubyCommand": "",
  "formatter": "auto",
  "linters": null,
  "bundleGemfile": "",
  "testTimeout": 300,
  "branch": "",
  "pullDiagnosticsOn": "both",
  "useBundlerCompose": false,
  "bypassTypechecker": false,
[...]

PR #3669, opened 2025-07-17
Avoid requiring add-ons that are copied into the workspace

vinistock:
### Motivation

Addresses https://github.com/Shopify/ruby-lsp/issues/3279#issuecomment-3080154096

If users have gems copied inside their workspace without pointing their BUNDLE_PATH to it, requiring project add-ons will require incorrect versions of gems by accident - leading to unpredictable behaviour.

In #3279, it seems that the workspace had gems in `vendor/bundle` despite `BUNDLE_PATH` not pointing to it. In that situation, we get the reported bug. This is a pretty odd setup, but we shouldn't crash in hard to debug ways.

### Implementation

The essence of the problem is discovering an add-on for the same gem twice. The correct discovery **always** comes from `Gem.find_files`, so I propose that we create an array of glob patterns to reject from the list of correctly discovered add-ons.

Then we reject any occurrences inside of the workspace that match the glob pattern, regardless of the gem version. Hopefully, this can avoid any form of requiring multiple versions of the same add-on.

### Automated Tests

Added a test that reproduces the issue.

### DIFF
--- a/lib/ruby_lsp/addon.rb
+++ b/lib/ruby_lsp/addon.rb
@@ -57,11 +57,26 @@ module RubyLsp
 
         if include_project_addons
           project_addons = Dir.glob("#{global_state.workspace_path}/**/ruby_lsp/**/addon.rb")
-
-          # Ignore add-ons from dependencies if the bundle is stored inside the project. We already found those with
-          # `Gem.find_files`
           bundle_path = Bundler.bundle_path.to_s
-          project_addons.reject! { |path| path.start_with?(bundle_path) }
+          gems_dir = Bundler.bundle_path.join("gems")
+
+          # Create an array of rejection glob patterns to ignore add-ons already discovered through Gem.find_files if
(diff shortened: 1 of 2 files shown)


vinistock:
* **#3669** <a href="https://app.graphite.dev/github/pr/Shopify/ruby-lsp/3669?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/Shopify/ruby-lsp/3669?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`
<hr />

You must have a Graphite account in order to use the merge queue. Sign up using [this link](https://app.graphite.dev/invite/github/Shopify?ref=merge-queue-instructions-in-stack-comment&prId=3669).

<h2></h2>
<sub>

</sub>

<sub>

</sub>

<h2></h2>

KaanOzkan (approved):

connorshea:
I can confirm this fixes the problem, thank you!

PR #3696, opened 2025-07-30
Add information on enabling the Ruby LSP using builtin vim.lsp instea…

rezhaTanuharja:
**Background**

With the release of Neovim 0.11, more and more users move away from the nvim-lspconfig plugin in favor of the built-in method of enabling LSP. This method can be quite confusing for new users:
- where to put the LSP config table within my nvim config?
- what should the LSP config table contain?

**Summary**

Add information on how to enable the Ruby LSP using the built-in method
- provide a config example of enabling the LSP that should work out of the box if users have `ruby-lsp` installed
- align the example with the example in the nvim-lspconfig section so users can compare the two ways

### DIFF
--- a/jekyll/editors.markdown
+++ b/jekyll/editors.markdown
@@ -192,6 +192,42 @@ folder is used.
 
 See [this issue][mason-abi] for further information.
 
+### Built-In vim.lsp
+
+**Note**: Ensure that you are using Neovim 0.11 or newer.
+
+You can also configure the Ruby LSP without the nvim-lspconfig plugin.
+Create an `lsp` directory inside your config directory and create a file `ruby-lsp.lua` inside it
+with the following content:
+
+```lua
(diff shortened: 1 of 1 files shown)


rezhaTanuharja:
I have signed the CLA!

jekyll/editors.markdown
```diff
@@ -192,6 +192,42 @@ folder is used.
 
 See [this issue][mason-abi] for further information.
 
+### Built-In vim.lsp
+
+**Note**: Ensure that you are using Neovim 0.11 or newer.
```
vinistock: We can use Jekyll's note syntax


```suggestion
{ .note }
Ensure that you are using NeoVim 0.11 or newer.
```

vinistock (commented):
Thanks for contributing to the documentation! I just have a minor comment

PR #3698, opened 2025-07-30
Handle `Errno::EPERM` errors when composing bundle

egiurleo:


egiurleo:
* **#3698** <a href="https://app.graphite.dev/github/pr/Shopify/ruby-lsp/3698?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/Shopify/ruby-lsp/3698?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#3699** <a href="https://app.graphite.dev/github/pr/Shopify/ruby-lsp/3699?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`
<hr />

You must have a Graphite account in order to use the merge queue. Sign up using [this link](https://app.graphite.dev/invite/github/Shopify?ref=merge-queue-instructions-in-stack-comment&prId=3698).

<h2></h2>
<sub>

</sub>

<sub>

</sub>

<h2></h2>

lib/ruby_lsp/server.rb
```diff
@@ -1388,6 +1388,9 @@ def compose_bundle(message)
       rescue Bundler::LockfileError => e
         send_message(Error.new(id: id, code: BUNDLE_COMPOSE_FAILED_CODE, message: e.message))
         return
+      rescue Errno::EPERM
+        # If the user doesn't have permission to perform read operations, we can't compose the bundle
+        return
```
egiurleo, line 1393: Let me know if it makes sense to return here -- my understanding is we can't compose the bundle without more permissions.

vinistock: It will depend on why we're getting the permission issue. When you invoke `Bundler.default_lockfile`, it will keep searching for a `Gemfile` and `Gemfile.lock` pair in parent directories until it reaches home.

Are we 100% sure that the permission issue is coming from reading the lockfile and not from trying to read the parent directories?

The issue could be that there's simply no bundle in the project and the user didn't give permissions to the editor for accessing any of the parent directories. In that case, returning early would not be correct because we can still compose the bundle as long as there's no lockfile and the user has permissions for the current directory.

PR #3706, opened 2025-08-07
Adds support for attr_* method references

### LINKED ISSUE #2668: Support `attr_*` for References
Following up from https://github.com/Shopify/ruby-lsp/pull/2650, we should also support References requests for `attr_reader`, `attr_writer` and `attr_accessor`.

tommycrumrine:
### Motivation
Shows references for methods defined with `attr_reader`, `attr_writer`, `attr_accessor`

Re-opening stale PR https://github.com/Shopify/ruby-lsp/pull/2848. Sorry this fell off my radar!

Closes #2668

### Implementation

`attr_reader`, `attr_writer`, `attr_accessor` are just methods, so on call node if we detect one of these methods, and one of the arguments match the target method name, add the reference.

Additionally, I have ensured it [works with empty and string arguments](https://github.com/Shopify/ruby-lsp/pull/2848#discussion_r1905879584) as well as added tests for these cases.

### Automated Tests

Yes!

### Manual Tests
Screenshot 2024-11-15 at 1 43 07 PM


lib/ruby_indexer/lib/ruby_indexer/reference_finder.rb
```diff
@@ -282,17 +282,53 @@ def on_instance_variable_target_node_enter(node)
 
     #: (Prism::CallNode node) -> void
     def on_call_node_enter(node)
-      if @target.is_a?(MethodTarget) && (name = node.name.to_s) == @target.method_name
-        @references << Reference.new(
-          name,
-          node.message_loc, #: as !nil
```
egiurleo, line 288: Does type checking pass without this `#: as !nil`?

lib/ruby_indexer/lib/ruby_indexer/reference_finder.rb
```diff
@@ -282,17 +282,53 @@ def on_instance_variable_target_node_enter(node)
...
+      else
+        false
+      end
+    end
+
+    #: (Prism::CallNode node) -> Array[String]
+    def unescaped_argument_names(node)
+      return [] if node.arguments.nil?
+
+      node.arguments.arguments.select { |arg| arg.respond_to?(:unescaped) }.map(&:unescaped)
```
egiurleo, line 314: Can `node.arguments.arguments` be nil?

egiurleo (commented):
Thanks for re-opening your PR and apologies for the slow response on our end! I just left two quick questions. I'll also re-run CI for you to make sure the tests look good.

lib/ruby_indexer/lib/ruby_indexer/reference_finder.rb
tommycrumrine: Opps, yep something is definitely off here, will look into it. I'm a little new to the type checking 😅

egiurleo: All good, let me know if you need any support on it.

lib/ruby_indexer/lib/ruby_indexer/reference_finder.rb
```diff
@@ -282,17 +282,53 @@ def on_instance_variable_target_node_enter(node)
...
+      else
+        false
+      end
+    end
+
+    #: (Prism::CallNode node) -> Array[String]
+    def unescaped_argument_names(node)
+      return [] if node.arguments&.arguments.nil?
+
+      node.arguments.arguments.select { |arg| arg.respond_to?(:unescaped) }.map(&:unescaped)
```
vinistock, line 314:
```suggestion
      node.arguments.arguments.filter_map do |arg|
        case arg
        when Prism::StringNode
          arg.unescaped
        when Prism::SymbolNode
          arg.value
        end
      end
```

lib/ruby_indexer/lib/ruby_indexer/reference_finder.rb
vinistock: I think we do need to cast it because `message_loc` can be `nil`. I checked out the branch and type checking is failing without the narrows.

vinistock (commented):
This is working well for when we ask for references on a call to an attribute, but when trying to find references on the attribute definition itself, we are accidentally searching for the `attr_reader` call instead of the symbol/string names.

```ruby
class Foo
  attr_reader :bar # here we search for `attr_reader` instead of bar

  def something
    bar # here it works well
  end
end
```

We need to adjust the target in [references](https://github.com/Shopify/ruby-lsp/blob/f38329fccca1c0eaa45149f08f72655bc3cae781/lib/ruby_lsp/requests/references.rb#L50).

Basically, we need to:

1. Start accepting `SymbolNode` as a possible target
2. Check if the `node_context.call_node` is an attribute related call when we get a `SymbolNode` back
3. Construct the target object using the symbol value instead of the call node's name

PR #3711, opened 2025-08-12
Disable spring when running tests through the explorer

vinistock:
### Motivation

We stopped requiring our custom LSP reporters through `RUBYOPT` because the environment variable caused child processes to accidentally require it too, which leads to unexpected events being received by the explorer.

Unfortunately, when using Spring, the `-r` argument we are now using isn't passed down to the server. When running `bin/rails test`, the Spring server that will run the test doesn't have the reporters required and so the results are not sent to the explorer.

The ideal solution would be to have Spring support some form of environment injection through something other than `RUBYOPT`, so that we can continue to not mess up with child processes while requiring what we need. For now, I think it's fair to disable spring. It's better to have tests run a little slower than not have the reporting work.

### Implementation

Started disabling spring for now.

### DIFF
--- a/vscode/src/streamingRunner.ts
+++ b/vscode/src/streamingRunner.ts
@@ -142,6 +142,7 @@ export class StreamingRunner implements vscode.Disposable {
       terminal = vscode.window.createTerminal({
         name,
         cwd,
+        env: { DISABLE_SPRING: "1" },
       });
     }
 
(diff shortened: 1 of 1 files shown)


vinistock:
* **#3711** <a href="https://app.graphite.dev/github/pr/Shopify/ruby-lsp/3711?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/Shopify/ruby-lsp/3711?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`
<hr />

You must have a Graphite account in order to use the merge queue. Sign up using [this link](https://app.graphite.dev/invite/github/Shopify?ref=merge-queue-instructions-in-stack-comment&prId=3711).

<h2></h2>
<sub>

</sub>

<sub>

</sub>

<h2></h2>

alexcrocha (approved):

PR #3713, opened 2025-08-13
Add VS Code rubyLsp.serverPath configuration

### LINKED ISSUE #3709: Allow running a server from a local repo in standalone mode

### Use case

Currently, developers working on ruby-lsp server itself cannot easily run their local changes in standalone mode. When debugging the extension in a target workspace without a Gemfile, the extension loads the installed gem instead of the local development copy. It will switch to the local copy but only when working within ruby-lsp itself (which has a Gemfile). This forces developers to either create custom Gemfiles (which changes the server mode) or use workarounds like symlinks.

### Description

Add a configuration setting that allows developers to specify a local file path to the ruby-lsp server executable. The main goal would be to use this setting in standalone mode, though it may make sense to apply this to the normal mode as well. This would keep things consistent and simplify setting up a local server in normal mode.

### Implementation

Add a new _rubyLsp.serverPath_ setting that accepts a file path to a local ruby-lsp executable. Modify the _getLspExecutables_ method to check this setting first, and if set, use the specified path instead of resolving to the installed gem. Ensure this setting cannot be used simultaneously with _rubyLsp.branch_ to prevent conflicts. The setting should be validated to ensure the path exists and points to a valid executable or ruby-lsp repo.

See #3690 for the initial discussion.

jconley88:
### Motivation

Closes  #3709 

#### VS Code Extension
- Added a `serverPath` configuration setting to package.json
- when `serverPath` configuration is present in `getLspExecutables`, set `--path` as an arg to the `ruby-lsp` executable
- The serverPath option supports relative and absolute paths to both the ruby-lsp repo root directory as well as the executable file

#### Ruby LSP Server
- Capture `--path` arg as an option on `ruby-lsp` execution
- When setting up the Gemfile in `setup_bundler.rb`, add the configured absolute path to `gem 'ruby-lsp', path: "..."`

In every location where the path is considered an error is raised if the branch is also set.

### Automated Tests

I followed the test pattern for the `--branch` option when determining what to test. I have added tests for Bundler setup, but not for VS Code nor for option parsing in the ruby-lsp executable. I would be happy to add additional testing if I missed anything.

### Manual Tests

1. Modify the VS Code "Run Extension" launch configuration to set the target workspace to a ruby project of your choice.
    ```json
    {
	    "name": "Run Extension",
	    "type": "extensionHost",
	    "request": "launch",
	    "args": [
                    "/path/to/your/target/workspace",
		    "--extensionDevelopmentPath=${workspaceFolder}"
	    ],
	    "outFiles": [
		    "${workspaceFolder}/out/**/*.js"
	    ],
	    "preLaunchTask": "${defaultBuildTask}"
    }
    ```
2. From the "Run and Debug" side bar, run the extension
3. In the newly opened debug VS Code instance
     1.  In the settings you should be able to see the new option: `rubyLsp.serverPath`.
     2. Set the`serverPath` to your local ruby-lsp repo
     3. *Restart the entire debug instance to properly pick up on the changes
     4. Inspect the [path/to/target/workspace]/.ruby-lsp/Gemfile
         1. You should see the `path:` designation
     5. It's a little more challenging to fully verify that your code is actually running. You may need to modify some code, add some output or change the version number so that you can inspect it in the status bar
     6. Repeat the above with relative vs absolute paths, the ruby-lsp root path vs executable path, and simultaneously setting the branch.

*There are multiple existing issues with the config file watcher and auto-reload when a configuration setting is changed
1. In current production (v0.26.1), the server restarts but does not respect `rubyLsp.branch` nor `rubyLsp.serverPath`
2. In current development (v0.26.2), the server either does not restart at all or at least, the output is not shown for a restart
3. Changing back and forth from a server that supports serverPath to a server that does not while running a vscode instance that does support serverPath likely would not work well anyhow due to the feature mismatch

### DIFF
--- a/exe/ruby-lsp
+++ b/exe/ruby-lsp
@@ -29,6 +29,13 @@ parser = OptionParser.new do |opts|
     options[:branch] = branch
   end
 
+  opts.on(
+    "--path [PATH]",
+    "Launch the Ruby LSP using a local PATH rather than the release version",
+  ) do |path|
+    options[:path] = path
+  end
+
   opts.on("--doctor", "Run troubleshooting steps") do
     options[:doctor] = true
   end
@@ -51,6 +58,11 @@ rescue OptionParser::InvalidOption => e
   warn(e)
   warn("")
   warn(parser.help)
+  exit(1)
+end
+
+if options[:branch] && options[:path]
+  warn('Invalid options: --branch and --path cannot both be set.')
   exit(1)
 end
 
--- a/jekyll/contributing.markdown
+++ b/jekyll/contributing.markdown
@@ -58,6 +58,20 @@ with a debugger. Note that the debug mode applies only until the editor is close
 Caveat: since you are debugging the language server instance that is currently running in your own editor, features will
 not be available if the execution is currently suspended at a breakpoint.
 
+#### Configuring the server version
+
+When developing the Ruby LSP server, you may want to test your changes in your own Ruby projects to ensure they work correctly in real-world scenarios.
+
+The running server, even in debug mode, will default to the installed release version*. You can use the `rubyLsp.serverPath` configuration setting in the target workspace to start your local copy instead. Make sure to restart the language server after making changes to pick up your updates.
+
+```jsonc
+{
+  "rubyLsp.serverPath": "/path/to/your/ruby-lsp-clone"
+}
+```
+
+*Note: There are some exceptions to this. Most notably, when the ruby-lsp repository is opened in VS Code with the extension active, it will run the server contained within the repository.
+
 #### Understanding Prism ASTs
 
 The Ruby LSP uses Prism to parse and understand Ruby code. When working on a feature, it's very common to need to
--- a/lib/ruby_lsp/setup_bundler.rb
+++ b/lib/ruby_lsp/setup_bundler.rb
@@ -35,7 +35,13 @@ module RubyLsp
     def initialize(project_path, **options)
       @project_path = project_path
       @branch = options[:branch] #: String?
+      @path = options[:path] #: String?
       @launcher = options[:launcher] #: bool?
+
+      if @branch && !@branch.empty? && @path && !@path.empty?
+        raise ArgumentError, "Branch and path options are mutually exclusive. Please specify only one."
+      end
+
       patch_thor_to_print_progress_to_stderr! if @launcher
 
       # Regular bundle paths
@@ -165,7 +171,13 @@ module RubyLsp
 
       unless @dependencies["ruby-lsp"]
         ruby_lsp_entry = +'gem "ruby-lsp", require: false, group: :development'
-        ruby_lsp_entry << ", github: \"Shopify/ruby-lsp\", branch: \"#{@branch}\"" if @branch
+        if @branch && !@branch.empty?
+          ruby_lsp_entry << ", github: \"Shopify/ruby-lsp\", branch: \"#{@branch}\""
+        end
+        if @path && !@path.empty?
+          absolute_path = File.expand_path(@path, @project_path)
+          ruby_lsp_entry << ", path: \"#{absolute_path}\""
+        end
         parts << ruby_lsp_entry
       end
 
--- a/vscode/package.json
+++ b/vscode/package.json
@@ -467,6 +467,11 @@
           "type": "string",
           "default": ""
         },
+        "rubyLsp.serverPath": {
+          "description": "Absolute or workspace-relative path to a local ruby-lsp repository or its ruby-lsp executable. Only supported if not using bundleGemfile",
+          "type": "string",
+          "default": ""
+        },
         "rubyLsp.pullDiagnosticsOn": {
           "description": "When to pull diagnostics from the server (on change, save or both). Selecting 'save' may significantly improve performance on large files",
           "type": "string",
--- a/vscode/src/client.ts
+++ b/vscode/src/client.ts
@@ -1,4 +1,5 @@
 import path from "path";
+import fs from "fs";
 import os from "os";
 import { performance as Perf } from "perf_hooks";
 
@@ -85,6 +86,7 @@ function getLspExecutables(workspaceFolder: vscode.WorkspaceFolder, env: NodeJS.
   const customBundleGemfile: string = config.get("bundleGemfile")!;
   const useBundlerCompose: boolean = config.get("useBundlerCompose")!;
   const bypassTypechecker: boolean = config.get("bypassTypechecker")!;
+  const serverPath: string = config.get("serverPath")!;
 
   const executableOptions: ExecutableOptions = {
     cwd: workspaceFolder.uri.fsPath,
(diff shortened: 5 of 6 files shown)


exe/ruby-lsp
```diff
@@ -29,6 +29,13 @@ parser = OptionParser.new do |opts|
     options[:branch] = branch
   end
 
+  opts.on(
+    "--path [PATH]",
```
vinistock: How about we call this `--lsp-path` to prevent any confusion with a potential ability to configure the workspace path or some other path?

jekyll/contributing.markdown
```diff
@@ -58,6 +58,20 @@ with a debugger. Note that the debug mode applies only until the editor is close
 Caveat: since you are debugging the language server instance that is currently running in your own editor, features will
 not be available if the execution is currently suspended at a breakpoint.
 
+#### Configuring the server version
+
+When developing the Ruby LSP server, you may want to test your changes in your own Ruby projects to ensure they work correctly in real-world scenarios.
+
+The running server, even in debug mode, will default to the installed release version*. You can use the `rubyLsp.serverPath` configuration setting in the target workspace to start your local copy instead. Make sure to restart the language server after making changes to pick up your updates.
```
vinistock: We automatically restart the server when any `rubyLsp.` settings change, so you actually don't need a manual restart.

```suggestion
The running server, even in debug mode, will default to the installed release version*. You can use the `rubyLsp.serverPath` configuration setting in the target workspace to start your local copy instead.
```

lib/ruby_lsp/setup_bundler.rb
```diff
@@ -35,7 +35,13 @@ def stdout
     def initialize(project_path, **options)
       @project_path = project_path
       @branch = options[:branch] #: String?
+      @path = options[:path] #: String?
       @launcher = options[:launcher] #: bool?
+
+      if @branch && !@branch.empty? && @path && !@path.empty?
```
vinistock: We're currently handling this both in the executable and here. I think we can skip this validation.

lib/ruby_lsp/setup_bundler.rb
```diff
@@ -165,7 +171,13 @@ def write_custom_gemfile
 
       unless @dependencies["ruby-lsp"]
         ruby_lsp_entry = +'gem "ruby-lsp", require: false, group: :development'
-        ruby_lsp_entry << ", github: \"Shopify/ruby-lsp\", branch: \"#{@branch}\"" if @branch
+        if @branch && !@branch.empty?
+          ruby_lsp_entry << ", github: \"Shopify/ruby-lsp\", branch: \"#{@branch}\""
+        end
+        if @path && !@path.empty?
+          absolute_path = File.expand_path(@path, @project_path)
```
vinistock: Is this path expansion correct? The `@project_path` is the workspace currently being worked on and `@path` is the path to the LSP, which are potentially unrelated.

I think we can set the expectation that the path to the LSP clone must always be an absolute path and if it's not we error early in the executable.

vscode/src/client.ts
```diff
@@ -119,13 +121,40 @@ function getLspExecutables(workspaceFolder: vscode.WorkspaceFolder, env: NodeJS.
...
+    const args = [];
     const workspacePath = workspaceFolder.uri.fsPath;
-    const command =
-      path.basename(workspacePath) === "ruby-lsp" && os.platform() !== "win32"
-        ? path.join(workspacePath, "exe", "ruby-lsp")
-        : "ruby-lsp";
+    let command: string;
 
-    const args = [];
+    if (serverPath.length > 0 && branch.length > 0) {
```
vinistock: Similarly here, I think we can do without this validation since the executable already errors in this case. These specific settings are for the development of the LSP only and don't impact end user experience, so we don't need to worry too much about the failure scenarios.

vscode/package.json
```diff
@@ -467,6 +467,11 @@
           "type": "string",
           "default": ""
         },
+        "rubyLsp.serverPath": {
+          "description": "Absolute or workspace-relative path to a local ruby-lsp repository or its ruby-lsp executable. Only supported if not using bundleGemfile",
+          "type": "string",
+          "default": ""
```
vinistock: I think we can avoid setting a default here, so that when nothing is configured the value is `undefined` and we can simply check `if (value)` instead of checking for the string's length.

vscode/src/client.ts
```diff
@@ -119,13 +121,40 @@ function getLspExecutables(workspaceFolder: vscode.WorkspaceFolder, env: NodeJS.
...
+        if (stat.isDirectory()) {
+          command = os.platform() !== "win32" ? path.join(absoluteServerPath, "exe", "ruby-lsp") : "ruby-lsp";
+        } else {
+          command = absoluteServerPath;
+        }
+      } else {
+        throw new Error(
+          `The configured rubyLsp.serverPath "${serverPath}" does not exist at "${absoluteServerPath}". `,
+        );
+      }
```
vinistock: A few things about this section:

- We don't need to change the command, it will still be `ruby-lsp`, we just need to append the new CLI flag
- In VS Code extensions, we cannot ever use `fs` because that bypasses the VS Code file system API. Everything has to happen through `vscode.workspace.fs`

That said, since these options are for the development of the LSP itself only, I think we can skip checking these arguments here and move everything to the server (that also ensures that the checks work for any editor connecting to the LSP).

Let's check if the path is absolute and if it exists there and simplify this part.

vscode/package.json
```diff
@@ -467,6 +467,11 @@
           "type": "string",
           "default": ""
         },
+        "rubyLsp.serverPath": {
+          "description": "Absolute or workspace-relative path to a local ruby-lsp repository or its ruby-lsp executable. Only supported if not using bundleGemfile",
```
vinistock: Let's make this only absolute.

vscode/src/client.ts
jconley88: I can't test it without changing the executable path. The extension is running the code with the serverPath configuration but without changing the executable directory, the server won't recognize the new --lsp-path option. I wasn't sure if you meant this, but I also removed the ability to specify the executable directly. I was already on the fence about it and i think it's just unnecessary complexity.

jconley88:
Thanks for all of the feedback and good point about the unnecessary validation! I think I addressed everything. I left the validation that makes sure the configured executable path in `getLspExectuables` is named `ruby-lsp`. This shouldn't technically still be validated when serverPath is specified, but it kept the logic simpler. Let me know if you disagree, I missed any feedback or you've thought of anything else. I'm happy to continue iterating.

jconley88:
Hey, I'm just checking in on this in case it got lost. Is there anything else that I can do to get this in a good state for review and/or merge?

rafaelfranca (approved):

jconley88:
I will review the failing checks soon. Thank you for approving the changes!

vscode/src/client.ts
jconley88: I believe that I addressed everything, but I've forgotten some context. My previous comment seems unnecessary and makes me think I was confused about your suggestion.

jconley88:
I resolved the feedback comments from @vinistock accidentally, but have since unresolved them. I was surprised that I had access to do that and got confused about whose responsibility it was. My apologies.

jconley88:
I believe that everything is up to date and ready for re-review. I rebased on main but kept the commits separate for easier review. The first commit is pre-feedback, the second addresses the feedback and the 3rd gets it up to date for CI.

rafaelfranca:
@vinistock given https://github.com/Shopify/ruby-lsp/security/advisories/GHSA-c4r5-fxqw-vh93. I assume we don't want this anymore right?

jconley88:
I absolutely understand if this is closed because of that CVE or just isn't a priority. There was no malicious intent here if this is related. I was just looking for a way to contribute and this solved a legitimate blocker when I was working in dev. I'm happy to resolve the conflicts and refactor if it's still desired. Thanks for the previous attention and feedback, regardless. Cheers!

vinistock:
Indeed. This was going to be another development convenience, that I myself suggested adding, but in light of the recent CVE, I think it's best to avoid it for now.

@jconley88 thank you for your work on this. I appreciate it!

PR #3731, opened 2025-08-30
Improve go to relevant file performance

### LINKED ISSUE #3327: Improve GoToRelevantFile performance
### Ruby LSP Information

Parent issue https://github.com/Shopify/ruby-lsp/issues/2966

The GoToRelevantFile function added in https://github.com/Shopify/ruby-lsp/pull/2985 works well with small size repos, but it has a noticeable lag when used on a larger codebase. 

Benchmark measurements on the shopify core codebase (400k files count) on my machine

(In order: user CPU time, system CPU time, the sum of the user and system CPU times, elapsed real time)
```
  0.875498   0.938839   1.814337 (  2.438297)
  0.204476   0.908439   1.112915 (  1.319023)
  1.396747   1.401042   2.797789 (  3.252911)
  3.044789   4.278284   7.323073 (  6.205184)
  2.496724   4.009404   6.506128 (  5.499605)
  2.575118   4.419145   6.994263 (  5.775319)
  2.427883   3.813284   6.241167 (  5.337139)
  2.646134   3.806451   6.452585 (  5.554977)
  1.610006   3.839775   5.449781 (  4.531290)
  3.278136   4.539844   7.817980 (  6.581827)
```

Different random file are used each time. This is the measurement of the time it takes to run the `GoToRelevantFile#perform` only.

domingo2000:
By doing a walk over the tree of directories in the workpsace we improve the dir Glob time because it only checks for the closest test directory folder, instead of searching in all the files of the workspace.

### Motivation

Closes #3327

### Implementation

As discussed [here](https://github.com/Shopify/ruby-lsp/issues/3327#issuecomment-3119997342) the implementation consist in a tree walk over the directory structure. By doing this walk until the workspace path, we only check for test files in the most closer ancestor folder that contains tests, the locality of tests in projects using engines in rails, or Packwerck, or in general in ruby projects, make the search correct most of the times. 

By checking fewer directories the performance of the Glob that search for files is improved drastically, making the go to relevant file feels instant in big projects.

### Automated Tests

Changed the current test implementation by changing the mocks, because the Glob calls are different now that we walk up over the directory structure.

### Manual Tests

1. Rubocop project

https://github.com/user-attachments/assets/3bfe86d5-232d-45cd-86db-b9e05d9a3d63

2. Rails project

https://github.com/user-attachments/assets/8baf00bd-bf1b-491b-aa2e-2bc3499c07dc

3. Buk Core project (> 27K ruby files)

https://github.com/user-attachments/assets/f4a817bf-3dab-4710-ad94-31921e94261a

### DIFF
--- a/lib/ruby_lsp/requests/go_to_relevant_file.rb
+++ b/lib/ruby_lsp/requests/go_to_relevant_file.rb
@@ -35,10 +35,59 @@ module RubyLsp
 
       #: -> Array[String]
       def find_relevant_paths
-        candidate_paths = Dir.glob(File.join("**", relevant_filename_pattern))
+        search_roots = determine_search_roots
+        candidate_paths = []
+
+        search_roots.each do |root|
+          glob_pattern = File.join(root, "**", relevant_filename_pattern)
+          candidate_paths.concat(Dir.glob(glob_pattern))
+        end
+
+        candidate_paths.map! { |path| path.delete_prefix(@workspace_path).delete_prefix("/") }
+
+        candidate_paths.uniq!
         return [] if candidate_paths.empty?
 
         find_most_similar_with_jaccard(candidate_paths).map { File.join(@workspace_path, _1) }
+      end
+
+      # Determine the search roots based on the closest test directories.
+      # This scopes the search to reduce the number of files that need to be checked.
+      #: -> Array[String]
+      def determine_search_roots
+        current_path = File.join(@workspace_path, @path)
+
+        current_dir = File.dirname(current_path)
+        while current_dir.start_with?(@workspace_path)
+          dir_basename = File.basename(current_dir)
+
+          # If current directory is a test directory, return its parent as search root
+          if TEST_KEYWORDS.any? { |keyword| dir_basename.include?(keyword) }
+            parent = File.dirname(current_dir)
+            return parent.start_with?(@workspace_path) ? [parent] : [@workspace_path]
+          end
+
+          # Search the test directories by walking up the directory tree
+          begin
+            entries = Dir.entries(current_dir).reject { |entry| entry.start_with?(".") }
+            test_dir_found = entries.any? do |entry|
+              full_path = File.join(current_dir, entry)
+              File.directory?(full_path) && TEST_KEYWORDS.any? { |keyword| entry == keyword }
+            end
+
+            return [current_dir] if test_dir_found
+          rescue Errno::EACCES, Errno::ENOENT
+            # Skip directories we can't read
+          end
+
+          # Move up one level
+          parent_dir = File.dirname(current_dir)
+          break if parent_dir == current_dir
+
+          current_dir = parent_dir
+        end
+
+        [@workspace_path]
       end
 
       #: -> String
--- a/test/requests/go_to_relevant_file_test.rb
+++ b/test/requests/go_to_relevant_file_test.rb
@@ -5,6 +5,7 @@ require "test_helper"
 
 class GoToRelevantFileTest < Minitest::Test
   def test_when_input_is_test_file_returns_array_of_implementation_file_locations
+    stub_directory_structure("/workspace/test/requests", has_test_dir: false)
     stub_glob_pattern("**/go_to_relevant_file.rb", ["lib/ruby_lsp/requests/go_to_relevant_file.rb"])
 
     test_file_path = "/workspace/test/requests/go_to_relevant_file_test.rb"
@@ -15,8 +16,12 @@ class GoToRelevantFileTest < Minitest::Test
   end
 
   def test_when_input_is_implementation_file_returns_array_of_test_file_locations
-    pattern =
-      "**/{{test_,spec_,integration_test_}go_to_relevant_file,go_to_relevant_file{_test,_spec,_integration_test}}.rb"
+    stub_directory_structure("/workspace/lib/ruby_lsp/requests", has_test_dir: false)
+    stub_directory_structure("/workspace/lib/ruby_lsp", has_test_dir: false)
+    stub_directory_structure("/workspace/lib", has_test_dir: false)
+    stub_directory_structure("/workspace", has_test_dir: true)
+
+    pattern = "**/{{test_,spec_,integration_test_}go_to_relevant_file,go_to_relevant_file{_test,_spec,_integration_test}}.rb"
     stub_glob_pattern(pattern, ["test/requests/go_to_relevant_file_test.rb"])
 
     impl_path = "/workspace/lib/ruby_lsp/requests/go_to_relevant_file.rb"
@@ -149,6 +154,21 @@ class GoToRelevantFileTest < Minitest::Test
   private
 
   def stub_glob_pattern(pattern, matches)
-    Dir.stubs(:glob).with(pattern).returns(matches)
+    absolute_pattern = pattern.start_with?("/") ? pattern : "/workspace/#{pattern}"
+    absolute_matches = matches.map { |match| match.start_with?("/") ? match : "/workspace/#{match}" }
+    Dir.stubs(:glob).with(absolute_pattern).returns(absolute_matches)
+  end
+
+  def stub_directory_structure(dir_path, has_test_dir:)
+    File.stubs(:directory?).with(File.join(dir_path, "other_dir")).returns(true)
+
+    if has_test_dir
+      Dir.stubs(:entries).with(dir_path).returns([".", "..", "test", "lib", "other_dir"])
+      File.stubs(:directory?).with(File.join(dir_path, "test")).returns(true)
+      File.stubs(:directory?).with(File.join(dir_path, "lib")).returns(true)
+    else
+      Dir.stubs(:entries).with(dir_path).returns([".", "..", "some_file.rb", "other_dir"])
+      File.stubs(:directory?).with(File.join(dir_path, "some_file.rb")).returns(false)
+    end
   end
 end


lib/ruby_lsp/requests/go_to_relevant_file.rb
```diff
@@ -35,12 +35,61 @@ def perform
...
+        candidate_paths = []
+
+        search_roots.each do |root|
+          glob_pattern = File.join(root, "**", relevant_filename_pattern)
+          candidate_paths.concat(Dir.glob(glob_pattern))
+        end
+
+        candidate_paths.map! { |path| path.delete_prefix(@workspace_path).delete_prefix("/") }
+
+        candidate_paths.uniq!
```
jenny-codes: Nit: This can be streamlined to 

```ruby
candidate_paths  = determine_search_roots
  .map { |root| File.join(root, "**", relevant_filename_pattern) }
  .flat_map { |pattern| Dir.glob(pattern) }
  .uniq
  .map { |path| path.delete_prefix(@workspace_path).delete_prefix("/") }
```

lib/ruby_lsp/requests/go_to_relevant_file.rb
```diff
@@ -35,12 +35,61 @@ def perform
...
         return [] if candidate_paths.empty?
 
         find_most_similar_with_jaccard(candidate_paths).map { File.join(@workspace_path, _1) }
       end
 
+      # Determine the search roots based on the closest test directories.
+      # This scopes the search to reduce the number of files that need to be checked.
+      #: -> Array[String]
+      def determine_search_roots
```
jenny-codes: Why is this an array? I am missing in the implementation where we are returning more than one value

lib/ruby_lsp/requests/go_to_relevant_file.rb
```diff
@@ -35,12 +35,61 @@ def perform
...
+
+          # Search the test directories by walking up the directory tree
+          begin
+            entries = Dir.entries(current_dir).reject { |entry| entry.start_with?(".") }
+            test_dir_found = entries.any? do |entry|
+              full_path = File.join(current_dir, entry)
+              File.directory?(full_path) && TEST_KEYWORDS.any? { |keyword| entry == keyword }
+            end
+
+            return [current_dir] if test_dir_found
```
jenny-codes: Nit: I think it can be simplified to

```ruby
contains_test_dir = Dir
  .entries(current_dir)
  .filter { |entry| TEST_KEYWORDS.include?(entry) }
  .any? { |entry| File.directory?(File.join(current_dir, entry)) }

return [current_dir] if contains_test_dir
```
The logic stays mostly the same, but I suggest remove the hidden file check as it would already be covered by the keyword match.

lib/ruby_lsp/requests/go_to_relevant_file.rb
```diff
@@ -35,12 +35,61 @@ def perform
...
       end
 
+      # Determine the search roots based on the closest test directories.
+      # This scopes the search to reduce the number of files that need to be checked.
+      #: -> Array[String]
+      def determine_search_roots
+        current_path = File.join(@workspace_path, @path)
+
+        current_dir = File.dirname(current_path)
+        while current_dir.start_with?(@workspace_path)
```
jenny-codes: Maybe the condition here should be "while the current directory is not the workspace_path"? Because as soon as the current_dir is the workspace_path we should just return the workspace_path anyway. It would simplify the logic on line 68 too as we won't have to check if the parent is out of the workspace path anymore.

lib/ruby_lsp/requests/go_to_relevant_file.rb
```diff
@@ -35,12 +35,61 @@ def perform
...
+          dir_basename = File.basename(current_dir)
+
+          # If current directory is a test directory, return its parent as search root
+          if TEST_KEYWORDS.any? { |keyword| dir_basename.include?(keyword) }
```
jenny-codes: Here you use `include?` to check for the test directory but at line 75 you are checking with an exact match. Why is that? And should should they be consistent?

test/requests/go_to_relevant_file_test.rb
```diff
@@ -15,8 +16,12 @@ def test_when_input_is_test_file_returns_array_of_implementation_file_locations
   end
 
   def test_when_input_is_implementation_file_returns_array_of_test_file_locations
-    pattern =
-      "**/{{test_,spec_,integration_test_}go_to_relevant_file,go_to_relevant_file{_test,_spec,_integration_test}}.rb"
+    stub_directory_structure("/workspace/lib/ruby_lsp/requests", has_test_dir: false)
+    stub_directory_structure("/workspace/lib/ruby_lsp", has_test_dir: false)
+    stub_directory_structure("/workspace/lib", has_test_dir: false)
+    stub_directory_structure("/workspace", has_test_dir: true)
```
jenny-codes: Can you walk me through why those stubs are needed? The only thing that was changed in the implementation was the Dir.glob pattern no?

jenny-codes (commented):
I like the idea but left some comments!

lib/ruby_lsp/requests/go_to_relevant_file.rb
domingo2000: It should be an exact match. First i thought the include? should catch more cases, but then i found that there are some directory names that could not be from tests like `specification_folder` that could make us miss the `spec` folder incorrectly.

Nice catch, i will change it.

lib/ruby_lsp/requests/go_to_relevant_file.rb
domingo2000: You are right, search _root_ will always be one, first i thought in candidate folders, but later changed to be the root.

test/requests/go_to_relevant_file_test.rb
domingo2000: Yes, if you have a better strategy to stub the Globs to actually test it, without being so implementation specific, i am happy to receive feedback. 

I did not want to pollute some /fixtures folder to just test this functionality.

lib/ruby_lsp/requests/go_to_relevant_file.rb
domingo2000: Good idea, i will change it.

domingo2000:
Thanks for the feedback @jenny-codes! I corrected all revisions, If you have any more feedback let me know.

Also checked again manually on the rubocop, rails and Buk monorepo to be sure it works well 👍🏻 .

lib/ruby_lsp/requests/go_to_relevant_file.rb
```diff
@@ -35,12 +35,53 @@ def perform
...
+              .any? { |entry| File.directory?(File.join(current_dir, entry)) }
+
+            return current_dir if contains_test_dir
+          rescue Errno::EACCES, Errno::ENOENT
+            # Skip directories we can't read
+          end
+
+          # Move up one level
+          parent_dir = File.dirname(current_dir)
+          break if parent_dir == current_dir
```
jenny-codes: We don't need the break statement here. The loop will already exit when `parent_dir == current_dir`

lib/ruby_lsp/requests/go_to_relevant_file.rb
```diff
@@ -35,12 +35,53 @@ def perform
 
       #: -> Array[String]
       def find_relevant_paths
-        candidate_paths = Dir.glob(File.join("**", relevant_filename_pattern))
+        pattern = File.join(search_root, "**", relevant_filename_pattern)
+        candidate_paths = Dir.glob(pattern)
+          .uniq
```
jenny-codes: We shouldn't need `.uniq` here

lib/ruby_lsp/requests/go_to_relevant_file.rb
```diff
@@ -35,12 +35,53 @@ def perform
...
         return [] if candidate_paths.empty?
 
         find_most_similar_with_jaccard(candidate_paths).map { File.join(@workspace_path, _1) }
       end
 
+      # Determine the search roots based on the closest test directories.
+      # This scopes the search to reduce the number of files that need to be checked.
+      #: -> String
+      def search_root
+        current_path = File.join(@workspace_path, @path)
```
jenny-codes: On second look–why do we need the workspace_path here? This method should be able to function only on the relative path. By removing it we can simplify the code in this method as well as removing the cleanup on line 41.

test/requests/go_to_relevant_file_test.rb
jenny-codes: Do the current tests pass without any modifications? if not they should pass with modifications to the searching pattern. I think for this feature we can create a new test "test_search_in_closest_directory". And let me look at the codebase to see if there are any existing patterns we can use to simulate file structures

domingo2000:
I changed to the relative path implementation. It works fine with manual testing. I will try to change the test to use a fixture folder in the filesystem instead of stubbing the responses to test the implementation without harcoding it.

domingo2000:
@jenny-codes I ended up testing the feature using a temporary directory strategy. That way the stubs are not implementation specific, the test is autocontained, and we don't need a real fixture folder.

test/requests/go_to_relevant_file_test.rb
```diff
@@ -4,151 +4,83 @@
...
+      lib_dir  = File.join(@workspace, "lib/ruby_lsp/requests")
+      test_dir = File.join(@workspace, "test/requests")
+      FileUtils.mkdir_p(lib_dir)
+      FileUtils.mkdir_p(test_dir)
+
+      impl_file = File.join(lib_dir, "go_to_relevant_file.rb")
+      test_file = File.join(test_dir, "go_to_relevant_file_test.rb")
+      File.write(impl_file, "# impl")
+      File.write(test_file, "# test")
+
```
jenny-codes: What is the reason that we need to write contents to the files?

domingo2000: Yeah, right, we only need to create the files, changed 👍🏻 .

domingo2000:
@jenny-codes, changed the test to only create the files, and added the "test within root" test to check that we don't get tests outside the search root.

domingo2000:
@jenny-codes I have no permissions to merge to shopify upstream, just pinging in case you forgot, no hurry on this PR.

jenny-codes:
@domingo2000 thank you!

domingo2000:
Thank to you!, I really use this feature a lot, and have used a lot your initial implementation.

PR #3739, opened 2025-09-07
Document `Request.register_watched_files` for addons

Earlopain:
### Motivation

https://github.com/Shopify/ruby-lsp/pull/2982 added this neat helper method. Can this be considered public api?


alexcrocha (approved):
Thank you!

PR #3763, opened 2025-09-23
docs: Update Sublime Text installation docs

guille:
### Motivation

The Sublime Text documentation is slightly outdated on two points:

1. The recommendation for disabling diagnostics. I couldn't find a reason why it was this way, but diagnostics work fine with current versions of the editor and the LSP plugin.
2. The selector on which files to use ruby-lsp for. Since ruby-lsp now supports ERB files, I have extended it to include them.

### Dependencies

The LSP plugin needs to be at least on version `4070-2.6.0` ([link](https://github.com/sublimelsp/LSP/releases/tag/4070-2.6.0)) for ERB to be supported properly. In earlier versions, the file doesn't have the right `language_id` in the request and the LSP server errors out in parsing it. See https://github.com/sublimelsp/LSP/pull/2642

### DIFF
--- a/jekyll/editors.markdown
+++ b/jekyll/editors.markdown
@@ -210,7 +210,7 @@ return {
   cmd = { "ruby-lsp" } -- or { "bundle", "exec", "ruby-lsp" },
 
   root_markers = { "Gemfile", ".git" },
-  
+
   init_options = {
     formatter = 'standard',
     linters = { 'standard' },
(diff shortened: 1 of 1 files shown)


guille:
Signed CLA

rwols:
Just have patience.

vinistock:
Thanks for the PR. We can definitely merge these changes. We're just backlogged with a lot of reviews/issues to go through and didn't get to this one in time. Also, I was on vacation.

If you wish to, reopen this so that we can review and ship.

PR #3783, opened 2025-10-13
Fix infinite recursion in constant alias resolution

### LINKED ISSUE #3782: Stack overflow when resolving self-referential constant alias
### Description

When a module defines a constant alias that references another constant of the same name in an outer namespace (for example, `B = B::C` inside `A::D`),  the Ruby LSP indexer enters infinite recursion while resolving the alias.  

This leads to a `SystemStackError` ("stack level too deep") during constant resolution.

### Ruby LSP Information

Following code is going to cause SystemStackError exception during reference search:

```ruby
  module A
    module B
      class C
      end
    end
  end

  module A
    module D
      B = B::C
    end
  end
```

Example log:
```
#<Thread:0x0000000125355e00 /Users/pkondzior/Projects/ruby-lsp/lib/ruby_lsp/base_server.rb:155 run> terminated with exception (report_on_exception is true):
/Users/pkondzior/Projects/ruby-lsp/lib/ruby_indexer/lib/ruby_indexer/index.rb:427:in 'Integer#downto': stack level too deep (SystemStackError)
	from /Users/pkondzior/Projects/ruby-lsp/lib/ruby_indexer/lib/ruby_indexer/index.rb:427:in 'RubyIndexer::Index#follow_aliased_namespace'
	from /Users/pkondzior/Projects/ruby-lsp/lib/ruby_indexer/lib/ruby_indexer/index.rb:1034:in 'RubyIndexer::Index#direct_or_aliased_constant'
	from /Users/pkondzior/Projects/ruby-lsp/lib/ruby_indexer/lib/ruby_indexer/index.rb:332:in 'RubyIndexer::Index#resolve'
	from /Users/pkondzior/Projects/ruby-lsp/lib/ruby_indexer/lib/ruby_indexer/index.rb:910:in 'RubyIndexer::Index#resolve_alias'
	from /Users/pkondzior/Projects/ruby-lsp/lib/ruby_indexer/lib/ruby_indexer/index.rb:436:in 'block in RubyIndexer::Index#follow_aliased_namespace'
	from /Users/pkondzior/Projects/ruby-lsp/lib/ruby_indexer/lib/ruby_indexer/index.rb:427:in 'Integer#downto'
	from /Users/pkondzior/Projects/ruby-lsp/lib/ruby_indexer/lib/ruby_indexer/index.rb:427:in 'RubyIndexer::Index#follow_aliased_namespace'
	from /Users/pkondzior/Projects/ruby-lsp/lib/ruby_indexer/lib/ruby_indexer/index.rb:1034:in 'RubyIndexer::Index#direct_or_aliased_constant'
	 ... 8542 levels...
	from /Users/pkondzior/Projects/ruby-lsp/lib/ruby_lsp/requests/references.rb:63:in 'RubyLsp::Requests::References#perform'
	from /Users/pkondzior/Projects/ruby-lsp/lib/ruby_lsp/server.rb:783:in 'RubyLsp::Server#text_document_references'
	from /Users/pkondzior/Projects/ruby-lsp/lib/ruby_lsp/server.rb:78:in 'RubyLsp::Server#process_message'
	from /Users/pkondzior/Projects/ruby-lsp/lib/ruby_lsp/base_server.rb:167:in 'block in RubyLsp::BaseServer#new_worker'
```

### Reproduction steps
1. Run this code in irb console where ruby-lsp is available
```ruby
require "uri"
require "ruby_lsp/internal"

index = RubyIndexer::Index.new

src = <<~RUBY
  module A
    module B
      class C; end
    end
  end

  module A
    module D
      B = B::C
    end
  end
RUBY

uri = URI.parse("file:///test.rb")
index.index_single(uri, src)

# Triggers stack overflow
index.resolve("A::D::B", [])
```

pkondzior:
### Motivation

Closes #3782

Fixes a `SystemStackError` that occurred when resolving certain constant aliases.  When a nested module defines a constant alias referencing another constant of the same short name in an outer namespace (for example, `B = B::C` inside `A::D`), the Ruby LSP indexer entered infinite recursion because `seen_names` was not propagated to `follow_aliased_namespace`.

### Implementation

Pass `seen_names` from `direct_or_aliased_constant` into `follow_aliased_namespace` to enable proper cycle detection.

### Automated Tests

Added a regression test `test_resolving_self_referential_constant_alias` covering the `B = B::C` case.

### Manual Tests

1. Open VS Code with Ruby LSP enabled.  
2. Create a file `test.rb`:
```ruby
module A
 module B
   class C
   end
 end
end

module A
 module D
   B = B::C
 end
end
```
3. Trigger Go to Definition or hover over B inside A::D.
4.Expected: No crash. B correctly resolves to A::B::C.

### DIFF
--- a/lib/ruby_indexer/lib/ruby_indexer/index.rb
+++ b/lib/ruby_indexer/lib/ruby_indexer/index.rb
@@ -1030,13 +1030,17 @@ module RubyIndexer
     #: (String full_name, Array[String] seen_names) -> Array[Entry::Constant | Entry::ConstantAlias | Entry::Namespace | Entry::UnresolvedConstantAlias]?
     def direct_or_aliased_constant(full_name, seen_names)
       if (entries = @entries[full_name])
-        return entries.map { |e| e.is_a?(Entry::UnresolvedConstantAlias) ? resolve_alias(e, seen_names) : e }
+        return entries.map do |e|
+          e.is_a?(Entry::UnresolvedConstantAlias) ? resolve_alias(e, seen_names) : e
+        end #: as Array[Entry::Constant | Entry::ConstantAlias | Entry::Namespace | Entry::UnresolvedConstantAlias])?
       end
 
       aliased = follow_aliased_namespace(full_name, seen_names)
       return if full_name == aliased || seen_names.include?(aliased)
 
-      @entries[aliased]&.map { |e| e.is_a?(Entry::UnresolvedConstantAlias) ? resolve_alias(e, seen_names) : e }
+      @entries[aliased]&.map do |e|
+        e.is_a?(Entry::UnresolvedConstantAlias) ? resolve_alias(e, seen_names) : e
+      end #: as Array[Entry::Constant | Entry::ConstantAlias | Entry::Namespace | Entry::UnresolvedConstantAlias])?
     end
 
     # Attempt to resolve a given unresolved method alias. This method returns the resolved alias if we managed to
--- a/lib/ruby_indexer/test/index_test.rb
+++ b/lib/ruby_indexer/test/index_test.rb
@@ -1168,10 +1168,11 @@ module RubyIndexer
         end
       RUBY
 
-      b_entry = @index.resolve("A::D::B", [])&.first #: as !nil
-      assert_equal(10, b_entry.location.start_line)
-      assert_equal("A::B::C", b_entry.target)
-      assert_instance_of(RubyIndexer::Entry::ConstantAlias, b_entry)
+      entry = @index.resolve("A::D::B", [])&.first #: as Entry::ConstantAlias
+
+      assert_kind_of(RubyIndexer::Entry::ConstantAlias, entry)
+      assert_equal(10, entry.location.start_line)
+      assert_equal("A::B::C", entry.target)
     end
 
     def test_resolving_qualified_references


lib/ruby_indexer/test/index_test.rb
```diff
@@ -1152,6 +1152,26 @@ class Foo
...
+
+        module A
+          module D
+            B = B::C
+          end
+        end
+      RUBY
+
+      foo_entry = @index.resolve("A::D::B", []) #: as !nil
+      # assert_equal(2, foo_entry.location.start_line)
```
vinistock: Did you mean to uncomment this? We're not currently asserting anything in this test.

It would be nice to verify that:

1. `A::D::B` is the `B` constant on line 9 of the Ruby snippet
2. That the target of the `B` alias is `A::B`

pkondzior: That was incomplete test, I've used it to trigger the issue but forgot to finish assertions. Addressed

lib/ruby_indexer/test/index_test.rb
```diff
@@ -1152,6 +1152,27 @@ class Foo
...
+        module A
+          module D
+            B = B::C
+          end
+        end
+      RUBY
+
+      b_entry = @index.resolve("A::D::B", [])&.first #: as !nil
+      assert_equal(10, b_entry.location.start_line)
+      assert_equal("B::C", b_entry.target)
```
vinistock: Shouldn't the target here be `A::B::C`?

pkondzior: Not sure, iI think this is another issue because alias resolution is still unable to solve this:
```
@index.send(:resolve_alias, b_entry, [])
=> #<RubyIndexer::Entry::UnresolvedConstantAlias:0x0000000122a91e60
 @comments="",
 @location=#<RubyIndexer::Location:0x0000000122a92040 @end_column=12, @end_line=10, @start_column=4, @start_line=10>,
 @name="A::D::B",
 @nesting=["A", "D"],
 @target="B::C",
 @uri=#<URI::Generic file:///fake/path/foo.rb>,
 @visibility=:public>
 ```

Comment (pkondzior) [reply lib/ruby_indexer/test/index_test.rb]: Seems like my change is only partially fixing the issue,  investigating

Comment (pkondzior) [reply lib/ruby_indexer/test/index_test.rb]: There ware two more issues that had to be fixed:
1. follow_aliased_namespace had to respect seen_names and skip entries which have been seen
2. direct_or_aliased_constant had to follow only not seen or non circular aliases

Comment (vinistock) [reply lib/ruby_indexer/test/index_test.rb]: Nice

File: lib/ruby_indexer/lib/ruby_indexer/index.rb [hunk 3]
```diff
@@ -1023,13 +1026,17 @@ def build_non_redundant_full_name(name, nesting)
...
+        return entries.map { |e| e.is_a?(Entry::UnresolvedConstantAlias) ? resolve_alias(e, seen_names) : e }
+      end
+
+      aliased = follow_aliased_namespace(full_name, seen_names)
+      return if full_name == aliased || seen_names.include?(aliased)
 
-      entries&.map do |e|
-        e.is_a?(Entry::UnresolvedConstantAlias) ? resolve_alias(e, seen_names) : e
-      end #: as Array[Entry::Constant | Entry::ConstantAlias | Entry::Namespace | Entry::UnresolvedConstantAlias])?
+      @entries[aliased]&.map { |e| e.is_a?(Entry::UnresolvedConstantAlias) ? resolve_alias(e, seen_names) : e }
```
vinistock: You'll need to cast these returns to match the method signature as we had before - since the method return is more specific than the general `Entry` types.

You can see more details in the linting errors.

pkondzior: Ah, i see.

pkondzior: I've ran locally `srb tc` and looks good

vinistock (approved):
Thank you for the PR!

PR #3790, opened 2025-10-20
Support multiple test file matches in go to relevant file

### LINKED ISSUE #3789: Open all directly associated test files when using the `go to tests` functionality

### Use case

We currently have the ability to go directly to any class' corresponding test file at the click of a button. It would be great to expand this functionality in cases where the tests are split across multiple files.

### Description

For example, given `app/models/user.rb`, we can use "Go to test" to open `test/models/user_test.rb`.

However, if the `User` model has become overly bloated and we split the unit tests (i.e. the same level of test abstraction) into `CreateUser`, `UpdateUser`, `DeleteUser` and `OtherUser` test files, we cannot open all of the associated files at once. This functionality would be extremely helpful not just in terms of DX but, also, opening all of the associated files would reduce the risk of a test file going "unseen" and therefore unmaintained.

### Implementation

_No response_

alexcrocha:
### Motivation

Closes #3789

"Go to relevant file" currently fails when tests for an implementation file are split into multiple test files within a subdirectory (e.g., `user.rb` having tests like `user/creation_test.rb`). This makes it harder to discover and maintain these split tests. This PR extends the feature to find all associated tests within a matching subdirectory.

Instead of opening all found files at once, we present a list for the developer to select from. This gives developers control over which file to open rather than flooding their editor with tabs.

<details><summary>Example image</summary>
 
image

</details> 

### Implementation

The implementation was changed to search using multiple glob patterns instead of just one. This involved refactoring the old `relevant_filename_pattern` method into `relevant_filename_patterns` to return an array of patterns.

This enables two key improvements:

**1. Finding tests in matching subdirectories:**

When triggering "Go to relevant file" on `app/models/user.rb`, the search now includes patterns that look inside a `user/` subdirectory, finding files like:

- `test/models/user_test.rb` (existing behavior)
- `test/models/user/create_user_test.rb` (**new**)
- `test/models/user/test_update_user.rb` (**new**)

**2. Navigating back from nested tests:**

The reverse navigation now also works for nested files. For example, from `test/requests/go_to_relevant_file/go_to_relevant_file_a_test.rb`, it will correctly find `lib/requests/go_to_relevant_file.rb`. This is done by checking if the parent directory's name is included in the test file's name.

 ### Automated Tests 

New tests have been added to cover these scenarios.

 ### Manual Tests

1.  In a test project, create the following file structure:

		.
		├── app
		│	└── models
		│		└── user.rb
		└── test
			└── models
				└── user
					├── create_user_test.rb
					└── update_user_test.rb
                
2.  Open `app/models/user.rb`.
3.  Click on the "Go to relevant file" button.
4.  **Expected:** A VS Code quick pick should appear with both `create_user_test.rb` and `update_user_test.rb` as options.
5.  Now open `test/models/user/create_user_test.rb`.
6.  Click on the "Go to relevant file" button.
7.  **Expected:** The editor should navigate directly to `app/models/user.rb`.

### DIFF
--- a/lib/ruby_lsp/requests/go_to_relevant_file.rb
+++ b/lib/ruby_lsp/requests/go_to_relevant_file.rb
@@ -35,8 +35,11 @@ module RubyLsp
 
       #: -> Array[String]
       def find_relevant_paths
-        pattern = File.join(search_root, "**", relevant_filename_pattern)
-        candidate_paths = Dir.glob(pattern)
+        patterns = relevant_filename_patterns
+
+        candidate_paths = patterns.flat_map do |pattern|
+          Dir.glob(File.join(search_root, "**", pattern))
+        end
 
         return [] if candidate_paths.empty?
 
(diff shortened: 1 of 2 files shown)


alexcrocha:
* **#3790** <a href="https://app.graphite.dev/github/pr/Shopify/ruby-lsp/3790?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/Shopify/ruby-lsp/3790?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`
<hr />

You must have a Graphite account in order to use the merge queue. Sign up using [this link](https://app.graphite.dev/invite/github/Shopify?ref=merge-queue-instructions-in-stack-comment&prId=3790).

<h2></h2>
<sub>

</sub>

<sub>

</sub>

<h2></h2>

Issue #3019, opened 2025-01-07
Version 0.23 breaks NeoVim integration


datanoise:
### Description

While typing Neovim sends incomplete code to ruby-lsp which results in errors like this:

`
[ERROR][2025-01-07 12:37:03] ...m/lsp/client.lua:1041	"LSP[ruby_lsp]"	"on_error"	{ code = "NO_RESULT_CALLBACK_FOUND", err = { id = 16, jsonrpc = "2.0", result = { items = { { message = "unexpected end-of-input, assuming it is closing the parent top level context", range = { ["end"] = { character = 0, line = 160 }, start = { character = 3, line = 159 } }, severity = 1, source = "Prism" }, { message = "expected an `end` to close the `class` statement", range = { ["end"] = { character = 0, line = 160 }, start = { character = 0, line = 160 } }, severity = 1, source = "Prism" }, { message = "mismatched indentations at 'end' with 'if' at 71", range = { ["end"] = { character = 5, line = 80 }, start = { character = 2, line = 80 } }, severity = 2, source = "Prism" }, { message = "mismatched indentations at 'end' with 'def' at 70", range = { ["end"] = { character = 3, line = 159 }, start = { character = 0, line = 159 } }, severity = 2, source = "Prism" } }, kind = "full" } } }
`

It is impossible to type now, because it seems these errors pop up on each key stroke. Reverting to version `0.22.1` fixes the issue.

vinistock:
Thank you for sharing that issue, it's very helpful. Could either of you try something?

If you edit the LSP code in place (`BUNDLE_GEMFILE=.ruby-lsp/Gemfile bundle open`) and remove [this line](https://github.com/Shopify/ruby-lsp/blob/cf1235e3125a4f9bbbc6d2c72b111803e00dec47/lib/ruby_lsp/base_server.rb#L166), does the issue go away?

Reading through those threads, what I understood is that a request cancelled by the client successfully has to continue returning the special cancelled error code even if the client decides to retry the cancelled request, but I want to be sure.

datanoise:
No change for me. the same errors are being reported.

adam12:
Assuming Neovim 0.10.3? There were a few LSP changes made.

- lsp: Retrigger diagnostics request on server cancellation (https://github.com/neovim/neovim/pull/31345) (https://github.com/neovim/neovim/pull/31427)
- lsp: Cancel pending requests before refreshing (https://github.com/neovim/neovim/pull/31500)

[...]

PR #3791, opened 2025-10-24
Restore request cancellation with correct implementation

### LINKED ISSUE #3063: Implement proper request cancellation
We currently have request cancellation implemented incorrectly. We are pushing the requests and their cancellation notification all in the incoming queue. Since `Thread::Queue#pop` pops items in the order in which they were pushed, we always see the cancellation _after_ we already ran the request, which completely defeats the whole purpose of cancelling it.

We tried fixing that in #2939, but it caused issues for NeoVim #3019. We need to review our understanding of what needs to happen with a cancellation and ensure it's working properly.

What we need to clarify is if a request is cancelled, what should the server return once we get to that request? Current understanding is the error code `RequestCancelled`, but maybe it shouldn't return anything and should only return that error is the request is tried? In essence, what are the expectations from the server for the first response for a cancelled request and any subsequent ones?

Relevant spec page: https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#cancelRequest

janko:
### Motivation

Closes #3063

The project symbol search is currently slow in Zed, because it sends a `workspace/symbol` request for every new letter typed in, and attempts to cancel previous requests aren't successful, as Ruby LSP currently doesn't actually cancel requests (it enqueues `$/cancelRequest` messages instead of processing them synchronously).

### Implementation

Fix for request cancellation was already attempted in https://github.com/Shopify/ruby-lsp/issues/3063, but it caused issues with Neovim (https://github.com/Shopify/ruby-lsp/issues/3019), so it was ultimately reverted.

The problem wasn't actually in the new error responses, it was that the attempted fix contained a bug, where **it was returning a regular response *in addition to* the error response**. So, request cancellation would result in *two* responses being returned for the same request, which rightfully tripped Neovim up.

The bug was in attempting to skip the regular response using `next`. In addition to advancing iteration of nearest loop, `next` also breaks out of the nearest closure, whichever comes first. In this case, the surrounding `Mutex#synchronize` block was the nearest, so it broke out of that, and proceeded to execute `#process_message`. This resulted in regular response being sent back even if the cancelled response was previously sent.

To avoid introducing a boolean local variable, I extracted the handling logic into a method, so that I can break out using `return`.

### Automated Tests

I updated the tests to match the original implementation. However, I didn't know how to reliably assert that the 2nd response isn't getting sent, because I don't know how to wait until the worker becomes idle. I feel like I don't have enough control over the concurrency.

### Manual Tests

I tested this in Zed, and everything looked correct in the RCP message log.

### DIFF
--- a/lib/ruby_lsp/base_server.rb
+++ b/lib/ruby_lsp/base_server.rb
@@ -83,7 +83,7 @@ module RubyLsp
         # The following requests need to be executed in the main thread directly to avoid concurrency issues. Everything
         # else is pushed into the incoming queue
         case method
-        when "initialize", "initialized", "rubyLsp/diagnoseState"
+        when "initialize", "initialized", "rubyLsp/diagnoseState", "$/cancelRequest"
           process_message(message)
         when "shutdown"
           @global_state.synchronize do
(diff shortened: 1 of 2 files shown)


janko:
Maybe someone can help test this in Neovim? @adam12, @natematykiewicz, or @dzirtusss?

I think it's enough to add the following to the Gemfile:

```rb
gem "ruby-lsp", github: "janko/ruby-lsp", branch: "restore-request-cancellation"
```

adam12:
I'll try to test this week.

adam12:
I've observed no issues running this branch over the last few days. :+1:

lib/ruby_lsp/base_server.rb
```diff
@@ -154,20 +154,28 @@ def fail_request_and_notify(id, message, type: Constant::MessageType::INFO)
...
-              next
-            end
-          end
+          handle_incoming_message(message)
+        end
+      end
+    end
 
-          process_message(message)
+    def handle_incoming_message(message)
```
vinistock, line 163: We're missing the sig here
```suggestion
    #: (Hash[Symbol, untyped]) -> void
    def handle_incoming_message(message)
```

janko: Added, thanks 👍🏻

vinistock:
Good to ship once the CLA bot is happy 👍

janko:
I have signed the CLA!

vinistock:

PR #3792, opened 2025-10-25
Speed up workspace symbol search by filtering before fuzzy matching

### LINKED ISSUE #2660: Provide a setting to automatically exclude all gems

janko:
### Motivation

Closes #2660

I want workspace symbol searches to be faster so that I use them more often in my code editor.

### Implementation

Currently, fuzzy matching is performed on all symbols, and then dependency/private entries are filtered out. With the assumption that filtering is faster than fuzzy matching, I flipped this around, by first filtering out dependency/private entires, and then doing the fuzzy matching on the remainder.

This turns out to be **~2-4x faster** in my benchmarks.

### Automated Tests

I didn't update automated tests.

### DIFF
--- a/lib/ruby_indexer/lib/ruby_indexer/index.rb
+++ b/lib/ruby_indexer/lib/ruby_indexer/index.rb
@@ -195,12 +195,13 @@ module RubyIndexer
     end
 
     # Fuzzy searches index entries based on Jaro-Winkler similarity. If no query is provided, all entries are returned
-    #: (String? query) -> Array[Entry]
-    def fuzzy_search(query)
+    #: (String? query) ?{ (Entry) -> bool? } -> Array[Entry]
+    def fuzzy_search(query, &condition)
       unless query
         entries = @entries.filter_map do |_name, entries|
           next if entries.first.is_a?(Entry::SingletonClass)
 
+          entries = entries.select(&condition) if condition
           entries
         end
 
@@ -211,6 +212,9 @@ module RubyIndexer
 
       results = @entries.filter_map do |name, entries|
         next if entries.first.is_a?(Entry::SingletonClass)
+
+        entries = entries.select(&condition) if condition
+        next if entries.empty?
 
         similarity = DidYouMean::JaroWinkler.distance(name.gsub("::", "").downcase, normalized_query)
         [entries, -similarity] if similarity > ENTRY_SIMILARITY_THRESHOLD
--- a/lib/ruby_lsp/requests/workspace_symbol.rb
+++ b/lib/ruby_lsp/requests/workspace_symbol.rb
@@ -20,17 +20,7 @@ module RubyLsp
       # @override
       #: -> Array[Interface::WorkspaceSymbol]
       def perform
-        @index.fuzzy_search(@query).filter_map do |entry|
-          uri = entry.uri
-          file_path = uri.full_path
-
-          # We only show symbols declared in the workspace
-          in_dependencies = file_path && !not_in_dependencies?(file_path)
-          next if in_dependencies
-
-          # We should never show private symbols when searching the entire workspace
-          next if entry.private?
-
+        fuzzy_search.filter_map do |entry|
           kind = kind_for_entry(entry)
           loc = entry.location
 
@@ -44,13 +34,31 @@ module RubyLsp
             container_name: container.join("::"),
             kind: kind,
             location: Interface::Location.new(
-              uri: uri.to_s,
+              uri: entry.uri.to_s,
               range:  Interface::Range.new(
                 start: Interface::Position.new(line: loc.start_line - 1, character: loc.start_column),
                 end: Interface::Position.new(line: loc.end_line - 1, character: loc.end_column),
               ),
             ),
           )
+        end
+      end
+
+      private
+
+      #: -> Array[RubyIndexer::Entry]
+      def fuzzy_search
+        @index.fuzzy_search(@query) do |entry|
+          file_path = entry.uri.full_path
+
+          # We only show symbols declared in the workspace
+          in_dependencies = file_path && !not_in_dependencies?(file_path)
+          next if in_dependencies
+
+          # We should never show private symbols when searching the entire workspace
+          next if entry.private?
+
+          true
         end
       end
     end


lib/ruby_lsp/requests/workspace_symbol.rb
```diff
@@ -53,6 +43,23 @@ def perform
           )
         end
       end
+
+      private
+
+      def fuzzy_search
```
pkondzior, line 50: Why this indirection ?

lib/ruby_indexer/lib/ruby_indexer/index.rb
```diff
@@ -196,11 +196,12 @@ def prefix_search(query, nesting = nil)
 
     # Fuzzy searches index entries based on Jaro-Winkler similarity. If no query is provided, all entries are returned
     #: (String? query) -> Array[Entry]
-    def fuzzy_search(query)
+    def fuzzy_search(query, &condition)
       unless query
         entries = @entries.filter_map do |_name, entries|
           next if entries.first.is_a?(Entry::SingletonClass)
 
+          entries = entries.select(&condition) if condition
```
pkondzior, line 204: what is condition doing ?

lib/ruby_lsp/requests/workspace_symbol.rb
```diff
@@ -53,6 +43,23 @@ def perform
...
+        @index.fuzzy_search(@query) do |entry|
+          file_path = entry.uri.raw_path
```
pkondzior: I am not sure if this is the right approach, we could rather cache this on entry level instead of skipping unscapping parser

lib/ruby_indexer/lib/ruby_indexer/uri.rb
```diff
@@ -31,6 +31,7 @@ def from_path(path:, fragment: nil, scheme: "file", load_path_entry: nil)
         end
 
         uri = build(scheme: scheme, path: escaped_path, fragment: fragment)
+        uri.raw_path = path
```
pkondzior: I dont think we should do that, we should rather split whole work into platform specifci path unscaping/escaping and then use the right code paths on the right platform, also the result can be cached on the entry level

lib/ruby_lsp/requests/workspace_symbol.rb
janko: What indirection are you referring to? I needed to pass a block to `Index#fuzzy_search`, and instead of

```rb
entries = @index.fuzzy_search(@query) do
  # ...
end
entries.flat_map do |entry|
  # ...
end
```

I chose to extract the first method invocation together with its block into a separate method.

lib/ruby_indexer/lib/ruby_indexer/index.rb
janko: It's calling the block passed to it by `WorkspaceSymbol#fuzzy_search`, in order to perform file filtering before the fuzzy match, as described in the PR.

lib/ruby_lsp/requests/workspace_symbol.rb
janko: Why do the unescaping at all? We had the full path when building the URI, why reconstruct it?

lib/ruby_indexer/lib/ruby_indexer/uri.rb
janko: The unescaping was done regardless of the platform, and that was the main bottleneck in my profiling. The secondary bottleneck was the `Regex#match?` in the Windows code path, that could arguably be skipped non non-Windows platforms. But since I skipped the reconstruction of the file path entirely, there was no need.

vinistock: Is there a difference between `raw_path` and simply `path` (already present in the URI object)?

lib/ruby_lsp/requests/workspace_symbol.rb
vinistock: It won't work properly because the URI path and file_path aren't the same thing on Windows and our comparison for `not_in_dependencies` will fail.

```ruby
uri = URI("file:///C:/ruby/something.rb")

# The URI's path is not a valid file path
uri.path # => "/C:/ruby/something.rb"

# It's the handling of `to_standardized_path` that turns it into the correct one
uri.to_standardized_path # => "C:/ruby/something.rb"
```

lib/ruby_indexer/lib/ruby_indexer/uri.rb
janko: Yes, `path` is the URI-escaped version of file path, while `raw_path` is the unchanged file path. For example, `path` will keep any spaces in the path percent-encoded:

```rb
uri = URI::Generic.from_path(path: "/Users/janko/Library/Application Support")
uri.path     # => "/Users/janko/Library/Application%20Support"
uri.raw_path # => "/Users/janko/Library/Application Support"
```

The problem is that with the URI-escaping we lose the information of the valid file path that was passed to `URI::Generic.from_path`, so `URI::Generic#to_standardized_path` needs to reconstruct it.

I proposed storing the original valid file path to avoid having to do that work.

lib/ruby_lsp/requests/workspace_symbol.rb
janko: Correct, but my understanding is that it's because of the conversion done in `URI::Generic.from_path`. The path that's passed to that method should be correct even Windows, shouldn't it? That's what's stored in `raw_path`.

janko:
@vinistock As proposed in https://github.com/Shopify/ruby-lsp/pull/3806#issuecomment-3512062729, I removed the `URI::Generic#to_standardized_path` optimizations and kept only doing filtering before fuzzy matching.

Contrary to my initial benchmarks, this change alone seems to provide a **~2-4x speedup**, so it does make sense on its own 🙂

vinistock (approved):
Thanks for the contribution!

PR #3799, opened 2025-10-27
Fix stack StackLevelError during self referential constant resolution

### LINKED ISSUE #3798: Stack level too deep on self referential constant alias
### Description

Self referential Constant alias are causing SystemStackError

### Ruby LSP Information

Following code snippet is going to cause stack level too deep 

### Reproduction steps

1. Open VS Code editor
2. Create test.rb with following content
```ruby
module RealClass
  CONSTANT = {}
end

module Foo
  SomeClass = ::SomeClass
  RealClass = ::RealClass

  UNRESOLVED = SomeClass::CONSTANT
  CONSTANT = RealClass::CONSTANT
end


RealClass
Foo::UNRESOLVED
```
3. Hoover onto `Foo::UNRESOLVED`
4. Go to Output, select Ruby LSP and observer there is SystemStackError exception:
```log
2025-10-27 12:09:38.460 [info] (test) #<Thread:0x00000001205de6e0 /Users/pkondzior/Projects/ruby-lsp/vendor/bundle/ruby/3.4.0/bundler/gems/ruby-lsp-5333bb8d45ea/lib/ruby_lsp/base_server.rb:155 run> terminated with exception (report_on_exception is true):

2025-10-27 12:09:38.468 [info] (test) /Users/pkondzior/Projects/ruby-lsp/vendor/bundle/ruby/3.4.0/bundler/gems/ruby-lsp-5333bb8d45ea/lib/ruby_indexer/lib/ruby_indexer/index.rb:424:in 'Integer#downto': stack level too deep (SystemStackError
```

pkondzior:
### Motivation

Closes #3798

Following code snipper is going to cause StackLevelError and will kill RubyLSP process

### Implementation

`ruby-lsp` is trying to create ConstantAlias from UnresolvedConstantAlias during constant resolution, I am checking if the target matches entry in cases when unresolved entry is same as target entry we should bail out from resolution because thats the same unresolved entry.

### Automated Tests

I've created `RubyIndexer::ConstantTest#test_self_referential_constant_alias` that is giving coverage for this case

### Manual Tests
1. Bootstrap VS Code ruby-lsp from this branch
2. Paste following code into editor
```Ruby
module RealClass
  CONSTANT = {}
end

module Foo
  SomeClass = ::SomeClass
  RealClass = ::RealClass

  UNRESOLVED = SomeClass::CONSTANT
  CONSTANT = RealClass::CONSTANT
end


RealClass
Foo::UNRESOLVED
```
3. Working on the file should not cause any errors on Ruby LSP in the output

### DIFF
--- a/lib/ruby_indexer/test/constant_test.rb
+++ b/lib/ruby_indexer/test/constant_test.rb
@@ -318,28 +318,6 @@ module RubyIndexer
       assert_instance_of(Entry::Constant, constant)
     end
 
-    def test_self_referential_constant_alias
-      index(<<~RUBY)
-        module RealClass
-          CONSTANT = {}
-        end
-
-        module Foo
-          SomeClass = ::SomeClass
-          RealClass = ::RealClass
-
-          UNRESOLVED = SomeClass::CONSTANT
-          CONSTANT = RealClass::CONSTANT
-        end
-      RUBY
-
-      assert_entry("RealClass", Entry::Module, "/fake/path/foo.rb:0-0:2-3")
-      assert_no_entry("SomeClass")
-      assert_entry("Foo", Entry::Module, "/fake/path/foo.rb:4-0:10-3")
-      assert_entry("RealClass::CONSTANT", Entry::Constant, "/fake/path/foo.rb:1-2:1-15")
-      assert_entry("Foo::UNRESOLVED", Entry::UnresolvedConstantAlias, "/fake/path/foo.rb:8-2:8-34")
-    end
-
     def test_indexing_or_and_operator_nodes
       index(<<~RUBY)
         A ||= 1
--- a/lib/ruby_indexer/lib/ruby_indexer/index.rb
+++ b/lib/ruby_indexer/lib/ruby_indexer/index.rb
@@ -903,6 +903,7 @@ module RubyIndexer
       return entry if seen_names.include?(alias_name)
 
       seen_names << alias_name
+
       target = resolve(entry.target, entry.nesting, seen_names)
       return entry unless target
 
--- a/lib/ruby_indexer/test/index_test.rb
+++ b/lib/ruby_indexer/test/index_test.rb
@@ -1175,6 +1175,23 @@ module RubyIndexer
       assert_equal("A::B::C", entry.target)
     end
 
+    def test_resolving_non_existing_self_referential_constant_alias
+      index(<<~RUBY)
+        module Foo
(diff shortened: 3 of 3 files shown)


lib/ruby_indexer/test/constant_test.rb
```diff
@@ -318,6 +318,28 @@ module A
       assert_instance_of(Entry::Constant, constant)
     end
 
+    def test_self_referential_constant_alias
```
vinistock: This test seems to pass even without the fix. It's because we're not invoking resolution as part of it. Also, I think just the unresolved portion of the code is enough to trigger the problem.

Can we please move this test to `index_test.rb` (where the rest of resolution is tested) and simplify the example to this:

```ruby
def test_self_referential_constant_alias
  index(<<~RUBY)
    module Foo
      SomeClass = ::SomeClass
      UNRESOLVED = SomeClass::CONSTANT
    end
  RUBY

  # try to resolve `Foo::UNRESOLVED` first
  entry = @index.resolve("Foo::UNRESOLVED", []) #: as !nil
  # add some asserts about what gets resolved
  
  # then try to resolve the alias' target inside of the `Foo` namespace
  entry = @index.resolve("SomeClass::Constant", ["Foo"]) #: as !nil
  # add some asserts about what gets resolved
end
```

This fails as expected without the fix.

pkondzior: Indeed, this has to go the way you have presented, in other [PR](https://github.com/Shopify/ruby-lsp/pull/3784/files#diff-5a27f6ba48b00a6aaadd1d4b91b5ac4b3e2ff6a180cb887dd9f3d012b55566e0R388) where this code has been introduced I am doing constant resolution at the end of the indexing step therefor the Index#resolve calls go through unresolved entries automatically, here we need that explicit harness.  I will adjust the code

pkondzior: Addressed in https://github.com/Shopify/ruby-lsp/pull/3799/commits/f8e5153b65e92174ceb3e095fc2cac875aab4a07

vinistock: Yeah, no resolution happens in the indexing step. Due to performance reasons for large codebases, all resolution is lazy.

We're currently in the process of taking a broader look into indexing for improving correctness and performance, so this should be improved soon.

vinistock (approved):

PR #3805, opened 2025-10-28
Make URI parsing more platform aware

pkondzior:
### Motivation

Related to #2660 

Makes URI parsing and normalization platform-aware, by splitting code paths into separate methods.

Overall on macos this brings following speedups:
- textDocument/diagnostic faster by 3.784 %
- workspace/symbol faster by 38.197 %

### Implementation

1. Split path handling into:
- `from_win_path` (Windows)  
- `from_unix_path` (Unix)  
- `from_path` now auto-selects based on platform.
2. Added platform-specific methods:  
- `to_standardized_win_path` (Windows)  
- `to_standardized_unix_path` (Unix)
- `to_standardized_path` now auto-selects based on platform.
3. Centralized unsafe regex `UNSAFE_REGEX` , previously it was always instantiated on `from_path` method call 
4. Improved escaping/unescaping and handling of Windows long paths (`//?/`).
5. Extended Sorbet RBI with `from_path`, `to_standardized_path`, and `full_path` signatures.

### Automated Tests

- Updated tests to cover both platforms and renamed method calls.

### DIFF
--- a/lib/ruby_indexer/lib/ruby_indexer/uri.rb
+++ b/lib/ruby_indexer/lib/ruby_indexer/uri.rb
@@ -12,22 +12,22 @@ module URI
     # NOTE: We also define this in the shim
     PARSER = const_defined?(:RFC2396_PARSER) ? RFC2396_PARSER : DEFAULT_PARSER
 
+    # This unsafe regex is the same one used in the URI::RFC2396_REGEXP class with the exception of the fact that we
+    # do not include colon as a safe character. VS Code URIs always escape colons and we need to ensure we do the
+    # same to avoid inconsistencies in our URIs, which are used to identify resources
+    UNSAFE_REGEX = %r{[^\-_.!~*'()a-zA-Z\d;/?@&=+$,\[\]]}
+
     class << self
       #: (path: String, ?fragment: String?, ?scheme: String, ?load_path_entry: String?) -> URI::Generic
-      def from_path(path:, fragment: nil, scheme: "file", load_path_entry: nil)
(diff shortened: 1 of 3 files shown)


lib/ruby_indexer/lib/ruby_indexer/uri.rb
```diff
@@ -52,21 +67,37 @@ def add_require_path_from_load_entry(load_path_entry)
...
-    def to_standardized_path
+    # On Windows, when we're getting the file system path back from the URI, we need to remove the leading forward
+    # slash
+    def to_standardized_win_path
       parsed_path = path
+
       return unless parsed_path
 
+      # we can bail out parsing if there is nothing to unescape
+      return parsed_path unless parsed_path.match?(/%[0-9A-Fa-f]{2}/)
```
vinistock, line 78: Why can we avoid unescaping here?

lib/ruby_indexer/lib/ruby_indexer/uri.rb
```diff
@@ -52,21 +67,37 @@ def add_require_path_from_load_entry(load_path_entry)
...
     end
 
-    alias_method :full_path, :to_standardized_path
+    #: -> String?
+    def to_standardized_unix_path
+      unescaped_path = path
+      return unless unescaped_path
+
+      # we can bail out parsing if there is nothing to be unescaped
+      return unescaped_path unless unescaped_path.match?(/%[0-9A-Fa-f]{2}/)
```
vinistock, line 95: Same question, why does this not need to be escaped?

PR #3806, opened 2025-10-29
Add Entry#in_dependencies? and refactor dependency checks

pkondzior:
### Motivation

Related #2660

This simple caching of uri path, file_name and dependency check on entry level provides significant performance improvements and on workspace/symbol makes it faster by 60.616 %

### Implementation

This commit introduces a new method `Entry#in_dependencies?` to encapsulate dependency path checks within the `Entry` class and refactors existing dependency checks across the codebase to use it. The logic for determining whether a file is inside dependencies was moved from `RubyLsp::Requests::Support::Common` into `RubyLsp::Utils`.

### Automated Tests

I wasn't sure if I should introduce any entry testing above what is being used now. If this will be required and change is considering useful for the project looking forward for recommendation how this should be tested on unit level (entry.rb does have its direct tests)

### DIFF
--- a/lib/ruby_indexer/lib/ruby_indexer/entry.rb
+++ b/lib/ruby_indexer/lib/ruby_indexer/entry.rb
@@ -41,20 +41,31 @@ module RubyIndexer
       @visibility == :private
     end
 
+    #: -> bool?
+    def in_dependencies?
+      @in_dependencies ||= if file_path
+        !RubyLsp.not_in_dependencies?(
+          file_path, #: as String
+        )
+      else
+        false
+      end #: bool?
+    end
+
     #: -> String
     def file_name
-      if @uri.scheme == "untitled"
+      @file_name ||= if @uri.scheme == "untitled"
         @uri.opaque #: as !nil
       else
         File.basename(
           file_path, #: as !nil
         )
-      end
+      end #: String?
     end
 
     #: -> String?
     def file_path
-      @uri.full_path
+      @file_path ||= @uri.full_path #: String?
     end
 
     #: -> String
--- a/lib/ruby_lsp/listeners/definition.rb
+++ b/lib/ruby_lsp/listeners/definition.rb
@@ -316,8 +316,8 @@ module RubyLsp
 
         methods.each do |target_method|
           uri = target_method.uri
-          full_path = uri.full_path
-          next if @sorbet_level.true_or_higher? && (!full_path || not_in_dependencies?(full_path))
+
+          next if @sorbet_level.true_or_higher? && !target_method.in_dependencies?
 
           @response_builder << Interface::LocationLink.new(
             target_uri: uri.to_s,
@@ -390,9 +390,8 @@ module RubyLsp
           # additional behavior on top of jumping to RBIs. The only sigil where Sorbet cannot handle constants is typed
           # ignore
           uri = entry.uri
-          full_path = uri.full_path
 
-          if !@sorbet_level.ignore? && (!full_path || not_in_dependencies?(full_path))
+          if !@sorbet_level.ignore? && !entry.in_dependencies?
             next
           end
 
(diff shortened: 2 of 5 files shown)


lib/ruby_indexer/lib/ruby_indexer/entry.rb
```diff
@@ -41,20 +41,31 @@ def private?
       @visibility == :private
     end
 
+    #: -> bool?
+    def in_dependencies?
```
vinistock, line 45: Maybe we can invert the logic here since we almost always want to check if something is defined in the workspace and end up negating the result everywhere.

How about we make this into `in_workspace?` and then we can use it more directly?

lib/ruby_indexer/lib/ruby_indexer/entry.rb
```diff
@@ -41,20 +41,31 @@ def private?
...
+        )
+      else
+        false
+      end #: bool?
+    end
+
     #: -> String
     def file_name
-      if @uri.scheme == "untitled"
+      @file_name ||= if @uri.scheme == "untitled"
```
vinistock, line 57: Does the file name have significant impact on performance? I can see `Uri#full_path` being slow, but I'm surprised that `File.basename` introduces significant overhead.

vinistock (commented):
@janko does this change achieve similar results as #3792?

If so, I think we can move forward with this one

janko:
@vinistock I'm not noticing any speedup in workspace symbol searches with these changes alone. This is because fuzzy matching still runs before filtering, so dependency checks will run only on a small subset of entries, while fuzzy matching will run on all entries (and is the bottleneck).

We could combine flipping the order of fuzzy matching and filtering from https://github.com/Shopify/ruby-lsp/pull/3792 with the dependency check speedup from this PR. When I try it locally, then workspace symbol search is about 20% faster than #3792.

<details>
<summary>Changes</summary>

```diff
diff --git a/lib/ruby_indexer/lib/ruby_indexer/index.rb b/lib/ruby_indexer/lib/ruby_indexer/index.rb
index ca4b18f2..deb91532 100644
--- a/lib/ruby_indexer/lib/ruby_indexer/index.rb
+++ b/lib/ruby_indexer/lib/ruby_indexer/index.rb
@@ -196,11 +196,12 @@ module RubyIndexer
 
     # Fuzzy searches index entries based on Jaro-Winkler similarity. If no query is provided, all entries are returned
     #: (String? query) -> Array[Entry]
-    def fuzzy_search(query)
+    def fuzzy_search(query, &condition)
       unless query
         entries = @entries.filter_map do |_name, entries|
           next if entries.first.is_a?(Entry::SingletonClass)
 
+          entries = entries.select(&condition) if condition
           entries
         end
 
@@ -212,6 +213,9 @@ module RubyIndexer
       results = @entries.filter_map do |name, entries|
         next if entries.first.is_a?(Entry::SingletonClass)
 
+        entries = entries.select(&condition) if condition
+        next if entries.empty?
+
         similarity = DidYouMean::JaroWinkler.distance(name.gsub("::", "").downcase, normalized_query)
         [entries, -similarity] if similarity > ENTRY_SIMILARITY_THRESHOLD
       end
diff --git a/lib/ruby_lsp/requests/workspace_symbol.rb b/lib/ruby_lsp/requests/workspace_symbol.rb
index 7c6186e4..b1654db8 100644
--- a/lib/ruby_lsp/requests/workspace_symbol.rb
+++ b/lib/ruby_lsp/requests/workspace_symbol.rb
@@ -20,15 +20,17 @@ module RubyLsp
       # @override
       #: -> Array[Interface::WorkspaceSymbol]
       def perform
-        @index.fuzzy_search(@query).filter_map do |entry|
-          uri = entry.uri
-
+        @index.fuzzy_search(@query) do |entry|
           # We only show symbols declared in the workspace
           next if entry.in_dependencies?
 
           # We should never show private symbols when searching the entire workspace
           next if entry.private?
 
+          true
+        end.map do |entry|
+          uri = entry.uri
+
           kind = kind_for_entry(entry)
           loc = entry.location
```
</deatils>

vinistock:
Ah, right. Let's move forward with this one for the URI related speed up and make #3792 only about filtering - especially if the speed up is even higher.

PR #3807, opened 2025-10-29
Add support for Package URL format for source links

paracycle:
### Motivation

Early on in the lifetime of this project we'd created our own source URI format to target locations inside gems. However, we'd missed the fact that a [Package URL format](https://github.com/package-url/purl-spec) already exists as a well adopted specification and addresses every need that we have.

### Implementation

I implemented the Package URL parsing in parallel to parsing the existing Source URI comments. Once Tapioca starts generating Package URL comments, and once we feel confident that it has wide adoption, we can stop handling Source URI comments in the future.

### Automated Tests

Added new expectation tests for document link with Package URL format source links.


ruby-lsp.gemspec
```diff
@@ -20,6 +20,7 @@ Gem::Specification.new do |s|
 
   # Dependencies must be kept in sync with the checks in the extension side on workspace.ts
   s.add_dependency("language_server-protocol", "~> 3.17.0")
+  s.add_dependency("packageurl-ruby")
```
vinistock: I really like the idea of using the Package URL spec, but we've been trying really hard to reduce runtime dependencies because the more we have, the larger the surface area for trouble with the LSP auto updates.

I actually had begun working on removing `language_server-protocol` too, so that we could minimize our dependencies to just the parsers - but I haven't finished it yet.

Do you think it would be too messy to implement a subclass of `URI::Generic` for this? Based on the examples on the PackageURL repo, they are all valid URIs, they just need some extra handling to separate certain bits from the `opaque`.

paracycle: Thanks. Since the upstream implementation is a single file, I decided to vendor the implementation into the Ruby LSP source tree with the license intact.

sorbet/rbi/gems/packageurl-ruby@0.2.0.rbi
vinistock, line 1: Do we still need this RBI?

paracycle: Oops, I'll remove.

PR #3813, opened 2025-11-03
Implement `workspace/didChangeConfiguration` and `workspace/configuration`

### LINKED ISSUE #3785: Support `workspace/didChangeConfiguration`

### Use case

I would make it possible to use a per-project configuration with eglot on Emacs using a default way to do that via `.dir-locals.el`.

### Description

I have been trying to set up Ruby LSP via eglot with a per-project settings (using StandardRB addon) via `.dir-locals.el`, as described [here](https://www.gnu.org/software/emacs/manual/html_node/eglot/Project_002dspecific-configuration.html). However, the formatting did not work as the server kept telling me it does not support `:documentFormattingProvider`. I found that this is because eglot sends `workspace/didChangeConfiguration` after having initialized the language server, but Ruby LSP does not support such reconfiguration.

I found https://github.com/Shopify/ruby-lsp/issues/205 where this was considered, but dropped, because it was not needed for VSCode.

### Implementation

I think with some guidance, I can implement it. It seems that what's mainly needed in response to this message is to update the global state and send new capabilities to the client. Although there might be some other details I'm not able to predict right now.

katafrakt:
### Motivation

Closes #3785

`workspace/didChangeConfiguration` and `workspace/configuration` are part of the spec. They are used to dynamically change the configuration of the server after initialization. With `eglot`, they are used to provide per-project configuration via `.dir-locals.el` files (see: https://www.gnu.org/software/emacs/manual/html_node/eglot/Project_002dspecific-configuration.html).

### Implementation

This adds support for the pull model of dynamic configuration changes, as described here: https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#workspace_configuration

Upon receiving `workspace/didChangeConfiguration` message the Ruby LSP server sends `workspace/configuration` to the client. The client replies. This reply is different than other replies handled by Ruby LSP so far, as it does not include `method` and needs to be associated with the request via `id`. This required implementing a collection of server-sent requests, to be able to match.

The `result` of the reply is not a hash, instead it's an array of hashes, so it needed to be handled separately as well.

Technically, upon receiving `workspace/configuration` the server should check if it should register or unregister some capabilities. I intended to do that too, but it started to become messy and also I did not have a way to properly test it, as `eglot` does not support dynamic registration of capabilities.

### Automated Tests

TBH it's a bit short on tests. I wasn't sure where/how to add more. I welcome all the suggestions in that area, because it feels a bit undertested.

### Manual Tests

I was able to swap the formatter and linter from Rubocop to StandardRB in `eglot` using this `.dir-locals.el`:

```emacs-lisp
 ((nil
   . ((eglot-workspace-configuration
       . (:rubyLsp
          (:formatter "standard" :linters ["standard"]
           :enabledFeatures (:codeActions t :diagnostics t :formatting t)))))))
```

### DIFF
--- a/lib/ruby_lsp/server.rb
+++ b/lib/ruby_lsp/server.rb
@@ -298,9 +298,6 @@ module RubyLsp
           rename_provider: rename_provider,
           references_provider: !@global_state.has_type_checker,
           document_range_formatting_provider: true,
-          workspace: {
-            configuration: true,
-          },
           experimental: {
             addon_detection: true,
             compose_bundle: true,
--- a/lib/ruby_lsp/client_capabilities.rb
+++ b/lib/ruby_lsp/client_capabilities.rb
@@ -11,7 +11,8 @@ module RubyLsp
       :window_show_message_supports_extra_properties,
       :supports_progress,
       :supports_diagnostic_refresh,
-      :supports_code_lens_refresh
+      :supports_code_lens_refresh,
+      :supports_workspace_configuration
 
     #: -> void
     def initialize
@@ -38,6 +39,9 @@ module RubyLsp
 
       # The editor supports server initiated refresh for code lenses
       @supports_code_lens_refresh = false #: bool
+
+      # The editor supports querying for its configuration via `workspace/configuration`
+      @supports_workspace_configuration = false #:bool
     end
 
     #: (Hash[Symbol, untyped] capabilities) -> void
@@ -66,6 +70,7 @@ module RubyLsp
 
       @supports_diagnostic_refresh = workspace_capabilities.dig(:diagnostics, :refreshSupport) || false
       @supports_code_lens_refresh = workspace_capabilities.dig(:codeLens, :refreshSupport) || false
+      @supports_workspace_configuration = workspace_capabilities[:configuration]
     end
 
     #: -> bool


lib/ruby_lsp/server.rb
```diff
@@ -196,6 +206,21 @@ def load_addons(include_project_addons: true)
...
+      # This assumes that the workspace configuration is under "rubyLsp" key, which seems
+      # to be the standard naming convention.
+      send_message(Request.workspace_configuration(
+        @current_request_id, section: "rubyLsp"
+      ))
+    end
+
+    def workspace_configuration_received(message)
+      options = { initializationOptions: message[:result]&.first }
+      messages_to_send = @global_state.apply_options(options)
```
katafrakt, line 220: This is a hack to reuse `apply_options`, which expects `initializationOptions`, but perhaps should be properly extracted to a method just accepting the options to apply.

Also, this just takes first hash from the reply's `result`, which works here, because we only asked for one configuration item, but technically is not exactly in line with the spec.

lib/ruby_lsp/server.rb
```diff
@@ -196,6 +206,21 @@ def load_addons(include_project_addons: true)
 
     private
 
+    #: (Hash[Symbol, untyped] message) -> void
+    def workspace_configuration_did_change(message)
```
vinistock, line 210: My understanding based on the [spec](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#didChangeConfigurationParams) is that the parameters of `workspace/didChangeConfiguration` already include which settings are supposed to be changed.

Why do we need to make a request to the client? Could we use what we receive as parameters instead?

lib/ruby_lsp/base_server.rb
```diff
@@ -21,7 +22,15 @@ def initialize(**options)
       @outgoing_dispatcher = Thread.new do
         unless @test_mode
           while (message = @outgoing_queue.pop)
-            @global_state.synchronize { @writer.write(message.to_hash) }
+            @global_state.synchronize do
```
vinistock, line 25: We already process one response for server->client requests [here](https://github.com/Shopify/ruby-lsp/blob/f38329fccca1c0eaa45149f08f72655bc3cae781/lib/ruby_lsp/server.rb#L159).

Are we able to reuse that and avoid changing the base server?

lib/ruby_lsp/server.rb
vinistock: Yeah, this method is coupling information that we get during initialization (like the workspace folder) with other settings. It would indeed be nice to split into two methods:

1. One that handles information required by the spec (workspace folders, capabilities and so on)
2. Another that handles only the Ruby LSP's specific settings (what gets passed as `initializationOptions`)

Beyond just improving the design, invoking this twice may have weird consequences, like accidentally changing the negotiated encoding, so I don't think we can reuse it as is.

For example, if the editor and server negotiated UTF-8 initially, invoking this method without passing the `capabilities -> general -> positionEncodings` data will result in the server choosing UTF-16 (respecting the spec's default encoding), which would mean that editor and server would be trying to communicate using different encodings - leading to crashes, documents being in an invalid state and so on.

vinistock (changes requested):

lib/ruby_lsp/server.rb
katafrakt: The way I understand the spec is that `workspace/didChangeConfiguration` can include the payload, but it's soft-deprecated behaviour, and I chose not to implement it.

> The workspace/configuration request is sent from the server to the client to fetch configuration settings from the client.  [...] This pull model replaces the old push model were the client signaled configuration change via an event.

https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#workspace_configuration

I think it's a strategic decision for the maintaining team to choose which approach to support and I will be happy to adjust this PR to whatever you decide (push, pull or both).

lib/ruby_lsp/base_server.rb
katafrakt: This implementation relies on the response having `method` inside the payload. In this case it does not, which is why I needed to match it to the request by it's ID to know what method it is. Here are relevant pieces from my session:

```
[jsonrpc] e[17:49:30.206] <-- workspace/configuration[2] {"id":2,"method":"workspace/configuration","params":{"items":[{"section":"rubyLsp"}]},"jsonrpc":"2.0"}
[jsonrpc] e[17:49:30.207] --> workspace/configuration[2] {"jsonrpc":"2.0","id":2,"result":[{"formatter":"standard","linters":["standard"],"enabledFeatures":{"codeActions":true,"diagnostics":true,"formatting":true}}]}
```

Perhaps there's a better way to do the reply-to-request matching I did not find?

lib/ruby_lsp/server.rb
```diff
@@ -273,6 +298,9 @@ def run_initialize(message)
           rename_provider: rename_provider,
           references_provider: !@global_state.has_type_checker,
           document_range_formatting_provider: true,
+          workspace: {
+            configuration: true,
```
vinistock: I don't see this in the [server capabilities](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#serverCapabilities). Is this referring to the client capability instead?

If so, we need to remember whether the client supports the `worskpace/configuration` request in `GlobalState`. We can't send the request if the client doesn't inform us that it's supported.

lib/ruby_lsp/server.rb
vinistock: Got it. That makes sense, but then let's only implement the pull model rather than pulling inside of the old event.

We'll probably need to fire this request upon receiving the `initialize` notification, so that we can read any other configurations.

Unfortunately, it's a bit wasteful for editors that include workspace settings as part of the initial boot. I wonder if there's any information we can use to send this request selectively.

lib/ruby_lsp/base_server.rb
vinistock: I see. In this case, I think we may need a new method for sending a request that eagerly registers it in the base server as something that's waiting for a response.

And maybe we can use a block to attach behaviour to the response ahead of time. Something like this:

```ruby
send_request(...) do |response|
  # do something with the response
end

# in the base server

@response_handlers = {} #: Hash[Integer | String, { untyped } -> void]

def send_request(..., &block)
  send_message(Request.new(...))

  @response_handlers[id, block]
end
```

This way, it's easier to see what is expected to happen when we receive a response back. As an example, if we had to send multiple requests for `workspace/configuration`, it would be hard to know what needs to happen on each response without assigning the behaviour upfront.

lib/ruby_lsp/server.rb
katafrakt: The reason why I implemented it in response to `didChangeConfiguration` is that Node's LS testbed [seems to be doing that](https://github.com/microsoft/vscode-languageserver-node/blob/cd32d3e505e8ec8df660bc8e328d5f650673c2d8/testbed/server/src/server.ts#L240-L247) - however I don't see it in actually real server implementation, so I'm not sure what's going on with this.

You are right about this being unnecessary for majority of LSP connections that do not want to dynamically reconfigure anything. Maybe it's not worth implementing after all?

lib/ruby_lsp/server.rb
katafrakt: This is indeed client capability. I got confused. Thanks for catching that. I fixed it in d0ddbf39fb8e9b075190aeed9717af7cb1ea6a8f.

katafrakt:
Following https://github.com/Shopify/ruby-lsp/pull/3813#discussion_r2519779114 I'm going to close this. It does not seem to provide value on its own and from the protocol description it's not clear how this should be handled.

PR #3818, opened 2025-11-10
Enable "toggle block style" refactor when there's no selection

rolandcrosby-columntax:
### Motivation

The "toggle block style" refactor action currently only works if there is a selection in the editor which covers a call node and its associated block. When there's no selection but the cursor is inside a block, it would be nice to be able to toggle the style of the block that the cursor is inside.

### Implementation

* In the `CodeActions` request: add a quick check using `@document.locate_node` to conditionally offer the "toggle block style" action if the cursor is inside a `BlockNode`
* In the `CodeActionResolve` request: add a similar call to find the current block node when there's no selection. The actual rewrite also needs the `CallNode` associated with the block, and I couldn't figure out a good way to do this without adding a second `RubyDocument.locate` call. (When you call `locate_node` to get the BlockNode, you don't get any reference to the `CallNode` that the `BlockNode` is attached to. You also can't just look for a `CallNode` covering the cursor - that could return a `CallNode` that's actually inside the block that you want to toggle.)

### Automated Tests

Added some expectation tests:
- verify that the code action is offered when the cursor is inside a block and isn't offered when the cursor is outside a block
- verify that the rewrite action toggles the expected block(s) depending on the cursor position

### Manual Tests

* Place your cursor outside a block and trigger code actions => none should be available. 
* Place your cursor inside a block and trigger code actions => "Refactor: Toggle block style" should be available, and should toggle the style of the block that the cursor is inside (and any nested blocks)

### DIFF
--- a/test/expectations/code_action_resolve/aref_call_aref_assign_without_selection.exp.json
+++ b/test/expectations/code_action_resolve/aref_call_aref_assign_without_selection.exp.json
@@ -1,0 +1,47 @@
+{
+    "params": {
+        "kind": "refactor.rewrite",
+        "title": "Refactor: Toggle block style",
+        "data": {
+            "range": {
+                "start": {
+                    "line": 0,
+                    "character": 29
+                },
+                "end": {
+                    "line": 0,
+                    "character": 29
+                }
+            },
+            "uri": "file:///fake"
+        }
+    },
+    "result": {
+        "title": "Refactor: Toggle block style",
+        "edit": {
+            "documentChanges": [
+                {
+                    "textDocument": {
+                        "uri": "file:///fake",
+                        "version": null
+                    },
+                    "edits": [
+                        {
+                            "range": {
+                                "start": {
+                                    "line": 0,
+                                    "character": 26
+                                },
+                                "end": {
+                                    "line": 0,
+                                    "character": 58
+                                }
+                            },
+                            "newText": "do |a|\n  a[\"field\"] == \"expected\"\nend"
+                        }
+                    ]
+                }
+            ]
+        }
+    }
+}
--- a/test/expectations/code_action_resolve/nested_block_calls_without_selection.exp.json
+++ b/test/expectations/code_action_resolve/nested_block_calls_without_selection.exp.json
@@ -1,0 +1,47 @@
+{
+    "params": {
+        "kind": "refactor.rewrite",
+        "title": "Refactor: Toggle block style",
+        "data": {
+            "range": {
+                "start": {
+                    "line": 1,
+                    "character": 30
+                },
+                "end": {
+                    "line": 1,
+                    "character": 30
+                }
+            },
+            "uri": "file:///fake"
+        }
+    },
+    "result": {
+        "title": "Refactor: Toggle block style",
+        "edit": {
+            "documentChanges": [
+                {
+                    "textDocument": {
+                        "uri": "file:///fake",
+                        "version": null
+                    },
+                    "edits": [
+                        {
+                            "range": {
+                                "start": {
+                                    "line": 0,
+                                    "character": 29
+                                },
+                                "end": {
+                                    "line": 3,
+                                    "character": 3
+                                }
+                            },
+                            "newText": "{ |a| nested_call(fourth_call).each { |b| } }"
+                        }
+                    ]
+                }
+            ]
+        }
+    }
+}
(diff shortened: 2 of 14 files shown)


rolandcrosby-columntax:
I have signed the CLA!

lib/ruby_lsp/requests/code_actions.rb
```diff
@@ -113,6 +109,25 @@ def attribute_actions
           ),
         ]
       end
+
+      #: -> Array[Interface::CodeAction]
+      def toggle_block_style_action
+        if @range[:start] == @range[:end]
+          block_context = @document.locate_node(
```
vinistock, line 116: Locating nodes is expensive because it needs to traverse the AST. Let's start locating the node under the cursor with no `node_types` filtering and then we can pass the node to this method and `attribute_actions`.

That way, both can check what type of node is under the cursor without traversing twice.

vinistock (commented):
Thanks for the PR, this is great. Let's just ensure we're not traversing the AST multiple times for discovering code actions and we should be good to ship

lib/ruby_lsp/requests/code_actions.rb
rolandcrosby-columntax: Hmm, that could be slightly complicated, because the two calls to `locate` have different `node_types` filters. `attribute_actions` is trying to find `INSTANCE_VARIABLE_NODES` covering the cursor, while `toggle_block_style_action` is looking for `BlockNode`s. Additionally, in `attribute_actions`, there are two different cases where we call `locate` with the range start position: (a) the case where there's no selection, and (b) the case where there's nothing fully contained within the selection that matches `INSTANCE_VARIABLE_NODES`.

Since `locate` happens to build an array of `ClassNode`, `DefNode`, `BlockNode`, etc. in the `NodeContext`'s `@nesting_nodes`, we could potentially reuse the `attribute_actions` locate result for `toggle_block_style_action` as follows:
- Add `attr_reader :nesting_nodes` to `NodeContext`
- In `CodeActions#perform`, if there's no selection, call `@document.locate_node(@range[:start], node_types: INSTANCE_VARIABLE_NODES)` (the filter that we need in `attribute_actions`)
- Pass the returned `NodeContext` to `attribute_actions` and `toggle_block_style_action` as a nilable param
- In `toggle_block_style_action`, if that param is provided, only return the toggle action if the context's `nesting_nodes` contains a `BlockNode`
- In the `if node.nil?` branch of `attribute_actions`, use the NodeContext passed by the caller if it's present; if not, we're in case (b) above and need to invoke `locate`

This optimization would only work because the conditional actions only care about these two kinds of nodes; if additional conditional code actions were added which needed to match other node types, we'd be back where we started and would need multiple `locate` calls. What do you think we should do?

vinistock: We should be able to just remove the `node_types` filter and return early manually in each method. For example:

```ruby
node_context = @document.locate_node(@range[:start])

code_actions.concat(toggle_block_style_action(node))
code_actions.concat(attribute_actions(node))

# ...

def toggle_block_style_action(context)
  return [] unless context.node.is_a?(Prism::BlockNode)

  # ...
end

def attribute_actions(node)
   return [] unless INSTANCE_VARIABLE_NODES.include?(node_context.node.class)
  # ...
end
```

Would something like this work?

rolandcrosby-columntax: @vinistock No, that wouldn't work. `locate_node` returns the innermost node at the cursor that matches one of the provided `node_types` -- or if `node_types` is unset, it returns whatever the innermost node is. Your proposed change would break the existing behavior of `attribute_actions`, which needs to check if there is a node at the cursor that is one of the `INSTANCE_VARIABLE_NODES`, regardless of whether that node has children of other types.

Concretely: In the existing implementation, if you had the expression `@foo = "something"` and the cursor were positioned in the string, the `locate_node` call in `attribute_actions` would return an `InstanceVariableWriteNode`. But if we removed the filter, `locate_node` would just return the `StringNode` containing the cursor, so we wouldn't detect that we should offer any attribute actions. Similarly, if we didn't filter to `BlockNode` in `toggle_block_style_action`, we'd always just find the innermost node of any kind, rather than any block containing the cursor.

vinistock: Yes, you're right. I think it might be worth trying to design a special way of locating nodes that remember all parents just for code actions in the future.

That way, we can traverse once and match all applicable actions with the full picture.

vinistock:
Huh, I'm not sure why CI got stuck. @rolandcrosby-columntax could you please rebase to kick off CI again?

rolandcrosby-columntax:
@vinistock Done. Thanks for reviewing, and let me know if you need anything else to get this merged!

PR #3840, opened 2025-11-25
Upgrade Bundler to v4.0.0.beta2

### LINKED ISSUE #3830: Test the LSP with Bundler 4
The pre-release for Bundler 4 is already out and we should give it a try for the LSP to identify any issues with our integrations.

We use a pretty considerable amount of APIs from Bundler, so there's a chance something might break with the upgrade.

vinistock:
### Motivation

Closes #3830

Upgrade to Bundler v4.0.0.beta2. We have direct integrations with Bundler APIs, so the sooner we start testing, the sooner we can identify any issues.

### Implementation

The upgrade was actually fairly smooth. I left comments on the changes about why they needed to happen.


vinistock:
* **#3840** <a href="https://app.graphite.com/github/pr/Shopify/ruby-lsp/3840?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/Shopify/ruby-lsp/3840?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`
<hr />

You must have a Graphite account in order to use the merge queue. Sign up using [this link](https://app.graphite.com/invite/github/Shopify?ref=merge-queue-instructions-in-stack-comment&prId=3840).

<h2></h2>
<sub>

</sub>

<sub>

</sub>

<h2></h2>

Gemfile.lock
```diff
@@ -131,12 +131,11 @@ GEM
 PLATFORMS
   arm64-darwin
   x64-mingw-ucrt
-  x64-mingw32
```
vinistock, line 134: This platform is getting deprecated in favour of the one above https://github.com/ruby/rubygems/pull/9023

lib/ruby_lsp/setup_bundler.rb
```diff
@@ -282,12 +282,21 @@ def run_bundle_install_directly(env, force_install: false)
 
       return update(env) if @needs_update_path.exist?
 
-      # The ENV can only be merged after checking if an update is required because we depend on the original value of
-      # ENV["BUNDLE_GEMFILE"], which gets overridden after the merge
```
vinistock, line 286: This comment was duplicated in the lines above.

lib/ruby_lsp/setup_bundler.rb
```diff
@@ -282,12 +282,21 @@ def run_bundle_install_directly(env, force_install: false)
...
       FileUtils.touch(@needs_update_path) if needs_update
 
       $stderr.puts("Ruby LSP> Checking if the composed bundle is satisfied...")
-      missing_gems = bundle_check
+
+      begin
+        missing_gems = bundle_check
+      rescue Errno::EPIPE, Bundler::HTTPError
+        # These are errors cases where we cannot recover
+        raise
```
vinistock, line 293: We were not re-raising here and instead trying to bundle install again if something failed during bundle check. I refactored the code to scope the rescue to only bundle check and made sure we're re-raising in the cases where we cannot try to recover.

lib/ruby_lsp/setup_bundler.rb
```diff
@@ -489,11 +494,16 @@ def rails_app?
     end
 
     #: -> void
-    def patch_thor_to_print_progress_to_stderr!
-      return unless defined?(Bundler::Thor::Shell::Basic)
+    def force_output_to_stderr!
+      # Bundler and RubyGems have different UI objects used for printing. We need to ensure that both are configured to
+      # print only to stderr or else they'll break the connection with the editor
+      Gem::DefaultUserInteraction.ui = Gem::StreamUI.new($stdin, $stderr, $stderr, true)
```
vinistock: We were not redirecting rubygems output to stderr, only Bundler output. With version 4, our tests to verify that we don't print to stdout began failing.

We can't print to stdout as that breaks the client/server communication, so here I override the default rubygems UI with both output IOs as `stderr`​.

lib/ruby_lsp/setup_bundler.rb
```diff
@@ -489,11 +494,16 @@ def rails_app?
...
+
+      ui = Bundler.ui
+      ui.output_stream = :stderr if ui.respond_to?(:output_stream=)
```
vinistock, line 504: There's this new `output_stream`​ concept in Bundler UI where we can select where to print to. This is not available on every version of Bundler, so we need to check if it's supported.

test/integration_test.rb
```diff
@@ -244,7 +244,7 @@ def test_launch_mode_with_full_bundle
   def test_launch_mode_with_no_gemfile_and_bundle_path
     in_temp_dir do |dir|
       Bundler.with_unbundled_env do
-        system("bundle config --local path #{File.join("vendor", "bundle")}")
+        system("bundle", "config", "set", "--local", "path", File.join("vendor", "bundle"))
```
vinistock, line 249: Invoking `bundle config`​ without `set`​, `get`​ or `list`​ is now a deprecation.

test/setup_bundler_test.rb
```diff
@@ -760,7 +760,7 @@ def test_progress_is_printed_to_stderr
             compose.setup!
           end
 
-          assert_match(/Bundle complete! [\d]+ Gemfile dependencies, [\d]+ gems now installed/, stderr)
+          assert_match(/Installing prism [.\d]+ with native extensions/, stderr)
```
vinistock: The output changed a bit, so I adapted the test.

test/setup_bundler_test.rb
```diff
@@ -869,29 +869,6 @@ def test_only_returns_environment_if_bundle_was_composed_ahead_of_time
     end
   end
 
-  def test_ignores_bundle_bin
```
vinistock, line 872: Based on what I was able to gather, setting `BUNDLE_BIN`​ no longer automatically creates binstubs when running `bundle install`​. You need to run `bundle binstubs --all`​ to achieve that same behaviour.

I had trouble finding a related pull request, but after investigating the code, this is what I think happened.

lib/ruby_lsp/setup_bundler.rb
st0012: Imagine the chaos we may have on API branching when bundler 5 is out 😄

test/setup_bundler_test.rb
```diff
@@ -190,10 +196,8 @@ def test_creates_composed_bundle_if_no_gemfile
     # Create a temporary directory with no Gemfile or Gemfile.lock
     Dir.mktmpdir do |dir|
       Dir.chdir(dir) do
-        bundle_gemfile = Pathname.new(".ruby-lsp").expand_path(dir) + "Gemfile"
-
         Bundler.with_unbundled_env do
-          stub_bundle_with_env(bundle_env(dir, bundle_gemfile.to_s))
+          stub_bundle_with_env(bundle_env(dir, ".ruby-lsp/Gemfile"))
```
alexcrocha: I am sensing a deeper reason for this change but I don't see what it is. Could you explain?

vinistock: With Bundler 4, some of the weaknesses of the way we are testing our composed bundle logic have surfaced. As I'm trying to ensure that the tests pass both locally and on CI, I'm trying to make some improvements along the way.

There's actually no deeper meaning to this change other than the API was non sensical. We had a private method that accepted both relative and absolute paths and had extra logic to handle both cases. This is completely unnecessary since we're the ones invoking the method anyway. We can just invoke it using one style every time and then the extra logic can be removed.

PR #3844, opened 2025-11-28
Exclude magic comments from documentation

### LINKED ISSUE #3775: Ignore "frozen_string_literal" pragma comments when indexing code documentation

### Use case

I want to see documentation for a Ruby class, but currently the Ruby LSP sometimes aggregates pragma comments along with everything else.

### Description

I have a class like this in my app:

```rb
# frozen_string_literal: true
class Foo < Bar
  def whatever
    puts 'qux'
  end
end
```

If I hover over a reference to this class elsewhere in the codebase, it will show "frozen_string_literal: true" as the documentation in the hover.

Image

In some cases, this can lead to it showing up repeatedly for a given class, like here for the pg_search gem:

Image

### Implementation

If the only comment that is found for a class is `# frozen_string_literal: true`, the comment should be discarded.

There are cases where a comment may intentionally have this line/string in it, and so it should probably only discard the comment if that's the _entire_ comment. I guess you could also discard it if it's the first line in a comment, for example if my code was like this:

```rb
# frozen_string_literal: true
# This class is used for blah blah blah blah.
class Foo < Bar
  def whatever
    puts 'qux'
  end
end
```

I realize you generally should have an empty line after the pragma, before any code/real comments (and many codebases have rubocop enforce that anyway), but I can't control it for third party gems, and so I still think this is worth considering to fix.

thomasmarshall:
### Motivation

Closes #3775.

This PR fixes an issue whereby magic comments were not excluded when lazily loaded.

### Implementation

I copied the implementation for eagerly loaded comments from [here](https://github.com/Shopify/ruby-lsp/blob/9e7450dfe690a399aa25dff4327a61090e71c826/lib/ruby_indexer/lib/ruby_indexer/declaration_listener.rb#L738).

In order to do this, I needed the entries to have access to the `magic_comment_regex` configuration. ~I opted to pass the index through to the entries, but I'm not sure if that's the best approach—is it part of the public API that plugins use?~ I'm passing just the configuration through rather than the index.

### Automated Tests

I have added a test and confirmed it failed before implementing the fix.

### Manual Tests

1. Add a magic comment above a constant
2. Hover over a reference to that constant to see if the magic comment appears as documentation

| Before | After | Real comments still appear |
| - | - | - |
| Screenshot 2025-11-28 at 16 48 06 | Screenshot 2025-11-28 at 16 48 33 | Screenshot 2025-11-28 at 16 49 07 |

### DIFF
--- a/lib/ruby_indexer/lib/ruby_indexer/entry.rb
+++ b/lib/ruby_indexer/lib/ruby_indexer/entry.rb
@@ -86,7 +86,7 @@ module RubyIndexer
           correct_group.filter_map do |comment|
             content = comment.slice.chomp
 
-            if content.valid_encoding?
+            if content.valid_encoding? && !content.match?(@configuration.magic_comment_regex)
               content.delete_prefix!("#")
               content.delete_prefix!(" ")
               content
--- a/lib/ruby_indexer/test/classes_and_modules_test.rb
+++ b/lib/ruby_indexer/test/classes_and_modules_test.rb
@@ -766,5 +766,25 @@ module RubyIndexer
         FileUtils.rm(path)
       end
     end
+
+    def test_lazy_comments_ignores_magic_comments
+      path = File.join(Dir.pwd, "lib", "foo.rb")
+      source = <<~RUBY
+        # frozen_string_literal: true
+
+        class Foo
+        end
+      RUBY
+      File.write(path, source)
+      @index.index_single(URI::Generic.from_path(path: path), source, collect_comments: false)
+
+      entry = @index["Foo"]&.first #: as !nil
+
+      begin
+        assert_empty(entry.comments)
+      ensure
+        FileUtils.rm(path)
+      end
+    end
   end
 end


lib/ruby_indexer/lib/ruby_indexer/entry.rb
```diff
@@ -3,6 +3,9 @@
 
 module RubyIndexer
   class Entry
+    #: Configuration
+    attr_reader :configuration
```
vinistock: I think we're not using this attr_reader. If that's the case, let's remove it.

vinistock (approved):
Nice

lib/ruby_indexer/lib/ruby_indexer/entry.rb
thomasmarshall: I tried to remove this, but then I realized it's used when constructing a `ConstantAlias` from an `UnresolvedConstantAlias` here: https://github.com/Shopify/ruby-lsp/blob/68ae7f2079457f0a4951ef973fa1209f858e2fa6/lib/ruby_indexer/lib/ruby_indexer/entry.rb#L385-L390

vinistock: Ah, right. Yeah, in that case, ignore my comment!

PR #3851, opened 2025-12-04
Fix discovery of test methods in specs

### LINKED ISSUE #3722: Tests defined as test_ methods in spec-style describe blocks are not discovered
Test methods (`test_` prefix) in a describe block are not discovered:

```ruby
describe MyGem do
    def test_foo
    end
end
```

thomasmarshall:
### Motivation

Closes #3722.

Methods starting `test_` are picked up my Minitest but were not being discovered by the LSP.

```rb
describe "MySpec" do
  def test_foo
    assert true
  end
end
```

### Implementation

This commit adds a `#on_def_node_enter` handler to register test items for methods starting `test_` in describe blocks. It should not match non-test methods (e.g. helper methods), nor should it match methods named `test_` in non-spec blocks.

### Automated Tests

I have added a new testfor this fix.

### Manual Tests

1. Add a spec with a `test_` method
2. See that it is discovered by the LSP

| Before | After |
| - | -  |
| Screenshot 2025-12-04 at 13 03 13 | Screenshot 2025-12-04 at 13 04 10 |

### DIFF
--- a/test/requests/discover_tests_test.rb
+++ b/test/requests/discover_tests_test.rb
@@ -836,10 +836,14 @@ module RubyLsp
         class NotASpec
           def test_ignored; end
         end
+
+        class NotDescribeSpec < Minitest::Spec
+          def test_not_duplicate; end
+        end
       RUBY
 
       with_minitest_spec_configured(source) do |items|
-        assert_equal(["MySpec"], items.map { |i| i[:id] })
+        assert_equal(["MySpec", "NotDescribeSpec"], items.map { |i| i[:id] })
         assert_equal(
           ["MySpec#test_foo", "MySpec::nested"],
           items.dig(0, :children).map { |i| i[:id] },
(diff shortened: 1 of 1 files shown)


vinistock:
I think this may cause duplicate test items in this case below. The reason is because both the `TestStyle` and `SpecStyle` listeners run together and so if we always handle `test_` methods, it will end up duplicating.

```ruby
class MySpec < Minitest::Spec
  def test_something; end # is this one duplicated?
end
```

Can you please check if we're doing the right thing in this case?

thomasmarshall:
Hmm in this case it's not duplicated (because we check `current_group.is_a?(DescribeGroup)`) but I found a case where it is duplicated, where we have a `describe` inside a spec class:

```rb
class MySpec < Minitest::Spec
  describe 'something' do
    def test_something; end # this one is duplicated
  end
end
```

I'll work on a fix.

PR #3859, opened 2025-12-09
Add detection reasons to auto-detection log messages

### LINKED ISSUE #3845: Log entries for auto-detected libraries could indicate how they were detected

### Use case

When trying to debug why a certain library was detected (ie. Rubocop) it would be nice to indicate how the detection happened.

### Description

When starting ruby-lsp in a project, I encountered the following log entries:

```
[2025-11-28 20:29:40] [LOG] Auto detected formatter: rubocop_internal
[2025-11-28 20:29:40] [LOG] Auto detected linters: rubocop_internal
[2025-11-28 20:29:40] [LOG] Detected test library: rails
[2025-11-28 20:29:40] [LOG] Ruby LSP detected this is a Sorbet project and will defer to the Sorbet LSP for some functionality
```

The problem is that none of these are true. The project doesn't use Rubocop (albeit not directly), nor does it use Sorbet.

I actually don't think this is a problem in ruby-lsp. I use a small shim script to run my local fork correctly and I _assume_ that it's detecting some of these as the ruby-lsp project.

But it would be great to know why they were detected. ie.

```
Auto detected formatter: rubocop_internal (presence of rubocop.yml in project root)
```

### Implementation

_No response_

adam12:
### Motivation

Closes #3845

RubyLSP will helpfully log the formatter and linter detected, but not indicate
how or why it came to that conclusion. This can make it difficult to ascertain
whether the detection was correct.

### Implementation

The detection methods for formatters, linters, test libraries, and type checkers
were updated to return not just the detected tool, but also a string indicating
the reason for the detection. The logging statements were updated to include
this reason.

A regular class for the detection result (vs `Data`) was chosen for Ruby compatibility.

### Automated Tests

No tests were addeed specifically for this logging change.

### Manual Tests

When running the LSP, you can observe the log messages to see the detection reasons.

### DIFF
--- a/lib/ruby_lsp/global_state.rb
+++ b/lib/ruby_lsp/global_state.rb
@@ -2,6 +2,21 @@
 # frozen_string_literal: true
 
 module RubyLsp
+  # Holds the detected value and the reason for detection
+  class DetectionResult
+    #: String
+    attr_reader :value
+
+    #: String
+    attr_reader :reason
+
+    #: (String value, String reason) -> void
+    def initialize(value, reason)
+      @value = value
+      @reason = reason
+    end
+  end
+
   class GlobalState
     #: String
     attr_reader :test_library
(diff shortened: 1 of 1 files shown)


lib/ruby_lsp/global_state.rb
```diff
@@ -228,60 +253,73 @@ def supports_watching_files
...
+        DetectionResult.new("unknown", "no test library detected")
       end
     end
 
-    #: (Array[String] dependencies) -> bool
+    # Returns [detected, reason] tuple for typechecker detection
+    #: (Array[String] dependencies) -> [bool, String]
     def detect_typechecker(dependencies)
-      return false if ENV["RUBY_LSP_BYPASS_TYPECHECKER"]
+      if ENV["RUBY_LSP_BYPASS_TYPECHECKER"]
```
vinistock: For consistency with the other methods, can we also return a `DetectionResult` here?

adam12: I applied this change.

rafaelfranca (approved):

vinistock (approved):

PR #3863, opened 2025-12-11
Expose `NodeContext#nesting_nodes` attribute reader

janko:
I needed this for implementing https://github.com/Shopify/ruby-lsp-rails/pull/669, where I had to check whether a node is within an instance method of a subclass of `ActiveRecord::Migration[x.y]` (see [relevant code](https://github.com/janko/ruby-lsp-rails/blob/298dd02fa98f6d983e1e80f1fbdaa88f5fbaccea/lib/ruby_lsp/ruby_lsp_rails/completion.rb#L128-L144)).

Since the latter isn't a direct class reference, I couldn't use the indexer to check the superclass of the migration class, as that was returning a fallback `::Object`.

### DIFF
--- a/lib/ruby_lsp/node_context.rb
+++ b/lib/ruby_lsp/node_context.rb
@@ -44,11 +44,7 @@ module RubyLsp
       locals = []
 
       nesting_nodes.each do |node|
-        if node.is_a?(Prism::ClassNode) || node.is_a?(Prism::ModuleNode) || node.is_a?(Prism::SingletonClassNode) ||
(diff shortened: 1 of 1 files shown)


sucicfilip:
I would also possibly benefit from this for this PR: https://github.com/Shopify/ruby-lsp-rails/pull/671

vinistock:
Let's please hold off merging this for now. We're about to start the work to integrate the new indexer into the LSP and there will be breaking API changes for add-ons.

Exposing more APIs right now will likely just mean more breaking changes.

PR #3878, opened 2025-12-31
add support for rv version manager

### LINKED ISSUE #3861: Support `rv` as a version manager.

### Use case

I'd like to use Ruby LSP with https://rv.dev/

### Description

Support `rv`.

### Implementation

_No response_

matheuscumpian:
### Motivation
Adding support for rv as a version manager on vscode extension

closes #3861

### DIFF
--- a/vscode/src/ruby/rv.ts
+++ b/vscode/src/ruby/rv.ts
@@ -1,0 +1,46 @@
+import * as vscode from "vscode";
+
+import { VersionManager, ActivationResult } from "./versionManager";
+
+// Manage your Ruby environment with rv
+//
+// Learn more: https://github.com/spinel-coop/rv
+export class Rv extends VersionManager {
+  async activate(): Promise<ActivationResult> {
+    const rvExec = await this.findRv();
+    const parsedResult = await this.runEnvActivationScript(`${rvExec} ruby run --`);
+
+    return {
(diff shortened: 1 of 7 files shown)


vscode/src/ruby/rv.ts
```diff
@@ -0,0 +1,46 @@
...
+    const config = vscode.workspace.getConfiguration("rubyLsp");
+    const configuredRvPath = config.get<string | undefined>("rubyVersionManager.rvExecutablePath");
+
+    if (configuredRvPath) {
+      return this.ensureRvExistsAt(configuredRvPath);
+    } else {
+      const possiblePaths = [
+        vscode.Uri.file("/home/linuxbrew/.linuxbrew/bin"),
+        vscode.Uri.file("/usr/local/bin"),
+        vscode.Uri.file("/opt/homebrew/bin"),
```
rafaelfranca: Should we be using `joinPath` like we are doing in all other places we build paths?

matheuscumpian: I noticed you've created a new PR (#3894) that refactors a lot of the version managers code. Should I wait for that to merge and then align this PR with the new pattern?

rafaelfranca: Let's finish this one and then I rebase mine.

rafaelfranca (approved):

PR #3884, opened 2026-01-03
Disable stale workflow on forks

andyw8:
To avoid contributors receiving these kinds of notifications:

> The "Close stale issues & pull requests" workflow in andyw8/ruby-lsp has been disabled

### DIFF
--- a/.github/workflows/stale_check.yml
+++ b/.github/workflows/stale_check.yml
@@ -9,6 +9,7 @@ permissions:
 jobs:
   stale:
     runs-on: ubuntu-latest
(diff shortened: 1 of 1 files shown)


Morriar:
This action doesn't exist anymore.

vinistock:
It does exist. We turned it off for PRs, but it still runs on issues: https://github.com/Shopify/ruby-lsp/blob/main/.github/workflows/stale_check.yml

@andyw8 this will need a rebase because the action changed names.

andyw8:
I've rebased.

vinistock:
Thank you!

PR #3893, opened 2026-01-06
Upgrade to Ruby v4.0.0

vinistock:
### Motivation

Upgrade our development Ruby version to v4.0.0.

I had to apply some minor fixes, like changing `pathname` to `open3` in source comment tests because `pathname` is no longer returned by `Gem::Specification.default_stubs`.

I also upgraded some dev gems.


vinistock:
* **#3893** <a href="https://app.graphite.com/github/pr/Shopify/ruby-lsp/3893?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/Shopify/ruby-lsp/3893?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`
<hr />

You must have a Graphite account in order to use the merge queue. Sign up using [this link](https://app.graphite.com/invite/github/Shopify?ref=merge-queue-instructions-in-stack-comment&prId=3893).

<h2></h2>
<sub>

</sub>

<sub>

</sub>

<h2></h2>

lib/ruby_indexer/test/configuration_test.rb
```diff
@@ -233,18 +233,16 @@ def test_does_not_fail_if_there_are_missing_specs_due_to_platform_constraints
...
+              system("bundle install")
 
-            _stdout, stderr = capture_subprocess_io do
               script = [
                 "require \"ruby_lsp/internal\"",
                 "RubyIndexer::Configuration.new.indexable_uris",
               ].join(";")
 
-              system("bundle exec ruby -e '#{script}'")
+              assert(system("bundle exec ruby -e '#{script}'"))
```
vinistock, line 244: I had to change this because there's a new warning coming from `rbs` that it needs to include `tsort` as a dependency.

st0012 (approved):

.github/workflows/ci.yml
```diff
@@ -44,7 +44,7 @@ jobs:
       fail-fast: false
       matrix:
         os: [ubuntu-latest, macos-latest, windows-latest]
-        ruby: ["3.2", "3.3", "3.4"]
```
rafaelfranca, line 47: Did we increase the minimum version? If not, we should keep testing with 3.2, 3.1 and 3.0 according with https://github.com/Shopify/ruby-lsp/blob/5f912a9bda3bd1d62a3faecd6678e09252a1758c/ruby-lsp.gemspec#L26

Otherwise people will report bugs that we are not even aware and have no way to test.

vinistock: Our minimum Ruby version requirement is in a bit of an odd spot at the moment and I haven't had the time to do a thorough review of how to improve the situation yet.

### The problem

The main problem is the coordination between the extension and the server's launch mechanism. Consider this scenario:

- We bump the minimum requirement of the server to 3.2
- Afterwards, we realize there's a change we need to make to the launch coordination between server and extension. We push a breaking change to the server and extension
- Users of Ruby <= 3.2 can't install the latest server, so when the extension attempts to launch it, an old incompatible version of the launch mechanism gets triggered and breaks the integration

This happened before in the past when we ended support for Ruby 2.7. Versions of the server that supported 2.7 existed, but because of backwards incompatible changes in the launch coordination, none of the 2.7 users were able to use the extension in any capacity.

### Why the tested versions don't match the minimum requirement

Of course, it's not good practice to not test against supported versions. The problem is that gems we depend on for development and needed to bump already dropped support for the older rubies (like tapioca, rbi and spoom).

I wanted to take the time to revisit the extension and server launch coordination, so that we can design an approach where bumping the minimum required version does not get us into the situation where we can't make breaking changes. Unfortunately, I haven't gotten to it yet and so the choice was to not bump the minimum for now, but remove the older rubies from CI so that we could update our dependencies.

PR #3894, opened 2026-01-08
Refactor VersionManager detection logic

rafaelfranca:
### Motivation

Looking at https://github.com/Shopify/ruby-lsp/pull/3878 I noticed that adding a new version manager was requiring a lot of duplication of logic. While working on removing that I also noticed that even the existing manager were duplicating logic, so I took the opportunity to refactor all the managers.

### Implementation

I moved the logic to detect the version manager to their classes. I also used a hash lookup instead of the exisint `switch` statement full of duplicated logic.

### DIFF
--- a/vscode/src/ruby.ts
+++ b/vscode/src/ruby.ts
@@ -45,7 +45,7 @@ interface ManagerClass {
   detect: (workspaceFolder: vscode.WorkspaceFolder, outputChannel: WorkspaceChannel) => Promise<DetectionResult>;
 }
 
-const VERSION_MANAGERS: Record<ManagerIdentifier, ManagerClass> = {
+const VERSION_MANAGERS: Record<ManagerIdentifier, ManagerClass | undefined> = {
   [ManagerIdentifier.Shadowenv]: Shadowenv,
   [ManagerIdentifier.Asdf]: Asdf,
   [ManagerIdentifier.Chruby]: Chruby,
@@ -55,8 +55,8 @@ const VERSION_MANAGERS: Record<ManagerIdentifier, ManagerClass> = {
   [ManagerIdentifier.Mise]: Mise,
   [ManagerIdentifier.RubyInstaller]: RubyInstaller,
   [ManagerIdentifier.Custom]: Custom,
-  [ManagerIdentifier.Auto]: None, // Auto is handled specially
   [ManagerIdentifier.None]: None, // None is last as the fallback
+  [ManagerIdentifier.Auto]: undefined, // Auto is handled specially
 };
 
 export class Ruby implements RubyInterface {
@@ -294,6 +294,15 @@ export class Ruby implements RubyInterface {
 
   private async runManagerActivation() {
     const ManagerClass = VERSION_MANAGERS[this.versionManager.identifier];
+
+    if (!ManagerClass) {
+      throw new Error(
+        `BUG: No ManagerClass found for identifier '${this.versionManager.identifier}'. ` +
+          `This indicates either: (1) discoverVersionManager() failed to detect any version manager and left identifier as 'auto', ` +
+          `or (2) a new ManagerIdentifier enum value was added without updating VERSION_MANAGERS.`,
+      );
+    }
+
     const manager = new ManagerClass(
       this.workspaceFolder,
       this.outputChannel,


vscode/src/test/suite/ruby/asdf.test.ts
```diff
@@ -136,32 +136,4 @@ suite("Asdf", () => {
     assert.strictEqual(yjit, true);
     assert.strictEqual(env.ANY, "true");
   });
-
-  test("Uses ASDF executable in PATH if script and Homebrew executable are not available", async () => {
```
vinistock, line 140: Why was this test removed? We still want to try to invoke `asdf` without an absolute path as a last resort.

This works for non-standard configurations where the user launches the editor through the terminal.

vscode/src/ruby.ts
```diff
@@ -35,6 +35,31 @@ export interface ManagerConfiguration {
...
+    context: vscode.ExtensionContext,
+    manuallySelectRuby: () => Promise<void>,
+    ...args: any[]
+  ): VersionManager;
+  detect?: () => Promise<vscode.Uri | undefined>;
+}
+
+const VERSION_MANAGERS: Record<ManagerIdentifier, ManagerClass> = {
+  [ManagerIdentifier.Asdf]: Asdf,
+  [ManagerIdentifier.Auto]: None, // Auto is handled specially
```
vinistock: How about we remove this and treat the `null` as a special case explicitly?

Otherwise, it's easy to accidentally access the hash for `Auto` and invoke methods on `None`, which may silently have the wrong behaviour.

vscode/src/ruby/chruby.ts
```diff
@@ -17,6 +17,11 @@ class RubyActivationCancellationError extends Error {}
 // A tool to change the current Ruby version
 // Learn more: https://github.com/postmodern/chruby
 export class Chruby extends VersionManager {
+  static async detect(workspaceFolder: vscode.WorkspaceFolder): Promise<vscode.Uri | undefined> {
+    const exists = await VersionManager.toolExists("chruby", workspaceFolder);
+    return exists ? vscode.Uri.file("chruby") : undefined;
```
vinistock: Is the idea to invoke `chruby` here, but then we need to build a fake URI to conform to the method signature?

vscode/src/ruby/versionManager.ts
```diff
@@ -19,6 +19,8 @@ export const VALUE_SEPARATOR = "RUBY_LSP_VS";
 export const FIELD_SEPARATOR = "RUBY_LSP_FS";
 
 export abstract class VersionManager {
+  private static readonly shell = process.env.SHELL?.replace(/(\s+)/g, "\\$1");
```
vinistock: This is actually not correct and something I've been meaning to fix. This only works when the editor is opened from the terminal.

We need to use `vscode.env.shell` to get the user's configured default shell.

vscode/src/ruby/versionManager.ts
```diff
@@ -150,4 +150,37 @@ export abstract class VersionManager {
 
     return execName;
   }
+
+  // Common helper to find a version manager executable with configuration support
+  protected async findVersionManagerUri(
+    managerName: string,
```
vinistock: What do you think of manually setting the name for all managers? This doesn't change, so that way we can avoid this parameter.

E.g.:

```typescript
class VersionManager {
  abstract protected name: string;
}

class Rbenv extends VersionManager {
  protected name = "Rbenv"
}
```

vscode/src/ruby/rv.ts
```diff
@@ -1,14 +1,13 @@
...
 import { pathToUri } from "../common";
-import { WorkspaceChannel } from "../workspaceChannel";
+import { Mise } from "./mise";
 
 // Manage your Ruby environment with rv
 //
 // Learn more: https://github.com/spinel-coop/rv
-export class Rv extends VersionManager {
-  private static getPossiblePaths(): vscode.Uri[] {
+export class Rv extends Mise {
```
vinistock, line 9: This one I find a bit weird. I would instead extract another common parent rather than having two different version managers inherit from each other.

vinistock:
Since this is a considerable refactor, we should cut a preview release of the extension once we converge on the implementation to catch any issues before the stable release.

vscode/src/ruby/chruby.ts
rafaelfranca: Yes. I refactored this in another branch, so it will not needed anymore to return fake Uris

vscode/src/ruby/rv.ts
rafaelfranca: I agree, but we have this same pattern with the Chruby and RubyInstaller, so I just simplified the hierarch chain of this one

vscode/src/ruby/chruby.ts
rafaelfranca: I applied the refactoring to this branch

vscode/src/ruby/versionManager.ts
rafaelfranca: I fixed this in 9d80320255e5bc4e3cecd1875b4fe6e285a30113, but not sure how I feel about changing behavior in a refactoring change. Do you want me to extract this commit to a different PR?

vscode/src/test/suite/ruby/asdf.test.ts
rafaelfranca: Added it back

vscode/src/ruby/versionManager.ts
rafaelfranca: Extracted in https://github.com/Shopify/ruby-lsp/pull/3917

rafaelfranca:
* **#3918** <a href="https://app.graphite.com/github/pr/Shopify/ruby-lsp/3918?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#3894** <a href="https://app.graphite.com/github/pr/Shopify/ruby-lsp/3894?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/Shopify/ruby-lsp/3894?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#3917** <a href="https://app.graphite.com/github/pr/Shopify/ruby-lsp/3917?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`
<hr />

You must have a Graphite account in order to use the merge queue. Sign up using [this link](https://app.graphite.com/invite/github/Shopify?ref=merge-queue-instructions-in-stack-comment&prId=3894).

<h2></h2>
<sub>

</sub>

<sub>

</sub>

<h2></h2>

PR #3902, opened 2026-01-15
Fix syntax highlighting for line-ending forward slashes

### LINKED ISSUE #1786: Syntax highlighting fails when `/` is at the end of the line
### Operating System

MacOS

### Ruby version

2.7.2

### Project has a bundle

- [ ] Has bundle

no manager

### Description

For a file like:

```ruby
my_number = 4 /
  4
my_other_number = 4 /
  1
puts(my_number) # correctly outputs 1
puts(my_other_number) # correctly outputs 4
```

In VSCode, with the extension enabled, I see:
image

Note that it's highlighting from the end of line 1 to the end of line 3, with a color that looks like the one used for regexes:

image

reese:
### Motivation

Closes #1786

This fixes an issue with syntax highlighting where slashes at the end of lines, such as for division across lines, strings with slashes at the end of lines, etc. were being read as regex contents. This would cause many lines of the source to be rendered in the same color, essentially removing any practical syntax highlighting. This would continue until another backslash was seen.

### Implementation

This change makes two changes to the regex grammar. The first (the addition of `(?!\\s*$)`) is a negative lookahead to reject `/` when followed only by whitespace until end of line. The second change removed the `| $` alternative that incorrectly allowed matching at end of line without a closing `/`.

### Automated Tests

See the added automated tests.

### Manual Tests

I opened up a VSCode window with these changes.

Before ❌ 

Screenshot 2026-01-15 at 1 07 19 PM

After ✅ 

Screenshot 2026-01-15 at 1 07 57 PM

### DIFF
--- a/vscode/grammars/ruby.cson.json
+++ b/vscode/grammars/ruby.cson.json
@@ -540,7 +540,7 @@
       ]
     },
     {
-      "begin": "(?x)\n(?<![\\w)])((/))(?![?*+])\n(?=\n  (?:\\\\/|[^/])*+          # Do NOT change the order\n  /[eimnosux]*\\s*\n  (?:\n    [)\\]}#.,?:]|\\|\\||&&|<=>|=>|==|=~|!~|!=|;|$|\n    if|else|elsif|then|do|end|unless|while|until|or|and\n  )\n  |\n  $\n)",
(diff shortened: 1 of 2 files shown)


reese:
I have signed the CLA!

PR #3913, opened 2026-01-21
Fix infinite loop when collecting transitive excluded gems

### LINKED ISSUE #3912: 100% CPU usage, circular dependency on indexer
https://github.com/Shopify/ruby-lsp/blob/7eb2eb7830636259335dd0c86f89834fbdcba430/lib/ruby_indexer/lib/ruby_indexer/configuration.rb#L230-L239

The variable `excluded` is iterated and modified at the same time, producing a circular loop.

This can be fixed by renaming the original `excluded` to something different.

rafaelfranca:
The code was modifying the `excluded` array while iterating over it with `.each`. In Ruby, `.each` continues to process newly added elements, which can cause excessive iterations or infinite loops when circular dependencies are present in the gem graph.

This caused 100% CPU usage in some environments.

Fixes #3912

### DIFF
--- a/lib/ruby_indexer/lib/ruby_indexer/configuration.rb
+++ b/lib/ruby_indexer/lib/ruby_indexer/configuration.rb
@@ -227,7 +227,7 @@ module RubyIndexer
       others.uniq!
       others.map!(&:name)
 
-      excluded.each do |dependency|
+      transitive_excluded = excluded.each_with_object([]) do |dependency, acc|
         next unless dependency.runtime?
 
         spec = dependency.to_spec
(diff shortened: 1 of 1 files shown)


lib/ruby_indexer/lib/ruby_indexer/configuration.rb
```diff
@@ -227,7 +227,10 @@ def initial_excluded_gems
       others.uniq!
       others.map!(&:name)
 
-      excluded.each do |dependency|
+      # Collect transitive dependencies in a separate array to avoid modifying `excluded` while iterating over it.
+      # Modifying an array during `.each` iteration causes Ruby to process newly added elements, which can lead to
+      # excessive iterations or infinite loops with circular dependencies. See: https://github.com/Shopify/ruby-lsp/issues/3912
```
vinistock: Nitpick, but I'd remove this comment block.

sequielo:
That was fast @rafaelfranca @vinistock :rocket: 

Thank you very much gentlemen, greetings from Argentina.

Will confirm it's working in the next release

PR #3928, opened 2026-01-27
[DOC] Add security documentation

st0012:
### Motivation

Ruby LSP, by design, can execute code (e.g. run `bundle install`) when the user simply opening the project in VS Code under trusted mode. This is necessary for usability and it's a common practice among language servers and their extensions. But the risk should be clearly communicated with users.

### Implementation

Add a page to document known (but not exhaustive) code executions that could be performed by Ruby LSP so users can be aware of the potential risks.

### DIFF
--- a/jekyll/security.markdown
+++ b/jekyll/security.markdown
@@ -10,7 +10,7 @@ This page documents potential risks when using the Ruby LSP VS Code extension an
 
 ## Trust Model
 
-**Ruby LSP assumes that all code in your workspace is trusted.**
+**Ruby LSP assumes that all code in your workspace (including dependencies) is trusted.**
 
 When you open a project with Ruby LSP, the extension and language server will execute code from that project as part of
 normal operation. This is fundamentally similar to running `bundle install` in that project directory.


jekyll/security.markdown
```diff
@@ -0,0 +1,51 @@
...
+## Recommendations
+
+1. **Only open trusted projects** - Treat opening a project in VS Code with Ruby LSP the same as running `bundle install`
+   in that directory
+2. **Be cautious with unfamiliar add-ons** - Add-ons have full access to your system when activated
+
+## Reporting Security Issues
+
+If you discover a security vulnerability in Ruby LSP, please report it through
+[GitHub Security Advisories](https://github.com/Shopify/ruby-lsp/security/advisories/new) rather than opening a public
```
rafaelfranca, line 53: I think only admins to the repo can do that, no?

jekyll/security.markdown
```diff
@@ -0,0 +1,51 @@
...
+
+If you are working with code you do not fully trust, you should be aware of the potential risks documented below.
+
+## Code Execution Vectors
+
+The following is a non-exhaustive list of ways that Ruby LSP may execute code from your workspace:
+
+### Bundle Installation
+
+Ruby LSP automatically runs `bundle install` (or `bundle check`) when starting up or when detecting changes to your
```
vinistock: I think we can summarize this as "bundler operations" given that we also run `bundle update`.

jekyll/security.markdown
```diff
@@ -0,0 +1,51 @@
...
+Gemfile. This will:
+
+- Execute any code in your Gemfile (Gemfiles are Ruby code)
+- Install gems specified in the Gemfile, which may include native extensions that execute during installation
+- Run any post-install hooks defined by gems
```
vinistock, line 31: Does RubyGems allow for post-install hooks? I thought you could only show messages.

jekyll/security.markdown
```diff
@@ -0,0 +1,51 @@
...
+
+Ruby LSP has an add-on system that automatically discovers and loads add-ons from:
+
+- Gems in your bundle that contain `ruby_lsp/**/addon.rb` files
+- Files matching `ruby_lsp/**/addon.rb` anywhere in your workspace
+
+Add-ons are loaded via `require` and their `activate` method is called, allowing them to execute arbitrary Ruby code.
+This is by design - add-ons can spawn processes, make network requests, or perform any other operation.
+
+## Recommendations
```
vinistock, line 43: Another recommendation, or at least we should mention it somewhere in the docs, that the extension and server will not run any code if the workspace is not trusted by VS Code.

jekyll/security.markdown
st0012: It's not a direct post-install hook. But you can use `spec.rdoc_options` to specify what RDoc should run to generate your gem's documentation for `ri` (which will be invoked by rubygems), including running your rdoc options etc.

jekyll/security.markdown
st0012: Added

jekyll/security.markdown
```diff
@@ -0,0 +1,54 @@
...
+nav_order: 25
+---
+
+# Security
+
+This page documents potential risks when using the Ruby LSP VS Code extension and the Ruby LSP language server with untrusted code.
+
+## Trust Model
+
+**Ruby LSP assumes that all code in your workspace is trusted.**
```
vinistock: Maybe we should be explicit that this includes dependencies too.
```suggestion
**Ruby LSP assumes that all code in your workspace (including dependencies) is trusted.**
```

jekyll/security.markdown
vinistock: Really? I had no idea.

Issue #3937, opened 2026-02-05
ruby-lsp downgrades during auto-update


modille:
### Ruby LSP Information

- **Editor**: Neovim (via nvim-lspconfig)
- **Ruby**: 3.3.4 (via devbox/nix)
- **Bundler**: 2.5.16
- **ruby-lsp**: 0.26.5 (expected) / 0.11.2 (after downgrade)
- **ruby-lsp-rails**: 0.4.8 (expected) / 0.2.6 (after downgrade)

### Reproduction steps

1. Start with a project that does not have `rbs` in its Gemfile (I think this is key based on the workaround I found below?)
2. Remove any existing ruby-lsp setup and install the latest (0.26.5):
```bash
$ rm -rf .ruby-lsp/

$ gem uninstall --all --executables ruby-lsp ruby-lsp-rails
Successfully uninstalled ruby-lsp-rails-0.4.8
Removing ruby-lsp
Removing ruby-lsp-check
Removing ruby-lsp-launcher
Removing ruby-lsp-test-exec
Successfully uninstalled ruby-lsp-0.26.5

$ gem install ruby-lsp ruby-lsp-rails
Fetching ruby-lsp-0.26.5.gem
Successfully installed ruby-lsp-0.26.5
Parsing documentation for ruby-lsp-0.26.5
Installing ri documentation for ruby-lsp-0.26.5
Done installing documentation for ruby-lsp after 0 seconds
Fetching ruby-lsp-rails-0.4.8.gem
Successfully installed ruby-lsp-rails-0.4.8
Parsing documentation for ruby-lsp-rails-0.4.8
Installing ri documentation for ruby-lsp-rails-0.4.8
Done installing documentation for ruby-lsp-rails after 0 seconds
2 gems installed
```
3. Open editor to trigger ruby-lsp initialization
4. Verify only 0.26.5 is installed and composed bundle shows 0.26.5:
```bash
$ gem list ruby-lsp
*** LOCAL GEMS ***
ruby-lsp (0.26.5)
ruby-lsp-rails (0.4.8)

$ grep "ruby-lsp (" .ruby-lsp/Gemfile.lock
    ruby-lsp (0.26.5)
      ruby-lsp (>= 0.26.0, < 0.27.0)
```
5. Close and reopen editor to trigger the auto-update cycle. Check `:LspLog` output
6. Note "ruby-lsp version regressed from 0.26.5 to 0.11.2" in LSP log output and be sad 😢

<details>

[...]

PR #3938, opened 2026-02-05
Add rbs to composed bundle update commands

modille:
### Motivation

Closes https://github.com/Shopify/ruby-lsp/issues/3937

### Implementation

I added `rbs` to the gems array for the `bundle update` command in both the launcher and non-launcher paths.

> [!NOTE]
> Should `prism` be in both of these places? And should `--conservative` be consistent in both places? 

### Automated Tests

I updated existing tests to expect rbs in the bundle update commands:
- test_uses_custom_bundle_path_when_no_gemfile_lock_is_present
- test_update_using_bundler_internals

### Manual Tests

I suspect this fixes the issue based on the manual workaround I found and documented in  https://github.com/Shopify/ruby-lsp/issues/3937

1. Create a project with a Gemfile that does NOT include rbs as a direct dependency
2. Trigger an auto-update
4. Observe that the `bundle update` command now includes rbs and that ruby-lsp is not downgraded

### DIFF
--- a/lib/ruby_lsp/setup_bundler.rb
+++ b/lib/ruby_lsp/setup_bundler.rb
@@ -331,7 +331,7 @@ module RubyLsp
     def update(env)
       # Try to auto upgrade the gems we depend on, unless they are in the Gemfile as that would result in undesired
       # source control changes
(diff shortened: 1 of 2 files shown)


modille:
I have signed the CLA!

st0012 (approved):
Thank you!

PR #3940, opened 2026-02-05
Fix Ruby 4.0 CI failure on ubuntu-latest

st0012:
## Motivation

The `ubuntu-latest` runner image (`20260126.10+`) ships a pre-installed Ruby 4.0.1 that causes native extension builds and bundler operations to fail. The `chmod +t` workaround from https://github.com/ruby/rubygems/issues/9284 is insufficient — the pre-installed Ruby has deeper issues beyond the sticky bit.

## Implementation

Remove the pre-installed Ruby 4.0 from `/opt/hostedtoolcache` before `setup-ruby` runs, forcing it to download a working build from ruby-builder.

## Automated Tests

Existing CI matrix covers this — the Ruby 4.0 on ubuntu-latest job now passes.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -53,6 +53,14 @@ jobs:
         with:
           submodules: "recursive"
 
+      # Workaround: The ubuntu-latest runner image (20260126.10+) ships a pre-installed Ruby 4.0.1
+      # that causes native extension builds and bundler operations to fail. Removing it forces
+      # setup-ruby to download a working build from ruby-builder instead.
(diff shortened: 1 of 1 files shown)


st0012:
It's the `ubuntu-latest` images's problem. It ships with Ruby 4.0.1 but that version (the version is ships, not Ruby 4.0.1) seems to have issues that prevent native gem installation.
If we use the Ruby built by `setup-ruby` then it'd work. But since it sees that the OS already has Ruby 4.0.1 installed, it doesn't do anything. So the workaround is to remove the cached Ruby from the image to force `setup-ruby` to install 4.0.1 again.

PR #3945, opened 2026-02-07
Skip disable line action for self-resolving cops

sucicfilip:
### Motivation

When `Layout/EmptyComment` is reported, the "Disable Layout/EmptyComment for this line" code action adds a `# rubocop:disable` inline comment, which makes the comment no longer empty, resolving the original offense. This then triggers `Lint/RedundantCopDisableDirective`, so the user ends up in a loop.

### Implementation

Added a `SELF_RESOLVING_DISABLE_COPS` constant to `RuboCopDiagnostic` listing cops where an inline disable would resolve the offense itself. The disable line action is skipped for these cops. Currently only includes `Layout/EmptyComment` but the list can be extended if other similar cases come up.

### Automated Tests

Added a test in `code_actions_formatting_test.rb` that verifies no disable action is offered for `Layout/EmptyComment`.

### Manual Tests

1. Open a Ruby file with an empty comment (`#` with nothing after it)
2. Hover over the `Layout/EmptyComment` diagnostic
3. Verify "Disable Layout/EmptyComment for this line" is no longer offered as a code action

### DIFF
--- a/lib/ruby_lsp/requests/support/rubocop_diagnostic.rb
+++ b/lib/ruby_lsp/requests/support/rubocop_diagnostic.rb
@@ -16,9 +16,9 @@ module RubyLsp
 
         # Cops where adding a `rubocop:disable` inline comment would itself resolve the offense,
         # causing Lint/RedundantCopDisableDirective to flag the disable as unnecessary.
-        SELF_RESOLVING_DISABLE_COPS = [
+        SELF_RESOLVING_DISABLE_COPS = Set.new([
           "Layout/EmptyComment",
-        ].freeze #: Array[String]
+        ]).freeze #: Set[String]
 
         ENHANCED_DOC_URL = begin
           gem("rubocop", ">= 1.64.0")


rafaelfranca (approved):

test/requests/code_actions_formatting_test.rb
```diff
@@ -59,8 +77,41 @@ def assert_fixtures_match(name, diagnostic_code, code_action_title)
...
+    # transform edits from lsp to the format RubyLsp::Document wants them
+    # this doesn't work with multiple edits if any edits add lines
+    edits = selected_action.dig(:edit, :documentChanges).flat_map do |doc_change|
+      doc_change[:edits].map do |edit|
+        { range: edit[:range], text: edit[:newText] }
+      end
+    end
+
+    document.push_edits(edits, version: 2)
+    # if document.source != expected
```
vinistock: Can we remove this?

rafaelfranca: Yeah. I thought the team kept it do debug, but I think we should remove. I'll do it

sucicfilip:
Thanks for the quick review :-)

PR #3948, opened 2026-02-09
Update devcontainer to use Ruby 4.0.1

rafaelfranca:
The devcontainer configuration was lagging behind to Ruby version.

This also changes development ruby version to 4.0.1.


.devcontainer/Dockerfile
```diff
@@ -0,0 +1,4 @@
+FROM mcr.microsoft.com/devcontainers/typescript-node:0-20
+
+# Install watchman
+RUN apt-get update && apt-get install -y watchman && rm -rf /var/lib/apt/lists/*
```
vinistock, line 4: Why do we need the `rm -rf` at the end?

.devcontainer/devcontainer.json
```diff
@@ -21,7 +23,15 @@
   // "postCreateCommand": "yarn install",
 
   // Configure tool-specific properties.
-  // "customizations": {},
+  "customizations": {
+    "vscode": {
+      "extensions": [
+          "shopify.ruby-extensions-pack",
+          "dbaeumer.vscode-eslint",
+          "streetsidesoftware.code-spell-checker"
```
vinistock: I'd vote to keep the spell checker one optional.

.devcontainer/devcontainer.json
```diff
@@ -3,13 +3,15 @@
...
   // Or use a Dockerfile or Docker Compose file. More info: https://containers.dev/guide/dockerfile
-  "image": "mcr.microsoft.com/devcontainers/typescript-node:0-20",
+  "build": {
+    "dockerfile": "Dockerfile"
+  },
   "features": {
     "ghcr.io/devcontainers/features/github-cli:1": {},
-    "ghcr.io/devcontainers/features/rust:1": {},
-    "ghcr.io/devcontainers/features/ruby:1": {}
+    "ghcr.io/rails/devcontainer/features/ruby": {
```
vinistock, line 11: I assume this already includes Rust?

.devcontainer/Dockerfile
rafaelfranca: Just to keep the image layer small. Otherwise it will keep a bunch of information that ins't needed. It is standard practice with installing apt packages in docker.

.devcontainer/devcontainer.json
rafaelfranca: Yes!

.devcontainer/devcontainer.json
rafaelfranca: Right. I use a different one, so I'm fine with removing this

vinistock:
I think the CI errors are fixed on main, you should just need a rebase.

PR #3950, opened 2026-02-10
Exclude bundled gems from test discovery

### LINKED ISSUE #3949: Test path pattern (TEST_PATH_PATTERN) discovers tests from .bundle/gems and vendor/bundle
### Description

Confirmed as a bug by @vinistock on [Ruby DX Discord](https://ruby-dx.slack.com/archives/C050D17S6FP/p1770666246587129).

#### VS Code Version

1.108.2

0.10.0

#### Ruby LSP Server Version

0.26.5

#### Ruby LSP Add-ons

- Standard Ruby (unknown)
- Ruby LSP RSpec (0.26.5)
- Ruby LSP Rails (0.4.8)
- RuboCop (1.82.1)

#### Ruby Version

3.4.7

#### Ruby Version Manager

rbenv

#### Installed Extensions

<details>

- copilot-chat (0.36.2)
- openapi-preview (2.3.1)
- rainbow-csv (3.24.1)
- remote-containers (0.442.0)
- ruby-lsp (0.10.0)
- ruby-lsp-go-to-spec (0.1.3)
- ruby-symbols (0.1.8)
- vim (1.32.4)
- vscode-containers (2.4.1)
- vscode-docker (2.0.0)
- vscode-pdf (0.1.6)
- vscode-standard-ruby (0.0.16)
- vscode-xml (0.29.0)
- vscode-xsd-treeview (0.0.5)
</details>

#### Ruby LSP Settings

<details>

##### Workspace

```json
{
  "addonSettings": {
    "Ruby LSP RSpec": {
      "rspecCommand": "bin/rspec"
    }
  },
  "rubyVersionManager": "rbenv",
  "indexing": {
    "excludedPatterns": [
      ".bundle/**/*"
    ]
  },
  "featureFlags": {
    "fullTestDiscovery": true
  }
}
```

##### User

```json
{
  "enabledFeatures": {
    "codeActions": true,
    "diagnostics": true,
    "documentHighlights": true,
    "documentLink": true,
    "documentSymbols": true,
    "foldingRanges": true,
    "formatting": true,
    "hover": true,
    "inlayHint": true,
    "onTypeFormatting": true,
    "selectionRanges": true,
    "semanticHighlighting": true,
    "completion": true,
    "codeLens": true,
    "definition": true,
    "workspaceSymbol": true,
    "signatureHelp": true,
    "typeHierarchy": true
  },
  "featuresConfiguration": {},
  "addonSettings": {
    "Ruby LSP RSpec": {
      "rspecCommand": "bin/rspec"
    }
  },
  "rubyVersionManager": "rbenv",
  "customRubyCommand": "",
  "formatter": "auto",
  "linters": null,
  "bundleGemfile": "",
  "testTimeout": 30,
  "branch": "",
  "pullDiagnosticsOn": "both",
  "useBundlerCompose": false,
  "bypassTypechecker": false,
  "rubyExecutablePath": "",
  "indexing": {
    "excludedPatterns": [
      ".bundle/**/*"
    ]
  },
  "erbSupport": true,
  "featureFlags": {
    "fullTestDiscovery": true
  },
  "sigOpacityLevel": "1"
}
```

</details>

### Reproduction steps

1. `bundle install --path .bundle/gems` or `bundle install --deployment` to install gems in the local repository.  Any path used could encounter this problem, but these two are common conventions across ruby projects.
2. Using VSCode as an IDE, navigate to the "testing" sidebar.
3. Click the Refresh tests icon at the top (looks like refresh)

Expected: only my `${workspaceRoot}/spec` directory is discovered, full of `rspec` tests.
Current: Tests from gems installed in `.bundle/gems` or `vendor/bundle` are also discovered.

Example showing `.bundle/gems`:
Image

### Suggested resolution
Option A: Make test patterns configurable (include and exclude).  Supply good defaults.
Option B: Ensure `.bundle` (and all subdirectories) and `vendor/bundle` and all subdirectories are hardcoded exclusions, much like the existing `TEST_PATH_PATTERN` functions.

rafaelfranca:
Test discovery was finding tests in .bundle/gems and vendor/bundle directories, which contain bundled gem code. This adds an exclude pattern to prevent these directories from being scanned during test discovery.

Fixes #3949.

### DIFF
--- a/vscode/src/testController.ts
+++ b/vscode/src/testController.ts
@@ -15,6 +15,13 @@ const asyncExec = promisify(exec);
 
 const NESTED_TEST_DIR_PATTERN = "**/{test,spec,features}/**/";
 const TEST_FILE_PATTERN = `${NESTED_TEST_DIR_PATTERN}{*_test.rb,test_*.rb,*_spec.rb,*.feature}`;
+const IGNORED_FOLDERS = [".bundle", "vendor/bundle", "node_modules", "tmp", "log"];
+const IGNORED_FOLDERS_EXCLUDE_PATTERN = `{${IGNORED_FOLDERS.map((folder) => `**/${folder}/**`).join(",")}}`;
+
+// Build a regex to match paths containing any of the ignored folders
+const IGNORED_FOLDERS_PATH_REGEX = new RegExp(
+  IGNORED_FOLDERS.map((folder) => `/${folder.replace(/[.*+?^${}()|[\]\\]/g, "\\$&")}/`).join("|"),
+);
 
 interface CodeLensData {
   type: string;
@@ -142,6 +149,10 @@ export class TestController {
       testFileWatcher,
       nestedTestDirWatcher,
       testFileWatcher.onDidCreate(async (uri) => {
+        if (this.isInIgnoredFolder(uri)) {
+          return;
+        }
+
         const workspace = vscode.workspace.getWorkspaceFolder(uri);
 
         if (!workspace || !vscode.workspace.workspaceFolders) {
@@ -160,6 +171,10 @@ export class TestController {
         await this.addTestItemsForFile(uri, workspace, initialCollection);
       }),
       testFileWatcher.onDidChange(async (uri) => {
+        if (this.isInIgnoredFolder(uri)) {
+          return;
+        }
+
         const item = await this.getParentTestItem(uri);
 
         if (item) {
@@ -172,6 +187,10 @@ export class TestController {
         }
       }),
       nestedTestDirWatcher.onDidDelete(async (uri) => {
+        if (this.isInIgnoredFolder(uri)) {
+          return;
+        }
+
         const pathParts = uri.fsPath.split(path.sep);
         if (pathParts.includes(".git")) {
           return;
@@ -184,6 +203,10 @@ export class TestController {
         }
       }),
       testFileWatcher.onDidDelete(async (uri) => {
+        if (this.isInIgnoredFolder(uri)) {
+          return;
+        }
+
         const item = await this.getParentTestItem(uri);
 
         if (item) {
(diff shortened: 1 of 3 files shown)


vscode/src/testController.ts
```diff
@@ -15,6 +15,7 @@ const asyncExec = promisify(exec);
 
 const NESTED_TEST_DIR_PATTERN = "**/{test,spec,features}/**/";
 const TEST_FILE_PATTERN = `${NESTED_TEST_DIR_PATTERN}{*_test.rb,test_*.rb,*_spec.rb,*.feature}`;
+const BUNDLED_GEMS_EXCLUDE_PATTERN = "{**/.bundle/**,**/vendor/bundle/**}";
```
vinistock: I agree with the approach, but I would generalize this into `EXCLUDED_PATTERNS` and not make it about dependencies.

That opens the possibility for us to allow exclusions to be configurable in the case of a repository that contains test file fixtures inside.

vscode/src/testController.ts
```diff
@@ -1146,6 +1163,11 @@ export class TestController {
     return false;
   }
 
+  private isInBundledGems(uri: vscode.Uri): boolean {
+    const pathParts = uri.fsPath.split(path.sep);
+    return pathParts.includes(".bundle") || pathParts.includes("vendor");
```
vinistock: Could we use the excluded patterns regex here?

Otherwise we're listing the directories to be excluded in two different places. Also, the patterns are slightly different (the regex excludes `vendor/bundle` whereas this excludes `vendor` completely).

vscode/src/testController.ts
vinistock: Also, I think we can squeeze better performance out of VS Code globbing files if we also exclude `node_modules`, `tmp` and `log` by default.

PR #3954, opened 2026-02-11
Upgrade to minintest 6

rafaelfranca:
The only think we needed to do was to change a test because `assert_equal(nil, nil)` is now an error on minitest.


bin/minitest
```diff
@@ -0,0 +1,16 @@
...
+# The application 'minitest' is installed as part of a gem, and
+# this file is here to facilitate running it.
+#
+
+ENV["BUNDLE_GEMFILE"] ||= File.expand_path("../Gemfile", __dir__)
+
+require "rubygems"
+require "bundler/setup"
+
+load Gem.bin_path("minitest", "minitest")
```
vinistock, line 16: Minitest has an executable now? Does that mean there's no longer a need for `bin/test`?

rafaelfranca: Yes, it does. I think it depends. `bin/test` also runs the indexer tests if you don't pass any argument. `bin/minitesr` would not do that, and to make matters worse, if the test file start with `test_` like one of our fixtures it will try to run and fail, and there is no easy way to configure `bin/minitest` to ignore those fixtures. That is why I kept it.

PR #3957, opened 2026-02-13
Stop preferring /opt/homebrew/bin/shadowenv

joshheinrichs-shopify:
We distribute shadowenv via Nix, but at least one person still has an old version installed on their machine via Homebrew. The version at the front of their $PATH and the one in Homebrew were far enough out of sync that they were incompatible. By just looking for shadowenv on $PATH we should be able to locate the correct version.

It looks like this was added to help with cases where $PATH is broken[1]. We could keep something as a fallback, but at least at Shopify, /opt/homebrew/bin isn't a good fallback anymore.

[1] b7dc6d2ca44781b68b47a0a574c319e5154f3e56

### DIFF
--- a/vscode/src/ruby/shadowenv.ts
+++ b/vscode/src/ruby/shadowenv.ts
@@ -21,7 +21,7 @@ export class Shadowenv extends VersionManager {
       );
     }
 
-    const shadowenvExec = await this.findExec([vscode.Uri.file("/opt/homebrew/bin")], "shadowenv");
+    const shadowenvExec = await this.findExec([], "shadowenv");
 
     try {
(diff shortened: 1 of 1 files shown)


vinistock:
The main reason the fallback was added is because if you open VS Code from an icon (rather than from the terminal), it does not inherit the shell environment and then it cannot find executables like `shadowenv`.

This was already an issue with the Homebrew version. Several developers open the editor from the icon and then then `/opt/homebrew/bin` wasn't in their `$PATH`, causing the environment activation to crash.

How can this problem be addressed in the Nix approach without blocking developers from opening their editors through the icon?

vinistock:
Superseded by https://github.com/Shopify/ruby-lsp/pull/4076

PR #3963, opened 2026-02-18
Fix incompatible addon version activation when Bundler.setup fails after retry

KaanOzkan:
When Bundler.setup raises GemNotFound or GitError and the retry has already been attempted (or install_error is set), the rescue block fell through without setting setup_error. This caused the server to boot thinking setup succeeded, bypassing the load_addons guard. Without Bundler constraining Gem.find_files, every installed version of every addon gem was discovered, leading to version mismatches (e.g., ruby-lsp-rails 0.4.2 loaded with ruby-lsp 0.26.5) and ArgumentError crashes.

The StandardError rescue already handled this correctly - this aligns the GemNotFound/GitError rescue to do the same.

### Motivation

Resolves errors that may occur from activating incorrect add-on versions. Instead of getting error reports we can not load the add-ons in the first place.

### Implementation

Set `setup_error` and adjust the `LOAD_PATH` similar to `StandardError` case.

### DIFF
--- a/exe/ruby-lsp-launcher
+++ b/exe/ruby-lsp-launcher
@@ -101,7 +101,7 @@ begin
     Bundler.setup
     $stderr.puts("Composed Bundle set up successfully")
   end
-rescue Bundler::GemNotFound, Bundler::GitError
+rescue Bundler::GemNotFound, Bundler::GitError => e
   # Sometimes, we successfully set up the bundle, but users either change their Gemfile or uninstall gems from an
   # external process. If there's no install error, but the gem is still not found, then we need to attempt to start from
   # scratch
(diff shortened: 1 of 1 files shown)


exe/ruby-lsp-launcher
```diff
@@ -113,6 +113,10 @@ rescue Bundler::GemNotFound, Bundler::GitError
       exec(Gem.ruby, __FILE__, *ARGV, "--retry")
     end
   end
+
+  setup_error = e
+  $stderr.puts("Failed to set up composed Bundle\n#{e.full_message}")
+  $LOAD_PATH.unshift(File.expand_path("../lib", __dir__))
```
KaanOzkan, line 119: Was there another benefit for trying to load addons in this case that I'm missing? It made sense to me that in case we don't trigger the`exec` we should enter degraded mode.

vinistock: No, we can't load add-ons if `Bundler.setup` failed indeed.

PR #3971, opened 2026-02-20
Clean up cancelled requests after processing them

vinistock:
### Motivation

We were growing the cancelled requests list indefinitely when the notification came in during process. We only cleaned up if the editor managed to cancel the request before we got to it.

### Implementation

We can fix the unbound memory growth by always ensuring that we delete from the list after processing.

### Automated Tests

Added a test to ensure that the list is always empty after the request got processed.

### DIFF
--- a/lib/ruby_lsp/base_server.rb
+++ b/lib/ruby_lsp/base_server.rb
@@ -165,18 +165,20 @@ module RubyLsp
 
       # Check if the request was cancelled before trying to process it
       @global_state.synchronize do
-        if id && @cancelled_requests.include?(id)
+        if id && @cancelled_requests.delete(id)
           send_message(Error.new(
             id: id,
             code: Constant::ErrorCodes::REQUEST_CANCELLED,
             message: "Request #{id} was cancelled",
           ))
-          @cancelled_requests.delete(id)
+
           return
         end
       end
 
       process_message(message)
+      # Ensure we remove the request if it got cancelled while it was being processed
+      @cancelled_requests.delete(id)
     end
 
     #: ((Result | Error | Notification | Request) message) -> void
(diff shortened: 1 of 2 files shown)


vinistock:
* **#3971** <a href="https://app.graphite.com/github/pr/Shopify/ruby-lsp/3971?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/Shopify/ruby-lsp/3971?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`
<hr />

You must have a Graphite account in order to use the merge queue. Sign up using [this link](https://app.graphite.com/invite/github/Shopify?ref=merge-queue-instructions-in-stack-comment&prId=3971).

<h2></h2>
<sub>

</sub>

<sub>

</sub>

<h2></h2>

lib/ruby_lsp/base_server.rb
```diff
@@ -165,18 +165,20 @@ def handle_incoming_message(message)
...
           ))
-          @cancelled_requests.delete(id)
+
           return
         end
       end
 
       process_message(message)
+      # Ensure we remove the request if it got cancelled while it was being processed
+      @cancelled_requests.delete(id)
```
alexcrocha, line 181: This is now outside the `@global_state.synchronize` block. Does it need to be wrapped again?
```suggestion
      @global_state.synchronize { @cancelled_requests.delete(id) }
```

vinistock: It doesn't. In the other case, we're interested in reading the value from the array and checking for its presence, so locking is important to ensure we don't have races.

For this last delete, all we care about is cleaning up the array so that it doesn't keep growing indefinitely. We're not taking any actions based on the presence of the value.

alexcrocha: I see, thanks!

PR #3979, opened 2026-02-26
Delete always-skipped GitHub action

amomchilov:
### Motivation

https://github.com/Shopify/ruby-lsp/actions/workflows/dependabot_update_rbis_run.yml


vinistock:
We do want to auto-update gem RBIs though. Why is this always skipped?

amomchilov:
@vinistock Yeah I'm looking into it

PR #3985, opened 2026-03-02
Apply lower bound `ruby-lsp` version constraint in composed bundle

st0012:
### Motivation

Without this constraint, Bundler could resolve to very old versions that cause errors. Reasons for choosing 0.18.0:

- Fixes resolution to versions like 0.11.2 which cause `NameError: uninitialized constant RubyLsp::RuboCop::Cop`
- First version with stable Prism 1.0 dependency (`~> 1.0`) — earlier versions used narrow pre-1.0 pins that caused dependency resolution issues
- Includes `rbs >= 3` dependency (added in v0.17.3) needed for core class indexing
- Not overly restrictive — v0.18.0 is from September 2024, leaving ~18 months of versions available

### Implementation

When the composed `.ruby-lsp/Gemfile` is generated for users who don't have `ruby-lsp` in their own Gemfile, apply a `>= 0.18.0` lower bound.

### DIFF
--- a/lib/ruby_lsp/setup_bundler.rb
+++ b/lib/ruby_lsp/setup_bundler.rb
@@ -34,6 +34,7 @@ module RubyLsp
     # Gems that should be kept up to date in the composed bundle. When updating, any of these gems that are not
     # already in the user's Gemfile will be updated together.
     GEMS_TO_UPDATE = ["ruby-lsp", "debug", "prism", "rbs"].freeze #: Array[String]
+    RUBY_LSP_MIN_VERSION = "0.18.0" #: String
 
     #: (String project_path, **untyped options) -> void
     def initialize(project_path, **options)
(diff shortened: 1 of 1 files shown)


lib/ruby_lsp/setup_bundler.rb
```diff
@@ -34,6 +34,7 @@ def stdout
     # Gems that should be kept up to date in the composed bundle. When updating, any of these gems that are not
     # already in the user's Gemfile will be updated together.
     GEMS_TO_UPDATE = ["ruby-lsp", "debug", "prism", "rbs"].freeze #: Array[String]
+    RUBY_LSP_MIN_VERSION = "0.18.0" #: String
```
vinistock: There's no need to annotate string literals.

```suggestion
    RUBY_LSP_MIN_VERSION = "0.18.0"
```

Issue #3639, opened 2025-07-02
Duplicate entries when receiving a `textDocument/didOpen` before indexing is complete


vinistock:
If the user opens the editor with files already opened, the editor might be able to get some `didOpen` notifications sent before indexing is complete. In this case, opening the documents will also index them, but we have no way to tell the indexing process to drop the URI.

We need to ensure that that indexing the same file twice without any modifications never results in duplicates because there are some scenarios that risk it:

- open a document before indexing is finished
- a file is modified on disk before indexing is finished
- a file is modified on disk and then immediately opened

andriytyurnikov:
Cross-referencing #4022, which removes one standing source of these duplicates — not the race described here, but a deterministic variant of the same root cause.

This issue is about the same file being indexed twice without a way to drop the earlier URI. There is a second path to the same outcome that does not need a race at all: if the URI we build during indexing is escaped differently from the URI the editor sends, the two never compare equal, so *every* file with an affected character is permanently indexed twice. @vinistock described exactly this in #3503:

> During initial indexing, the URIs used to store entries come from our own logic and escaping. After you open a document in the editor and modify it, then we are receiving the URI from the editor. Any mismatches between the two will result in inserting the same declarations for two different URIs.

That mismatch was real. Comparing `URI::Generic.from_path` against the `vscode-uri` package directly, 12 of 17 sample paths disagreed, because our safe set came from RFC 2396 and preserved `(`, `)`, `!`, `'`, `*`, `$`, `&`, `+`, `,`, `;`, `=` and `@`, all of which the editor percent-encodes. So any path containing one of those — `foo (copy).rb`, `a&b.rb`, `Rails.application, etc.rb` — produced two index entries every time it was opened. #4022 narrows the safe set to match, reaching 17/17.

This does not fix the timing problem in this issue, and it is orthogonal to the new indexer work. It just means that once that lands, escaping mismatches will not be quietly re-introducing the same duplicates underneath it.

Related: #3842 was closed as a duplicate of this issue. That report showed two definition results differing only by `//` versus `/` in the URI — a third route to the same symptom, path normalisation rather than escaping or timing. Worth noting that it is not covered by #4022 either, in case it needs tracking separately.

PR #3992, opened 2026-03-06
Start indexing and resolving Rubydex graph

vinistock:
### Motivation

This PR starts indexing and resolving the Rubydex graph on boot and for tests. Note that consuming incremental changes is not yet a part of the API, so this PR doesn't address that yet.

Our strategy, for smaller and easier to review PRs, will be to maintain both indexing mechanisms while we migrate each request one by one. That way, we don't have to review a massive amount of changes and we can rely on CI.

### Implementation

The implementation makes sure we're:

- Instantiating the graph and configuring the encoding and workspace path
- Indexing and resolving on boot and tests

### DIFF
--- a/lib/ruby_lsp/server.rb
+++ b/lib/ruby_lsp/server.rb
@@ -1228,12 +1228,18 @@ module RubyLsp
 
     #: -> void
     def perform_initial_indexing
+      progress("indexing-progress", message: "Indexing workspace...")
+      @global_state.graph.index_workspace
+
+      progress("indexing-progress", message: "Resolving graph...")
+      @global_state.graph.resolve
+
       # The begin progress invocation happens during `initialize`, so that the notification is sent before we are
       # stuck indexing files
       Thread.new do
         begin
           @global_state.index.index_all do |percentage|
-            progress("indexing-progress", percentage)
+            progress("indexing-progress", percentage: percentage)
             true
           rescue ClosedQueueError
             # Since we run indexing on a separate thread, it's possible to kill the server before indexing is complete.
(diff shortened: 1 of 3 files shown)


lib/ruby_lsp/server.rb
```diff
@@ -1228,12 +1228,18 @@ def shutdown
 
     #: -> void
     def perform_initial_indexing
+      progress("indexing-progress", message: "Indexing workspace...")
+      @global_state.graph.index_workspace
+
+      progress("indexing-progress", message: "Resolving graph...")
+      @global_state.graph.resolve
```
alexcrocha, line 1235: Should this be inside its own `Thread.new` block as well, like the existing indexing? Or is Rubydex handling it in some other way

vinistock: The old indexing happens inside a thread because it's too slow and we don't want to block requests while we're still indexing.

Rubydex is fast enough that we don't have to run indexing in a thread, which will simplify a lot of our code.

For example, right now, we need code to bust the code lens cache and trigger a refresh after indexing because we may have computed incorrect lenses if a request comes in before indexing is finished.

st0012: Right now it still takes 30s to index core using rubydex, and even if we cut it down to 20s I'd argue that it's still a noticeable wait.

I'd vote for keeping it in a thread for now, so to users the behaviour is simply "background indexing got faster", instead of actually seeing longer blocking on requests?

And after fully migrated to rubydex, we see how fast we can realistically get to and what cleanups will be worth the tradeoff.

vinistock: The asynchronous aspect of indexing is the source of various bugs, so I would prefer we invert the decision here. Remove the complexity and revisit if users actually complain it's too slow.

For example, #3639 is difficult to fix properly with async indexing. We have the code lens hack to prevent issues, but we also have a terrible hack in file watched events to keep sleeping and re-enqueuing until indexing is complete.

st0012: I see. If it clears out many bugs at once then let's do it then.
Do you have a fully list of such bugs/cleanups somewhere so we don't forget to do them or parellalize the work?

vinistock: I don't have a consolidated list. It's a mix between issues in this repo and things that will naturally go away once we remove the old index.

PR #4002, opened 2026-03-09
Improve test stability

soutaro:
This PR is to make some tests more stable.

* Replaced a fixed 10-second delay in `workspace.test.ts` with the new `retry` utility, allowing assertions to be retried for up to 30 seconds, improving reliability in the presence of asynchronous file system events.
* Increased test timeouts in `testController.test.ts` and `workspace.test.ts` to 30 and 60 seconds respectively, ensuring tests have sufficient time to complete under slow conditions.

### DIFF
--- a/vscode/src/test/suite/helpers.ts
+++ b/vscode/src/test/suite/helpers.ts
@@ -70,3 +70,21 @@ export function createContext() {
     },
   } as unknown as FakeContext;
 }
+
+// Retries the given assertion function until it passes or the timeout (in ms) is exceeded.
+// Polls every 500ms. If the assertion never passes, the last assertion error is thrown.
+export async function retryForDuration(timeoutMs: number, fn: () => void) {
+  const deadline = Date.now() + timeoutMs;
+
+  while (Date.now() < deadline) {
+    try {
+      fn();
+      return;
+    } catch (_error) {
+      await new Promise((resolve) => setTimeout(resolve, 500));
+    }
+  }
+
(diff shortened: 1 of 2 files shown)


vscode/src/test/suite/helpers.ts
```diff
@@ -70,3 +70,21 @@ export function createContext() {
     },
   } as unknown as FakeContext;
 }
+
+// Retries the given assertion function until it passes or the timeout (in ms) is exceeded.
+// Polls every 500ms. If the assertion never passes, the last assertion error is thrown.
+export async function retry(timeoutMs: number, fn: () => void) {
```
st0012: Let's make the name more explicit on what type of retry this does:

```suggestion
export async function retry_for_duration(duration: number, fn: () => void) {
```

soutaro: Changed the function name to `retryForDuration`.

soutaro:
Another test is failing...

```
  1 failing
  1) Launch integrations
       with launcher mode enabled:
     Error: Timeout of 120000ms exceeded. For async tests and hooks, ensure "done()" is called; if returning a Promise, ensure it resolves. (d:\a\ruby-lsp\ruby-lsp\vscode\out\test\suite\launch.test.js)
  	at listOnTimeout (node:internal/timers:588:17)
  	at process.processTimers (node:internal/timers:523:7)

Error: 1 tests failed.
```

PR #4004, opened 2026-03-12
Optimize test suite duration by caching test infrastructure

st0012:
## Summary
- Cache `RuboCopFormatter` at the class level in formatting and diagnostics expectation tests, avoiding repeated creation of `RuboCopRunner` instances (2 per test × 2,352 tests)
- Cache `GlobalState` at the class level in `ExpectationsTestRunner`, avoiding redundant initialization per test
- Skip addon loading in Definition and Hover expectation tests where addons aren't needed (`with_server(load_addons: true)` costs ~38ms vs ~2ms without)

In total, this reduces the CI duration by about 10min (~30 -> 20), which is a 33% reduction.

## Test plan
- [x] All existing tests pass (18,329 tests, same count, same 2 pre-existing nix-path failures)
- [x] No new failures or errors introduced
- [x] Compare CI test duration against main branch baseline

### DIFF
--- a/test/requests/definition_expectations_test.rb
+++ b/test/requests/definition_expectations_test.rb
@@ -7,15 +7,11 @@ require_relative "support/expectations_test_runner"
 class DefinitionExpectationsTest < ExpectationsTestRunner
   expectations_tests RubyLsp::Requests::Definition, "definition"
 
-  private
-
-  # Skip addon loading by default — only test_definition_addons needs addons
+  # Skip add-on loading by default — only test_definition_addons needs it
   def with_server(source = nil, uri = Kernel.URI("file:///fake.rb"), stub_no_typechecker: false, load_addons: false,
     &block)
     super
   end
-
-  public
 
   def run_expectations(source)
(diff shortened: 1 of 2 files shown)


test/requests/definition_expectations_test.rb
```diff
@@ -7,6 +7,12 @@
 class DefinitionExpectationsTest < ExpectationsTestRunner
   expectations_tests RubyLsp::Requests::Definition, "definition"
 
+  # Skip add-on loading by default — only test_definition_addons needs it
+  def with_server(source = nil, uri = Kernel.URI("file:///fake.rb"), stub_no_typechecker: false, load_addons: false,
```
vinistock, line 11: I'd rather we change the default to not load add-ons than creating these methods that only invoke super tbh.

st0012: I thought about this but then it could break all the addons using this in tests, which IMO would be a bad experience 😅

vinistock: Fair, but then I would vote for us to update all calls to pass `load_addons: false` tbh.

PR #4007, opened 2026-03-15
fix: Don't escape regex characters in test commands on Windows

### LINKED ISSUE #3759: CodeLens Run Test In Terminal doesn't work (on Windows) due to incorrectly escaped Regex
#### VS Code Version

1.104.1

0.9.32

#### Ruby LSP Server Version

0.26.1

#### Ruby LSP Add-ons

- RuboCop (unknown)

#### Ruby Version

3.1.4

#### Ruby Version Manager

none

#### Installed Extensions

<details>

- Lisp (0.1.12)
- code-spell-checker (4.2.6)
- color-highlight (2.8.0)
- errorlens (3.26.0)
- exercism-cli (0.2.2)
- gitlens (17.4.1)
- jikiscript (1.0.0)
- postcss (1.0.9)
- prettier-vscode (11.0.0)
- rails (0.22.0)
- rainbow-csv (3.22.0)
- react-native-ide (1.3.0)
- remote-explorer (0.5.0)
- remote-ssh (0.120.0)
- remote-ssh-edit (0.87.0)
- ruby-lsp (0.9.32)
- vscode-eslint (3.0.16)
- vscode-expo-tools (1.6.1)
- vscode-github-actions (0.27.2)
- vscode-graphql (0.13.2)
- vscode-graphql-syntax (1.3.8)
- vscode-tailwindcss (0.14.26)
- vscode-versionlens (1.22.4)
- vscode-yaml (1.18.0)
- vscode-zipfs (3.0.0)
- vsliveshare (1.0.5959)
</details>

#### Ruby LSP Settings

<details>

##### Workspace

```json
{
  "indexing": {
    "includedPatterns": [
      "lib/**/*.rb",
      "test/**/*.rb"
    ]
  }
}
```

##### User

```json
{
  "enabledFeatures": {
    "codeActions": true,
    "diagnostics": true,
    "documentHighlights": true,
    "documentLink": true,
    "documentSymbols": true,
    "foldingRanges": true,
    "formatting": false,
    "hover": true,
    "inlayHint": true,
    "onTypeFormatting": false,
    "selectionRanges": true,
    "semanticHighlighting": true,
    "completion": true,
    "codeLens": true,
    "definition": true,
    "workspaceSymbol": true,
    "signatureHelp": true,
    "typeHierarchy": true
  },
  "featuresConfiguration": {},
  "addonSettings": {},
  "rubyVersionManager": {
    "identifier": "none"
  },
  "customRubyCommand": "",
  "formatter": "auto",
  "linters": null,
  "bundleGemfile": "",
  "testTimeout": 30,
  "branch": "",
  "pullDiagnosticsOn": "both",
  "useBundlerCompose": false,
  "bypassTypechecker": false,
  "rubyExecutablePath": "",
  "indexing": {
    "includedPatterns": [
      "lib/**/*.rb",
      "test/**/*.rb"
    ]
  },
  "erbSupport": true,
  "featureFlags": {},
  "sigOpacityLevel": "1"
}
```

</details>

### Reproduction steps

1. Start the Ruby LSP using VS Code
2. Open a Ruby test file with a single test
3. Click "Run Test in Terminal"
4. Terminal reports no tests ran

```ruby
class TestNothing < Minitest::Test
  def test_nothing
    refute_nil true
  end
end
```

The full file can be tested:

```shell
C:\Users\Derk-Jan\...snip...>bundle exec ruby -rC:/Ruby31-x64/lib/ruby/gems/3.1.0/gems/ruby-lsp-0.26.1/lib/ruby_lsp/test_reporters/minitest_reporter.rb -rC:/Ruby31-x64/lib/ruby/gems/3.1.0/gems/ruby-lsp-0.26.1/lib/ruby_lsp/test_reporters/test_unit_reporter.rb -Itest c:/Users/Derk-Jan/...snip.../test/test_nothing.rb --name "/^TestNothing(#|::)/"
Run options: --name "/^TestNothing(#|::)/" --seed 56682

# Running:

.

Finished in 0.015750s, 63.4917 runs/s, 63.4917 assertions/s.
```

Simplifying this command we see:

```shell
bundle exec ruby -rPathToLspProvidedReporter.rb -rPathToAnotherLspProvidedReporter.rb -Itest PathToFile --name "/^TestClassName(#|::)/"
```

There are three paths:

- PathToLspProvidedReporter
- PathToAnotherLspProvidedReporter
- PathToFile

There is one regex matcher:

- /^TestNothing(#|::)/

For a single test, we get:

```shell
C:\Users\Derk-Jan\...snip...>bundle exec ruby -rC:/Ruby31-x64/lib/ruby/gems/3.1.0/gems/ruby-lsp-0.26.1/lib/ruby_lsp/test_reporters/minitest_reporter.rb -rC:/Ruby31-x64/lib/ruby/gems/3.1.0/gems/ruby-lsp-0.26.1/lib/ruby_lsp/test_reporters/test_unit_reporter.rb -Itest c:/Users/Derk-Jan/...snip.../test_nothing.rb --name "/^TestNothing#test_nothing\$/"
Run options: --name "/^TestNothing#test_nothing\\$/" --seed 33369

[...]

Finished in 0.010229s, 0.0000 runs/s, 0.0000 assertions/s.

0 runs, 0 assertions, 0 failures, 0 errors, 0 skips
Nothing ran for filter: /^TestNothing#test_nothing\$/
```

The issue can be seen in the `--name` argument:

```shell
/^TestNothing#test_nothing\\$/
```

Removing the extra slash makes the entire command work again.

BillionClaw:
Fixes #3759

Shellwords.escape() is designed for Unix shells and incorrectly escapes characters like `$` on Windows. This causes the "Run Test In Terminal" CodeLens feature to fail on Windows because the regex pattern becomes incorrectly escaped.

Changes:
- Added escape_for_shell() helper that uses Gem.win_platform? to detect Windows
- On Windows: returns text unescaped (Windows cmd parsing doesn't need same escaping)
- On Unix: continues using Shellwords.escape()

### DIFF
--- a/lib/ruby_lsp/listeners/code_lens.rb
+++ b/lib/ruby_lsp/listeners/code_lens.rb
@@ -236,35 +236,47 @@ module RubyLsp
           dynamic_stack = group_stack[last_dynamic_reference_index + 1..] #: as !nil
 
           if method_name
-            " --name " + "/::#{Shellwords.escape(dynamic_stack.join("::")) + "#" + Shellwords.escape(method_name)}$/"
+            " --name " + "/::#{escape_for_shell(dynamic_stack.join("::")) + "#" + escape_for_shell(method_name)}$/"
           else
             # When clicking on a CodeLens for `Test`, `(#|::)` will match all tests
             # that are registered on the class itself (matches after `#`) and all tests
             # that are nested inside of that class in other modules/classes (matches after `::`)
-            " --name " + "\"/::#{Shellwords.escape(dynamic_stack.join("::"))}(#|::)/\""
+            " --name " + "\"/::#{escape_for_shell(dynamic_stack.join("::"))}(#|::)/\""
           end
         elsif method_name
           # We know the entire path, do an exact match
-          " --name " + Shellwords.escape(group_stack.join("::")) + "#" + Shellwords.escape(method_name)
+          " --name " + escape_for_shell(group_stack.join("::")) + "#" + escape_for_shell(method_name)
         elsif spec_name
-          " --name " + "\"/^#{Shellwords.escape(group_stack.join("::"))}##{Shellwords.escape(spec_name)}$/\""
+          " --name " + "\"/^#{escape_for_shell(group_stack.join("::"))}##{escape_for_shell(spec_name)}$/\""
         else
           # Execute all tests of the selected class and tests in
           # modules/classes nested inside of that class
-          " --name " + "\"/^#{Shellwords.escape(group_stack.join("::"))}(#|::)/\""
+          " --name " + "\"/^#{escape_for_shell(group_stack.join("::"))}(#|::)/\""
         end
       end
 
       #: (Array[String] group_stack, String? method_name) -> String
       def generate_test_unit_command(group_stack, method_name)
         group_name = group_stack.last #: as !nil
-        command = " --testcase " + "/#{Shellwords.escape(group_name)}/"
+        command = " --testcase " + "/#{escape_for_shell(group_name)}/"
 
         if method_name
-          command += " --name " + Shellwords.escape(method_name)
+          command += " --name " + escape_for_shell(method_name)
         end
 
         command
+      end
+
+      #: (String text) -> String
+      def escape_for_shell(text)
+        if Gem.win_platform?
+          # On Windows, Shellwords.escape incorrectly escapes characters like `$`
+          # which breaks regex patterns. Windows doesn't need the same escaping
+          # as Unix shells for these characters.
+          text
+        else
+          Shellwords.escape(text)
+        end
       end
 
       #: (Prism::CallNode node, kind: Symbol) -> void


rafaelfranca:
Fix looks good. Can you sign the CLA please?

lib/ruby_lsp/listeners/code_lens.rb
```diff
@@ -236,37 +236,49 @@ def generate_minitest_command(group_stack, method_name, spec_name)
...
         command
       end
 
+      #: (String text) -> String
+      def escape_for_shell(text)
+        if Gem.win_platform?
+          # On Windows, Shellwords.escape incorrectly escapes characters like `$`
+          # which breaks regex patterns. Windows doesn't need the same escaping
+          # as Unix shells for these characters.
+          text
```
vinistock, line 276: Are you positive there's no need for any escaping? There are differences between PowerShell and exe and we have been bitten before with syntax errors that only happened in one or the other.

Would it be possible to add integration tests that generate the test commands and then attempt to run it on both PowerShell and exe? That would be super helpful for ensuring correctness.

rafaelfranca: All shells need escaping, but not for the arguments being passed to this method. They are all generated by our code, not based on user input, no? They would only need escaping if we were passing quoted strings as arguments to this method.

rafaelfranca: We should also make sure we avoid escaping in the calls that would receive a regexp. I think it is only the method name

rafaelfranca:
Thank you for the pull request. I think we should go with #4012 instead

rafaelfranca:
Nevermind, this fix need to happen in multiple place. I do think we should not be escaping on windows for the method name, but we should fix this problem in all places that need to be fixed.

lib/ruby_lsp/listeners/code_lens.rb
vinistock: This is escaping the user's test class names and test method names, which may include regex wild cards for dynamically defined classes.

For example:

```ruby
module Foo
  module variable
    class MyTest < Minitest::Test
    end
  end
end
```

This will generate a regex with a wild card to allow us to execute the test despite the dynamic portion of the nesting.

Are all characters involved valid for PowerShell and exe? That's the part I'm not sure.

For example, we started escaping the regex anchor `$` because not escaping it breaks execution on the fish shell.

At the end of the day, we need proper tests to verify that we generate can be correctly executed on Windows or we'll always be subject to a small modification not being syntax compatible with its shells.

onthebed:
Friendly bump -- this PR has been open for a while. Happy to make any changes if needed, or close it if no longer relevant.

onthebed:
Thanks for the review — I agree this branch is broader than it should be. I'll leave 4007 open and let #4012 carry the narrower fix instead.

breezeFur:
Re: the escaping question — the regex content still needs to be preserved, but `Shellwords.escape` is Unix-shell specific and is the part that breaks the generated command on Windows. So the goal here is not “no escaping at all”, it’s “avoid Unix-style quoting on Windows and keep the regex form intact.” I agree we should add integration coverage that exercises the generated command on both PowerShell and exe.

onthebed:
You're right to push on this. After revisiting it, I don't think the Windows helper in 4007 is the simplest viable fix. The narrower fix is to keep the regex fragments intact and only stop over-escaping the  pattern where we know we're passing a generated regex, which is what #4012 does. I'll leave this PR as the duplicate and let #4012 carry the change plus regression coverage.

PR #4014, opened 2026-03-18
Migrate type inferrer to use Rubydex

vinistock:
### Motivation

We need the type inferrer to operate on the Rubydex graph instead of the old index. Otherwise, we can't migrate requests properly.

### Implementation

Changed the inferrer to accept a graph and made all of the changes to maintain the existing behaviour.

### Automated Tests

I had to skip a hover test while our RBS indexer doesn't handle methods. I'll use `[RUBYDEX]` in all skips so that we can easily find them later.

### DIFF
--- a/lib/ruby_lsp/type_inferrer.rb
+++ b/lib/ruby_lsp/type_inferrer.rb
@@ -5,9 +5,9 @@ module RubyLsp
   # A minimalistic type checker to try to resolve types that can be inferred without requiring a type system or
   # annotations
   class TypeInferrer
-    #: (RubyIndexer::Index index) -> void
-    def initialize(index)
-      @index = index
+    #: (Rubydex::Graph) -> void
+    def initialize(graph)
+      @graph = graph
     end
 
     #: (NodeContext node_context) -> Type?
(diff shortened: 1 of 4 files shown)


lib/ruby_lsp/type_inferrer.rb
```diff
@@ -121,11 +122,11 @@ def guess_type(raw_receiver, nesting)
         .map(&:capitalize)
         .join
 
-      entries = @index.resolve(guessed_name, nesting) || @index.first_unqualified_const(guessed_name)
-      name = entries&.first&.name
-      return unless name
+      declaration = @graph.resolve_constant(guessed_name, nesting)
+      declaration ||= @graph.search(guessed_name).first
+      return unless declaration
```
alexcrocha, line 127: I notice different semantics in the APIs here. I am not very familiar with this part of the code, but I am assuming the intent is to have a best effort guess at what the type is as at this point we were not able to resolve it.

The question I have is, will `@index.first_unqualified_const(guessed_name)` and `@graph.search(guessed_name).first` return the same result, or is the heuristic different?

vinistock: It's indeed a bit different. This https://github.com/Shopify/rubydex/pull/673 will allow us to match the existing behaviour.

vinistock: To avoid being blocked, I will ship this PR and update the code once https://github.com/Shopify/rubydex/pull/673 is shipped.

PR #4015, opened 2026-03-18
Partially migrate go to definition to use Rubydex

vinistock:
### Motivation

This PR partially migrates go to definition to use Rubydex. It ignores constants for now as we need the visibility API to complete it. Everything else is migrated.

### Implementation

There's not much that's notable other than changing to use the new APIs, but I'll highlight two things:

1. I always disliked our `Common` module. I propose that we instead patch Rubydex's classes with LSP related things so that we can do things like `definition.to_lsp_range` (see the code). It feels a lot more elegant and, as long as we include `lsp` in the method names, it won't really conflict with anything
2. The logic to find the correct graph declaration for go to definition, hover and signature help will probably be super similar. I think we can extract a common implementation for the 3, but I held back for now so that we can take a more complete look into how the requests use the graph

### Automated Tests

I migrated our tests. The only thing to call out here is that Rust's ordering does not match the original indexer's ordering (due to how `HashMap` works).

This makes no functional difference, but tests expect a very specific order to assert things. Instead of paying the price to sort the result, which is unnecessary, I just sorted in the tests.

### DIFF
--- a/lib/ruby_lsp/internal.rb
+++ b/lib/ruby_lsp/internal.rb
@@ -31,6 +31,9 @@ require "securerandom"
 require "shellwords"
 require "set"
 
+# Rubydex LSP additions
+require "ruby_lsp/rubydex/definition"
+
 require "ruby-lsp"
 require "ruby_lsp/base_server"
 require "ruby_indexer/ruby_indexer"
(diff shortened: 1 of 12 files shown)


st0012:
>  It ignores constants for now as we need the visibility API to complete it. Everything else is migrated.

Is the missing API for replacing `entry.private?`? If that's the case, can we do the full migration already and skip the test related to that specific check?

vinistock:
@st0012 done! Skipped the visibility test and migrated the request.

PR #4018, opened 2026-03-19
Fix document links for source comments above sig blocks

KaanOzkan:
### Motivation

In RBI files, `# source://` comments above `sig` blocks were not clickable because `extract_document_link` only checked the line immediately before the `def` node. However, when a `sig` block sits between the comment and the `def`, the comment is no longer in the line immediately before.

### Implementation

Register for `on_call_node_enter` to record source comments found above `sig` calls, keyed by the sig's end line. Then in `extract_document_link`, fall back to the stashed sig comment when no direct comment is found. ~It's only registered in RBI files to prevent overhead but still provide value for Tapioca's gem RBIs.~

### Automated Tests

Added

### Manual Tests

I played around with it locally, didn't notice new performance issues.

### DIFF
--- a/lib/ruby_lsp/listeners/document_link.rb
+++ b/lib/ruby_lsp/listeners/document_link.rb
@@ -60,6 +60,7 @@ module RubyLsp
         @lines_to_comments = comments.to_h do |comment|
           [comment.location.end_line, comment]
         end #: Hash[Integer, Prism::Comment]
+        @sig_comments = {} #: Hash[Integer, Prism::Comment]
 
         dispatcher.register(
           self,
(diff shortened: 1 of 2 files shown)


lib/ruby_lsp/listeners/document_link.rb
```diff
@@ -69,6 +70,20 @@ def initialize(response_builder, uri, comments, dispatcher)
           :on_constant_write_node_enter,
           :on_constant_path_write_node_enter,
         )
+
+        # Only register for call nodes in RBI files — the callback is currently only used to handle
+        # `# source://` comments above `sig` blocks, which mostly appear in RBI files.
+        dispatcher.register(self, :on_call_node_enter) if path&.end_with?(".rbi")
```
vinistock: I think we don't have to limit this to RBI only. If someone wanted to add a package link to a Ruby file, I think it would be fair to be able to jump.

vinistock (approved):
The fix is good, but I'd remove the RBI restriction

lib/ruby_lsp/listeners/document_link.rb
KaanOzkan: I was wary of the overhead but I think it's still negligible. removed the limit.

PR #4019, opened 2026-03-19
Migrate workspace symbol to use Rubydex

vinistock:
### Motivation

This PR migrates workspace symbol to use Rubydex.

### Implementation

1. Started using `Graph#search` to match symbols
2. After doing some research, I realized that no language server excludes private/protected symbols from the results. Additionally, there's a proposal open to include `visibility` as a new field in workspace symbols. I think excluding based on visibility was not the right choice, so I removed that filter
3. I added the `to_lsp_kind` to our definition patches

### Automated Tests

I removed a redundant test and expanded the one about private/protected definitions.

### DIFF
--- a/lib/ruby_lsp/requests/workspace_symbol.rb
+++ b/lib/ruby_lsp/requests/workspace_symbol.rb
@@ -12,54 +12,32 @@ module RubyLsp
       #: (GlobalState global_state, String? query) -> void
       def initialize(global_state, query)
         super()
-        @global_state = global_state
         @query = query
-        @index = global_state.index #: RubyIndexer::Index
+        @graph = global_state.graph #: Rubydex::Graph
       end
 
       # @override
       #: -> Array[Interface::WorkspaceSymbol]
       def perform
-        fuzzy_search.filter_map do |entry|
-          kind = kind_for_entry(entry)
-          loc = entry.location
+        response = []
 
-          # We use the namespace as the container name, but we also use the full name as the regular name. The reason we
-          # do this is to allow people to search for fully qualified names (e.g.: `Foo::Bar`). If we only included the
-          # short name `Bar`, then searching for `Foo::Bar` would not return any results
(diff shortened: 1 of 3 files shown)


lib/ruby_lsp/requests/workspace_symbol.rb
```diff
@@ -12,54 +12,32 @@ class WorkspaceSymbol < Request
       #: (GlobalState global_state, String? query) -> void
       def initialize(global_state, query)
         super()
-        @global_state = global_state
         @query = query
-        @index = global_state.index #: RubyIndexer::Index
+        @graph = global_state.graph #: Rubydex::Graph
```
st0012, line 16: Since we don't need `@global_state` anymore, does it make sense to just pass `graph` in?

lib/ruby_lsp/requests/workspace_symbol.rb
```diff
@@ -12,54 +12,32 @@ class WorkspaceSymbol < Request
...
-            ),
-          )
-        end
-      end
-
-      private
+          declaration.definitions.each do |definition|
+            location = definition.location
+            uri = URI(location.uri)
+            file_path = uri.full_path
```
st0012, line 30: Not blocking, but does it make sense to make this `Definition#full_path`?

lib/ruby_lsp/requests/workspace_symbol.rb
vinistock: I would still keep all requests receiving the global state in case they need anything.

lib/ruby_lsp/requests/workspace_symbol.rb
vinistock: I think we should have Rubydex expose its own URI abstraction, so that `location.uri` returns an object instead of a string.

Something that would allow us to do `location.uri.to_file_path`. I've been treating that as an improvement for later though.

PR #4020, opened 2026-03-19
Unexpected Errno::ENAMETOOLONG on Windows

rellampec:
### Motivation

The ruby built (ruby executable path: `C:\ruby\Ruby32-x64\bin\ruby.exe`) of [RubyInstaller](https://rubyinstaller.org/) (for Windows machines) triggers an error on `Dir.glob` for wide inclusive patterns.

```ruby
Dir.glob("*/**/*")

# => <internal:dir>:ln:in `glob': Filename too long - frontends/webapp-frontend/src/features/featureMangement/features/featureTypes/components/FeatureTypeDraggableField/components/AreaConfiguration/components/AreaFilters/components/AreaFiltersDrawer/components/AreaByLocationPreview/hooks (Errno::ENAMETOOLONG)
```
- Apparently, their `x64` version hasn't been compiled by taking into account what follows further down.

In `ruby-lsp` this translates into:

1. Failed `index_all` ([ln](https://github.com/Shopify/ruby-lsp/blob/main/lib/ruby_indexer/lib/ruby_indexer/configuration.rb#L71): `Dir.glob(File.join(@workspace_path, pattern), flags)`).
2. Failed `load_addons` ([ln](https://github.com/Shopify/ruby-lsp/blob/main/lib/ruby_lsp/addon.rb#L59): `Dir.glob("#{global_state.workspace_path}/**/ruby_lsp/**/addon.rb")`).

Windows has a registry entry, under the path `Computer\HKEY_LOCAL_MACHINE\SYSTEM\CurrentControlSet\Control\FileSystem`, called `LongPathsEnabled` to _enabling support for file **paths over 260 characters**_, that defaults to `0` , but that can be changed to `1`.

There have been a couple of bugs reported in ruby-core:
- 2015 - [Bug #12551: Exception accessing file with long path on windows](https://bugs.ruby-lang.org/issues/12551) (Closed)
- 2022 - [Bug #18947: Unexpected Errno::ENAMETOOLONG on Windows](https://bugs.ruby-lang.org/issues/18947) (Open)

And one fix has been merged: [[Win32] long path name support [Bug #12551]#4505](https://github.com/ruby/ruby/pull/4505) ([article](https://gabusc.us/posts/windows-ruby-and-long-paths/) of the author of the fix).

However, the fix can only make its way to the executable if the Windows application manifest that is embedded to it during the built has this entry:

```xml
<ws2:longPathAware>true</ws2:longPathAware>
```

None of the ruby installer repos ([old](https://github.com/oneclick/rubyinstaller/issues?q=is%3Aissue%20LongPathsEnabled) or [new](https://github.com/oneclick/rubyinstaller2/issues?q=is%3Aissue%20LongPathsEnabled)) has any reported issue on this matter.

On the other hand, although `ruby-lsp` allows to declare an indexation configuration parameter with this settings key `rubyLsp.indexing: {excludedPatterns: []}`, excluding the root folder of the deepest part of the working directory does not prevent the error to popup. Although **such an error shouldn't be reason for `ruby-lsp` to stop indexing**, it does stop indexing.

[Vs Code extension Settings](https://shopify.github.io/ruby-lsp/vscode-extension#indexing-configuration):

```json
{
  "rubyLsp.indexing": {
    "excludedPatterns": [
      "**/test/**/*.rb",
      "frontends/**/*.rb",
      "vendor/**/*",
      ".bundle/**/*",
      "gitlab-ci/**/*",
      "deb/**/*",
      "graphql-docs/**/",
      "deployment/**/*"
    ]
}
```
- Observe that `"frontends/**/*.rb"` is excluded, and nevertheless the indexer ended up failing.

### Implementation

My changes involve all the below in order of priority:

1. More error logging (`Ruby LSP output`) on indexation and addons loading.
2. Let `load_addons` and `index_all` continue by logging any error when it occurs (instead of completely failing). This is implemented with error handling and logging.
3. Decouple/granulate `glob` on `load_addons` provided that it globs on per main space/root-folder basis (same approach as used in the indexer with `top_level_directories` ([line](https://github.com/Shopify/ruby-lsp/blob/087c505c5edc0c626f3f3b055b018f3ac922498b/lib/ruby_indexer/lib/ruby_indexer/configuration.rb#L42))
4. Ensure that patterns excluded from indexation aren't included during the scoping of included urls. This has been implemented decoupling top level directories in the `included_patterns` variable, and subtracting from the `Array` the `excluded_patterns`, before the `Dir.glob` is run.

### Automated Tests

I haven't, because I do not know how to do it, and I do not know if the changes I implemented are really part of the test suit.

### Manual Tests

1. From a local repo with the fork as origin, I implemented changes and pushed them to the remote.
2. With an aside folder (not in the working directory) for the `ruby-lsp` _Gemfile_, I pointed the `ruby-lsp` gem to the feature branch of this PR.
3. The VS Code is configured to use that Gemfile to bundle ruby-lsp.
4. I have also re-installed multiple times the gem that I built locally.
5. On each change, I used `reload window` on VS Code, and checked the `view > output` terminal.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -7,7 +7,6 @@ gemspec
 gem "bundler", "~> 4.0.0"
 gem "minitest", "~> 6.0"
 gem "test-unit", require: false
-gem "byebug"
 
 group :development do
   gem "debug", "~> 1.9", require: false
--- a/Gemfile
+++ b/Gemfile
@@ -7,6 +7,7 @@ gemspec
 gem "bundler", "~> 4.0.0"
 gem "minitest", "~> 6.0"
 gem "test-unit", require: false
+gem "byebug"
 
 group :development do
   gem "debug", "~> 1.9", require: false
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -11,8 +11,6 @@ GEM
   specs:
     ast (2.4.3)
     benchmark (0.5.0)
-    byebug (13.0.0)
-      reline (>= 0.6.0)
     date (3.5.1)
     debug (1.11.1)
       irb (~> 1.10)
@@ -143,7 +141,6 @@ PLATFORMS
 
 DEPENDENCIES
   bundler (~> 4.0.0)
-  byebug
   debug (~> 1.9)
   minitest (~> 6.0)
   mocha (~> 2.3)
@@ -163,7 +160,6 @@ DEPENDENCIES
 CHECKSUMS
   ast (2.4.3) sha256=954615157c1d6a382bc27d690d973195e79db7f55e9765ac7c481c60bdb4d383
   benchmark (0.5.0) sha256=465df122341aedcb81a2a24b4d3bd19b6c67c1530713fd533f3ff034e419236c
-  byebug (13.0.0) sha256=d2263efe751941ca520fa29744b71972d39cbc41839496706f5d9b22e92ae05d
   date (3.5.1) sha256=750d06384d7b9c15d562c76291407d89e368dda4d4fff957eb94962d325a0dc0
   debug (1.11.1) sha256=2e0b0ac6119f2207a6f8ac7d4a73ca8eb4e440f64da0a3136c30343146e952b6
   erb (6.0.1) sha256=28ecdd99c5472aebd5674d6061e3c6b0a45c049578b071e5a52c2a7f13c197e5
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -11,6 +11,8 @@ GEM
   specs:
     ast (2.4.3)
     benchmark (0.5.0)
+    byebug (13.0.0)
+      reline (>= 0.6.0)
     date (3.5.1)
     debug (1.11.1)
       irb (~> 1.10)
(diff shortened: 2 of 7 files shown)


vinistock:
Thank you for the PR. We are in the middle of adopting our new code indexer Rubydex, which completely replaces all of the parts of this PR and takes a different approach to listing files that doesn't go through glob. The new implementation also collects errors rather than stopping indexing.

It would be nice to re-test once the version with the new indexer is shipped. I believe the problem will be fixed.

rellampec:
@vinistock and the addons screening? would the new feature also solve the issue?

vinistock:
Sorry, indeed the add-ons will still be a problem. However, I'm lacking context to understand what is the right layer for the fix.

Is this a legitimate problem that may happen depending on the user's Windows machine or is this a Ruby bug that has not yet been fixed?

If it's the former, is it possible to write a test that fails on CI so that we don't regress?

rellampec:
>If it's the former, is it possible to write a test that fails on CI so that we don't regress?

@vinistock to my view it is a ruby installer bug ([RubyInstaller](https://rubyinstaller.org/)).

Not sure how to add the test on CI, but the description of this FR gives an example.

Do you need anything from my end?

Issue #3842, opened 2025-11-26
`textDocument/definition` returns duplicate results with inconsistent file paths


timvg80:
### Ruby LSP Information

**Editor:** Neovim 0.11.5  
**Ruby LSP version:** 0.26.4  
**Ruby version:** `ruby 3.4.2 (2025-02-15 revision d2930f8e7a) +PRISM [arm64-darwin24]`  
**Operating System:** macOS 15.6.1
**LSP client:** nvim-lspconfig

**Note:** This issue occurs in Neovim but not in VS Code.

### Reproduction steps

1. Open any Ruby file in Neovim with ruby-lsp configured
2. Place cursor on a module/class name
3. Enable debug logging with `:lua vim.lsp.set_log_level("debug")`
4. Trigger go-to-definition using `:lua vim.lsp.buf.definition()`
5. Check `:LspLog` after triggering definition

**Expected behavior:** Single definition result with correct file path

**Actual behavior:** Two identical definition results where the only difference is the file URI - one has a double slash (`//`) and one has a single slash (`/`)

**Code snippet or error message**

LSP log showing the duplicate response:
```
[DEBUG][2025-11-26 18:45:36] ...m/lsp/client.lua:674    "LSP[ruby_lsp]"    "client.request"    1    "textDocument/definition"    { position = { character = 10, line = 7 }, textDocument = { uri = "file:///Users/<path-to-directory>/<ruby-file>.rb" } }    <function 1>    5
[DEBUG][2025-11-26 18:45:36] .../vim/lsp/rpc.lua:277    "rpc.send"    { id = 11, jsonrpc = "2.0", method = "textDocument/definition", params = { position = { character = 10, line = 7 }, textDocument = { uri = "file:///Users/<path-to-directory>/<ruby-file>.rb" } } }
[...]

Issue #896, opened 2023-08-16
Explore further standardizing URI handling


vinistock:
Explore if it would be possible to completely stop using file paths and only deal in terms of URIs everywhere in the LSP.

Maybe our own `URI` class to abstract some of the stuff we need for the LSP. One thing we could do is using only the `scheme`, `path` and `opaque` parts for implementing `hash`, so that URI's with the same components always land on the same key in hashes and we don't need to use `to_s`.

PR #4022, opened 2026-03-20
Escape URIs like the editor does and harden Dir.glob against metacharacters in paths

### LINKED ISSUE #3503: Linux file system with square brackets in the directory/filename are failing to be indexed
https://github.com/Shopify/ruby-lsp/blame/6acf78b6f5e94c8fd13be5ad7a4c82d083233853/lib/ruby_indexer/lib/ruby_indexer/uri.rb#L21

The unsafe_regexp includes square brackets and I'm finding this is causing the ruby-lsp indexing to fail when it comes across a file path like the following "/home/streetp/index[0]/example.rb".

When I modify the ruby_indexer code and change the unsafe_regexp, removing the square brackets, to be ```unsafe_regex = %r{[^\-_.!~*'()a-zA-Z\d;/?@&=+$,]}```

Then it works. I haven't checked tested this for Windows/Mac - and I'm not aware of whether this may create issues but I know that I have these work filepaths on my system (and these generated filepaths that break ruby-lsp indexing are not things I can control and change).

### LINKED ISSUE #4023: Bracket/brace filenames break all LSP features
## Description

Files with `[`, `]`, `{`, or `}` in their names (e.g., `[id].rb`, `{slug}.rb`) break multiple ruby-lsp features. These filename patterns are common in Rails/Next.js routing conventions.

When such a file exists in the workspace, **all** LSP features fail for it — indexing, go-to-definition, references, hover, completion, diagnostics, code lens, etc.

## Root cause

Two independent issues compound:

1. **URI encoding mismatch**: `URI::Generic.from_path` includes `[]` in its safe character set, so brackets are left unescaped. However, VS Code's `vscode-uri` always encodes them (`[` → `%5B`, `]` → `%5D`). This mismatch means the server's URI for a file doesn't match the client's URI, so the document store can't find it.

2. **Glob metacharacters in paths**: `Dir.glob` treats `[]` as character classes and `{}` as alternation groups. When file or workspace paths containing these characters are interpolated into glob patterns (e.g., `Dir.glob(File.join(workspace_path, "**/*.rb"))`), the glob interprets them as pattern syntax rather than literal characters.

## Reproduction

1. Create a file named `[id].rb` in a Ruby project
2. Open the project in VS Code with ruby-lsp
3. Open `[id].rb` — observe that no LSP features work (no syntax highlighting from semantic tokens, no go-to-definition, no hover, etc.)

## Fix

PR #4022

andriytyurnikov:
Fixes #3503. Contributes to #3639.

## The problem

The server and the editor have to agree on how a file is spelled. A URI identifies a resource, so if indexing builds `file:///a/%5Bid%5D.rb` for a file the editor calls `file:///a/[id].rb`, the same file is tracked under two identities: indexed twice, reported twice, and never resolved to the document the editor has open.

There are two independent ways that agreement breaks today, and both are fixed here.

**1. Our escaping did not match the editor's.** `URI::Generic.from_path` used the RFC 2396 safe set. VS Code builds its URIs with `vscode-uri`, which only leaves the RFC 3986 unreserved set (`A-Za-z0-9-._~`) plus `/` unescaped. Checked against the package directly, **12 of 17 sample paths disagreed** — not just on brackets, but on `(`, `)`, `!`, `'`, `*`, `$`, `&`, `+`, `,`, `;`, `=` and `@`. Every one of those is a standing source of the duplication tracked in #3639.

**2. Paths were interpolated into glob patterns.** `Dir.glob` reads `[id]` as a character class and `{a,b}` as an alternation, so a workspace or file path containing them silently matches nothing.

When both bite, every feature fails for the affected file: indexing, definition, references, hover, completion, diagnostics, code lens. The reporter in #3503 hit it on a generated path they could not control (`/home/streetp/index[0]/example.rb`); the names are also routine under routing conventions (`app/models/[id].rb`, `{slug}.rb`).

## Related reports

This is one failure class with several faces. Fixing the spelling mismatch is what they have in common:

| Issue | Symptom |
|---|---|
| #3503 | Bracketed paths fail to index |
| #3639 | Duplicate entries when a `didOpen` races indexing |
| #3842 | Duplicate definitions in Neovim, `//` vs `/` |
| #3285 | Duplicate definitions on Windows, `c%3A` vs `C%3A` |
| #4024 | A bracketed workspace root indexes zero files |

#896 sketches the structural end state — URI identity based on components rather than on the exact string. This PR does not go there, but narrowing the encoding is a prerequisite either way: component-based equality still needs the components to agree.

### URI escaping

Narrowing the safe set to `[^\-_.~a-zA-Z\d/]` brings the sample from 12 of 17 disagreeing to **17 of 17 matching**, with every round trip intact. It also fixes `from_path` raising `URI::InvalidComponentError` for a path containing `?`, which used to be treated as safe and produced a string `URI::Generic.build` rejected.

### Glob hardening

- `Dir.glob`'s `base:` parameter in `references.rb`, `rename.rb`, `addon.rb`, `test_style.rb` and `go_to_relevant_file.rb`, so workspace and file paths are passed as literal paths rather than as patterns. This also works on Windows, where backslash escaping is unavailable.
- `RubyLsp.escape_glob_metacharacters` for user-provided content interpolated into a pattern: `require_relative` content in the completion listener, and file names in `go_to_relevant_file.rb`. The comma is escaped too, because `complete_require_relative` interpolates into a brace group where an unescaped comma splits the alternation.

### Expectation test harness

Adding a bracketed fixture surfaced a pre-existing bug. Seven expectation test classes built their URI as `URI("file://#{@_path}")` from a **relative** path, so `file://test/fixtures/foo.rb` parsed `test` as the URI authority and the path came out as `/fixtures/foo.rb` — the `test/` segment silently dropped.

`CodeLens` appends `-Itest` when the path matches `**/test/**/*`, so with that segment missing **five committed expectations asserted commands without `-Itest`**, the opposite of what a user gets for a file genuinely under `test/`. Those are regenerated here; across all 324 modified lines the only changes are the restored `/test/` segment and the flag it enables.

The runner also now looks expectation files up with `File.file?` instead of a glob, and sanitises fixture names when generating test method names.

## Tests

Every test below was verified to **fail when its fix is reverted**, not merely to pass with it.

| Test | Covers |
|---|---|
| `uri_test` — parity, unreserved characters, `?` | escaping matches `vscode-uri` |
| `store_test` — client-encoded URI vs server-built URI | the reported failure |
| `addon_test` — add-ons under `[id]/{slug}/` | `addon.rb` (escaping and glob in one path) |
| `references_test`, `rename_test` — bracketed workspace | `references.rb`, `rename.rb` |
| `resolve_test_commands_test` — real directory, unstubbed | `test_style.rb` `test_dir` expansion |
| `go_to_relevant_file_test` ×3 | file name either direction, and parent directory |
| `completion_test` ×2 | brackets and commas in `require_relative` |
| `test/fixtures/[id].rb` | the expectations runner |

The pre-existing `test_dir` tests stub `Dir.glob`, which bypasses the `base:` behaviour under test, so the new one hits the real file system. Without the sanitised method name, the fixture makes every expectation suite fail to parse with `def test_hover__[id]__does_not_raise`.

## Deliberately not included

**Windows drive letter case.** `vscode-uri` normalises the drive to lower case regardless of input; `from_path` preserves whatever it is given, so indexing (upper case, from `Dir.pwd`) and the editor (lower case) still disagree. That is #3285, which is currently closed but only half fixed — evidence and a repro are [in this comment](https://github.com/Shopify/ruby-lsp/issues/3285#issuecomment-5148209336). It is left out because two committed tests assert the current behaviour, and `server.rb` has case-sensitive `path.start_with?(@global_state.workspace_path)` guards that would have to change at the same time. Normalising only `from_path` would trade one Windows bug for another.

**Bracketed workspace roots.** `configuration.rb` still interpolates the workspace path, so a bracketed **root** indexes zero files. Converting it to `base:` triggers a `Dir.glob` bug: when `top_level_directories` is empty the pattern becomes `{}/**/*.rb`, and with `base:` that escapes the base and walks the file system from `/` — reproduced on Ruby 4.0.1 as `Errno::EPERM` on `/Library/Application Support/com.apple.TCC`. Tracked in #4024, which was auto-closed as stale and should be reopened. A follow-up branch fixes it by guarding the empty case.

**`//` normalisation** (#3842) is not addressed; this only covers URIs we build.

## Test plan

- [x] `bundle exec rake` — 330 + 18,565 runs, 0 failures, 0 errors
- [x] `bundle exec srb tc` — no errors
- [x] `bin/rubocop` — no offenses
- [x] `bin/tapioca check-shims` — no duplicates
- [x] Escaping compared against the `vscode-uri` package directly — 17/17

### DIFF
--- a/lib/ruby_indexer/test/uri_test.rb
+++ b/lib/ruby_indexer/test/uri_test.rb
@@ -81,5 +81,100 @@ module RubyIndexer
       assert_equal(path, uri.to_standardized_path)
       assert_equal("file:///path/with/unicode/%E6%96%87%E4%BB%B6.rb", uri.to_s)
     end
+
+    def test_from_path_with_brackets
+      uri = URI::Generic.from_path(path: "/some/path/[id].rb")
+      assert_equal("file:///some/path/%5Bid%5D.rb", uri.to_s)
+    end
+
+    def test_from_path_with_braces
+      uri = URI::Generic.from_path(path: "/some/path/{slug}.rb")
+      assert_equal("file:///some/path/%7Bslug%7D.rb", uri.to_s)
+    end
+
+    def test_round_trip_with_brackets
+      path = "/some/path/[id].rb"
+      uri = URI::Generic.from_path(path: path)
+      assert_equal(path, uri.to_standardized_path)
+    end
+
+    def test_round_trip_with_braces
+      path = "/some/path/{slug}.rb"
+      uri = URI::Generic.from_path(path: path)
+      assert_equal(path, uri.to_standardized_path)
+    end
+
+    def test_from_path_with_parentheses
+      uri = URI::Generic.from_path(path: "/some/path/(id).rb")
+      assert_equal("file:///some/path/%28id%29.rb", uri.to_s)
+    end
+
+    def test_round_trip_with_parentheses
+      path = "/some/path/(id).rb"
+      uri = URI::Generic.from_path(path: path)
+      assert_equal(path, uri.to_standardized_path)
+    end
+
+    def test_round_trip_with_spaces_inside_brackets
+      path = "/some/path/[id page].rb"
+      uri = URI::Generic.from_path(path: path)
+      assert_equal(path, uri.to_standardized_path)
+      assert_equal("file:///some/path/%5Bid%20page%5D.rb", uri.to_s)
+    end
+
+    def test_round_trip_with_spaces_inside_braces
+      path = "/some/path/{slug name}.rb"
+      uri = URI::Generic.from_path(path: path)
+      assert_equal(path, uri.to_standardized_path)
+      assert_equal("file:///some/path/%7Bslug%20name%7D.rb", uri.to_s)
+    end
+
+    def test_round_trip_with_spaces_inside_parentheses
+      path = "/some/path/file (copy).rb"
+      uri = URI::Generic.from_path(path: path)
+      assert_equal(path, uri.to_standardized_path)
+      assert_equal("file:///some/path/file%20%28copy%29.rb", uri.to_s)
+    end
+
+    # The editor and the server must escape paths identically, otherwise the same file is tracked under two different
+    # URIs and the index ends up with duplicate entries (Shopify/ruby-lsp#3639). These expectations were captured from
+    # the `vscode-uri` package, which is what VS Code uses to build the URIs it sends us
+    def test_from_path_escapes_every_character_the_editor_escapes
+      {
+        "/some/path/[id].rb" => "file:///some/path/%5Bid%5D.rb",
+        "/some/path/{slug}.rb" => "file:///some/path/%7Bslug%7D.rb",
+        "/some/path/(id).rb" => "file:///some/path/%28id%29.rb",
+        "/some/path/a&b.rb" => "file:///some/path/a%26b.rb",
+        "/some/path/a,b.rb" => "file:///some/path/a%2Cb.rb",
+        "/some/path/a+b.rb" => "file:///some/path/a%2Bb.rb",
+        "/some/path/a=b.rb" => "file:///some/path/a%3Db.rb",
+        "/some/path/a@b.rb" => "file:///some/path/a%40b.rb",
+        "/some/path/a;b.rb" => "file:///some/path/a%3Bb.rb",
+        "/some/path/a!b.rb" => "file:///some/path/a%21b.rb",
+        "/some/path/a*b.rb" => "file:///some/path/a%2Ab.rb",
+        "/some/path/a$b.rb" => "file:///some/path/a%24b.rb",
+        "/some/path/it's.rb" => "file:///some/path/it%27s.rb",
+        "/some/path/with space.rb" => "file:///some/path/with%20space.rb",
+      }.each do |path, expected|
+        uri = URI::Generic.from_path(path: path)
+        assert_equal(expected, uri.to_s)
+        assert_equal(path, uri.to_standardized_path)
+      end
+    end
+
+    def test_from_path_does_not_escape_unreserved_characters
+      path = "/some/path/a-b_c.d~e.rb"
+      uri = URI::Generic.from_path(path: path)
+      assert_equal("file:///some/path/a-b_c.d~e.rb", uri.to_s)
+      assert_equal(path, uri.to_standardized_path)
+    end
+
+    def test_from_path_with_question_mark
+      # A question mark used to be treated as safe, which made `URI::Generic.build` reject the escaped path for not
+      # being a valid absolute path component
+      uri = URI::Generic.from_path(path: "/some/path/what?.rb")
+      assert_equal("file:///some/path/what%3F.rb", uri.to_s)
+      assert_equal("/some/path/what?.rb", uri.to_standardized_path)
+    end
   end
 end
--- a/lib/ruby_lsp/listeners/completion.rb
+++ b/lib/ruby_lsp/listeners/completion.rb
@@ -457,10 +457,11 @@ module RubyLsp
         # if the path is not a directory, glob all possible next characters
         # for example ../somethi| (where | is the cursor position)
         # should find files for ../somethi*/
+        escaped_content = RubyLsp.escape_glob_metacharacters(content)
         path_query = if content.end_with?("/") || content.empty?
-          "#{content}**/*.rb"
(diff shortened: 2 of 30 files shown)


RoyLeviGit (approved):
Did not run the lsp with this change but took a close look and ran AI diagnostics on it and it looks good to me. Thank you for solving this issue.

lib/ruby_indexer/test/uri_test.rb
```diff
@@ -81,5 +81,64 @@ def test_from_path_with_unicode_characters
       assert_equal(path, uri.to_standardized_path)
       assert_equal("file:///path/with/unicode/%E6%96%87%E4%BB%B6.rb", uri.to_s)
     end
+
+    def test_from_path_with_brackets
+      uri = URI::Generic.from_path(path: "/some/path/[id].rb")
+      assert_match("%5B", uri.path)
+      assert_match("%5D", uri.path)
```
rafaelfranca:
```suggestion
```

lib/ruby_indexer/test/uri_test.rb
```diff
@@ -81,5 +81,64 @@ def test_from_path_with_unicode_characters
...
+      assert_equal("file:///some/path/%5Bid%5D.rb", uri.to_s)
+    end
+
+    def test_from_path_with_braces
+      uri = URI::Generic.from_path(path: "/some/path/{slug}.rb")
+      assert_match("%7B", uri.path)
+      assert_match("%7D", uri.path)
```
rafaelfranca:
```suggestion
```

lib/ruby_lsp/listeners/completion.rb
```diff
@@ -457,10 +457,11 @@ def complete_require_relative(node)
         # if the path is not a directory, glob all possible next characters
         # for example ../somethi| (where | is the cursor position)
         # should find files for ../somethi*/
+        escaped_content = content.gsub(/[\[\]{}*?\\]/) { |c| "\\#{c}" }
```
rafaelfranca: We should move this to a helper and reuse in the other place

lib/ruby_indexer/test/uri_test.rb
andriytyurnikov: @rafaelfranca huh? looks same to me

lib/ruby_indexer/test/uri_test.rb
andriytyurnikov: @rafaelfranca huh? looks same to me

lib/ruby_indexer/test/uri_test.rb
rafaelfranca: I'm asking those lines to be removed. There is no need for them when the assertion below is already asserting it.

lib/ruby_lsp/listeners/completion.rb
andriytyurnikov: @rafaelfranca up yours, I am not as familiar with project structure though

lib/ruby_indexer/test/uri_test.rb
andriytyurnikov: Thanks, that makes sense — the `assert_equal` on the full URI string already covers the encoding. Removed the redundant `assert_match` lines in both tests.

lib/ruby_lsp/listeners/completion.rb
andriytyurnikov: Good call! Extracted `escape_glob_metacharacters` as a module-level method on `RubyLsp` in `utils.rb`, and updated both `completion.rb` and `go_to_relevant_file.rb` to use it.

RoyLeviGit (approved):
Same as before, did not explicitly run this change but ran even more AI diagnostics on it including letting the AI suggest concerns and look at rest of repo for comparison and it can out approving PR and so will I :)

andriytyurnikov:
@rafaelfranca - rebased successfully

RoyLeviGit (approved):
Rebase

andriytyurnikov:
Rebased onto main; `addon.rb` conflicted with #4178. Escaping now matches `vscode-uri` exactly — 12 of 17 paths previously disagreed on `( ) & , + = @ ; ' ! * $`, not just brackets, each a standing source of #3639 duplicates. Every hardened call site now has a regression test, verified to fail when its fix is reverted. Rationale, plus a harness fix touching committed expectations, is in the rewritten description.

PR #4029, opened 2026-03-24
Partially migrate hover to use Rubydex

vinistock:
### Motivation

Partially migrate hover to use Rubydex (we're missing the Ruby API for method signatures and visibility to finish).

### Implementation

Very similar to go to definition, just switched to use the graph where applicable. I will continue to resist the urge to refactor pieces into a shared API until we have more requests and can extract a proper abstraction.

### Automated Tests

As we're doing the migration, I'm also taking the opportunity to stop indexing all RBS core definitions when it's not really necessary. Request tests shouldn't be verifying indexing logic, so we just need to ensure the request does the right thing.

The benefit is that our tests should be faster.

### DIFF
--- a/lib/ruby_lsp/requests/support/common.rb
+++ b/lib/ruby_lsp/requests/support/common.rb
@@ -64,6 +64,46 @@ module RubyLsp
           receiver.nil? || receiver.is_a?(Prism::SelfNode)
         end
 
+        #: (String, Enumerable[Rubydex::Definition], ?Integer?) -> Hash[Symbol, String]
+        def categorized_markdown_from_definitions(title, definitions, max_entries = nil)
+          markdown_title = "```ruby\n#{title}\n```"
+          file_links = []
+          content = +""
+          defs = max_entries ? definitions.take(max_entries) : definitions
+          defs.each do |definition|
+            # For Markdown links, we need 1 based display locations
+            loc = definition.location.to_display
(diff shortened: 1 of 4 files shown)


lib/ruby_lsp/requests/support/common.rb
```diff
@@ -64,6 +64,44 @@ def self_receiver?(node)
...
+            end
+
+            # The format for VS Code file URIs is `file:///path/to/file.rb#Lstart_line,start_column-end_line,end_column`
+            string_uri = "#{loc.uri}#L#{loc.start_line},#{loc.start_column}-#{loc.end_line},#{loc.end_column}"
+            file_links << "[#{file_name}](#{string_uri})"
+            content << "\n\n#{definition.comments.map { |comment| comment.string.delete_prefix("# ") }.join("\n")}" unless definition.comments.empty?
+          end
+
+          additional_entries_text = if max_entries && defs.length > max_entries
+            additional = defs.length - max_entries
```
alexcrocha: When `max_entries` is not nil, `defs.length` will always be the same as `max_entries` since:
`defs = max_entries ? definitions.take(max_entries) : definitions`

```suggestion
          additional_entries_text = if max_entries && definitions.count > max_entries
            additional = definitions.count - max_entries
```

vinistock: Good catch!

PR #4032, opened 2026-03-26
Update graph on edits

vinistock:
### Motivation

Start updating the graph when incremental edits happen. This PR handles both in-memory modifications (`textDocument/didChange`) and changes committed to disk (`workspace/didChangeWatchedFiles`).

### Implementation

Some key decisions:

1. Incremental indexing and resolution now happens immediately for in-memory edits. We received feedback from several users that not every editor fires all feature requests and so attaching indexing to them is not adequate (#3384). Additionally, Rubydex incremental indexing and resolution should be fast enough that we can do it every time
2. I removed `Store#push_edits`. That method didn't add much and I needed the document instance to access the source after modifications
3. Watched file notifications always send a string URI for the `change[:uri]`, so I ensured all of our tests accurately reflected that


lib/ruby_lsp/server.rb
```diff
@@ -414,8 +414,18 @@ def text_document_did_close(message)
     def text_document_did_change(message)
       params = message[:params]
       text_document = params[:textDocument]
+      string_uri = text_document[:uri].to_s
 
-      @store.push_edits(uri: text_document[:uri], edits: params[:contentChanges], version: text_document[:version])
+      document = @store.get(string_uri)
```
alexcrocha: nit: I know we're working with `untyped` here, but since `Store#get` signature is `(URI::Generic uri)` this doesn't feel right 🫣

test/server_test.rb
```diff
@@ -483,7 +483,7 @@ def test_did_change_watched_files_does_not_fail_for_non_existing_files
       params: {
         changes: [
           {
-            uri: URI::Generic.from_path(path: File.join(Dir.pwd, "lib", "non_existing.rb")).to_s,
+            uri: URI::Generic.from_path(path: File.join(Dir.pwd, "lib", "non_existing.rb").to_s),
```
alexcrocha:
```suggestion
            uri: URI::Generic.from_path(path: File.join(Dir.pwd, "lib", "non_existing.rb")).to_s,
```
Looks to me like it was correct before

lib/ruby_lsp/server.rb
vinistock: That was my mistake, we should indeed pass a `URI::Generic` here.

test/server_test.rb
vinistock: Yes, indeed. Fixed it!

lib/ruby_lsp/server.rb
```diff
@@ -415,7 +415,16 @@ def text_document_did_change(message)
       params = message[:params]
       text_document = params[:textDocument]
 
-      @store.push_edits(uri: text_document[:uri], edits: params[:contentChanges], version: text_document[:version])
+      document = @store.get(text_document[:uri])
+      document.push_edits(params[:contentChanges], version: text_document[:version])
+
+      language_id = document.language_id
+
+      if [:ruby, :rbs].include?(language_id)
```
st0012, line 423: Maybe Rubydex should expose a list of languages and extensions it supports?

lib/ruby_lsp/server.rb
```diff
@@ -1033,6 +1042,23 @@ def workspace_did_change_watched_files(message)
...
+      graph = @global_state.graph
+
+      # Handle deletions and accumulate additions and changes for indexing
+      additions_and_changes = changes.each_with_object([]) do |change, acc|
+        if change[:type] == Constant::FileChangeType::DELETED
+          graph.delete_document(change[:uri])
+        else
+          path = URI(change[:uri]).to_standardized_path
+          next if path.nil?
+          next unless File.directory?(path) || [".rb", ".rbs"].include?(File.extname(path))
```
st0012, line 1054: Same as the above. I think this could be something rubydex exposes.

lib/ruby_lsp/server.rb
vinistock: I would delay this decision a bit for us to understand a few things:

- Right now, we could start indexing ERB files since the Ruby LSP separates the Ruby content (so we'd treat it as just a plain Ruby file). Is that the right approach (flexibility) or should we build first-class support for ERB in Rubydex?
- Ruby is super conducive to creating DSL files with custom or no extensions (`.ru`, `Procfile`, `Rakefile`...). Trying to list everything on the toolkit side rather than on the client side might not be the right choice considering the amount of possibilities

In short, I think we need to spend a bit more thinking about this before committing to an approach. For now, let's move forward with the client choosing what to index.

PR #4033, opened 2026-03-26
Migrate rename to use Rubydex

vinistock:
### Motivation

Migrate rename to use Rubydex. We currently only support constants, but we can easily expand to global, class and instance variables once we finish adding indexing for them.

### Implementation

The main changes are actually simplifications. We don't need the `ReferenceFinder` because Rubydex tracks references. We can just access it directly, which makes the implementation more straight forward.

### Automated Tests

The rename tests were pretty messy and difficult to write. I refactored the tests a bit and added a couple more scenarios.

### DIFF
--- a/lib/ruby_lsp/requests/rename.rb
+++ b/lib/ruby_lsp/requests/rename.rb
@@ -22,6 +22,7 @@ module RubyLsp
       def initialize(global_state, store, document, params)
         super()
         @global_state = global_state
+        @graph = global_state.graph #: Rubydex::Graph
         @store = store
         @document = document
         @position = params[:position] #: Hash[Symbol, Integer]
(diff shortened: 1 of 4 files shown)


lib/ruby_lsp/requests/rename.rb
```diff
@@ -97,80 +95,76 @@ def collect_file_renames(fully_qualified_name, document_changes)
         # We also look for an associated test file and rename it too
         short_name = fully_qualified_name.split("::").last #: as !nil
 
-        @global_state.index[fully_qualified_name]&.each do |entry|
+        declaration = @graph[fully_qualified_name]
+        return unless declaration
```
alexcrocha: Should we pass the declaration itself to `collect_file_renames()` and avoid re-fetching it here?

vinistock: Yes, good idea!

PR #4040, opened 2026-03-31
Return MethodNotFound (-32601) for unrecognized LSP requests

### LINKED ISSUE #3981: Unrecognized LSP requests are silently dropped instead of returning MethodNotFound (-32601)
The `process_message` case statement in `lib/ruby_lsp/server.rb` has no `else` clause, so when the server receives a request for an unsupported method (e.g. `textDocument/prepareCallHierarchy`), it matches no branch, returns nil, and never sends a response. The LSP spec requires servers to respond to any request bearing an `id` — for unsupported methods, that's error code `-32601` (MethodNotFound). The silent drop causes well-behaved clients to hang indefinitely waiting for a response. Adding an `else` branch that calls `send_message(Error.new(id: message[:id], code: -32601, message: "Method not found: #{message[:method]}"))` when `message[:id]` is present would fix it.

IgorFroehner:
### Motivation

Fixes #3981 

The `process_message` case statement had no else clause, so requests for unsupported methods were silently dropped. This caused clients to hang indefinitely waiting for a response. The LSP spec requires servers to respond with error code -32601 for unsupported methods.

### Implementation

Just answer with Method not found when the message method don't match any of the switch statement branches.

### Automated Tests

I added one check to see if the message `RubyLsp::Error` was sent.

### Manual Tests

Start the server via bundle exec ruby-lsp and send a request with an unsupported method (e.g. textDocument/nonExistentMethod). Verify the server responds with {"error": {"code": -32601, "message": "Method not found: textDocument/nonExistentMethod"}} instead of silently dropping the request.

### DIFF
--- a/test/server_test.rb
+++ b/test/server_test.rb
@@ -1735,7 +1735,7 @@ class ServerTest < Minitest::Test
     })
 
     error = find_message(RubyLsp::Error)
-    assert_equal(Constant::ErrorCodes::METHOD_NOT_FOUND, error.code)
+    assert_equal(RubyLsp::Constant::ErrorCodes::METHOD_NOT_FOUND, error.code)
     assert_equal("Method not found: #{non_existent_method}", error.message)
   end
 


IgorFroehner:
I have signed the CLA!

lib/ruby_lsp/server.rb
```diff
@@ -120,6 +120,12 @@ def process_message(message)
         @global_state.synchronize { @cancelled_requests << message[:params][:id] }
       when nil
         process_response(message) if message[:result]
+      else
+        send_message(Error.new(
+          id: message[:id],
+          code: -32601,
+          message: "Method not found: #{message[:method]}",
+        )) if message[:id]
```
vinistock: Minor styling suggestion, but let's also please not hard code the error code. You can access the error code from `Constant::ErrorCodes::NAME_OF_ERROR`.

```suggestion
        id = message[:id]

        if id
          send_message(Error.new(
            id: id,
            code: -32601,
            message: "Method not found: #{message[:method]}",
          ))
        end
```

test/server_test.rb
```diff
@@ -1726,6 +1726,18 @@ def test_launch_bundle_compose_forwards_argv_to_launcher
...
 
+  def test_unrecognized_request_returns_method_not_found
+    @server.process_message({
+      id: 1,
+      method: "textDocument/nonExistentMethod",
+      params: {},
+    })
+
+    error = find_message(RubyLsp::Error)
+    assert_equal(-32601, error.code)
```
vinistock: Same here, let's use the constant rather than the hard coded value.

lib/ruby_lsp/server.rb
IgorFroehner: Nice suggestion, I did it here: https://github.com/Shopify/ruby-lsp/pull/4040/changes/8a8da0c1a9dc2cdddf32f39b2e1d3fa8676b40fa

IgorFroehner:
@vinistock can I get another review here?

vinistock:
There are some type checking errors. We probably just need the `RubyLsp::` prefix before `Constant` since the tests don't re-open the `RubyLsp` module.

PR #4062, opened 2026-04-15
Fix type inference for parent scopes and method definitions with receivers

vinistock:
### Motivation

In the original Ruby LSP indexer, we did not handle the resolution of parent scopes used in definitions. For example:

```ruby
module Bar; end

module Foo
  # This defines `Bar::Baz` because the `Bar` reference gets resolved to top level `Bar`
  # through the lexical scope. Rubydex does this correctly, but the Ruby LSP doesn't
  # run resolution on its node context and misunderstands this as `Foo::Bar::Baz`
  class Bar::Baz
  end
end
```

Similarly this example

```ruby
module Foo
  module Bar; end
end

module Baz
  include Foo

  # This doesn't define `Baz::Bar::Qux`. It has access to `Bar` through inheritance
  # and defines `Foo::Bar::Qux`.
  #
  # Note that references to `Baz::Bar::Qux` also work, but not because that's the fully
  # qualified name of the constant. `Baz::Bar` finds `Foo::Bar` through inheritance,
  # allowing you to refer to `Qux`
  class Bar::Qux; end
end
```

We also did not handle method definition receivers correctly, which don't advance the lexical scope, but modify the type of `self`. For example:

```ruby
class Bar; end

class Foo
  def Bar.baz
    # self here is singleton class of Bar (Bar::<Bar>). Therefore, the var reference
    # is accessing `Bar::<Bar>#@var`
    @var
  end
end
```

### Implementation

There are two main parts to fix, which can be found in the first commit:

1. Started using a new object to represent the surrounding method definition in node context, which includes the receiver
2. Stopped pushing `def self.foo` as a singleton entry in the node context nesting. Method receivers don't advance the lexical scope and do not influence constant resolution. They only influence the type of `self`

Then I made the necessary adjustments on the type inferrer.

### Automated Tests

The last commit includes a bunch of tests verifying that we're correctly handling these cases in definition and hover.

### DIFF
--- a/lib/ruby_lsp/node_context.rb
+++ b/lib/ruby_lsp/node_context.rb
@@ -82,6 +82,8 @@ module RubyLsp
           receiver = node.receiver
 
           surrounding_method = case receiver
+          when nil
+            MethodDef.new(node.name.to_s, "none")
           when Prism::SelfNode
             MethodDef.new(node.name.to_s, "self")
           when Prism::ConstantReadNode, Prism::ConstantPathNode
--- a/lib/ruby_lsp/node_context.rb
+++ b/lib/ruby_lsp/node_context.rb
@@ -5,6 +5,21 @@ module RubyLsp
   # This class allows listeners to access contextual information about a node in the AST, such as its parent,
   # its namespace nesting, and the surrounding CallNode (e.g. a method call).
   class NodeContext
+    # Represents the surrounding method definition context, tracking both the method name and its receiver
+    class MethodDef
+      #: String
+      attr_reader :name
+
+      #: String?
+      attr_reader :receiver
+
+      #: (String name, String? receiver) -> void
+      def initialize(name, receiver)
+        @name = name
+        @receiver = receiver
+      end
+    end
+
     #: Prism::Node?
     attr_reader :node, :parent
 
@@ -14,7 +29,7 @@ module RubyLsp
     #: Prism::CallNode?
     attr_reader :call_node
 
-    #: String?
+    #: MethodDef?
     attr_reader :surrounding_method
 
     #: (Prism::Node? node, Prism::Node? parent, Array[(Prism::ClassNode | Prism::ModuleNode | Prism::SingletonClassNode | Prism::DefNode | Prism::BlockNode | Prism::LambdaNode | Prism::ProgramNode)] nesting_nodes, Prism::CallNode? call_node) -> void
@@ -26,7 +41,7 @@ module RubyLsp
 
       nesting, surrounding_method = handle_nesting_nodes(nesting_nodes)
       @nesting = nesting #: Array[String]
-      @surrounding_method = surrounding_method #: String?
+      @surrounding_method = surrounding_method #: MethodDef?
     end
 
     #: -> String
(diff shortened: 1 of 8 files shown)


test/requests/definition_expectations_test.rb
```diff
@@ -1419,6 +1419,294 @@ def bar
...
+        params: { textDocument: { uri: uri }, position: { character: 6, line: 9 } },
+      )
+
+      # @var should resolve through Bar::Baz's singleton class, not Foo::Bar::Baz
+      response = server.pop_response.response
+      assert_equal(1, response.length)
+      assert_equal(6, response[0].range.start.line)
+    end
+  end
+
```
Morriar, line 1709: Should we add a few tests using `extend`?

lib/ruby_lsp/node_context.rb
```diff
@@ -64,10 +79,16 @@ def handle_nesting_nodes(nodes)
...
-          nesting << "<#{nesting.flat_map { |n| n.split("::") }.last}>"
+          receiver = node.receiver
+
+          surrounding_method = case receiver
+          when Prism::SelfNode
+            MethodDef.new(node.name.to_s, "self")
+          when Prism::ConstantReadNode, Prism::ConstantPathNode
+            MethodDef.new(node.name.to_s, receiver.slice)
+          else
+            MethodDef.new(node.name.to_s, nil)
```
Morriar, line 92: Does this mean we'll handle cases like `def foo.bar; baz; end` as if it didn't have a receiver?

test/requests/definition_expectations_test.rb
vinistock: There's no logic in go to definition that's specific to `extend` and I've been trying to not mix Rubydex related tests with the request related tests.

For example, we already have tests to assert that we follow ancestors when searching for instance variable definitions. If that wasn't working for `extend`, that would be a bug in Rubydex and so I believe the test belongs there.

lib/ruby_lsp/node_context.rb
vinistock: Yup. It's not correct, but I didn't think it was worth the complexity to add a special case to represent an unknown type of `self`.

Method definition receivers other than `self` are already not super common and dynamic ones are even more rare.

Morriar: Could we just do this:

```ruby
when nil
  MethodDef.new(node.name.to_s, "(self)")
else
  MethodDef.new(node.name.to_s, nil)
end
```

So we keep track of implicit nil vs. something else?

lib/ruby_lsp/type_inferrer.rb
```diff
@@ -135,15 +135,90 @@ def self_receiver_handling(node_context)
...
+      surrounding_method = node_context.surrounding_method
+
+      if surrounding_method
+        receiver_name = surrounding_method.receiver
+
+        case receiver_name
+        when "self"
+          # `def self.foo` — self is the singleton of the enclosing class/module
+          return resolve_singleton_type_from_nesting(nesting)
+        when nil
```
Morriar: And then here we check on the implicit `self`:

```suggestion
        when "(self)"  
```

lib/ruby_lsp/node_context.rb
vinistock: Yeah, we can do that. I used `none` as the sentinel value instead of `(self)` to avoid confusion with `self`.

Morriar (approved):

test/type_inferrer_test.rb
```diff
@@ -44,6 +44,96 @@ def self.bar
...
+        class Foo
+          def Bar.baz
+            @@hello1
+          end
+        end
+      RUBY
+
+      # Class variables follow lexical scope, not the method receiver
+      assert_equal("Foo", @type_inferrer.infer_receiver_type(node_context).name)
+    end
```
Morriar, line 135: We may need to add a test with a dynamic value: `def foo.bar`

vinistock: I added one! Line 80.

PR #4066, opened 2026-04-17
Migrate find references to Rubydex

vinistock:
### Motivation

Migrate find references to use Rubydex. There are two things missing, which I encoded in the tests:

1. Indexing instance variable references (I have a branch going)
2. Exposing method reference receivers, so that we can be more precise about references

### Implementation

I migrated the request in a way that it's already prepared to account for instance, class and global variables as soon as we add support for it in Rubydex.

### Automated Tests

Added a bunch of new tests trying to encode all sorts of behaviours that the previous ones didn't. For example:

- Ensuring that the request only shows references if you have the cursor exactly on top of names
- The placeholder behaviour for global, class and instance variables
- Different variations of method references


lib/ruby_lsp/requests/references.rb
```diff
@@ -9,126 +9,226 @@ module Requests
     class References < Request
       include Support::Common
 
+      MAX_NUMBER_OF_METHOD_CANDIDATES_WITHOUT_RECEIVER = 30
```
Morriar, line 12: Why do we need this now?

vinistock: To avoid returning an excessive number of results back to the editor. We have the same for other features.

When we don't know the type of the receiver, like when searching for a common name like `new`, we may find an insane amount of references.

Morriar: But was it different with the old indexer?

vinistock: No, we were missing it in the old indexer.

Morriar (approved):

PR #4072, opened 2026-04-21
Migrate type hierarchy to use Rubydex

vinistock:
### Motivation

This PR migrates type hierarchy to use Rubydex.

### Implementation

I moved all constant resolution to the prepare type hierarchy request, which pays the price once for resolving and then forwards the fully qualified name to super/sub types through the flexible `data` attribute. I also ensured that users can request singleton hierarchy.

The super types request is considerably simpler because the fully qualified name is always received through the `data` attribute.

The rest of the implementation ensures that we are handling singleton classes correctly because of their implicit nature. For example:

```ruby
class Foo
  class << self
  end
end
```

The `Foo::<Foo>` declaration inherits from `Object::<Object>`, which exists in the graph, but there's no definition to back it up (it's implicit). In those cases, we need to find the definition of the original attached object, so that we can show something useful to the user.

### Automated Tests

I added a bunch more tests verifying that we behave correctly under a variety of conditions.


test/requests/prepare_type_hierarchy_test.rb
```diff
@@ -14,8 +14,8 @@ class Foo; end
         textDocument: { uri: uri },
         position: { line: 0, character: 1 },
       })
-      result = server.pop_response.response
 
+      result = server.pop_response.response
```
Morriar, line 18: Do we need these changes? Now the blank line is not at the same place than the tests you add below.

test/requests/prepare_type_hierarchy_test.rb
```diff
@@ -63,12 +63,69 @@ class Bar; end
...
+      end
+    RUBY
+
+    with_server(source) do |server, uri|
+      server.process_message(id: 1, method: "textDocument/prepareTypeHierarchy", params: {
+        textDocument: { uri: uri },
+        position: { line: 2, character: 6 },
+      })
+      result = server.pop_response.response
+
```
Morriar, line 123: Like here the blank line is different now

lib/ruby_lsp/requests/prepare_type_hierarchy.rb
```diff
@@ -5,9 +5,7 @@ module RubyLsp
   module Requests
     # The [prepare type hierarchy
     # request](https://microsoft.github.io/language-server-protocol/specification#textDocument_prepareTypeHierarchy)
-    # displays the list of ancestors (supertypes) and descendants (subtypes) for the selected type.
-    #
-    # Currently only supports supertypes due to a limitation of the index.
+    # displays the list of direct ancestors (supertypes) and descendants (subtypes) for the selected type.
```
Morriar, line 8: We delete this: `Currently only supports supertypes due to a limitation of the index.` but it doesn't look like we support subtypes yet?

vinistock: We don't, but it's not because of indexing limitations. Also, I think the absence of the `subtypes` request (which makes it not appear in the documentation) is enough to communicate that we don't yet support it.

This comment is likely to be forgotten once we add it.

test/requests/prepare_type_hierarchy_test.rb
vinistock: I standardized on the same style. Claude was adding these lines and I kept fixing it, then I understood it was inferring the style from the rest of this test.

Morriar (approved):

PR #4087, opened 2026-04-28
Decouple Shadowenv test from specific dev setup

vinistock:
### Motivation

Step towards #2877

We were actually bitten (even more than once) by the same problem that affects contributors. Our extension tests are too tied to specific machine setups, so if anything changes, they fail locally while passing on CI.

Currently, shadowenv tests are failing locally. This PR decouples the test from the real shadowenv executable and re-writes everything to test the logic without depending on machine setup.

### Implementation

The idea is to simply use stubs to verify the different situations that can happen. We no longer actually download and invoke the executable.

PR #4088, opened 2026-04-28
Decouple Ruby tests from specific machine setup

vinistock:
### Motivation

Another step towards https://github.com/Shopify/ruby-lsp/issues/2877

Our Ruby test was also coupled with specific machine setups and was in fact failing locally due to recent changes. This PR decouples the test from specific setups, so that contributors (and our own team) can always run these.

### Implementation

I extracted a better helper for stubbing VS Code settings since this is repeated pattern in our tests. Then refactored the Ruby activation tests to verify logic with stubs rather than shelling out.

### DIFF
--- a/vscode/src/test/suite/helpers.ts
+++ b/vscode/src/test/suite/helpers.ts
@@ -3,6 +3,7 @@ import os from "os";
 import fs from "fs";
 
 import * as vscode from "vscode";
+import sinon from "sinon";
 
 import { MAJOR, MINOR, RUBY_VERSION } from "../rubyVersion";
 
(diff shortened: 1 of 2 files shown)


vinistock:
* **#4088** <a href="https://app.graphite.com/github/pr/Shopify/ruby-lsp/4088?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/Shopify/ruby-lsp/4088?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#4087** <a href="https://app.graphite.com/github/pr/Shopify/ruby-lsp/4087?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`
<hr />

You must have a Graphite account in order to use the merge queue. Sign up using [this link](https://app.graphite.com/invite/github/Shopify?ref=merge-queue-instructions-in-stack-comment&prId=4088).

<h2></h2>
<sub>

</sub>

<sub>

</sub>

<h2></h2>

vscode/src/test/suite/ruby.test.ts
```diff
@@ -163,19 +163,24 @@ suite("Ruby environment activation", () => {
...
-    sandbox.stub(vscode.workspace, "getConfiguration").returns({
-      get: (name: string) => {
-        if (name === "rubyVersionManager") {
-          return { identifier: manager };
-        } else if (name === "bundleGemfile") {
-          return "";
-        }
+    const envStub = [
+      "3.3.5",
+      "~/.gem/ruby/3.3.5,/opt/rubies/3.3.5/lib/ruby/gems/3.3.0",
```
alexcrocha: I noticed some of the values are repeated, could they be constants? Or defaults as part of a helper function?

alexcrocha (approved):

vscode/src/test/suite/helpers.ts
```diff
@@ -57,6 +58,43 @@ export const LSP_WORKSPACE_FOLDER: vscode.WorkspaceFolder = {
...
+      // violates Proxy invariants. Instead, build a delegating object that overrides `get` and forwards everything
+      // else to the real configuration (preserving `this` via bind so private state access still works).
+      const wrapper: vscode.WorkspaceConfiguration = {
+        get(key: string, defaultValue?: unknown) {
+          if (key in sectionStubs) {
+            return sectionStubs[key];
+          }
+          return defaultValue === undefined ? real.get(key) : real.get(key, defaultValue);
+        },
+        has: real.has.bind(real),
```
st0012: Do we want to make sure `has` behave the same way `get` does to prevent mismatching results in the future?

vscode/src/test/suite/ruby.test.ts
```diff
@@ -39,70 +39,70 @@ suite("Ruby environment activation", () => {
...
-        if (name === "rubyVersionManager") {
-          return { identifier: ManagerIdentifier.Rbenv };
-        } else if (name === "bundleGemfile") {
-          return "";
-        }
-
-        return undefined;
+    stubWorkspaceConfiguration(sandbox, {
+      rubyLsp: {
+        rubyVersionManager: { identifier: ManagerIdentifier.Rbenv },
```
st0012, line 113: Does this stub out the `rbenvExecutablePath` config test runner may have?

st0012 (approved):

vscode/src/test/suite/helpers.ts
vinistock: Good catch.

vscode/src/test/suite/ruby.test.ts
vinistock: That never happens because of the access patterns we have in the code and the fact that stubs is just a hash of the string name for the setting.

For example, we would have something that looks like this:

```typescript
{
  "rubyVersionManager": { identifier: ManagerIdentifier.Rbenv }
}
```

If you ask for `settings.get("rubyVersionManager.rbenvExecutablePath")`, we won't find it in the stubs hash and will end up invoking the real implementation.

This would only be a concern for settings that we read and then try to access fields on. It's definitely possible, but it complicates the stubs a bit so I'll leave that as a future improvement if we actually need it.

PR #4091, opened 2026-05-01
Use method receivers to narrow reference search results

vinistock:
### Motivation

Use the method reference receivers to narrow reference search results and show more precise entries.

### Implementation

The idea is to check if the receiver is one of the expected names we are searching for.

### Automated Tests

Modified some of the existing scenarios and added another one.

### DIFF
--- a/test/requests/references_test.rb
+++ b/test/requests/references_test.rb
@@ -71,43 +71,72 @@ class ReferencesTest < Minitest::Test
     assert_includes(ref_lines, 5) # reference
   end
 
-  def test_does_not_match_simple_names_when_method_receiver_is_known
+  AMBIGUOUS_BAR_SOURCE = <<~RUBY
+    class Foo
+      def self.bar
+      end
+    end
+
+    class Qux
+      def bar
+      end
+    end
+
+    class Other
+      def self.bar
+      end
+    end
+
+    it = Qux.new
+    it.bar
+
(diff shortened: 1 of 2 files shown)


test/requests/references_test.rb
```diff
@@ -71,43 +71,72 @@ def bar
...
+    refs = find_references(AMBIGUOUS_BAR_SOURCE, { line: 18, character: 4 }, include_declarations: true)
     ref_lines = refs.map { |r| r.range.start.line }
+
     assert_includes(ref_lines, 1) # def self.bar declaration
-    assert_includes(ref_lines, 13) # Foo.bar call site
+    assert_includes(ref_lines, 18) # Foo.bar call site
     refute_includes(ref_lines, 6) # def bar (Qux) must not appear
+    refute_includes(ref_lines, 11) # def self.bar (Other) must not appear
+    refute_includes(ref_lines, 19) # Other.bar call site is filtered out by its resolved receiver
+    assert_includes(ref_lines, 16)
```
Morriar: Let's add a comment here like the others?

```suggestion
    assert_includes(ref_lines, 16) # it.bar included because receiver is unresolved
```

Morriar: Also maybe we could sort by line?

Morriar (approved):

PR #4096, opened 2026-05-07
Partially migrate completion to Rubydex

vinistock:
### Motivation

This PR partially migrates completion to use Rubydex. There are still some crashes and bugs that I identified, but at least it gets us half way there.

### Implementation

The previous way completion was implemented was honestly not very good or correct, so unfortunately this is basically a complete re-write using Rubydex as the completion engine.

It might honestly be easier to open this branch and read the code in its entirety.

### Automated Tests

Added a few more tests.

### DIFF
--- a/lib/ruby_lsp/rubydex/declaration.rb
+++ b/lib/ruby_lsp/rubydex/declaration.rb
@@ -2,12 +2,13 @@
 # frozen_string_literal: true
 
 module Rubydex
+  # @abstract
   class Declaration
     # Detail text shown on a `TypeHierarchyItem` for this declaration. Hints at multiplicity
     # when the declaration spans more than one re-open; otherwise falls back to the primary
     # definition's file name so users can quickly see where the type comes from.
     #
-    #: () -> String?
+    #: -> String?
     def lsp_type_hierarchy_detail
       defs = definitions
       count = defs.count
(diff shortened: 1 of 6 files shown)


lib/ruby_lsp/rubydex/declaration.rb
```diff
@@ -2,6 +2,7 @@
 # frozen_string_literal: true
 
 module Rubydex
+  # @abstract
```
Morriar, line 5: This should go in Rubydex

lib/ruby_lsp/rubydex/declaration.rb
```diff
@@ -20,8 +21,15 @@ def lsp_type_hierarchy_detail
...
     end
+
+    # @abstract
+    #: () -> Integer
+    def to_lsp_completion_kind
+      raise RubyLsp::AbstractMethodInvokedError
+    end
   end
 
+  # @abstract
```
Morriar, line 32: Same here

lib/ruby_lsp/rubydex/declaration.rb
```diff
@@ -45,4 +53,84 @@ def direct_supertypes
       seen.values
     end
   end
+
+  class Class
+    # @override
+    #: () -> Integer
```
Morriar: Nit, you don't need the empty parens:
```suggestion
    #: -> Integer
```

Morriar (approved):

lib/ruby_lsp/rubydex/declaration.rb
vinistock: Agreed. I'll leave it here for now so that type checking works, but I'll open a PR to mark them as abstract in the exported RBI.

lib/ruby_lsp/rubydex/declaration.rb
vinistock: Agreed. I'll leave it here for now so that type checking works, but I'll open a PR to mark them as abstract in the exported RBI.

PR #4097, opened 2026-05-07
Handle visibility in hover

vinistock:
### Motivation

This PR starts using the new visibility APIs to handle hover correctly.

### Implementation

This PR is based on the go to definition branch, so I extracted the helpers into `Common` and used it for both.

### Automated Tests

Added tests ensuring we only show hover for reachable methods and constants.

### DIFF
--- a/lib/ruby_lsp/requests/support/common.rb
+++ b/lib/ruby_lsp/requests/support/common.rb
@@ -77,6 +77,46 @@ module RubyLsp
           receiver.nil? || receiver.is_a?(Prism::SelfNode)
         end
 
+        # Returns true when a constant declaration is reachable from the call site. Private constants are only
+        # reachable from within the namespace where they are defined.
+        #
+        #: (Rubydex::Declaration declaration, String value, NodeContext node_context) -> bool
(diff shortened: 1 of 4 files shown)


lib/ruby_lsp/requests/support/common.rb
```diff
@@ -77,6 +77,46 @@ def self_receiver?(node)
           receiver.nil? || receiver.is_a?(Prism::SelfNode)
         end
 
+        # Returns true when a constant declaration is reachable from the call site. Private constants are only
+        # reachable from within the namespace where they are defined.
+        #
+        #: (Rubydex::Declaration declaration, String value, NodeContext node_context) -> bool
+        def constant_reachable_from_call_site?(declaration, value, node_context)
+          return true unless declaration.is_a?(Rubydex::Visibility) && declaration.private?
```
st0012, line 85: If declaration is a constant declaration, how could it be a `Rubydex::Visibility` too 🤔

vinistock: Because `Rubydex::Visibility` is a module. All constant types include it.

Morriar (approved):

PR #4100, opened 2026-05-08
Use Rubydex for method completion resolve

vinistock:
### Motivation

Almost exactly the same change as we did for hover, but this time for completion resolve. We show documentation for methods including their signatures.

### Implementation

Just used the created helpers to show the method's signature.

### Automated Tests

Simplified the RBS test, but kept the rest the same.

### DIFF
--- a/lib/ruby_lsp/requests/completion_resolve.rb
+++ b/lib/ruby_lsp/requests/completion_resolve.rb
@@ -53,9 +53,10 @@ module RubyLsp
         guessed_type = @item.dig(:data, :guessed_type)
         title = @item[:label].dup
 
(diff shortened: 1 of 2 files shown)


lib/ruby_lsp/requests/completion_resolve.rb
```diff
@@ -51,11 +51,12 @@ def perform
         return @item unless declaration
 
         guessed_type = @item.dig(:data, :guessed_type)
-        title = @item[:label].dup
+        title = +@item[:label].dup
```
Morriar: Why do we need `+` here? Isn't `String#dup` always returning an unfrozen string?

```suggestion
        title = @item[:label].dup
```

Morriar (approved):

Issue #4119, opened 2026-05-25
Problem with projects using different ruby versions opened in one workspace in VSC


ToniShingly:
#### VS Code Version

1.121.0

#### Ruby LSP Extension Version

0.10.4

#### Ruby LSP Server Version

0.26.9

#### Ruby LSP Add-ons

- Ruby LSP Rails (0.4.8)
- RuboCop (unknown)

#### Ruby Version

3.0.2

#### Ruby Version Manager

rbenv

#### Installed Extensions

<details>
<summary>Click to expand</summary>

- LiveServer (5.7.10)
- debugpy (2026.6.0)
- docker (0.18.0)
- es7-react-js-snippets (4.4.3)
- go (0.52.2)
- makefile-tools (0.12.17)
- markdown-preview-enhanced (0.8.27)
- prisma (31.11.0)
- python (2026.4.0)
- remote-containers (0.459.0)
- remote-wsl (0.104.3)
- ruby-lsp (0.10.4)
- todo-tree (0.0.226)
- vs-code-prettier-eslint (6.0.0)
- vscode-containers (2.4.4)
- vscode-docker (2.0.0)
- vscode-eslint (3.0.24)
- vscode-pylance (2026.2.1)
- vscode-tailwindcss (0.14.29)
</details>

#### Ruby LSP Settings

<details>
<summary>Click to expand</summary>

##### Workspace

```json
{}
```

##### User

```json
{
  "enabledFeatures": {
    "codeActions": true,
    "diagnostics": true,
    "documentHighlights": true,
    "documentLink": true,
    "documentSymbols": true,
    "foldingRanges": true,
    "formatting": true,
    "hover": true,
    "inlayHint": true,
    "onTypeFormatting": true,
    "selectionRanges": true,
    "semanticHighlighting": true,
    "completion": true,
    "codeLens": true,
    "definition": true,
    "workspaceSymbol": true,
    "signatureHelp": true,
    "typeHierarchy": true
  },
  "featuresConfiguration": {},
  "addonSettings": {},
  "rubyVersionManager": {
    "identifier": "rbenv"
  },
  "customRubyCommand": "",
  "formatter": "auto",
  "linters": null,
  "bundleGemfile": "",
  "testTimeout": 30,
  "pullDiagnosticsOn": "both",
  "useBundlerCompose": false,
  "bypassTypechecker": false,
  "rubyExecutablePath": "",
  "indexing": {},
  "erbSupport": true,
[...]

Issue #4114, opened 2026-05-22
Settings no longer allowed in workspace


g-arjones:
Since https://github.com/Shopify/ruby-lsp/commit/d9a58274ef2b35e5e80eb2200718b47a98cb06fd and https://github.com/Shopify/ruby-lsp/commit/e78a80cf497d699822aeab29b7620a18734155a4 ,`rubyLsp.customRubyCommand` and `rubyLsp.bundleGemfile` are no longer allowed in the workspace configuration. How are users supposed to deal with projects having different activation commands and Gemfiles in different paths?

These are exactly the kind of settings you want to apply to workspaces.

i10416:
I believe this is because `rubyLsp.customRubyCommand` is machine scope.

https://code.visualstudio.com/api/references/contribution-points#:~:text=A%20configuration%20setting%20can%20have%20one%20of%20the%20following%20possible%20scopes

> If no scope is declared, the default is window.

The default scope is `window`, but https://github.com/Shopify/ruby-lsp/commit/d9a58274ef2b35e5e80eb2200718b47a98cb06fd changes every those scopes to `machine`

https://github.com/Shopify/ruby-lsp/commit/d9a58274ef2b35e5e80eb2200718b47a98cb06fd prevents untrusted workspace from executing arbitrary command via customRubyCommand, so it would be safe to set the scope to `window`.

amarkowitz:
This issue is affecting our team, too.

Splines:
As my issue #4124 was flagged as duplicate of this, I just want to give an overview again here of the bug we face and a workaround, which is unfortunately downgrading the VSCode extension.

### 🎈 Bug Description

The Ruby LSP Extension does not start correctly anymore when I open any Ruby file. In the extension logs, we see this

```sh
2026-06-04 12:27:01.431 [info] (mampf) Running command: ` && ruby -EUTF-8:UTF-8 '/home/mampf/.vscode-server/extensions/shopify.ruby-lsp-0.10.4/activation.rb'` in /workspaces/mampf using shell: /usr/bin/zsh
```

This fails due to the `&&` in the beginning. Probably, the first part before the `&&` got somehow lost or was set to a null value. This leads to this error during startup:

```
Automatic Ruby environment activation with custom failed: Command failed: && ruby -EUTF-8:UTF-8 '/home/mampf/.vscode-server/extensions/shopify.ruby-lsp-0.10.4/activation.rb' zsh:1: parse error near `&&'
```

[...]

PR #4120, opened 2026-05-26
Restore window scope for several workspace specific settings

i10416:
### Motivation
~With capabilities.untrustedWorkspaces.restrictedConfigurations~ It seems impossible to set customRubyCommand at workspace level because VS Code prevents the extension from reading the machine scope options in workspace's .vscode/settings.json.

It fails with `Automatic Ruby environment activation with custom failed: Command failed: && ruby -EUTF-8:UTF-8 '/path/to/.vscode/extensions/shopify.ruby-lsp-0.10.4/activation.rb' zsh:1: parse error near &&' ` because the command here becomes ` && ruby - ...`.

https://github.com/Shopify/ruby-lsp/blob/4c109e356e3ece372b51be8866dce841ff1f487f/vscode/src/ruby/custom.ts#L12

It is sometimes case with me to find two ruby projects use different tools to manage ruby.
It is also quite likely that different projects have bundleGemfile in different locations.

This change allows users to change some workspace specific settings for each workspace.

Related to
- https://github.com/Shopify/ruby-lsp/issues/4119
- https://github.com/Shopify/ruby-lsp/issues/4114

### Implementation
Change option scope from "machine" to "machine-configurable"

### Manual Tests
Add the following .vscode/settings.json to your workspace  and enable the extension.

```json
{
  "rubyLsp.rubyVersionManager": {
    "identifier": "custom"
  },
  "rubyLsp.customRubyCommand": "{any command of your choice}"
}
```

### DIFF
--- a/vscode/package.json
+++ b/vscode/package.json
@@ -440,7 +440,7 @@
         "rubyLsp.customRubyCommand": {
           "description": "A shell command to activate the right Ruby version or add a custom Ruby bin folder to the PATH. Only used if rubyVersionManager is set to 'custom'",
           "type": "string",
-          "scope": "machine"
+          "scope": "machine-overridable"
         },
         "rubyLsp.formatter": {
           "description": "Which tool the Ruby LSP should use for formatting files",
--- a/vscode/package.json
+++ b/vscode/package.json
@@ -131,7 +131,7 @@
       {
         "command": "rubyLsp.selectRubyVersionManager",
         "title": "Select Ruby version manager",
-        "category": "Ruby LSP"
+        "category": "Ruby LSP",
       },
       {
         "command": "rubyLsp.toggleFeatures",
@@ -389,6 +389,7 @@
         },
         "rubyLsp.rubyVersionManager": {
           "type": "object",
+          "scope": "window",
           "properties": {
             "identifier": {
               "description": "The Ruby version manager to use",
@@ -440,7 +441,7 @@
         "rubyLsp.customRubyCommand": {
           "description": "A shell command to activate the right Ruby version or add a custom Ruby bin folder to the PATH. Only used if rubyVersionManager is set to 'custom'",
           "type": "string",
-          "scope": "machine-overridable"
+          "scope": "window"
         },
         "rubyLsp.formatter": {
           "description": "Which tool the Ruby LSP should use for formatting files",
@@ -478,7 +479,7 @@
         "rubyLsp.bundleGemfile": {
           "description": "Relative or absolute path to the Gemfile to use for bundling the Ruby LSP server. Do not use this if you're working on a monorepo or your project's Gemfile is in a subdirectory (look into multiroot workspaces instead). Only necessary when using a separate Gemfile for the Ruby LSP",
           "type": "string",
-          "scope": "machine",
+          "scope": "window",
           "default": ""
         },
         "rubyLsp.testTimeout": {
@@ -509,7 +510,7 @@
         "rubyLsp.rubyExecutablePath": {
           "description": "Path to the Ruby installation. This is used as a fallback if version manager activation fails",
           "type": "string",
-          "scope": "machine"
+          "scope": "window"
         },
         "rubyLsp.indexing": {
           "description": "Indexing configurations. Modifying these will impact which declarations are available for definition, completion and other features",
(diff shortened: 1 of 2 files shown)


i10416:
Just signed CLA

i10416:
BTW, this is never `true` because `workspace.getConfiguration(...).get(...)` returns default value for some primitive types;

https://github.com/Shopify/ruby-lsp/blob/4c109e356e3ece372b51be8866dce841ff1f487f/vscode/src/ruby/custom.ts#L26

https://github.com/microsoft/vscode/issues/123957#issuecomment-859425120

It must use inspect method to detect null value.

vscode/package.json
```diff
@@ -440,7 +440,7 @@
         "rubyLsp.customRubyCommand": {
           "description": "A shell command to activate the right Ruby version or add a custom Ruby bin folder to the PATH. Only used if rubyVersionManager is set to 'custom'",
           "type": "string",
-          "scope": "machine"
+          "scope": "machine-overridable"
```
i10416: or `window`?

https://code.visualstudio.com/api/references/contribution-points#:~:text=A%20configuration%20setting%20can%20have%20one%20of%20the%20following%20possible%20scopes:

vscode/package.json
```diff
@@ -389,6 +389,7 @@
         },
         "rubyLsp.rubyVersionManager": {
           "type": "object",
+          "scope": "window",
```
i10416, line 392: Ruby version manager could be different depending on workspaces.

vscode/package.json
```diff
@@ -440,7 +441,7 @@
         "rubyLsp.customRubyCommand": {
           "description": "A shell command to activate the right Ruby version or add a custom Ruby bin folder to the PATH. Only used if rubyVersionManager is set to 'custom'",
           "type": "string",
-          "scope": "machine"
+          "scope": "window"
```
i10416, line 444: It is safe to use window scope because untrusted workspace cannot execute command thanks to capabilities.untrustedWorkspaces options.

i10416:
@vinistock Hello, Vinicius. Thank you for your effort to maintain and improve this project. Recently, I tried VS Code ruby-lsp, but I'm faced with the problem due to workspace specific setting scope. Could you take a look at this PR and review the changes? I already signed CLA.

vscode/package.json
rafaelfranca: I'm not sure if this one makes sense to not be machine dependent. I know some people use a custom ruby command to add their own activation script, but I don't think this is something we want to support going forward.

`capabilities.untrustedWorkspaces` can help, but I don't think people inspect the contents of the `.vscode` folder of repositories they clone from the internet in order to click that "trust this workspace" button.

vscode/package.json
```diff
@@ -478,7 +479,7 @@
         "rubyLsp.bundleGemfile": {
           "description": "Relative or absolute path to the Gemfile to use for bundling the Ruby LSP server. Do not use this if you're working on a monorepo or your project's Gemfile is in a subdirectory (look into multiroot workspaces instead). Only necessary when using a separate Gemfile for the Ruby LSP",
           "type": "string",
-          "scope": "machine",
+          "scope": "window",
```
rafaelfranca, line 482: I think this should be window. @alexcrocha what was the reason to make it machine dependent? It is for sure a per-project config.

vscode/package.json
i10416: When someone resorts to custom rubyVersionManager, it is likely that they need something adhoc with maximum flexiblity and it is not unusual that solution differs from one project to another.

If Ruby LSP forces machine scope here, they have to maintain a lines of single shell script with mixed concerns from different, independent projects. I believe it makes Ruby LSP less convenient to use.

I feel clicking the "trust this workspace" button should be treated as "I know what I'm doing", and if they seriously care about security, they can prepare sandboxed environment outside of VS Code (and I think it would be the way to go as it is more generic solution not tied to IDE).

I will investigate if there is VS Code API that allows develops to customize the trust workspace popup to notice them that users should be pay extra attention to .vscode directory.

vscode/package.json
alexcrocha: Yes, I agree. It shouldn't have been bundled with the other changes and we should relax it

HaraldNordgren:
Hi @i10416 @rafaelfranca, any update on this?

It would help Homebrew/brew too. Right now `customRubyCommand` and `bundleGemfile` being machine-scoped forces every contributor to copy them into their personal user settings instead of the repo's own `.vscode/settings.json`.

PR #4126, opened 2026-06-04
Use env shebangs for RubyGems installs

joshheinrichs-shopify:
### Motivation

RubyGems bin wrappers hardcode the Ruby used at install time. With Nix-style Ruby installs, that path can later be garbage-collected or replaced, leaving wrappers like ruby-lsp and bundle pointing at a stale store path.

### Implementation

Ruby LSP already activates the project Ruby environment before installing or launching the server, so install the bootstrap ruby-lsp gem and any missing Bundler versions with RubyGems' env-shebang option. This lets generated wrappers resolve ruby from the activated PATH instead.

Bundler project binstubs already use env shebangs by default; this makes the RubyGems bin wrappers Ruby LSP relies on behave similarly.

### Automated Tests

Yes.

### Manual Tests

In a clean project, ruby-lsp and bundler should have `#!/usr/bin/env ruby` as the shebang.

### DIFF
--- a/lib/ruby_lsp/setup_bundler.rb
+++ b/lib/ruby_lsp/setup_bundler.rb
@@ -445,7 +445,7 @@ module RubyLsp
       requirement = Gem::Requirement.new(@bundler_version.to_s)
       return if Gem::Specification.any? { |s| s.name == "bundler" && requirement =~ s.version }
 
-      Gem.install("bundler", @bundler_version.to_s)
+      Gem.install("bundler", @bundler_version.to_s, env_shebang: true)
     end
 
     #: -> bool
--- a/test/setup_bundler_test.rb
+++ b/test/setup_bundler_test.rb
@@ -655,6 +655,38 @@ class SetupBundlerTest < Minitest::Test
     end
   end
 
+  def test_installs_missing_bundler_with_env_shebang
+    in_temp_dir do |dir|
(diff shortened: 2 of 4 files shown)


pcasaretto (approved):
Thanks for the fix, Josh. I tested this branch locally and this looks good to me.

Validation performed:

- Ruby-side regression coverage passes: `test/setup_bundler_test.rb` completed with 41 runs, 189 assertions, and 0 failures/errors when run in a Nix shell with `libyaml`/`pkg-config` available for `psych`.
- VS Code extension code compiles successfully.
- The PR-specific VS Code assertions for installing/updating `ruby-lsp` with `--env-shebang` pass. I had to run from a short `/tmp` worktree because the deeper macOS path hit the Unix socket path length limit in vscode-test; unrelated Chruby/debugger environment tests still failed in that local setup, but the assertions added by this PR passed.
- Manual RubyGems verification confirms the generated `ruby-lsp` wrapper starts with `#!/usr/bin/env ruby` when installed with `--env-shebang`.

This addresses the Nix/Ruby refresh failure mode we saw where generated binstubs captured an absolute stale Nix store Ruby path. The main boundary to keep in mind is that this prevents newly installed/updated wrappers from going stale; existing stale binstubs still need to be regenerated by an install/update/pristine path. That is an acceptable scope for this PR.

:shipit:

rafaelfranca (approved):

PR #4133, opened 2026-06-11
Support "on type formatting" in Zed

### LINKED ISSUE #2971: Code completion for blocks is unhelpful due to cursor placement
#### VS Code Version

1.95.3

0.8.16

#### Ruby LSP Server Version

0.22.1

#### Ruby LSP Addons

- Ruby LSP Rails

#### Ruby Version

3.3.6

#### Ruby Version Manager

custom

#### Installed Extensions

<details>

- amx-netlinx (0.6.1)
- better-toml (0.3.2)
- copy-text (0.4.9)
- cpptools (1.22.11)
- csharp (2.55.29)
- elixir-ls (0.24.2)
- gitlens (16.0.5)
- go (0.42.1)
- haml (1.4.1)
- isort (2023.10.1)
- material-icon-theme (5.15.0)
- platformio-ide (3.3.3)
- postcss (2.0.0)
- python (2024.20.0)
- ruby-lsp (0.8.16)
- rust-analyzer (0.3.2212)
- vscode-ansi (1.1.7)
- vscode-dotnet-runtime (2.2.3)
- vscode-eslint (3.0.10)
- vscode-npm-script (0.3.29)
- vscode-projects-plus (1.24.0)
- vscode-zipfs (3.0.0)
</details>

#### Ruby LSP Settings

<details>

##### Workspace

```json
{}
```

##### User

```json
{
  "enabledFeatures": {
    "codeActions": true,
    "diagnostics": true,
    "documentHighlights": true,
    "documentLink": true,
    "documentSymbols": true,
    "foldingRanges": true,
    "formatting": true,
    "hover": true,
    "inlayHint": true,
    "onTypeFormatting": true,
    "selectionRanges": true,
    "semanticHighlighting": true,
    "completion": true,
    "codeLens": true,
    "definition": true,
    "workspaceSymbol": true,
    "signatureHelp": true,
    "typeHierarchy": true
  },
  "featuresConfiguration": {},
  "addonSettings": {},
  "rubyVersionManager": {
    "identifier": "custom"
  },
  "customRubyCommand": "rbenv init - fish | source",
  "formatter": "auto",
  "linters": null,
  "bundleGemfile": "",
  "testTimeout": 30,
  "branch": "",
  "pullDiagnosticsOn": "both",
  "useBundlerCompose": false,
  "bypassTypechecker": false,
  "rubyExecutablePath": "",
  "indexing": {},
  "erbSupport": true,
  "featureFlags": {}
}
```

</details>

### Reproduction steps

ruby-lsp auto-completes ruby blocks which is potentially a great feature, but the implementation is not quite there. Examples:

* Type `method do |`. ruby-lsp autocompletes the argument pipes to `method do ||`, but the cursor is *after* the closing pipe, so attempting to finish typing e.g. `|a|` results in `method do ||a||`. You have to remember to arrow left before typing the argument.
* Type `def method_name<enter>`. ruby-lsp autocompletes the method structure with an empty line and `end`, but the cursor is after `end` so you have to remember to arrow up before typing the method body. Same with `do ... end` blocks.

This behavior IMO is unexpected because in other cases where VSCode autocompletes, the cursor is placed where you would expect. E.g. if you type `"` or `(` you get `""` or `()` but the cursor is between the two characters and you can immediately continue typing.

I really like the completion feature in concept, but since the cursor is placed in the wrong spot it slows me down to the point where disabling completion and typing the whole block myself is faster.

janko:
### Motivation

Zed doesn't support snippet syntax (`$0`) in `document/onTypeFormatting` responses, nor is it possible for its Ruby extension to implement it. Because the cursor position isn't right after edits, Zed's Ruby extension currently has `onTypeFormatting` disabled by default for Ruby LSP. 

Fixes https://github.com/Shopify/ruby-lsp/issues/2971

### Implementation

For "end" and "heredoc end", we can work around this by inserting them in the line *below* the cursor. This prevents the cursor from moving after the inserted text, avoiding the need to rewind it. We can't support this at the EOF, as there is no next line, but that's the rarer scenario.

This won't work for formatting on the same line, because Zed cannot rewind the cursor. Zed already supports auto-closing brackets, so we skip closing curly braces. We also skip pipe auto-close, because it's better not to add anything than to end up with incorrect cursor placement.

### Automated Tests

I've added unit tests.

### Manual Tests

To test this in Zed, you first need to enable on type formatting in Zed settings:

```
{
  "lsp": {
    "ruby-lsp": {
      "initialization_options": {
        "enabledFeatures": {
          "onTypeFormatting": true
        }
      }
    }
  }
}
```

### DIFF
--- a/lib/ruby_lsp/requests/on_type_formatting.rb
+++ b/lib/ruby_lsp/requests/on_type_formatting.rb
@@ -75,6 +75,8 @@ module RubyLsp
 
       #: -> void
       def handle_pipe
+        return unless supports_snippet_anchor?
+
         current_line = @lines[@position[:line]]
         return unless /((?<=do)|(?<={))\s+\|/.match?(current_line)
 
@@ -106,6 +108,7 @@ module RubyLsp
 
       #: -> void
       def handle_curly_brace
+        return unless supports_snippet_anchor?
         return unless /".*#\{/.match?(@previous_line)
 
         add_edit_with_text("}")
@@ -129,9 +132,13 @@ module RubyLsp
         next_line = @lines[@position[:line] + 1]
 
         if current_line.nil? || current_line.strip.empty? || current_line.include?(")") || current_line.include?("]")
-          add_edit_with_text("\n")
-          add_edit_with_text("#{indents}end")
-          move_cursor_to(@position[:line], @indentation + 2)
+          if supports_snippet_anchor?
+            add_edit_with_text("\n")
+            add_edit_with_text("#{indents}end")
+            move_cursor_to(@position[:line], @indentation + 2)
+          elsif next_line
+            add_edit_with_text("#{indents}end\n", { line: @position[:line] + 1, character: 0 })
+          end
         elsif next_line.nil? || next_line.strip.empty?
           add_edit_with_text("#{indents}end\n", { line: @position[:line] + 1, character: @position[:character] })
           move_cursor_to(@position[:line] - 1, @indentation + @previous_line.size + 1)
@@ -141,9 +148,15 @@ module RubyLsp
       #: (String delimiter) -> void
       def handle_heredoc_end(delimiter)
         indents = " " * @indentation
-        add_edit_with_text("\n")
-        add_edit_with_text("#{indents}#{delimiter}")
-        move_cursor_to(@position[:line], @indentation + 2)
+        next_line = @lines[@position[:line] + 1]
+
+        if supports_snippet_anchor?
+          add_edit_with_text("\n")
+          add_edit_with_text("#{indents}#{delimiter}")
+          move_cursor_to(@position[:line], @indentation + 2)
+        elsif next_line
+          add_edit_with_text("#{indents}#{delimiter}\n", { line: @position[:line] + 1, character: 0 })
+        end
       end
 
       #: (String spaces) -> void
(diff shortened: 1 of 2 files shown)


JacobNWolf:
Very eager to see this merged!

vinistock:
Thank you for the PR. Can't we completely do without the cursor movement if we switch to the strategy you proposed for Zed? Even for other editors?

Also, I do want to point out that the curly brace formatting is for occurrences inside interpolation `"#{"` and not for bare `{}` which VS Code also auto-closes.

janko:
> Can't we completely do without the cursor movement if we switch to the strategy you proposed for Zed? Even for other editors?

Well, considering that I haven't found a way to do inline formatting (e.g. `|`), or make `end` auto-insertion work at the end of the file, at this stage we cannot drop the cursor movement implementation.

> Also, I do want to point out that the curly brace formatting is for occurrences inside interpolation "#{" and not for bare {} which VS Code also auto-closes.

Oh, didn't come to my mind, that makes sense 👍🏻

janko:
The `#{` can probably be autocompleted by Zed by modifying `autoclose_before` setting to include double quotes and backticks, and similar for VSCode. The `|` symbol can't seem to reliably be autoclosed by the editor, so we're back to the language server with cursor movement.

PR #4153, opened 2026-07-03
Add Reveal In Explorer and trigger continuous test reruns on any workspace file save

Code-Farmer-Ming:
### Motivation

There are two recurring workflow gaps in the current Test Explorer experience:

1. It is not always easy to jump from the current test context back to the explorer tree.
2. Continuous testing is too narrow when it only reruns on test-file edits, since many failures are caused by changes outside test files (application code, config, shared helpers, etc.).

This PR addresses both by improving navigation and broadening rerun triggers to better match real-world development flow.

### Implementation

1. Added Reveal In Explorer action  
- Introduced a dedicated Reveal In Explorer action so a test can be quickly focused in the Test Explorer at any time.
- Aligned command/type naming consistently across generation, resolution, and extension command wiring.

2. Expanded Continuous Test triggers to workspace-wide saves  
- Updated continuous mode so reruns are triggered by saves to any file in the target workspace, not only test files.
- Preserved workspace scoping and non-file filtering to avoid cross-workspace noise and invalid triggers.

### Automated Tests

- Updated and extended test coverage for Reveal In Explorer behavior and command resolution.
- Updated CodeLens-related assertions to account for the additional test action.
- Verified compile/test behavior for the modified extension paths; unrelated environment-dependent failures remain outside the scope of this change.

### Manual Tests

1. Reveal In Explorer  
- Open a test file and wait for discovery.
- Trigger Reveal In Explorer from test actions/CodeLens.
- Confirm the corresponding test is focused/revealed in Test Explorer.

2. Continuous Test (workspace-wide trigger)  
- Start a continuous run for a test or test group.
- Save a non-test file in the same workspace.
- Confirm the continuous run reruns automatically.
- Save a test file and confirm rerun still works.
- In multi-root setup, verify only saves in the target workspace trigger reruns.


Code-Farmer-Ming:
I have signed the CLA!

PR #4177, opened 2026-07-27
Re-regenerate global binstubs when Ruby version changes

vinistock:
### Motivation

RubyGems installs executable binstubs that sometimes bake the Ruby version. When the Ruby version for the project is updated, if the version manager reuses that path for any reason, we crash on boot because the baked version doesn't match the one actually being used.

### Implementation

The fix is to invoke `gem pristine ruby-lsp --only-executables --env-shebang` when Ruby gets upgraded. To achieve that, we need to remember a slightly more complex update state, so that this is triggered not only when a certain amount of time passed, but when Ruby gets an upgrade too.

FYI: I cut [v0.10.5-preview](https://github.com/Shopify/ruby-lsp/releases/tag/vscode-ruby-lsp-v0.10.5-preview) to test out the changes before cutting a stable release.

### DIFF
--- a/vscode/package.json
+++ b/vscode/package.json
@@ -2,7 +2,7 @@
   "name": "ruby-lsp",
   "displayName": "Ruby LSP",
   "description": "VS Code plugin for connecting with the Ruby LSP",
-  "version": "0.10.4",
+  "version": "0.10.5",
   "publisher": "Shopify",
   "repository": {
     "type": "git",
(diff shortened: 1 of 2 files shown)


rafaelfranca (approved):

rafaelfranca (approved):

rafaelfranca (approved):

st0012 (approved):

larister (approved):
Thanks for sorting this @vinistock!

PR #4182, opened 2026-07-30
Fix multiline ERB comment parsing

### LINKED ISSUE #4181: Multi-line `<%# %>` ERB comments are parsed as Ruby, so semantic highlighting colors the prose as method calls
### Ruby LSP Information

  - ruby-lsp 0.26.10
  - shopify.ruby-lsp VS Code extension 0.10.5
  - VS Code 1.128.1
  - ruby 4.0.6 (2026-07-14 revision 03b6d3f889) +PRISM [arm64-darwin25]

### Reproduction steps

Only the **first** line of a multi-line `<%# … %>` ERB comment is treated as a comment. Every following line is handed to Prism as Ruby, so the prose parses as a chain of method calls and semantic highlighting repaints each word with the `method` token type.

The TextMate grammar is correct here — `erb.cson.json` scopes the whole block as `comment.block.erb` — so this is only the semantic-token layer painting over it. On screen the first line stays comment-colored and the rest of the block turns code-colored.

1. Open a `.erb` file in VS Code with the Ruby LSP extension active.
2. Add a single-line `<%# … %>` comment and a multi-line one (snippet below).
3. Look at the multi-line comment: line 1 is comment-colored, every later line is colored as code.
4. Optionally request `textDocument/semanticTokens/full` for the document (`languageId: "erb"`, default `initializationOptions`) to see the tokens directly.

```erb
<%# A single-line comment stays a comment. %>

<%# The first line of this comment is fine,
    but every following line is parsed as Ruby
    and each word here becomes a method token %>

<div>markup</div>
```

Image

**Expected:** no semantic tokens inside either comment.

**Actual:** the single-line comment yields 0 tokens; the three-line block yields 13, all `method`:

```
  ok          L1-1 (1 line)   tokens inside: 0
  MISCOLORED  L3-5 (3 lines)  tokens inside: 13
        L4 "but"       -> method
        L4 "every"     -> method
        L4 "following" -> method
        L4 "line"      -> method
        L4 "is"        -> method
        L4 "parsed"    -> method
        L4 "as"        -> method
        L5 "each"      -> method
        L5 "word"      -> method
        L5 "here"      -> method
        L5 "becomes"   -> method
        L5 "a"         -> method
        L5 "token"     -> method
```

Two details that make this easy to recognize:

- L3 (the first comment line) is clean — the problem starts on line 2 of the block.
- Words that happen to be Ruby keywords produce no token and so keep the comment color (`and` on L5). In a real comment most words turn code-colored while the occasional keyword stays green, which looks arbitrary until you know the cause.

It scales with block length: a 6-line comment in one of our views produces 53 tokens.

### Cause

`ERBScanner#scan_char` in `lib/ruby_lsp/erb_document.rb` (present on `main` as of filing):

```ruby
when "<"
  if next_char == "%"
    @inside_ruby = true
    @current_pos += 1
    push_char("  ")

    if next_char == "=" && @source[@current_pos + 2] == "="
      @current_pos += 2
      push_char("  ")
    elsif next_char == "=" || next_char == "-"
      @current_pos += 1
      push_char(" ")
    end
```

There is no branch for `#`, so `<%#` opens an ordinary Ruby region and the `#` is pushed into the Ruby buffer as a plain character. `when "\n"` then appends the newline to `@ruby` without closing the region or tracking that a `#` comment is open.

The extracted Ruby for the snippet above is therefore:

```ruby
    # The first line of this comment is fine,
    but every following line is parsed as Ruby
    and each word here becomes a method token
```

The `#` comments out line 1 only; lines 2-3 are live code as far as Prism is concerned.

A fix would be for the scanner to recognize `<%#` and blank the region through to the matching `%>`, the way it already blanks the tag delimiters, rather than emitting it as Ruby.

### Workaround

For anyone hitting this, disabling semantic tokens for ERB only, which keeps them for `.rb` files:

```json
"[erb]": { "editor.semanticHighlighting.enabled": false }
```

Pybsama:
### Motivation

Closes #4181.

Multiline `<%# ... %>` comments currently leak every line after the first into
the virtual Ruby source. Prism can then parse prose as method calls, which
produces incorrect semantic highlighting and can affect other features that use
the ERB AST.

### Implementation

- Recognize the `#` marker when opening an ERB tag and track an explicit ERB
  comment state.
- Replace comment content with equal-length spaces in both virtual sources while
  preserving LF and CRLF sequences and source offsets.
- Clear the comment state on both `%>` and `-%>` closing paths, then resume
  normal host-language and Ruby scanning.
- Ignore apparent `<%` openers inside an active ERB comment.

### Automated Tests

Added document-level regression tests for:

- multiline comments with LF and `-%>`, followed by real Ruby;
- multiline comments with CRLF and `%>`, followed by both HTML and real Ruby;
- exact virtual-source masking, newline preservation, syntax validity, and
  scanner recovery after each closing form.

Validation performed locally:

- `bin/test test/erb_document_test.rb`
- `bundle exec rake`
- `bundle exec rubocop`
- `bundle exec srb tc`
- `bin/tapioca check-shims`

### Manual Tests

I did not run a VS Code UI test. To verify manually, open an ERB document with
semantic highlighting enabled, add a multiline `<%# ... %>` block followed by
HTML and `<%= ruby_code %>`, and confirm that the comment remains entirely
comment-colored while highlighting resumes for the following content.

### DIFF
--- a/lib/ruby_lsp/erb_document.rb
+++ b/lib/ruby_lsp/erb_document.rb
@@ -82,6 +82,7 @@ module RubyLsp
         @ruby = +"" #: String
         @current_pos = 0 #: Integer
         @inside_ruby = false #: bool
+        @inside_erb_comment = false #: bool
       end
 
       #: -> void
@@ -100,12 +101,16 @@ module RubyLsp
 
         case char
         when "<"
-          if next_char == "%"
+          if !@inside_erb_comment && next_char == "%"
             @inside_ruby = true
             @current_pos += 1
             push_char("  ")
 
-            if next_char == "=" && @source[@current_pos + 2] == "="
+            if next_char == "#"
+              @inside_erb_comment = true
+              @current_pos += 1
+              push_char(" ")
+            elsif next_char == "=" && @source[@current_pos + 2] == "="
               @current_pos += 2
               push_char("  ")
             elsif next_char == "=" || next_char == "-"
@@ -121,6 +126,7 @@ module RubyLsp
           if @inside_ruby && next_char == "%" &&
               @source[@current_pos + 2] == ">"
             @current_pos += 2
+            @inside_erb_comment = false
             push_char("   ")
             @inside_ruby = false
           else
@@ -130,9 +136,10 @@ module RubyLsp
           end
         when "%"
           if @inside_ruby && next_char == ">"
-            @inside_ruby = false
             @current_pos += 1
+            @inside_erb_comment = false
             push_char("  ")
+            @inside_ruby = false
           else
             push_char(
               char, #: as !nil
@@ -159,7 +166,10 @@ module RubyLsp
 
       #: (String char) -> void
       def push_char(char)
-        if @inside_ruby
+        if @inside_erb_comment
+          @ruby << " " * char.length
+          @host_language << " " * char.length
+        elsif @inside_ruby
           @ruby << char
           @host_language << " " * char.length
         else
--- a/test/erb_document_test.rb
+++ b/test/erb_document_test.rb
@@ -73,6 +73,53 @@ class ERBDocumentTest < Minitest::Test
     assert_equal("   \r\nbar   ", document.parse_result.source.source)
   end
 
+  def test_multiline_erb_comments_are_excluded_from_virtual_sources
+    source = +"<%# first line\ncomment_method(:argument)\n-%>\n<%= visible_call %>"
+    document = RubyLsp::ERBDocument.new(
+      source: source,
+      version: 1,
+      uri: URI("file:///foo.erb"),
+      global_state: @global_state,
+    )
+
+    document.parse!
+
+    expected_ruby = source.gsub(/[^\r\n]/, " ")
+    visible_call = "visible_call"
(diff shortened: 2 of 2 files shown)


Pybsama:
Revalidated this patch against current `main` (`2057d0b1`) on 2026-08-31:

- Regression test first fails on unmodified current `main`: 11 runs, 27 assertions, 2 failures. With this patch: 11 runs, 29 assertions, 0 failures.
- `bundle exec srb tc`, `bin/tapioca check-shims`, and `bundle exec rubocop` pass.
- Full `bundle exec rake` passes: indexer 318 runs / 1,650,408 assertions; main suite 18,543 runs / 59,691 assertions; 0 failures, 0 errors, 14 skips.
- The focused ERB document test passed 50 consecutive runs.
- Semgrep ran 44 Ruby rules over the changed files with 0 findings.
- The patch merges cleanly with current `main`; the rebased validation commit produces the same stable patch ID and merge tree as the existing PR patch.

I am keeping the PR as a draft because the contributor CLA requires the contributor's own legal attestation.

PR #4201, opened 2026-09-06
Allow add-ons to register and execute commands

pvcresin:
### Motivation

Ruby LSP add-ons can contribute Code Lenses and other editor features, but there is currently no generic way for an add-on to register and handle the commands referenced by those features. This forces add-ons that need server-side command handling to provide additional editor integration.

### Implementation

- Add `Addon#commands` and `Addon#execute_command` hooks.
- Add `Addon#command_id`, which appends a UUID generated for each add-on instance to logical command identifiers.
- Register the scoped add-on command IDs through the standard `client/registerCapability` request after add-ons are loaded.
- Handle the standard `workspace/executeCommand` request by resolving scoped IDs back to logical commands and routing them to the owning add-on.
- Support multiple Ruby LSP server instances in multi-root workspaces without command name collisions.
- Respect the `workspace.executeCommand.dynamicRegistration` capability reported by the client and ignore errored add-ons.
- Document the API and add protocol/server coverage.

Add-ons are loaded after the initialize response, so dynamic registration keeps the existing add-on lifecycle unchanged. The VS Code extension does not require changes because its existing `vscode-languageclient` dependency handles standard dynamic execute-command registration.

### Automated Tests

- `test/requests/execute_command_test.rb` (registration, command ID uniqueness, and routing)
- `test/global_state_test.rb`
- Related Server, Addon, and Code Lens tests
- RuboCop
- Sorbet

### Manual Tests

Not run in a full VS Code session. The new server tests verify the dynamic registration payload and command dispatch. An end-to-end test can be performed with an add-on that returns a command from `commands`, implements `execute_command`, and exposes a Code Lens using `command_id`.

### DIFF
--- a/lib/ruby_lsp/addon.rb
+++ b/lib/ruby_lsp/addon.rb
@@ -280,5 +280,17 @@ module RubyLsp
     def resolve_test_commands(items)
       []
     end
+
+    # Returns the commands provided by the add-on
+    # @overridable
+    #: -> Array[String]
+    def commands
+      []
+    end
+
+    # Executes a command provided by the add-on
+    # @overridable
+    #: (String command, Array[untyped] arguments) -> untyped
+    def execute_command(command, arguments); end
   end
 end
--- a/test/requests/execute_command_test.rb
+++ b/test/requests/execute_command_test.rb
@@ -1,0 +1,109 @@
+# typed: true
+# frozen_string_literal: true
+
+require "test_helper"
+
+module RubyLsp
+  class ExecuteCommandTest < Minitest::Test
+    def setup
+      @addon_class = Class.new(Addon) do
+        def activate(global_state, outgoing_queue); end
+        def deactivate; end
+
+        def name
+          "Command Add-on"
+        end
+
+        def version
+          "0.1.0"
+        end
+
+        def commands
+          ["commandAddon.echo"]
+        end
+
+        def execute_command(command, arguments)
+          { command: command, arguments: arguments }
+        end
+      end
+
+      Addon.addon_classes.delete(@addon_class)
+    end
+
+    def teardown
+      Addon.addons.select { |addon| addon.is_a?(@addon_class) }.each(&:deactivate)
+      Addon.addons.delete_if { |addon| addon.is_a?(@addon_class) }
+    end
+
+    def test_executes_an_addon_command
+      Addon.addons << @addon_class.new
+
+      with_server(load_addons: false) do |server, _uri|
+        server.process_message(
+          id: 1,
+          method: "workspace/executeCommand",
+          params: {
+            command: "commandAddon.echo",
+            arguments: ["hello"],
+          },
+        )
+
+        result = server.pop_response
+        assert_instance_of(Result, result)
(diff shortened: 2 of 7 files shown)


pvcresin:
I have signed the CLA!

pvcresin:
I have signed the CLA!

soutaro (changes requested):
Thanks for working on this! There’s a multi-root workspace case we need to handle. Ruby LSP runs a separate server for each workspace folder, so two folders loading the same add-on will register the same command names, which results in a registration error in VS Code.

Could we make the command names unique per server instance? One option is a UUID suffix, with a helper that add-ons can use when building Code Lenses and other command references. Ruby LSP could map those names back to the original commands when dispatching to the add-on. I tested this approach in VS Code and confirmed that commands reached the correct server, including after restarting one of them.

pvcresin:
@soutaro 
Thanks for pointing this out! I updated the implementation in 41dda40.
Each add-on instance now gets a UUID-based command ID through Addon#command_id. Ruby LSP registers these scoped IDs and resolves them back to the original logical command before dispatching to Addon#execute_command. This prevents command collisions between Ruby LSP server instances in multi-root workspaces.
I also added tests covering command ID uniqueness and routing commands to the correct add-on instance, and updated the add-on documentation. The targeted tests, RuboCop, and Sorbet checks all pass.

lib/ruby_lsp/addon.rb
```diff
@@ -1,6 +1,8 @@
 # typed: strict
 # frozen_string_literal: true
 
+require "securerandom"
```
soutaro, line 4: I think this `require` can be deleted, since we usually load `ruby_lsp` library to use `addon`.

test/requests/execute_command_test.rb
```diff
@@ -0,0 +1,150 @@
...
+        assert_equal(Constant::ErrorCodes::INVALID_PARAMS, error.code)
+        assert_equal("Unknown command: commandAddon.missing", error.message)
+      end
+    end
+
+    def test_registers_addon_commands_after_initialization
+      addon = @addon_class.new
+      Addon.addons << addon
+
+      server = Server.new(test_mode: true)
```
soutaro, line 124: Is there any reason not to use `with_server(load_addons: false)` here?

soutaro (commented):
Thanks! There are a few minor things, but the PR looks good overall.

lib/ruby_lsp/addon.rb
pvcresin: Removed the explicit require "securerandom" in c0bf20a.

test/requests/execute_command_test.rb
pvcresin: Thank you for the suggestion. 
I updated the test in bc946d4 to use `with_server(load_addons: false)`.
The existing capability setup and stubs remain in place, while the shared helper now handles server cleanup, so the manual ensure and run_shutdown are no longer necessary.

pvcresin:
@soutaro All done with the handling work.

PR #4204, opened 2026-09-13
Replace `rv ruby run` command with `rv run`

thedanbob:
### Motivation

`rv ruby run` was deprecated in rv 0.5.3 and has now been removed in 0.7.0.

### Implementation

Replaced `rv ruby run` with `rv run`.

### Automated Tests

Updated `rv.test.ts` with the new command.

### Manual Tests

Install rv 0.7.0 and set it as the ruby version manager.


thedanbob:
I have signed the CLA

